PR #76, opened 2024-01-08
feat: Polymorphic function definitions

mark-koch:
Nested generic functions are not supported yet


mark-koch:
Implemented in #522

PR #116, opened 2024-01-18
chore: Fix pre-commit config, fix linting

aborgna-q:
The `mypy` step in `pre-commit` didn't get updated after the name change.

`pre-commit run --all-files` threw multiple errors, so I updated those too.
(I had to update the mypy check version to avoid some missing type errors).


.pre-commit-config.yaml
```diff
@@ -1,63 +1,65 @@
...
-    - id: debug-statements
+      - id: check-added-large-files
+      - id: check-case-conflict
+      - id: check-executables-have-shebangs
+      - id: check-merge-conflict
+      - id: check-toml
+      - id: check-vcs-permalinks
+      - id: check-yaml
+      - id: detect-private-key
+      # - id: end-of-file-fixer
```
qartik, line 13: I commented out eof fixer as it conflicts with the string comparison checks we do in `.err` files, similar to trailing whitespace below.

qartik:
I see that mypy and ruff are being run on the CI, but it might be nice to just run this pre-commit config there instead of separate checks.

Also, I noticed `poetry install` doesn't seem to install some necessary dependencies such as pytest, pytket and tket2. We should document those in the README under developer instructions. I am new to poetry, I wonder if it allows additional deps analogous to `pip install .[dev]` that we did before.

aborgna-q:
There's a `[tool.poetry.group.dev.dependencies]` section for dev dependencies.

I think #118 adds tket2 as an optional dep. You need to pass `--extras=pytket` to poetry to install them.

ss2165 (approved):

PR #121, opened 2024-01-19
doc: Update readme

mark-koch:


README.md
```diff
@@ -1,24 +1,61 @@
 # Guppy
 
-## About
+Guppy is a quantum programming language that is fully embedded into Python.
+It allows you to write high-level hybrid quantum programs with classical control flow and mid-circuit measurements using Pythonic syntax:
+
+```python
+from guppylang import guppy
```
mark-koch, line 7: Strictly speaking, we also need to import `Qubit` or do `from __future__ import annotations`.

Do we want to add that here?

ss2165 (approved):

PR #125, opened 2024-01-19
feat: add top-level imports

### LINKED ISSUE #100: "Guppy prelude" for runnable scripts
The `planqc` program examples in guppy_runner assume that we have automatically include some elements like the `@guppy` decorator, `Qubit`, etc.

These could be explicitly defined as a "guppy program prelude" (not to be confused with Hugr's prelude), that the runner can automatically include before running the script.

Is there something else that should be included there?

### LINKED ISSUE #123: Top level re-export of decorator + common functionality

### LINKED ISSUE #127: fix example in readme with quantum module import

ss2165:
Closes #123 
Closes #100 
Closes #127

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,14 +4,13 @@ Guppy is a quantum programming language that is fully embedded into Python.
 It allows you to write high-level hybrid quantum programs with classical control flow and mid-circuit measurements using Pythonic syntax:
 
 ```python
-from guppylang import guppy, Qubit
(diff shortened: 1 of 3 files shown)


aborgna-q:
I'd say we can close #100 with this.
Having the scripts do something like `from guppylang import *` seems reasonable enough.

mark-koch (approved):

PR #140, opened 2024-01-25
fix: Stop exiting interpreter on error

### LINKED ISSUE #109: Don't exit python interpreter on compiler error (kills kernel).
or make it easy to configure

mark-koch:
Instead, let the exception go through and use `sys.excepthook` to customise how the error message is printed.

Fixes #109

### DIFF
--- a/guppylang/error.py
+++ b/guppylang/error.py
@@ -2,16 +2,15 @@ import ast
 import functools
 import sys
 import textwrap
-from collections.abc import Callable, Sequence
+from collections.abc import Callable, Iterator, Sequence
+from contextlib import contextmanager
 from dataclasses import dataclass, field
+from types import TracebackType
 from typing import Any, TypeVar, cast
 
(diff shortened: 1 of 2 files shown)


guppylang/error.py
```diff
@@ -169,27 +180,40 @@ def format_source_location(
...
+            try:
+                return f(*args, **kwargs)
+            except GuppyError as err:
+                # For normal usage, this `try` block is not necessary since the
+                # excepthook is automatically invoked when the exception (which is being
+                # reraised below) is not handled. However, when running tests, we have
+                # to manually invoke the hook to print the error message, since the
+                # tests always have to capture exceptions.
+                if "pytest" in sys.modules:
+                    hook(type(err), err, err.__traceback__)
```
ss2165, line 216: https://docs.pytest.org/en/latest/example/simple.html#pytest-current-test-environment-variable

this seems like a more stable test?

ss2165 (approved):
thanks 👍

PR #145, opened 2024-02-02
feat: Improve import system

### LINKED ISSUE #133: Implicit module imports
Following the changes in #105, it would be nice if we could avoid `gupply.load(...)` calls and instead read of the imported Guppy modules from the Python environment.

```python
from guppy import Qubit

@guppy
def foo(q: Qubit) -> Qubit:
   ...

guppy.compile()  # Sees that `Qubit` is in scope, so automatically imports the corresponding `quantum` module
```

### LINKED ISSUE #134: Rethink the import system
* qualified/unqualified?
* open/closed?
* renaming?

mark-koch:
Closes #133 and closes #134

### DIFF
--- a/guppylang/checker/core.py
+++ b/guppylang/checker/core.py
@@ -4,7 +4,7 @@ import itertools
 from abc import ABC, abstractmethod
 from collections.abc import Iterable, Iterator
 from dataclasses import dataclass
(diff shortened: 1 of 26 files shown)


mark-koch:
Superseded by #432

Issue #142, opened 2024-01-26
Should `Qubit` type be lowercase `qubit`?


ss2165:
to match other "builtins" `int, float, bool, list, linst` etc.

croyzor:
Pros:
* More pythonic
* Reinforces that `qubit` is a built in thing to guppy

Cons:
* Syntax highlighting is a bit worse. E.g.:
```python
f(q : Qubit) # Highlighted like a type
f(q : qubit) # Plain
```

PR #165, opened 2024-03-06
refactor!: Make `qubit` type lower case

croyzor:
Addresses #142 

---

Also: update 27 test cases from
```python
from guppylang.hugr.tys import Qubit
```
to
```python
from guppylang.prelude.quantum import qubit
```
and some whitespace trimming


guppylang/prelude/quantum.py
```diff
@@ -21,89 +21,110 @@ def quantum_op(op_name: str) -> ops.OpType:
...
     linear=True,
 )
-class Qubit:
+class qubit:
     pass
 
 
 @guppy.hugr_op(quantum, quantum_op("H"))
-def h(q: Qubit) -> Qubit: ...
+def h(q: qubit) -> qubit:
```
croyzor, line 29: Weird, the pre-commit hook made me change this with `ruff format`... Version mismatch? Does the ci have a specific config that allows this?

qartik: The pre-commit config needs to be updated, it's stuck at ruff v0.1.13. Try `pre-commit autoupdate`.

qartik: Though it seems the github action is using a slightly older version too https://github.com/ChartBoost/ruff-action/blob/main/.pre-commit-config.yaml#L17

Perhaps it's best to run the pre-commit config on the CI instead of relying on a github action that will be out of our control.

croyzor: "on our end" meaning in local dev environments?

qartik: Yes, at https://github.com/CQCL/guppylang/blob/main/.pre-commit-config.yaml#L28

ss2165 (approved):
🙏

PR #166, opened 2024-03-06
chore: Update pre-commit packages

croyzor:


qartik (approved):

qartik:
One suggestion is to perhaps pin the same version of ruff in the github action, so we don't run into this inconsistency again. See bottom of https://github.com/ChartBoost/ruff-action#ruff-action

croyzor:
I assumed we'd want to keep up to date with developments in ruff

PR #174, opened 2024-03-19
feat: New type representation with parameters

mark-koch:
Most of the diff lines are just moving things to new files, so it's not as bad as it looks :D (see comments below)

These are the main new things:
* Prepare support for types and functions that are generic over constant values (e.g. bounded nats):
  - Generic types and functions are now defined in terms of `Parameter`s  and `Argument`s that can be either types or constants (see `tys/param.py` and `tys/arg.py`).
  - Implementations for `ConstParam` and `ConstArg` will follow in the future
* Improved pretty printing of types (see `tys/printing.py`)
* Add a notion of `TypeDefinition` (see `tys/definition.py`) that replaces the ad-hoc creation of Python classes to define new types
* `BoolType` is no longer a `SumType`. This was a hugr implementation detail and not relevant for the Guppy type system.

Drive by renaming:
* Rename function type members from `args/return` to `inputs/output` since `args` is already used now
* Rename `GuppyType` to `Type`

### DIFF
--- a/guppylang/tys/common.py
+++ b/guppylang/tys/common.py
@@ -27,7 +27,7 @@ class Transformer(ABC):
     """Abstract base class for a type visitor that transforms types."""
 
     @abstractmethod
-    def transform(self, arg: Any) -> Any | None:
+    def transform(self, arg: Any, /) -> Any | None:
         """Transforms the object.
 
         Return a transformed type or `None` to continue the recursive visit.
@@ -38,7 +38,7 @@ class Visitable(ABC):
     """Abstract base class for objects that can be recursively visited."""
 
     @abstractmethod
-    def visit(self, visitor: Visitor) -> None:
+    def visit(self, visitor: Visitor, /) -> None:
         """Accepts a visitor on the object.
 
         Implementors of this method should first invoke the visitor on the object
@@ -51,7 +51,7 @@ class Transformable(Visitable, ABC, Generic[T]):
     """Abstract base class for objects that can be recursively transformed."""
 
     @abstractmethod
-    def transform(self, transformer: Transformer) -> T:
+    def transform(self, transformer: Transformer, /) -> T:
         """Accepts a transformer on the object.
 
         Implementors of this method should first invoke the transformer on the object


guppylang/custom.py
```diff
@@ -53,7 +50,10 @@ def __init__(
     @property  # type: ignore[override]
     def ty(self) -> FunctionType:
         if self._ty is None:
-            return UnknownFunctionType()
+            # If we don't have a specified type, then the extension writer has to
+            # provide their own type-checking code. Therefore, it doesn't matter which
+            # type we return here since it will never be inspected.
+            return FunctionType([], NoneType())
```
mark-koch, line 56: I got rid of `UnknownFunctionType` since it was very hacky. It's fine to return an arbitrary function type here since the actual checking logic is delegated to the extension writer

guppylang/decorator.py
```diff
@@ -142,48 +142,9 @@ def type(
 
         def dec(c: type) -> type:
             _name = name or c.__name__
-
-            @dataclass(frozen=True)
-            class NewType(GuppyType):
```
mark-koch, line 147: Getting rid of this dynamic class creation is one of the big wins of this PR

guppylang/error.py
```diff
@@ -84,58 +82,6 @@ class InternalGuppyError(Exception):
     """Exception for internal problems during compilation."""
 
 
-class UndefinedPort(OutPortV):
-    """Dummy port for undefined variables.
-
-    Raises an `InternalGuppyError` if one tries to access one of its properties.
-    """
```
mark-koch, line 91: This class was actually unused

guppylang/tys/parsing.py
mark-koch, line 1: Most of this file was copied from the old `gtypes.py`

guppylang/tys/subst.py
mark-koch, line 1: Most of this file was copied from the old `gtypes.py`

guppylang/tys/ty.py
mark-koch, line 1: Most of this file was copied from the old `gtypes.py`

aborgna-q (approved):
Lots of changes, but LGTM :+1:

guppylang/tys/common.py
```diff
@@ -0,0 +1,60 @@
...
+    @abstractmethod
+    def to_hugr(self) -> T:
+        """Computes the Hugr representation of the object."""
+
+
+class Visitor(ABC):
+    """Abstract base class for a type visitor that transforms types."""
+
+    @abstractmethod
+    def visit(self, arg: Any, /) -> bool:
```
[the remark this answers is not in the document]
mark-koch: Indeed, they should be everywhere 👍

guppylang/tys/printing.py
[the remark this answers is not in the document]
mark-koch: Some additional parentheses in some cases to disambiguate, for example `(T -> T) -> T` was incorrectly printed as `T -> T -> T` before.

Also, if multiple type variables have the name chosen by the user, they now get disambiguated as `T`, `T'1`, `T'2`, ...

PR #179, opened 2024-04-08
feat: Add a unified definition system

### LINKED ISSUE #170: Leave a pointer for @guppy-defined functions
Currently the decorator leaves a panicking function behind.
This
```python
@guppy
def my_fn(...) -> ...:
    ...
```
Turns to this (after registering the function in the guppy module)
```python
my_fn = lambda ... : raise GuppyError
```

Ideally, we would leave an object that contains a reference to the module and function name,
so we can use the identifier to extract the relevant node from the compiled hugr.
```python
@guppy(module)
def my_fn(...) -> ...:
    ...

assert my_fn.name == "my_fn"
assert my_fn.module == module
# This still panics
my_fn()
```

### LINKED ISSUE #175: Unify definitions
Currently, the `Globals` object stores values, parameters, and types in different namespaces. This list will grow further, once we add external constants etc. Also, they are only identified by name which is not ideal (in particular, instance methods are encoded as `type.func_name`).

Instead we should do the following:

* Add a `Definition` ABC with subclasses `TypeDef`, `FuncDef`, `FuncDecl`, `CustomFuncDef`, ...
* Each definition has a `DefId` that uniquely identifies it across modules
* `Globals` holds a single `dict[str, Definition]` together with a class member lookup `dict[TypeDef, dict[str, Definition]]`

mark-koch:
* Adds a new `Definition` ABC for all user-defined things on module-level. Currently, these are functions, types, and type vars.
* All definition related files are moved to the submodule `guppylang.definition`
* Definitions are identified by a globally unique id instead of their name. The `Globals` container is restructured accordingly.
* The `int(x)`, `bool(x)`, `qubit()`, etc. functions are replaced with calls to the `__new__` constructor of the type.
* The module compilation code is greatly simplified by:
   -  Adding ABCs for the compilation actions: `ParsableDef -> CheckableDef -> CompilableDef -> CompiledDef`
   - Adding type aliases for compilation stages: `RawDef < ParsedDef < CheckedDef < CompiledDef`

Closes #175 and also closes #170 (as a drive-by)


guppylang/definition/function.py
mark-koch, line 1: This file was mostly copied from `guppylang/checker/func_checker.py` and `guppylang/compiler/func_compiler.py`

guppylang/prelude/builtins.py
mark-koch, line 1: Here, we replace the old `bool(x)`, `int(x)` etc functions with a `__new__(x)` constructor

guppylang/definition/function.py
```diff
@@ -0,0 +1,147 @@
...
+@dataclass(frozen=True)
+class RawFunctionDef(ParsableDef):
+    """A raw function definition provided by the user.
+
+    The raw definition stores exactly what the user has written (i.e. the AST), without
+    any additional checking or parsing. Furthermore, we store the values of the Python
+    variables in scope at the point of definition.
+    """
+
+    defined_at: ast.FunctionDef
```
mark-koch, line 29: A thought: Maybe we should store a reference to the actual Python function and do the AST generation in `parse()` instead? I think this would fit better with traditional compiler lingo?

guppylang/definition/common.py
```diff
@@ -0,0 +1,118 @@
...
+    name: str
+    defined_at: ast.FunctionDef | None
+
+    @property
+    @abstractmethod
+    def description(self) -> str:
+        """ """
+
+
+class ParsableDef(Definition, ABC):
```
croyzor, line 58: I'm not sure what the point is of inheriting `Definition, ABC` since `Definition` already inherits `ABC`?

guppylang/definition/common.py
```diff
@@ -0,0 +1,118 @@
...
+
+class CompilableDef(Definition, ABC):
+    """Abstract base class for definitions that still need to be compiled to Hugr.
+
+    The result of compilation should be a `CompiledDef` with a pointer to the Hugr node
+    that was created for this definition.
+    """
+
+    @abstractmethod
+    def compile(self, graph: Hugr, parent: Node) -> "CompiledDef":
```
croyzor, line 99: Is there something more instructive we can call `compile` and `compile_contents`? `compile_decl` and `compile_def`? Or `compile_outer` and `compile_inner`?

guppylang/definition/function.py
mark-koch: Mhm you mean "dereferencing" in terms of looking up the source code?

The parsed `ast.FunctionDef` is an AST. There is no way to go back to a callable function from this. At the moment, we're just throwing away the Python function.

I agree, the current version works. However, I think it could be nicer to do the actual parsing in the function called `parse` instead of doing it earlier in the `@guppy` decorator?

guppylang/definition/common.py
mark-koch: Fixed

guppylang/definition/common.py
mark-koch: Done

guppylang/definition/function.py
mark-koch: I implemented this in #181. I have to admit it's a bit unclean for nested function definitions. What do you think?

guppylang/prelude/_internal.py
```diff
@@ -271,14 +269,15 @@ def compile(self, args: list[OutPortV]) -> list[OutPortV]:
             float_value(0.0), get_type(self.node), self.dfg.node
         )
         zero = self.graph.add_load_constant(zero_const.out_port(0), self.dfg.node)
-        return Float.__ne__.compile_call(
+        [out] = Float.__ne__.compile_call(
```
croyzor, line 272: Why are this and `Float.__truediv__` now returning lists?

guppylang/checker/cfg_checker.py
```diff
@@ -131,7 +131,7 @@ def check_bb(
     if bb == cfg.entry_bb:
         assert len(bb.predecessors) == 0
         for x, use in bb.vars.used.items():
-            if x not in cfg.ass_before[bb] and x not in globals.values:
+            if x not in cfg.ass_before[bb] and x not in globals:
```
croyzor, line 134: I can't see any changes in the `cfg` directory, which makes me think that `ass_before` still has values in `DefAssignmentDomain == set[str]` rather than the `DefId` which is indexing `globals`?

guppylang/prelude/_internal.py
mark-koch: They also returned a list before, but the issue is that mypy can no longer infer this.

The problem is that `Float` now defines a decorated `__new__` method which throws off mypy, so it can only infer `Float: Any` and thus `Float.__ne__: Any` and `Float.__ne__.compile_call: Any`. By explicitly matching on the list, we can trick mypy into still accepting the code.

Generally, I think that refering back to methods defined on `Float` in this way is more of a hack and not ideal. Once module linking works properly, we can hopefully replace this with a regular Guppy function instead of this hacky `CustomCallCompiler`

guppylang/module.py
```diff
@@ -93,184 +95,129 @@ def load(self, m: Union[ModuleType, "GuppyModule"]) -> None:
...
+    ) -> dict[DefId, CheckedDef]:
+        """Helper method to parse and check raw definitions."""
+        raw_globals = globals | Globals(raw_defs, {}, {}, {})
+        parsed = {
+            defn.id: defn.parse(raw_globals) if isinstance(defn, ParsableDef) else defn
+            for defn in raw_defs.values()
+        }
+        parsed_globals = globals | Globals(parsed, {}, {}, {})
+        return {
+            defn.id: (
```
croyzor, line 158: It seems a bit of a code smell that we only use the values here - it seems we have a dictionary where all of the values contain their key

guppylang/definition/custom.py
```diff
@@ -1,129 +1,145 @@
...
-        if self._ty is None:
+        # TODO: This should be raised during checking, not compilation!
+        if not self.higher_order_value:
             raise GuppyError(
                 "This function does not support usage in a higher-order context",
                 node,
             )
 
-        if self._ty.parametrized:
+        if self.ty.parametrized:
```
croyzor, line 142: I'm not sure where this `self.ty` is getting set

guppylang/definition/function.py
```diff
@@ -0,0 +1,175 @@
...
+        dfg: DFContainer,
+        graph: Hugr,
+        globals: CompiledGlobals,
+        node: AstNode,
+    ) -> list[OutPortV]:
+        """Compiles a call to the function."""
+        # TODO: Hugr should probably allow us to pass type args to `Call`, so we can
+        #   avoid loading the function to manually add a `TypeApply`
+        if type_args:
+            func = graph.add_load_constant(self.hugr_node.out_port(0), dfg.node)
```
croyzor, line 152: This could be `self.load`

guppylang/compiler/expr_compiler.py
```diff
@@ -178,8 +180,8 @@ def visit_LocalCall(self, node: LocalCall) -> OutPortV:
         return self._pack_returns(rets)
 
     def visit_GlobalCall(self, node: GlobalCall) -> OutPortV:
-        func = self.globals[node.func.name]
-        assert isinstance(func, CompiledFunction)
+        func = self.globals[node.def_id]
```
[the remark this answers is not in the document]
croyzor: No it shouldn't!

guppylang/checker/core.py
```diff
@@ -34,109 +33,112 @@
...
     def __or__(self, other: "Globals") -> "Globals":
+        impls = {
+            def_id: self.impls.get(def_id, {}) | other.impls.get(def_id, {})
+            for def_id in self.impls.keys() | other.impls.keys()
+        }
         return Globals(
-            self.values | other.values,
-            self.type_defs | other.type_defs,
-            self.param_vars | other.param_vars,
+            {**self.defs, **other.defs},
```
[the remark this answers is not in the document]
mark-koch: Morally yes, but `Globals.defs` is an immutable mapping (since it has to be covariant), which doesn't support `|`

guppylang/checker/expr_checker.py
```diff
@@ -315,10 +316,20 @@ def visit_Name(self, node: ast.Name) -> tuple[ast.Name, Type]:
...
-            value = self.ctx.globals.values[x]
-            return with_loc(node, GlobalName(id=x, value=value)), value.ty
+        elif x in self.ctx.globals:
+            # Look-up what kind of definition it is
+            match self.ctx.globals[x]:
+                case ValueDef() as defn:
+                    return with_loc(node, GlobalName(id=x, def_id=defn.id)), defn.ty
+                # For types, we return their `__new__` constructor
+                case TypeDef() as defn if constr := self.ctx.globals.get_instance_func(
+                    defn, "__new__"
```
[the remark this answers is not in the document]
mark-koch: There is also `int(x)` which takes everthing that implements `__int__`, same for `float(x)` and `bool(x)`.

Before, these were defined as functions, but now the names is already taken by the type.

Python handles this in the same way: `int` is a class/type and "calling" a class executes the constructor

guppylang/definition/custom.py
mark-koch: It is set while parsing the `RawCustomFunctionDef`:

https://github.com/CQCL/guppylang/blob/05c5724962e6a5841ec2caf3f688ed9fbdad0808/guppylang/definition/custom.py#L69-L82

guppylang/checker/expr_checker.py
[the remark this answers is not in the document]
croyzor: Sure, but I'm only worried about the `Qubit` case because adding unnecessary `QAlloc`s (when checking type signatrues for example) will mess up any qubit resource estimates!

However, I'm fairly satisfied we're doing the right thing here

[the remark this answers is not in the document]
mark-koch: Ah I see. Yes, `Name` nodes in types are handled by different code in `tys.parsing`

croyzor (approved):

PR #181, opened 2024-04-10
chore: Move parsing logic from module to definition

mark-koch:


guppylang/checker/func_checker.py
```diff
@@ -102,7 +102,15 @@ def check_nested_func_def(
...
             func = ParsedFunctionDef(
-                def_id, func_def.name, func_def, func_ty, globals.python_scope
+                def_id,
+                func_def.name,
+                func_def,
+                func_ty,
+                # We don't have a Python function for nested definitions, so we insert a
+                # dummy function here. This should be fine since this should never be
+                # inspected after parsing and this definition is not leaked to the user.
+                lambda: None,
```
croyzor, line 112: This has got me thinking about the class hierarchy a bit. How much is it really the case that we want to be able to used a `Checked` or `Parsed` definition in place of a `Parsable` or `Raw` one? I'm hoping that they are all used in distinct phases. If we lose some of the inheritance (such that `Checked` and `Compiled` don't inherit `Raw`) we could get rid of this line?

mark-koch: Yeah I think this is better (see latest commit) 👍

croyzor (approved):

PR #191, opened 2024-04-11
ci: build wheels and publish to pypi

aborgna-q:
Builds the wheels on every push to main, and publishes the to pypi when pushing a new version tag


mark-koch (approved):
Thank you!

PR #194, opened 2024-04-11
docs: Add compiler API docs

mark-koch:
See https://cqcl.github.io/guppylang/api-docs/ for the deployed docs


guppylang/tys/ty.py
```diff
@@ -423,15 +423,21 @@ def transform(self, transformer: Transformer) -> "Type":
...
 
 
-# We define the `Type` type as a union of all `TypeBase` subclasses defined above. This
-# models an algebraic data type and enables exhaustiveness checking in pattern matches
-# etc.
-# Note that this might become obsolete in case the `@sealed` decorator is added:
-#  * https://peps.python.org/pep-0622/#sealed-classes-as-algebraic-data-types
-#  * https://github.com/johnthagen/sealed-typing-pep
+#: The type of parametrized Guppy types.
 ParametrizedType: TypeAlias = FunctionType | TupleType | SumType | OpaqueType
```
mark-koch, line 427: We have to add these `#:` docstrings to all type aliases and also do `from __future__ import annotations` everywhere to make it work. I'll leave that to another PR

docs/api-docs/conf.py
```diff
@@ -0,0 +1,105 @@
...
+    "sphinx.ext.viewcode",
+    "sphinx.ext.intersphinx",
+]
+
+html_theme = "sphinx_book_theme"
+
+html_title = "Guppy compiler development docs"
+
+html_theme_options = {
+    "repository_url": "https://github.com/CQCL/guppy",
```
aborgna-q, line 26:
```suggestion
    "repository_url": "https://github.com/CQCL/guppylang",
```

.github/workflows/docs.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          python-version: '3.10'
+          cache: "poetry"
+      - name: Install Guppy
+        run: poetry install --with docs
+      - name: Build docs
+        run: |
+          cd docs
+          ./build.sh
+          cd ..
```
aborgna-q, line 42: The PWD is local to each step
```suggestion
```

aborgna-q (approved):

PR #196, opened 2024-04-16
feat: Allow calling a tensor of functions

### LINKED ISSUE #141: Syntax for quantum function tensor product
Maybe `*` to indicate the product:

```python
q0, q1 = (h*x)(q0, q1)
# equivalent to:
q0 = h(q0)
q1 = x(q1)
```

Alternatives: 
- `,` comma like BRAT, but confusion with tuples
- `|`
- `@` usually matrix multiplication so confusion with composition rather than product

croyzor:
Allow calling functions as if using tensor product of operators, with tuple syntax: `(f, g)(a, b, c)`. This should work whenever all of the tuple elements are themselves callable, and otherwise raise an error.

- Calls of tuples compile down to a `TensorCall` which packs up the `LocalCall`s and `GlobalCall`s to the functions in the tensor product
- Some parts of typechecking have been generalised to allow more arguments than the function can process and return the leftovers. I did this instead of being concrete about arg numbers in order to facilitate use of the `*` operator for chaining calls like `(f, g)(*h(a, b))` but there might be an easier way. 

Resolves #141 

Update: This no longer adds any new nodes or types -- it just uses tuples and adds a couple of helper functions to deal with them


mark-koch:
I'll do a more detailed review later, but here are some first thoughts:

I'm not sure if constructing a `FunctionTensor` during CFG construction is the best move since this will only work if you call the tuple directly. For example, I belive the following doesn't work, right?
```python
@guppy
def foo() -> int:
    return 42

@guppy
def bar() -> bool:
    return True

@guppy
def baz() -> tuple[int, bool]:
    f = foo, bar
    return f()
```
In order to call "dynamic" function tuples, you'd have to wait until type checking a call `f(args)`, infer the type of `f`, and check if the type is callable, i.e. if
* it is function type, or
* it is a type that implements `__call__`, or
* it is tuple of callable types

If you made it such that callable types check against function types, then you wouldn't even need the `tensor` type since you could use `Callable` everywhere:

```python
@guppy
def tensor_foo_bar() -> Callable[[], tuple[int, bool]]:
    return foo, bar
```

guppylang/checker/expr_checker.py
```diff
@@ -470,11 +565,46 @@ def visit_Call(self, node: ast.Call) -> tuple[ast.expr, Type]:
...
+            call_nodes: list[ast.expr] = []
+            for func in node.func.elts:
+                args, return_ty, inst, remaining_args = synthesize_call_with_leftovers(
+                    func_ty, remaining_args, node, self.ctx
+                )
+                processed_args.extend(args)
+                if isinstance(return_ty, TupleType):
+                    return_tys.extend(return_ty.element_types)
+                else:
+                    return_tys.append(return_ty)
```
mark-koch, line 595: There is a helper function `type_to_row` that does this

guppylang/checker/expr_checker.py
```diff
@@ -231,14 +258,81 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
+                        )
+                    )
+                    f_processed = instantiate_poly(f, f_ty, inst)
+
+                    check_inst(f_ty, inst, node)
+                    # Expect that each function is a `CallableDef`
+                    if isinstance(f_processed, GlobalName):
+                        assert isinstance(
+                            self.ctx.globals[f_processed.def_id], CallableDef
+                        )
```
mark-koch, line 294: This needs to be a user error instead of assert

guppylang/checker/expr_checker.py
```diff
@@ -470,11 +565,46 @@ def visit_Call(self, node: ast.Call) -> tuple[ast.expr, Type]:
...
+                if isinstance(return_ty, TupleType):
+                    return_tys.extend(return_ty.element_types)
+                else:
+                    return_tys.append(return_ty)
+
+                if isinstance(func, GlobalName):
+                    assert isinstance(self.ctx.globals[func.def_id], CallableDef)
+                    call_nodes.append(
+                        GlobalCall(def_id=func.def_id, args=args, type_args=inst)
+                    )
```
mark-koch, line 601: Similar to the checking part, you need `self.ctx.globals[func.def_id].synthesize_call()` and replace the assert with a user error

guppylang/checker/expr_checker.py
```diff
@@ -470,11 +565,46 @@ def visit_Call(self, node: ast.Call) -> tuple[ast.expr, Type]:
...
+                else:
+                    call_nodes.append(LocalCall(func=func, args=args))
```
mark-koch, line 603: `TypeApply` is needed here as well

guppylang/compiler/expr_compiler.py
```diff
@@ -172,23 +179,74 @@ def _pack_returns(self, returns: list[OutPortV]) -> OutPortV:
...
+        if isinstance(func.ty, FunctionType):
+            call = self.graph.add_indirect_call(func, args)
+            rets = [call.out_port(i) for i in range(len(type_to_row(output)))]
+        elif isinstance(func.ty, TupleType) and parse_function_tensor(func.ty):
+            func_ports = self._unpack_tuple(func)
+            # Now we have to manage this hand-off
+            remaining_args = args
+            for func in func_ports:
+                outs, remaining_args = self._compile_tensor_with_leftovers(func, args)
+                rets.extend(outs)
```
mark-koch, line 205: This doesn't work with nested tuples

guppylang/tys/ty.py
```diff
@@ -520,3 +520,44 @@ def _unify_args(
...
+def function_tensor_signature(tys: list[FunctionType]) -> FunctionType:
+    """Compute the combined function signature of a list of functions"""
+    inputs: list[Type] = []
+    outputs: list[Type] = []
+    for fun_ty in tys:
+        inputs.extend(fun_ty.inputs)
+        if isinstance(fun_ty.output, TupleType):
+            outputs.extend(fun_ty.output.element_types)
+        else:
+            outputs.append(fun_ty.output)
```
mark-koch, line 552: You can use `type_to_row`

guppylang/tys/ty.py
```diff
@@ -520,3 +520,44 @@ def _unify_args(
...
+    return FunctionType(inputs, TupleType(outputs))
```
mark-koch, line 553: Use `row_to_type` to avoid creating unary output tuples

guppylang/hugr/hugr.py
```diff
@@ -491,11 +492,13 @@ def add_call(
         self, def_port: OutPortV, args: list[OutPortV], parent: Node | None = None
     ) -> VNode:
         """Adds a `Call` node to the graph."""
-        assert isinstance(def_port.ty, FunctionType)
+        func_ty = is_function(def_port.ty)
+        assert isinstance(func_ty, FunctionType)
```
mark-koch, line 496: So, `is_function` also accepts tuples. I think this is quite dangerous here since tuples in Hugr are not callable, so you can construct stuff that is invalid Hugr

tests/integration/test_tensor.py
```diff
@@ -0,0 +1,166 @@
...
+
+    @guppy(module)
+    def glo(x: int) -> int:
+        return x
+
+    @guppy(module)
+    def foo(x: int) -> int:
+        return glo(x)
+
+    validate(module.compile())
```
mark-koch, line 133: This is already tested in `test_call.py`

tests/integration/test_tensor.py
mark-koch, line 1: Could you add some tests for local tensor calls? Something like

```
f = foo, bar
f(x, y, z)
```

Also, some tests involving polymorphic functions would be good

guppylang/checker/expr_checker.py
```diff
@@ -188,13 +192,36 @@ def _synthesize(
...
-        for i, el in enumerate(node.elts):
-            node.elts[i], s = self.check(el, ty.element_types[i].substitute(subst))
-            subst |= s
-        return node, subst
+
+        # Tuples can either be inert python tuples or tuples of functions which
+        # can be called in guppy. The former thing is checkable, but in the
+        # latter case we should be able to synthesise function types for the
+        # elements. Check here whether the given type is a tuple of function
+        # types to work out which case we're in.
```
mark-koch, line 202: Is this distinction needed anymore? Can we just revert the code to how it was before?

mark-koch (commented):
I think this is looking quite good overall 👍 My main concern is that the call checking code is becoming a bit hard to read, covering so many cases. Also, I'm not sure if nested tuples are always handled correctly?

One suggestion to consider is that we could get rid of the special logic that is used to check tuple literals and instead always rely on the general version using `function_tensor_signature()` and `check_call()` / `synthesize_call()`? This would simplify the logic substantially and give us nesting for free

guppylang/checker/expr_checker.py
croyzor: It seems this is needed to make the row semantics of tuples work, i.e. having `(T) == T` for tuples of functions, but I think it's simpler to delete this and accept this behaviour

guppylang/checker/expr_checker.py
```diff
@@ -266,8 +266,9 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
             node.func = instantiate_poly(node.func, func_ty, inst)
             return with_loc(node, LocalCall(func=node.func, args=args)), return_ty
 
-        if isinstance(func_ty, TupleType) and parse_function_tensor(func_ty):
-            function_elements = parse_function_tensor(func_ty)
+        if isinstance(func_ty, TupleType) and (
+            function_elements := parse_function_tensor(func_ty)
+        ):
```
mark-koch, line 247: One thing to note: An empty `function_elements` lists also evaluate to `False` which means we can't call empty tuples. But I think that's fine, the expression `()()` isn't that useful :D

guppylang/checker/expr_checker.py
```diff
@@ -214,12 +214,17 @@ def visit_Tuple(self, node: ast.Tuple, ty: Type) -> tuple[ast.expr, Subst]:
...
-                # should have independent variables
-                subst = unify(fun_ty, elt_ty, {}) or {}
-                big_subst |= {}
+                if not isinstance(fun_ty, FunctionType):
+                    return self._fail(elt_ty, fun_ty, node.elts[i])
+                else:
+                    elem_tys.append(fun_ty)
+                    # Start with an empty substitution because the function types
+                    # should have independent variables
+                    subst = unify(fun_ty, elt_ty, {}) or {}
```
mark-koch:
```suggestion
                    subst = unify(fun_ty, elt_ty, {})
```

guppylang/checker/expr_checker.py
```diff
@@ -188,13 +192,25 @@ def _synthesize(
...
+        if not parse_function_tensor(ty):
+            subst: Subst = {}
+            for i, el in enumerate(node.elts):
+                node.elts[i], s = self.check(el, ty.element_types[i].substitute(subst))
+                subst |= s
+            return node, subst
+        else:
+            elem_tys: list[FunctionType] = []
+            for i, (elt, elt_ty) in enumerate(zip(node.elts, ty.element_types)):
+                node.elts[i], fun_ty = self._synthesize(elt, allow_free_vars=True)
```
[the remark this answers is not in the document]
mark-koch: Yes we can resolve them but we would need to go back to the place where variables were created and fill in the solution. There is currently no good way to do this

guppylang/checker/expr_checker.py
```diff
@@ -295,16 +295,12 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
-                        )
-                        call_nodes.append(
-                            GlobalCall(
-                                def_id=f_processed.def_id,
-                                args=processed_args,
-                                type_args=inst,
-                            )
+                        defn = self.ctx.globals[f_processed.def_id]
+                        assert isinstance(defn, CallableDef)
+                        call_node, subst = defn.check_call(
```
mark-koch: You're overriding the `subst` you got from `check_call_with_leftovers`. The substitution you get here should always be empty since `f_ty.output` shouldn't have any unsolved vars.

guppylang/checker/expr_checker.py
mark-koch: Agreed!

guppylang/tys/ty.py
```diff
@@ -520,3 +520,31 @@ def _unify_args(
...
+
+
+def function_tensor_signature(tys: list[FunctionType]) -> FunctionType:
+    """Compute the combined function signature of a list of functions"""
+    inputs: list[Type] = []
+    outputs: list[Type] = []
+    for fun_ty in tys:
+        inputs.extend(fun_ty.inputs)
+        outputs.extend(type_to_row(fun_ty.output))
+    return FunctionType(inputs, row_to_type(outputs))
```
mark-koch, line 550: I just noticed you're missing the parameters. You need to merge the params of each functions and shift the variable indices for each function you add to the right.

Maybe we should
```python
assert not fun_ty.parameterized
```
But then we have to catch the polymorphic case with a user error earlier

guppylang/checker/expr_checker.py
```diff
@@ -231,14 +234,41 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
             node.func = instantiate_poly(node.func, func_ty, inst)
             return with_loc(node, LocalCall(func=node.func, args=args)), return_ty
-        elif f := self.ctx.globals.get_instance_func(func_ty, "__call__"):
-            return f.check_call(node.args, ty, node, self.ctx)
+
+        if isinstance(func_ty, TupleType) and (
+            function_elements := parse_function_tensor(func_ty)
+        ):
+            assert isinstance(function_elements, list)
+            tensor_ty = function_tensor_signature(function_elements)
```
mark-koch, line 249: If we go with this suggestion, we should check for polymorphic functions at this point and error to the user

guppylang/checker/expr_checker.py
```diff
@@ -231,14 +234,41 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
+            big_subst: Subst = {}
+
+            processed_args, big_subst, inst = check_call(
+                tensor_ty, node.args, tensor_ty.output, node.func, self.ctx
+            )
+
+            if len(inst) > 0:
+                raise GuppyTypeError(
+                    "Polymorphic functions in tuples are not supported"
+                )
```
mark-koch, line 260: Then this can be an assert that `inst` is empty

guppylang/checker/expr_checker.py
mark-koch: (Same for the `synthesize_call` case)

tests/integration/test_tensor.py
mark-koch, line 1: Could you also add a test for calls of nested tuples?

mark-koch: Also, some tests for the error cases would be nice

mark-koch (commented):
Nice, this looks a lot cleaner!

guppylang/checker/expr_checker.py
```diff
@@ -635,18 +685,39 @@ def type_check_args(
...
+    Checks that all unification variables can be inferred.
+    """
+    exprs, subst, leftovers = type_check_args_with_leftovers(
+        inputs, func_ty, subst, ctx, node
+    )
+    check_leftovers_nil(len(exprs), leftovers, node)
+    return exprs, subst
+
+
+def type_check_args_with_leftovers(
```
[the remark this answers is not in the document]
croyzor: Reverted!

guppylang/checker/expr_checker.py
```diff
@@ -231,14 +234,44 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
             check_inst(func_ty, inst, node)
             node.func = instantiate_poly(node.func, func_ty, inst)
             return with_loc(node, LocalCall(func=node.func, args=args)), return_ty
-        elif f := self.ctx.globals.get_instance_func(func_ty, "__call__"):
-            return f.check_call(node.args, ty, node, self.ctx)
+
+        if isinstance(func_ty, TupleType) and (
+            function_elements := parse_function_tensor(func_ty)
+        ):
+            assert isinstance(function_elements, list)
```
mark-koch, line 248: nit: I this assert is no longer needed
```suggestion
```

guppylang/checker/expr_checker.py
```diff
@@ -231,14 +234,44 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
+            if any(f.parametrized for f in function_elements):
+                raise GuppyTypeError(
+                    "Polymorphic functions in tuples are not supported", node.func
+                )
+
+            tensor_ty = function_tensor_signature(function_elements)
+
+            processed_args, subst, inst = check_call(
+                tensor_ty, node.args, tensor_ty.output, node.func, self.ctx
+            )
```
mark-koch, line 258: I think if you do this instead you can get rid of the unification below
```suggestion
            processed_args, subst, inst = check_call(
                tensor_ty, node.args, ty, node, self.ctx
            )
            assert len(inst) == 0
            return return with_loc(node, TensorCall(...)), subst
```

guppylang/compiler/expr_compiler.py
```diff
@@ -183,6 +188,50 @@ def visit_LocalCall(self, node: LocalCall) -> OutPortV:
...
+            outs, remaining_args = self._compile_tensor_with_leftovers(
+                elem, remaining_args
+            )
+            rets.extend(outs)
+        assert remaining_args == []
+
+        if len(rets) == 1:
+            return rets[0]
+        else:
+            return self._pack_returns(rets, node.out_tys)
```
mark-koch, line 209: Why is this `len == 1` special case needed?

guppylang/checker/expr_checker.py
mark-koch: Also notice the updated location `node.func` to `node`. I think this should improve the error location in some of the golden tests

mark-koch (approved):
Just some final nits. The tests are looking good 👍

Feel free to merge afterwards

guppylang/compiler/expr_compiler.py
croyzor: It should be redundant, duplicating the logic of `pack_returns`. I'll remove it

PR #200, opened 2024-05-01
chore: Update ruff version used by CI

croyzor:


guppylang/checker/expr_checker.py
```diff
@@ -933,7 +933,7 @@ def eval_py_expr(node: PyExpr, ctx: Context) -> Any:
             "`py(...)` expression",
             e.node or node,
         ) from None
-    except Exception as e:  # noqa: BLE001
```
croyzor, line 936: ruff seems very confused by this. If we remove the `noqa`, the lint complains about this line. If we keep the lint, ruff complains that it's not used

aborgna-q (approved):
👍

PR #206, opened 2024-05-13
chore: Pass globals to check_instantiate

mark-koch:
This is required as a precursor for checking struct definitions


acl-cqc (approved):
👍

PR #210, opened 2024-05-13
test: Add struct tests

mark-koch:


tests/error/util.py
```diff
@@ -12,11 +12,9 @@
 
 def run_error_test(file, capsys):
     file = pathlib.Path(file)
-    spec = importlib.util.spec_from_file_location("test_module", file)
-    py_module = importlib.util.module_from_spec(spec)
 
     with pytest.raises(GuppyError):
-        spec.loader.exec_module(py_module)
+        importlib.import_module(f"tests.error.{file.parent.name}.{file.name}")
```
mark-koch, line 17: For some reason, inspect.getsource() doesn't work for classes that are loaded via importlib.util.module_from_spec(). Switching to importlib.import_module() seems to work though

tests/error/struct_errors/invalid_generic.err
```diff
@@ -0,0 +1,6 @@
+Guppy compilation failed. Error in file $FILE:14
+
+12:    @guppy.struct(module)
+13:    class MyStruct(Generic[X]):
+                              ^
+GuppyError: Not a parameter
```
aborgna-q, line 6: Could this mention that it's expecting a guppy type var instead of a python one?

mark-koch: That would be nice! It's a bigger change since we would need to hold on to the Python scope when defining a struct, but that's probably a good idea anyways

mark-koch: Ok, I looked into this and it's not super easy to detect. I'll leave this to another PR (see #215)

aborgna-q (approved):

PR #217, opened 2024-05-17
fix: Minor fixes

croyzor:
I tried adding pyright to my environment instead of mypy, and it highlighted these issues!

### DIFF
--- a/guppylang/definition/common.py
+++ b/guppylang/definition/common.py
@@ -49,14 +49,10 @@ class Definition(ABC):
     name: str
     defined_at: ast.AST | None
 
-    @property
-    @abstractmethod
-    def description(self) -> str:
-        """Description of this definition to be used in messages to the user.
-
-        The returned text should fit into messages of the following form: "expected
-        a function, but got {description of this definition} instead".
-        """
+    # Description of this definition to be used in messages to the user.
+    # The returned text should fit into messages of the following form: "expected
+    # a function, but got {description of this definition} instead".
+    description: str
 
 
 class ParsableDef(Definition):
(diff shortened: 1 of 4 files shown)


guppylang/definition/common.py
```diff
@@ -49,14 +49,10 @@ class Definition(ABC):
...
-    def description(self) -> str:
-        """Description of this definition to be used in messages to the user.
-
-        The returned text should fit into messages of the following form: "expected
-        a function, but got {description of this definition} instead".
-        """
+    # Description of this definition to be used in messages to the user.
+    # The returned text should fit into messages of the following form: "expected
+    # a function, but got {description of this definition} instead".
+    description: str
```
croyzor, line 55: N.B. in all of our subclasses of `Definition`, `description` is defined as a field

mark-koch: Does pyright complain about this? In the future we might have definitions where the description depends on the instance (for example if we unify struct and enum definitions into one class), so I would rather leave it more general

croyzor: To be clear: the complaint is that the parent has property called `description`, and the subclasses are shadowing that name with a `str` field. I _think_ to inherit properties the subclasses also need to look like:
```
class A(Definition):
    @property
    def definition() -> str:
        ...
```
but I could easily be wrong
```

mark-koch: I see. The thing is that fields on frozen dataclasses are observably equivalent to get-only properties, so it's the same interface. But it looks like pyright doesn't recognise this...

It's confusing that we use the same pattern for `TypeBase.linear` but it's not complaining there?

PR #219, opened 2024-05-17
fix: Serialisation of float values

### LINKED ISSUE #218: Float value serialisation is broken

mark-koch:
Fixes #218


qartik (approved):
This seems to work with `get_custom_value()`

PR #220, opened 2024-05-17
fix: Mypy tket2 error

mark-koch:
For some reason, mypy started complaining that it can no longer find `Tk2Circuit.to_hugr_json`.

I believe this was introduced in #201 but not flagged because of CI caching?


aborgna-q (approved):
Seems this is missing on the typing definition.
-.-'

It should be ok to add the ignore until we fix that in the next release.

PR #225, opened 2024-05-28
feat: Turn int and float into core types

### LINKED ISSUE #250: Unify numeric type representations

mark-koch:
This PR adds a new `NumericType` to the type hierarchy that now represents `int` and `float`.

Closes #250


guppylang/tys/builtin.py
```diff
@@ -79,6 +86,23 @@ def check_instantiate(
         return NoneType()
 
 
+@dataclass(frozen=True)
+class _NumericTypeDef(TypeDef):
+    """Type definition associated with the builtin `None` type.
```
croyzor, line 91: associated with the `Numeric` type!

guppylang/prelude/builtins.py
```diff
@@ -21,19 +21,28 @@
...
+)
 
 builtins = GuppyModule("builtins", import_builtins=False)
 
 T = guppy.type_var(builtins, "T")
 L = guppy.type_var(builtins, "L", linear=True)
 
 
+# Define the nat type so scripts can import it
+nat = nat_type_def
```
croyzor, line 43: I'd expect this to implement some numeric dunder methods, like `__add__`, `__eq__`, `__gt__`?

guppylang/prelude/_internal.py
```diff
@@ -145,16 +122,12 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
 
         for i in range(len(args)):
             args[i], ty = ExprSynthesizer(self.ctx).synthesize(args[i])
-            if isinstance(ty, OpaqueType) and ty.defn == self.ctx.globals["int"]:
+            if isinstance(ty, NumericType) and ty.kind == NumericType.Kind.Int:
```
croyzor, line 125: Shouldn't `nat` be valid too? Maybe bool?

croyzor: arguably anything that implements `__float__`...

croyzor (commented):
I'm surprised not to see any special handling for `nat` in the type checker, and that it doesn't have any methods implemented by default. Are we ever checking that `nat`s >= 0?

mark-koch:
My plan was to leave all of this to another PR so this one doesn't become too big. But I could also add it here if you'd prefer to look at everything?

guppylang/prelude/_internal.py
mark-koch: Yeah, we need a better way to figure out which types are automatically coreced (also applies to function tuples etc.). These `CustomCallChecker`s are quite hacky...

I'm tempted to just leave it as is until we have a proper solution?

croyzor:
I think the better way to split it up would be NumericType in this PR, then the one that adds the functionality also adds `nat`

guppylang/prelude/_internal.py
mark-koch: > arguably anything that implements __float__...

This only works as long as float is at the top of the numeric hierarchy. For example, if we add `complex` in the future, it might also have a `__float__` method but shouldn't be coreced automtically.

Also we'd need coercsion from `nat` to `int` etc...

croyzor: I think the better interim solution is that we coerce `NumericType` instead of just `int`, then come up with something better down the line

guppylang/prelude/_internal.py
```diff
@@ -273,6 +247,38 @@ def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
...
+    """
+
+    def _prepare_args(self, args: list[ast.expr]) -> list[ast.expr]:
+        # Cast all inputs to int
+        to_int = self.ctx.globals.get_instance_func(bool_type(), "__int__")
+        assert to_int is not None
+        return [to_int.synthesize_call([arg], arg, self.ctx)[0] for arg in args]
+
+    def _get_func(self) -> CallableDef:
+        # Get the int function with the same
```
croyzor, line 264:
```suggestion
        # Get the int function with the same name
```

guppylang/prelude/_internal.py
```diff
@@ -242,6 +247,38 @@ def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
...
+        func = self.ctx.globals.get_instance_func(int_type(), self.func.name)
+        assert func is not None
+        return func
+
+    def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
+        args, _, inst = synthesize_call(self.func.ty, args, self.node, self.ctx)
```
croyzor, line 270: I don't really understand what this call is doing?

mark-koch:
> What's the motivation for this change?

Mainly consistency with Python. Adding `__add__` etc to `bool` is easy to do and Imo we should follow Python as closely as possible unless we have a good reason not to. 

I don't feel super strongly though, so we could reconsider if numeric bools are controversial?

qartik:
If we are considering guppy as a strongly typed language, I think `bool` should not be implicitly treated as a numeric type. We could certainly provide casting/conversion functions for explicit coercion, if they don't exist already.

guppylang/prelude/_internal.py
mark-koch: It checks `args` against the function inputs and potentially infers some type args. The call to `self._get_func().synthesize_call()` below delegates to the corresponding integer function.

mark-koch:
@qartik I moved the bool change to another PR #263 and created issue #264 for discussion.

@croyzor Could you have another at this PR? Sorry for moving things around so much

guppylang/tys/builtin.py
```diff
@@ -145,6 +175,10 @@ def bool_type() -> OpaqueType:
     return OpaqueType([], bool_type_def)
 
 
+def int_type() -> NumericType:
+    return NumericType(NumericType.Kind.Int)
```
croyzor, line 179: Should this not contain some reference to `int_type_def`?

croyzor: Is this used? I see a couple of places it'd work in this PR but it's just being inlined

guppylang/prelude/_internal.py
```diff
@@ -116,16 +93,16 @@ def logic_op(op_name: str, args: list[tys.TypeArg] | None = None) -> ops.OpType:
...
+    if args is None:
+        args = num_params * [tys.TypeArg(tys.BoundedNatArg(n=NumericType.INT_WIDTH))]
     return ops.OpType(
-        ops.CustomOp(
-            extension=ext,
-            op_name=op_name,
-            args=num_params * [tys.TypeArg(tys.BoundedNatArg(n=INT_WIDTH))],
-            parent=UNDEFINED,
-        )
+        ops.CustomOp(extension=ext, op_name=op_name, args=args, parent=UNDEFINED)
```
croyzor, line 105: Nice 👍

croyzor (approved):

guppylang/tys/builtin.py
mark-koch: > Should this not contain some reference to int_type_def?

No, only the `int_type_def` has a reference to a `NumericType`. The inverse mapping from type to type def is implemeted here:
https://github.com/CQCL/guppylang/blob/cf8a5299e7b317de1f30f2c84c1b665bfba8c904/guppylang/checker/core.py#L82-L112

> Is this used? I see a couple of places it'd work in this PR but it's just being inlined

Good point, it's indeed unused. I'll remove it 👍

PR #247, opened 2024-06-18
feat: Return already-compiled hugrs from `GuppyModule.compile`

### LINKED ISSUE #246: Keep the Hugr around after compiling a module
Right now calling `GuppyModule.compile` throws an error if the module has already been compiled.

In some cases we may want to use a module multiple times, for example when generating multiple circuits from it.
Currently the work around is to manually keep the HUGR around, but it would be more straightforward if we can keep a reference to compiled hugr in the module.

```python
@guppy(module)
def my_func():
  pass

@guppy(module)
def my_other_func():
  pass

make_circuit(module, "my_func")
make_circuit(module, "my_other_func") # This cannot access the hugr anymore
```

aborgna-q:
Closes #246 

I'm not sure where to add a test for this, as it's not integration nor error testing.

### DIFF
--- a/guppylang/module.py
+++ b/guppylang/module.py
@@ -36,6 +36,10 @@ class GuppyModule:
     # Whether the module has already been compiled
     _compiled: bool
 
+    # If the hugr has already been compiled, keeps a reference that can be returned
+    # from `compile`.
+    _compiled_hugr: Hugr | None
+
     # Map of raw definitions in this module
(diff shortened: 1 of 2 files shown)


mark-koch:
> I'm not sure where to add a test for this, as it's not integration nor error testing.

Mhm it's probably fine to just add it to the `integration` module, maybe in a new file or in `test_basic.py`?

mark-koch (approved):

PR #254, opened 2024-06-19
feat: Add nat type

### LINKED ISSUE #228: Add a type for unsigned integers
Will be used to parameterise the size of arrays.

Should be implemented on top of the changes in https://github.com/CQCL/guppylang/pull/225

mark-koch:
Closes #228


mark-koch:
> I don't understand what how the width of nat and int are handled.

Guppy currently assumes that ints are 64 bit across the board, so `int_op` automatically inserts a log width of 6.

> I don't understand what is_to_u and iu_to_s are expected to do. If widths aren't changing aren't they the identity?

I guess they could be, but I'd prefer to error when trying to convert a negative int to nat, or when trying to convert a nat that is too large to fit into an int64.

PR #258, opened 2024-06-24
feat: Add array type

mark-koch:


guppylang/prelude/builtins.py
```diff
@@ -722,6 +725,18 @@ def reverse(self: linst[L]) -> linst[L]: ...
     def sort(self: list[T]) -> None: ...
 
 
+n = guppy.nat_var(builtins, "n")
+
+
+@guppy.extend_type(builtins, array_type_def)
+class Array:
+    @guppy.hugr_op(builtins, DummyOp("ArrayGet"))
+    def __getitem__(self: array[T, n], idx: int) -> T: ...
```
mark-koch, line 734: The type varable `T` is declared as non-linear, so it cannot be instantiated with `qubit` (see error tests).

We first need some changes to the type checker in order to fully support linear arrays

guppylang/prelude/builtins.py
```diff
@@ -57,11 +59,12 @@ class nat:
...
-class array:
-    """Class to import in order to use arrays."""
+_S = TypeVar("_S")
+_T = TypeVar("_T")
+
 
-    def __class_getitem__(cls, item):
-        return cls
+class array(Generic[_S, _T]):
+    """Class to import in order to use arrays."""
```
aborgna-q, line 67: The names are switched from the `Array` definition below (which uses `arrar[T, n]`).
Can you document which typevar is the array type and which one the length?

aborgna-q (approved):
The linear array errors look quite confusing, but I guess they are OK as temporary measure.

guppylang/prelude/builtins.py
mark-koch: This class is just for the purpose of mypy and ruff to stop them from complaining, so it really doesn't matter what type vars we use here. But you're right, I'll choose some more suggestive names 👍

PR #260, opened 2024-06-24
fix: Loading custom polymorphic function defs as values

### LINKED ISSUE #259: `CustomFunctionDef.load_with_args` is broken when type args are nonempty

mark-koch:
Fixes #259


guppylang/definition/custom.py
```diff
@@ -145,27 +145,20 @@ def load_with_args(
                 node,
             )
         assert len(self.ty.params) == len(type_args)
-
-        # Find the module node by walking up the hierarchy
-        module: Node = dfg.node
```
mark-koch, line 150: We can put the FunctionDef anywhere, no need to walk up to the module

PR #262, opened 2024-06-24
feat: Generate constructor methods for structs

### LINKED ISSUE #261: Auto-generate struct constructor methods

mark-koch:
Closes #261

### DIFF
--- a/guppylang/definition/custom.py
+++ b/guppylang/definition/custom.py
@@ -12,7 +12,7 @@ from guppylang.compiler.core import CompiledGlobals, DFContainer
 from guppylang.definition.common import ParsableDef
 from guppylang.definition.value import CompiledCallableDef
 from guppylang.error import GuppyError, InternalGuppyError
-from guppylang.hugr_builder.hugr import Hugr, Node, OutPortV
+from guppylang.hugr_builder.hugr import Hugr, OutPortV
 from guppylang.nodes import GlobalCall
 from guppylang.tys.subst import Inst, Subst
 from guppylang.tys.ty import FunctionType, NoneType, Type, type_to_row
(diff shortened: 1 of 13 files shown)


guppylang/module.py
```diff
@@ -180,9 +181,18 @@ def compile(self) -> Hugr:
         )
         self._globals = self._globals.update_defs(type_defs)
 
+        # Collect auto-generated methods
+        generated: dict[DefId, RawDef] = {}
+        for defn in type_defs.values():
+            if isinstance(defn, CheckedStructDef):
+                for method_def in defn.generated_methods:
+                    generated[method_def.id] = method_def
+                    self._globals.impls.setdefault(defn.id, {})
```
croyzor, line 190: I think this wants to happen in the outer loop, so that the impls aren't set to `{}` each time a method is added

guppylang/definition/struct.py
```diff
@@ -186,6 +193,33 @@ def check_instantiate(
...
+
+        class ConstructorCompiler(CustomCallCompiler):
+            """Compiler for the `__new__` constructor method of a struct."""
+
+            def compile(self, args: list[OutPortV]) -> list[OutPortV]:
+                return [self.graph.add_make_tuple(args).out_port(0)]
+
+        constructor_sig = FunctionType(
+            inputs=[f.ty for f in self.fields],
+            output=StructType([p.to_bound(i) for i, p in enumerate(self.params)], self),
```
croyzor, line 208: How does this work, I assume the list is the `args` from `ParametrizedTypeBase`, then `self` is the `defn` from `StructType`. So in the default constructors for python dataclasses, you provide the fields for all of the parents, outer to inner?

guppylang/definition/custom.py
```diff
@@ -97,10 +97,16 @@ def compile_call(
...
 @dataclass(frozen=True)
-class CustomFunctionDef(RawCustomFunctionDef, CompiledCallableDef):
+class CustomFunctionDef(CompiledCallableDef):
     """A custom function with parsed and checked signature."""
 
+    defined_at: AstNode
+    call_checker: "CustomCallChecker"
+    call_compiler: "CustomCallCompiler"
     ty: FunctionType
+    higher_order_value: bool
```
croyzor, line 107: When is this not true?

guppylang/module.py
mark-koch: `setdefault` only sets if the key doesn't exist yet

https://docs.python.org/3/library/stdtypes.html#dict.setdefault

guppylang/definition/struct.py
mark-koch: Yes, exactly. I could also use keyword args if it would be clearer?

guppylang/definition/custom.py
mark-koch: When the type of the function cannot be expressed in Guppy's type system.

For example, the `len(x)` function is a `CustomFunction` that checks if `x` implements `__len__`. We can't write down a type for `len`, so we can't use it as a value

guppylang/module.py
croyzor: Okay, I think this still wants to be in the outer loop, but is ultimately harmless then

croyzor (approved):

PR #266, opened 2024-06-25
fix: Add missing test file

mark-koch:
This file should have been added with #258

### DIFF
--- a/tests/error/test_array_errors.py
+++ b/tests/error/test_array_errors.py
@@ -1,0 +1,20 @@
+import pathlib
+import pytest
+
+from tests.error.util import run_error_test
+
+path = pathlib.Path(__file__).parent.resolve() / "array_errors"
+files = [
+    x
+    for x in path.iterdir()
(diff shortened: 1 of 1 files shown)


aborgna-q (approved):
👍

tests/error/test_array_errors.py
```diff
@@ -0,0 +1,20 @@
...
+import pytest
+
+from tests.error.util import run_error_test
+
+path = pathlib.Path(__file__).parent.resolve() / "array_errors"
+files = [
+    x
+    for x in path.iterdir()
+    if x.is_file()
+    if x.suffix == ".py" and x.name != "__init__.py"
```
aborgna-q, line 11: This is consistent with all the other error tests, but shouldn't the double if be an `and` chain instead?

mark-koch: Yes, that would be better. I'll make a PR to change it across the board

PR #267, opened 2024-06-25
ci: Create release-please PRs as drafts

aborgna-q:
To avoid spamming the release team with review requests.


mark-koch (approved):
Thanks 👍

PR #273, opened 2024-06-27
refactor: Turn linearity checking into separate compiler stage

### LINKED ISSUE #272: Turn linearity check into separate compiler stage

mark-koch:
Closes #272

### DIFF
--- a/guppylang/cfg/builder.py
+++ b/guppylang/cfg/builder.py
@@ -1,4 +1,5 @@
 import ast
+import copy
 import itertools
 from collections.abc import Iterator
 from typing import NamedTuple
@@ -270,7 +271,9 @@ class ExprBuilder(ast.NodeTransformer):
         # assignment statement and replace the expression with `x`.
         if not isinstance(node.target, ast.Name):
             raise InternalGuppyError(f"Unexpected assign target: {node.target}")
-        assign = ast.Assign(targets=[node.target], value=self.visit(node.value))
+        assign = ast.Assign(
+            targets=[copy.deepcopy(node.target)], value=self.visit(node.value)
+        )
         set_location_from(assign, node)
         self.bb.statements.append(assign)
         return node.target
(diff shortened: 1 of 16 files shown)


tests/error/linear_errors/break_unused.py
```diff
@@ -26,7 +26,7 @@ def foo(i: int) -> bool:
         if i == 0:
             break
         i -= 1
-        b ^= measure(q)
+        b &= measure(q)
```
mark-koch, line 29: for some reason `bool.__xor__` is missing (see #274).

The test worked before since the linearity violation triggered an error before we got to this point. Now we type check everything first, so this has to be fixed (similar for the remaining modified tests)

guppylang/cfg/builder.py
```diff
@@ -270,7 +271,9 @@ def visit_NamedExpr(self, node: ast.NamedExpr) -> ast.Name:
         # assignment statement and replace the expression with `x`.
         if not isinstance(node.target, ast.Name):
             raise InternalGuppyError(f"Unexpected assign target: {node.target}")
-        assign = ast.Assign(targets=[node.target], value=self.visit(node.value))
+        assign = ast.Assign(
+            targets=[copy.deepcopy(node.target)], value=self.visit(node.value)
+        )
```
aborgna-q, line 276: Is this a drive-by?

aborgna-q (approved):
Nice and tidy.

The comments on `_check_comprehension` are quite useful.

guppylang/cfg/builder.py
mark-koch: No this is required now that we annotate the LHS of assignments with types

https://github.com/CQCL/guppylang/blob/f0dec41f8e722cbbbf2a9f98a7980a85f67c9a56/guppylang/checker/stmt_checker.py#L49-L56

If we don't make a copy, the type checker will see that a type has been assigned when the variable is used the next time and won't bother checking again which leads incorrect behaviour

PR #283, opened 2024-07-04
tests: Add LLVM-based execution tests

croyzor:


croyzor:
> I would prefer for either:
> 
>     * add `execute_llvm` into the existing crate, perhaps renaming it. Add a feature for llvm execution.
> 
>     * Add a root Cargo.toml to put both crates into a workspace.

I agree, but AFAICT, workspaces don't work well for maturin projects, and poetry doesn't allow extras/groups to have subtle enough effects, such as setting a feature flag on a rust package.

My thinking is that the structure either looks like this, or I can roll `execute-llvm` into `validator` and install it by default
 
> I see you have omitted quantum operations. I suppose I would be OK merging this without them, but I do think they are important. They are quite easy with JIT: pull in qir-runner to your shared library and make the qir symbols available to the llvm ir. Copy the quantum extension from `guppy.rs` in hugr-llvm, implement a few ops(Measure,H,CX ?) by lowering to the appropriate qir symbol. Arrange for the qir initialise function to be called before the Hugr main. This can be done by emitting a "main" function that calls the Hugr main(after said initialisation). 💥. If you decide not to do this, please do create a followup issue.

I'll add a follow up issue

PR #290, opened 2024-07-08
feat: Linearity checking for places

mark-koch:
See #295 for context, the tests are at #293

The basic idea is that we implicitly "unfold" structs into tuples and then rely on the existing linearity checking logic. The example from #295 is checked as follows:

```python
@guppy.struct
class MyStruct:
   q1: qubit
   q2: qubit
   x: int

def main(s: MyStruct):
   s_q1, s_q2, s_x = s
   q = h(s_q1)
   t = h(s_q2)
   y = s_x + s_x
   use((s_q1, s_q2, s_x))  # Error
   ...
```

This is implemented using the `leaf_places()` iterator that gives us the leaf projections of places with (nested) struct types.

### DIFF
--- a/guppylang/cfg/analysis.py
+++ b/guppylang/cfg/analysis.py
@@ -2,7 +2,7 @@ from abc import ABC, abstractmethod
 from collections.abc import Iterable
 from typing import Generic, TypeVar
 
-from guppylang.cfg.bb import BB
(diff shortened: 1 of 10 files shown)


qartik:
LGTM.

qartik (approved):

PR #291, opened 2024-07-08
feat: Compile places and struct field accesses to Hugr

mark-koch:
See #295 for context, the tests are at #293

The idea is to implement a similar logic to one used for linearity checking in #290. As soon as we assign a variable with struct type, we immediately unpack the struct to make the individual fields available. We only repack once the whole struct value is requested.


guppylang/compiler/core.py
```diff
@@ -45,28 +21,49 @@ class DFContainer:
...
+    def __setitem__(self, place: Place, port: OutPortV) -> None:
+        # When assigning a struct value, we immediately unpack it recursively and only
+        # store the leaf wires.
+        is_return = isinstance(place, Variable) and is_return_var(place.name)
+        if isinstance(place.ty, StructType) and not is_return:
+            unpack = self.graph.add_unpack_tuple(port, self.node)
+            for field, field_port in zip(
+                place.ty.fields, unpack.out_ports, strict=False
+            ):
+                self[FieldAccess(place, field, None)] = field_port
```
aborgna-q, line 59: I think we should delete `self.locals[place.id]` here?
Otherwise reading the whole struct would still use the old value.

guppylang/compiler/core.py
```diff
@@ -45,28 +21,49 @@ class DFContainer:
...
+        Note that this mutates the Hugr since we might need to pack or unpack some
+        tuples to obtain a port for places that involve struct fields.
+        """
+        # First check, if we already have a wire for this place
+        if place.id in self.locals:
+            return self.locals[place.id]
+        # Otherwise, our only hope is that it's a struct value that we can rebuild by
+        # packing the wires of its constituting fields
+        if not isinstance(place.ty, StructType):
+            raise InternalGuppyError(f"Couldn't obtain a port for `{place}`")
```
aborgna-q, line 40: What about the inverse? Could we auto-unpack structs when requesting a `FieldAccess`?

aborgna-q (approved):
Nice, really ergonomic 👍

guppylang/compiler/core.py
mark-koch: Good catch! Here is a minimal example that currently fails validation because of this:

```python
@guppy.struct(module)
class MyStruct:
    x: int


@guppy(module)
def foo() -> MyStruct:
    s = 0
    s = MyStruct(42)
    return s  # Wires up the 0 value since `s = 0` is still in scope
```

I'll fix it and add a test

mark-koch: See b7bc393 for the test

guppylang/compiler/core.py
mark-koch: Yes, that would be an alternative way of doing it: Instead of unpacking on assignments, we could unpack when requesting a field.

I actually implemented that version first, but the logic is more complicated. You need to be smart about looking upwards/downwards in the place hierarchy in the right order, to avoid doing an exponential search for all the possible ways to unpack/repack a value. Also, when you assign to a field you need to invalidate and unpack all parent places.

I think the benefit of the approach you suggested is that it leads to fewer unnessecary `MakeTuple`/`UnpackTuple` pairs.

But I wanted to do the most basic thing possible for this PR. Though I still have the other version around if you want to have a look if you prefer it?

aborgna-q: No need to block this now. Nicer hugrs is not a priority for guppy (though it would simplify tket2's logic :) ).

aborgna-q (approved):

PR #293, opened 2024-07-08
tests: Add various struct linearity and field access tests

mark-koch:
See #295 for context


tests/error/comprehension_errors/guarded3.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:24
+
+22:    @guppy(module)
+23:    def foo(qs: linst[MyStruct]) -> linst[qubit]:
+24:        return [s.q2 for s in qs if bar(s.q1)]
+                            ^
+GuppyTypeError: Field `s.q2` with linear type `qubit` is not used on all control-flow paths of the list comprehension
```
croyzor, line 7: I think talking about "control-flow paths" of a list comprehension could be a bit misleading, not sure what the better wording is. I'm thinking `Field ... with linear type ... may not be used by this comprehension` but that's not great either

mark-koch: Ideally, you'd want to say that `s.q2` is not used if `bar(s.q1)` is false.

Our dataflow analysis can already exactly identify which branch leads to a non-use, so we are actually not far off from having this better error message. However, we need to extend our error framework so that it can point to this location in the message.

I think something like this would be nice:

```
23:    def foo(qs: linst[MyStruct]) -> linst[qubit]:
24:        return [s.q2 for s in qs if bar(s.q1)]
                            ^ Field `s.q2` with linear type `qubit` is not always used

24:        return [s.q2 for s in qs if bar(s.q1)]
                                       --------- Field is not used if this expression is `False`
```

tests/error/linear_errors/branch_use_field2.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:21
+
+19:    if b:
+20:        measure(s.q2)
+21:    return s
+              ^
+GuppyError: Field `s.q2` with linear type `qubit` was already used (at 20:16)
```
croyzor, line 7: Maybe "already used in some control flow paths"?

tests/error/linear_errors/field_copy2.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:17
+
+15:    @guppy(module)
+16:    def foo(s: MyStruct) -> tuple[MyStruct, qubit]:
+17:        return s, s.q
+                     ^^^
+GuppyError: Field `s.q` with linear type `qubit` was already used (at 17:11)
```
croyzor, line 7: Can we record the use site of linear variables and point to where they were used when we give these error messages? I guess this requires upgrading the error framework to handle multiple locations

tests/error/linear_errors/branch_use_field1.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:18
+
+16:    @guppy(module)
+17:    def foo(b: bool) -> bool:
+18:        s = MyStruct(qubit())
+           ^
+GuppyError: Field `s.q` with linear type `qubit` is not used on all control-flow paths
```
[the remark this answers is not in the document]
mark-koch: Mhm but 20 is an actual use? I think it makes more sense to point out that a variable is not used at the place where it is defined?

tests/error/linear_errors/branch_use_field2.err
mark-koch: Agreed, this would be better. However, we need to do some extra checks to determine if it's in "some" or "all" control-flow paths.

I'd prefer leaving this to another PR since it's not directly related to structs (the same error shows up if you use a regular variable instead of a struct access). See #306

tests/error/linear_errors/field_copy2.err
mark-koch: We're already tracking the point where the variable has been used (that's the `(at 17:11)` part in the message). But it would be better tyo show this with a separate span 👍

Added to #306

tests/error/comprehension_errors/guarded3.err
mark-koch: Added to issue #306

tests/error/linear_errors/if_both_unused_field.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:19
+
+17:    def foo(b: bool) -> int:
+18:        if b:
+19:            s = MyStruct(qubit())
+               ^
+GuppyError: Field `s.q` with linear type `qubit` is not used on all control-flow paths
```
croyzor, line 7: This is interesting. Instead of a linear variable in an outer context not being used in some control flow paths, there are variables created *within* certain control flow paths that are never used. I have no idea what the error should say

tests/error/linear_errors/if_both_unused_reassign_field.py
```diff
@@ -0,0 +1,26 @@
...
+module.load(quantum)
+
+
+@guppy.struct(module)
+class MyStruct:
+    q: qubit
+
+
+@guppy(module)
+def foo(b: bool, s: MyStruct) -> MyStruct:
```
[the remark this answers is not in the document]
mark-koch: Yes, once we can point to multiple error locations, we could do

```
Guppy compilation failed. Error in file $FILE:17

15:    @guppy(module)
16:    def foo(s: MyStruct) -> MyStruct:
               ^^^^^^^^^^^ Field `s.q` with linear type `qubit` is not used

19:        s.q = qubit()
           --- `s.q` reassigned without use of the previous value
```

Added to #306

tests/error/linear_errors/unused_field1.err
```diff
@@ -0,0 +1,6 @@
+Guppy compilation failed. Error in file $FILE:18
+
+16:    @guppy(module)
+17:    def foo(s: MyStruct) -> int:
+               ^^^^^^^^^^^
+GuppyError: Field `s.q` with linear type `qubit` is not used on all control-flow paths
```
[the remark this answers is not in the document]
mark-koch: The above errors where found while checking a single BB. This one is found via global dataflow analysis since there is an unconditional branch to the return BB.

We currently always add the `"on all control-flow paths"` suffix if errrors are found using the latter method, even if the branches are unconditional. I think this wouldn't be too hard to fix, but again adding it to #306 since it's not directly related to structs

tests/error/struct_errors/assign_call.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:20
+
+18:    @guppy(module)
+19:    def bar() -> None:
+20:        foo().x += 1
+           ^^^^^
+GuppyError: Assigning to this expression is not supported yet
```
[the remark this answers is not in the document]
mark-koch: Hmm I think once we have implemented proper reference semantics we should support this as long as there is no linearity violation. Why would it be controversial?

tests/error/struct_errors/invalid_attribute_access.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:15
+
+13:    @guppy(module)
+14:    def foo(s: MyStruct) -> None:
+15:        s.z
+           ^^^
+GuppyTypeError: Expression of type `MyStruct` has no attribute `z`
```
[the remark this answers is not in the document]
mark-koch: Yes 👍

https://docs.python.org/3/tutorial/classes.html#instance-objects

tests/error/linear_errors/if_both_unused_field.err
croyzor: Again, probably not for this PR

croyzor (approved):

tests/error/linear_errors/if_both_unused_field.err
mark-koch: Yeah this is the same problem as in https://github.com/CQCL/guppylang/pull/293#discussion_r1686164876

PR #289, opened 2024-07-08
feat: Add `Place` union and lower AST to it during checking

mark-koch:
See #295 for context.

The tests are at #293

PR #295, opened 2024-07-08
feat: Allow access to struct fields and mutation of linear ones

### LINKED ISSUE #156: User-defined types: Tracking issue
- [x] #155
- [x] #175 
- [x] #159 
- [x] #161
- [x] #261
- [x] #272
- [x] #276
- [x] #280

### LINKED ISSUE #276: Track linearity of individual struct fields

### LINKED ISSUE #280: Lower struct field access to Hugr

mark-koch:
This is the feature branch for enabling struct field access and mutation of linear fields.

Closes #276, closes #280, and closes #156. 

The difficult bit is that we want to track linearity of individual struct fields separately (akin to Rust's partial moves):

```python
@guppy.struct
class MyStruct:
   q1: qubit
   q2: qubit
   x: int

def main(s: MyStruct):
   q = h(s.q1)
   t = h(s.q2)    # Using s.q2 is fine, only s.q1 has been used
   y = s.x + s.x  # Classical fields can be used multiple times
   use(s)         # Error: Linear fields of s have already been used
   ...
```

This is the plan:

* We introduce a new notion called `Place`: A place is a description for a storage location of a local value that users can refer to in their program. Roughly, these are values that can be lowered to a static wire within the Hugr and are tracked separately when checking linearity.
* For the purposes of this PR, a place is either a local variable or a field projection of another place that holds a struct. I.e. places are paths `a.b.c.d` of zero or more nested struct accesses. In the future, indexing into an array etc will also become a place.
* During type checking, we figure out which AST nodes correspond to places and annotate them as such
* For linearity checking, we run a liveness analysis pass that tracks usage and assignment of places across the CFG. This way, linearity of different struct fields is tracked individually.
* When compiling to Hugr, we keep track of a mapping from places to wires/ports

Tracked PRs:

* #288: Precursor PR to generalise our program analysis framework to run on places in the future.
* #289: Adds the `Place` type and implements the type checking logic to turn `ast.Name` and `ast.Attribute` nodes into places.
* #290: Update linearity checker to operate on places instead of variables
* #291: Lower places to Hugr
* #292: Some missing pieces to handle nested functions correctlt
* #293


qartik:
Where does the notion of `place` come from?

mark-koch:
I copied the terminolgy from Rust: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/mir/struct.Place.html

croyzor (approved):

Issue #282, opened 2024-07-03
Tracking Issue: Inout values


mark-koch:
Development in PR #311

- [x] #313
- [x] #309
- [x] #312
- [x] #347
- [x] #314
- [x] #317
- [x] #319
- [x] #330
- [x] #332
- [x] #333
- [x] #338

Follow ups:
- #318
- #378
- #379
- #380

mark-koch:
Closed by #311

PR #311, opened 2024-07-24
feat: Inout arguments

mark-koch:
This is the dev branch for the inout argument feature (tracked by #282).

The idea is to allow explicit `@inout` annotations on function arguments that "give back" the passed value after the function returns:

```python
@guppy
def foo(q: qubit @inout) -> None: ...

@guppy
def bar(q1: qubit @inout, q2: qubit @inout) -> bool: ...

@guppy
def main() -> None:
   q1, q2 = qubit(), qubit()
   foo(q1)          # Desugars to `q1 = foo(q1)`
   x = bar(q1, q2)  # Desugars to `q1, q2, x = bar(q1, q2)`
   y = bar(q1, q1)  # Error: Linearity violation, q1 used twice
```

To enable this, we need to enforce that `@inout` arguments are not moved in the body of the function (apart from passing them in another `@inout` position). This means that the argument will always be bound to the same name and never aliased which allows us to desugar `@inout` functions like

```python
@guppy
def bar(q1: qubit, q2: qubit) -> bool:
   [body]
   return [expr]
```

into

```python
@guppy
def bar(q1: qubit, q2: qubit) -> tuple[qubit, qubit, bool]:
   [body]
   return q1, q2, [expr]  # Linearity checker needs to ensure that q1, q2 are unused
```

Note that we only allow `@inout` annotations on linear types, since they would be useless for classical ones (unless we also implement an ownership system for classical values). Supporting them would make the checking logic more complicated without providing any meaningful benefit.

Tracked PRs:
* #315
* #316
* #349
* #344
* #321
* #331
* #350
* #339 
* #340
* #351


mark-koch:
To fix this, we need to
* Only check against local variables here: https://github.com/CQCL/guppylang/blob/a7cc97a6c1af674972c3a37e28ef7e33ddf6fa3d/guppylang/checker/linearity_checker.py#L182
* Update Hugr generation to use the fine-grained place map generated by the linearity checker instead of the variable map from type checking

ss2165 (approved):
🔥 🔥 🔥 🔥 

big fan of the diligent tracking

PR #316, opened 2024-07-24
feat: Parse inout annotations in function signatures

### LINKED ISSUE #309: Parse inout annotation in function signatures

mark-koch:
Closes #309. See #282 for context.


tests/error/misc_errors/return_flag_callable.err
```diff
@@ -0,0 +1,6 @@
+Guppy compilation failed. Error in file $FILE:14
+
+12:    @guppy.declare(module)
+13:    def foo(f: Callable[[], qubit @inout]) -> None: ...
+                  ^^^^^^^^^^^^^^^^^^^^^^^^^^
+GuppyError: `@` type annotations are not allowed in this position
```
mark-koch, line 6: The broad error location is not ideal, but improving this requires a bigger refactor

guppylang/tys/parsing.py
```diff
@@ -13,14 +13,14 @@
...
-from guppylang.tys.ty import NoneType, NumericType, TupleType, Type
+from guppylang.tys.ty import InputFlags, NoneType, NumericType, TupleType, Type
 
 
 def arg_from_ast(
     node: AstNode,
     globals: Globals,
     param_var_mapping: dict[str, Parameter] | None = None,
-) -> Argument:
+) -> tuple[Argument, InputFlags]:
```
acl-cqc, line 23: arg! (argh!) Sorry but I got getting really confused here; what is an `Argument`? Is it a value Hugr static TypeArg? (It certainly looks like one) - but then we wouldn't want it to be `inout` or have InputFlags, would we?

I think `arg_from_ast` parses both TypeArgs (Hugr TypeArgs of *kind* TypeParam::Type), and ConstArgs (Hugr TypeArgs of other kinds). But, hang on, ConstArgs are not function parameters (that'd be like `def foo(x: 6)` !). And `@inout` is *only* appropriate for the top/outermost level argument to a `def`, or to `Callable`, yet we make life difficult for *all the other* `check_instantiate`s in order to fit in `Callable` (e.g. `list[int @inout]`) - I see 5 calls to check_no_flags, plus a couple that have no args to have any flags.
* I think passing the flag by subclassing could be an easier route, you *might* still have an equivalent of `check_no_flags` but I reckon it might just fall out of normal `check_all_args` or checking-against-TypeParam (that is: make a normal Hugr TypeParam::Type unable to accept an InoutTypeArg, but Callable has a special TypeParam::MaybeInOutType that can accept both InoutTypeArg and normal TypeArg).
* Alternatively....this'll be even more controversial....can you special-case Callable?

Generally I think I'd favour breaking the lengthy `arg_from_ast` up into `flagged_arg_from_ast` or (`maybe_`)`inout_arg_from_ast` that looks for BinOp with `@` and then (usually) falls back to a shorter `arg_from_ast` that does *not* support that case, and thus, returns a TypeArg (`| ConstArg`) only. (Or returns Argument without separate flags, that'd be a first/lesser step from what's in the PR now).

Which is to say that (my gut feeling is) you'd do better to push error checking downwards towards the parsing/leaves, rather than parsing whatever and then raising error messages (like, found function parameter of type 6). Not sure if that would help with the error locations. Then, say, `maybe_inout_arg_from_ast` would (if not BinOp) call `type_arg_from_ast` (that returns TypeArg, and does not support ConstArg), and then there is `const_or_type_arg_from_ast` (used for arguments to parameterized types), that kind of thing.

Note you've already gone this route with `type_from_arg` vs `type_with_flags_from_arg`, so it's not thaaaat radical....

guppylang/definition/ty.py
```diff
@@ -1,20 +1,24 @@
...
 from guppylang.tys.arg import Argument
 from guppylang.tys.param import Parameter, check_all_args
-from guppylang.tys.ty import OpaqueType, Type
+from guppylang.tys.ty import InputFlags, OpaqueType, Type
 
 if TYPE_CHECKING:
     from guppylang.checker.core import Globals
 
 
+FlaggedArgs: TypeAlias = Sequence[tuple[Argument, InputFlags]]
```
acl-cqc, line 19: Most uses have gone from `Sequence[Argument]` to `FlaggedArgs`, but you do have a couple of `tuple[Argument, InputFlags]` elsewhere. Maybe define `FlaggedArg = tuple[...]` and use that (so `Sequence[FlaggedArg]` in many places)....albeit, elsewhere I suggest a bigger redifinition of FlaggedArg ;)

guppylang/prelude/builtins.py
```diff
@@ -55,6 +55,16 @@ def py(*_args: Any) -> Any:
     raise GuppyError("`py` can only by used in a Guppy context")
 
 
+class _Inout:
+    """Dummy class to support `@inout` annotations."""
```
acl-cqc, line 59: This is to allow `def foo(x: int @inout)`, since Python (in its infinite wisdom *cough* *splutter* *boo*) disallows `def foo(x: @inout int)`. Obviously, it's fine to do *matrix multiplication* inside a type annotation but not invoke a decorator....

However, have you thought about the "pythonic" alternative, which I believe would be `def foo(x: Annotated[int, "inout"]`?`@inout` is a bastardization of python syntax anyway so what would users think? I admit `Annotated` is rather long - what about `def foo(x: Inout[int])` ? (No way, I hear you say - and I see the downside - maybe that lengthy Annotated isn't sooo bad, then....)

Not seriously recommending you change, but how different / how much simpler might that make things, would it impact the design at all?

guppylang/checker/func_checker.py
```diff
@@ -148,8 +148,13 @@ def check_signature(func_def: ast.FunctionDef, globals: Globals) -> FunctionType
     for inp in func_def.args.args:
         if inp.annotation is None:
             raise GuppyError("Argument type must be annotated", inp)
-        ty = type_from_ast(inp.annotation, globals, param_var_mapping)
-        inputs.append(FuncInput(ty, InputFlags.NoFlags))
+        ty, flags = type_with_flags_from_ast(inp.annotation, globals, param_var_mapping)
```
acl-cqc, line 151: This does seem a bit like, we have two places that define the set(s) of flags legal in the signature of a function - `_CallableTypeDef` also makes similar checks. If you parsed `type_with_flags_from_ast` for the return type too, you could call some single checking function (maybe it takes a `list[FlaggedArg]` and a single `FlaggedArg` and then `_CallableTypeDef` first separates out the last element of the list) that issues appropriate errors and returns some appropriate e.g. `(list[TypeArg],TypeArg)`

guppylang/tys/parsing.py
acl-cqc: Of course we should also think about likely future annotations. `list[MyStruct @frozen]` is plausible, true....

acl-cqc (commented):
Ok, so I've written some quite long comments suggesting a couple of big refactors. I *think* commoning up `check_signature` with `_CallableTypeDef.check_instantiate` should be done, unless it's much harder than it looks, and (if you do nothing else) then I'd do `FlaggedArgs` -> `FlaggedArg`.

But I'd also think seriously about breaking up `arg_from_ast`, and switching from `tuple[type, flags]` to a broader union (`TypeArg | ConstArg | InoutTypeArg`, say - maybe there are benefits in the approach taken by python's `Annotated`), even if you don't go that way.

guppylang/tys/parsing.py
```diff
@@ -122,15 +133,27 @@ def arg_from_ast(
 _type_param = TypeParam(0, "T", True)
 
 
+def type_with_flags_from_ast(
+    node: AstNode,
+    globals: Globals,
+    param_var_mapping: dict[str, Parameter] | None = None,
+) -> tuple[Type, InputFlags]:
+    """Turns an AST expression into a Guppy type possibly annotated with @flags."""
+    # Parse an argument and check that it's valid for a `TypeParam`
```
acl-cqc, line 142: Yes, it's amusing to think that we could have got this far with `x: 6 @inout` ;-)

guppylang/tys/parsing.py
mark-koch: > arg! (argh!) Sorry but I got getting really confused here; what is an Argument? Is it a value Hugr static TypeArg? 

Sorry for the confusion, yes `Argument` is Guppy's version of static arguments to types. E.g. the type `array[int, 5]` has a `TypeArg` `int` and a `ConstArg` `5`.

> More comments would really help, maybe also think about renaming Argument->StaticArg, say.

I added #358 to consider the rename

> And `@inout` is only appropriate for the top/outermost level argument to a def, or to Callable, yet we make life difficult for all the other check_instantiates in order to fit in Callable (e.g. `list[int @inout]`) [...] Alternatively....this'll be even more controversial....can you special-case Callable?

After reflecting a bit on this I think that special casing `Callable` is probaby the best solution in the near term. I initially wanted a general solution that can handle future flags like `@frozen` etc., but I agree it adds a lot of complexity and probably isn't worth it for now. I like your subclassing idea though, I'll probably go for that of we want more general flags in the future 👍

mark-koch: > Special-casing Callable might be fine in the short-term, tho, and maybe even in the longer term if it turns out we don't need similar specials for any of the others...

Agreed, I think I'll do that for now 👍

guppylang/definition/ty.py
mark-koch: This no longer applied after ff545a7

guppylang/prelude/builtins.py
mark-koch: I created  #359 for discussion.

> Not seriously recommending you change, but how different / how much simpler might that make things, would it impact the design at all?

I think this is mostly a syntax concern. The design wouldn't be affected much imo

guppylang/tys/parsing.py
mark-koch: Another benefit of special casing `Callable` is that we can now properly handle the list syntax `Callable[[Arg1, Arg2], Return]` without the flattening hack

guppylang/tys/builtin.py
```diff
@@ -35,28 +33,10 @@ class _CallableTypeDef(TypeDef):
...
-                raise GuppyError(
-                    f"Non-linear type `{ty}` cannot be annotated as `@inout`",
-                    loc,  # TODO: Better error location
-                )
-        if output_flags != InputFlags.NoFlags:
-            raise GuppyError(
-                "`@` type annotations are not allowed in this position", loc
-            )
-        return FunctionType([FuncInput(ty, flags) for ty, flags in inputs], output_ty)
+        # Callable types are constructed using special login in the type parser
```
acl-cqc, line 38:
```suggestion
        # Callable types are constructed using special logic in the type parser
```

acl-cqc: You could consider moving some of that code into CallableTypeDef itself, just so you could have a comment "Callable types are constructed using instantiate_func_type()" or something like that, but this is fine

guppylang/tys/parsing.py
```diff
@@ -109,27 +107,81 @@ def arg_from_ast(
...
-            set_location_from(stmt, loc=node)
-            shift_loc(
-                stmt,
-                delta_lineno=node.lineno - 1,  # -1 since lines start at 1
-                delta_col_offset=node.col_offset + 1,  # +1 to remove the `"`
-            )
-            return arg_from_ast(stmt.value, globals, param_var_mapping)
-        except (SyntaxError, ValueError):
-            raise GuppyError("Invalid Guppy type", node) from None
+        node = _parse_delayed_annotation(node.value, node)
```
acl-cqc, line 114: Yes, breaking that out is a good move :)

guppylang/tys/parsing.py
```diff
@@ -109,27 +107,81 @@ def arg_from_ast(
...
+        return stmt.value
+
+
+def _parse_callable_type(
+    args: list[ast.expr],
+    loc: AstNode,
+    globals: Globals,
+    param_var_mapping: dict[str, Parameter] | None,
+) -> FunctionType:
+    """Helper function to parse a `Callable[[<arguments>], <return types>]` type."""
```
acl-cqc, line 144:
```suggestion
    """Helper function to parse a `Callable[[<arguments>], <return type>]` type."""
```
Or are multiple return types allowed??

guppylang/tys/parsing.py
```diff
@@ -109,27 +107,81 @@ def arg_from_ast(
...
+    err = (
+        "Function types should be specified via "
+        "`Callable[[<arguments>], <return types>]`"
```
acl-cqc, line 147: Ditto
```suggestion
        "`Callable[[<arguments>], <return type>]`"
```
I wonder about also noting, this is standard python hyping!

guppylang/tys/parsing.py
```diff
@@ -138,10 +190,29 @@ def type_with_flags_from_ast(
...
     param_var_mapping: dict[str, Parameter] | None = None,
 ) -> tuple[Type, InputFlags]:
-    """Turns an AST expression into a Guppy type possibly annotated with @flags."""
-    # Parse an argument and check that it's valid for a `TypeParam`
-    arg, flags = arg_from_ast(node, globals, param_var_mapping)
-    return _type_param.check_arg(arg, node).ty, flags
+    """Turns an AST expression into a Guppy type with some optional @flags."""
+    # Check for `type @flag` annotations
+    if isinstance(node, ast.BinOp) and isinstance(node.op, ast.MatMult):
+        ty, flags = type_with_flags_from_ast(node.left, globals, param_var_mapping)
```
acl-cqc, line 196: Note this means `x @inout @inout` is also allowed and has the same effect as only once

acl-cqc (approved):
Thank you for changes @mark-koch, looks good to me! :). Improvement in error location is nice too.

I wonder about *also* being able to parse `Annotated[<type>, inout]` just so that there is an allowed "pythonic" syntax too (we claim guppy is pythonic after all! And then we can see what people do with it), but I suspect that's more effort than it's worth...

guppylang/tys/parsing.py
mark-koch: Indeed, I think that should be fine though

Issue #320, opened 2024-07-24
Turn result function into a proper AST node


mark-koch:
Stop relying on `CustomCallCompiler` which breaks the contract between declared type args and function signatures. This causes problems for #321

Issue #318, opened 2024-07-24
Improve linearity error messages involving inout arguments


mark-koch:
* https://github.com/CQCL/guppylang/pull/321#discussion_r1689945388
* https://github.com/CQCL/guppylang/pull/321#discussion_r1689946958

PR #321, opened 2024-07-24
feat:  Update linearity checking for calls with inout arguments

### LINKED ISSUE #317: Update linearity checking for calls with inout arguments

mark-koch:
Implicitly reassign places after a call in an `@inout` position. See #311 for context.

* Store `InputFlags` in `Variable` so we can find out if a variable is `@inout` during linearity checking
* Pass globals to the `BBLinearityChecker` so we can lookup function signatures to figure out if they contain inout args
* Handle `LocalCall` and `GlobalCall` nodes in `BBLinearityChecker` to reassign inout places

Notes:

* Tests involving the result function are failing due to #320
* Some of the error messages in this PR are not ideal. I want to adress that in a separate PR so this one doesn't get too big (see #318)
* Hugr validation is skipped in the tests for now until we have updated the lowering code (see #317)

Closes #317


tests/error/inout_errors/override_after_call.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:16
+
+14:    @guppy(module)
+15:    def test(q1: qubit, q2: qubit) -> tuple[qubit, qubit]:
+16:        q1 = foo(q1, q2)
+                    ^^
+GuppyError: Variable `q1` with linear type `qubit` is not used
```
mark-koch, line 7: This error is quite confusing. We need a special case for unused variables if the last assignment came implicitly from an inout. I'll try to adress that in a separate PR later (see #318).

Any ideas what the message should be here?

tests/error/inout_errors/unused_after_call.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:16
+
+14:    @guppy(module)
+15:    def test(q: qubit) -> None:
+16:       foo(q)
+              ^
+GuppyError: Variable `q` with linear type `qubit` is not used on all control-flow paths
```
mark-koch, line 7: Same problem here. Any ideas what a good error message would be?

aborgna-q: Can we detect when `q` doesn't get used in **any** control-flow paths?
Then we can assume the user though this is a consuming operation, and we can emit something like
```
GuppyError: Variable `q` with linear type `qubit` is not consumed by this call
```

aborgna-q (approved):
Nice!

I love than half the diff is tests :)

PR #326, opened 2024-07-25
feat: Allow redefinition of names in guppy modules

### LINKED ISSUE #307: redefining a function should be allowed
when I use guppy in a notebook and I re-evaluate a cell containing a `@guppy` function definition I get an error:

```
guppylang.error.GuppyError: ('Module `main` already contains a definition named `rus`', None)
```

I suggest that instead, the existing function definition should be replaced

doug-q:
closes #307

### DIFF
--- a/guppylang/module.py
+++ b/guppylang/module.py
@@ -5,7 +5,6 @@ from collections.abc import Callable, Mapping
 from types import ModuleType
 from typing import Any, Union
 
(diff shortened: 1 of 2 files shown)


mark-koch (approved):
I'm happy to merge this as is to improve the experience of notebook users. It matches the regular behaviour of Python anyways.

We can always revisit this later if it leads to problems down the road

PR #328, opened 2024-07-25
ci: Run CI for PRs targeting any branch

mark-koch:
To support working on dev branches like #311


ss2165 (approved):
Ok for now but if CI gets clogged we could revert to encouraging manual runs for stacked PRs

aborgna-q:
`tket2` uses a `run-ci-checks` PR label to force running all the checks. We could do something similar here; by default only run on main unless the PR opts-in

PR #329, opened 2024-07-25
feat: Add Ry gate to the guppy quantum prelude

croyzor:


mark-koch (commented):
This op doesn't exist yet:

https://github.com/CQCL/tket2/blob/7b14f181b21b37eb89d8d4f1228606a7c0476e87/tket2/src/ops.rs#L46-L68

Until then it might be better to define it as

```python
import math

@guppy(module)
def ry(q: qubit, theta: float) -> qubit:
   q = rx(q, py(math.pi / 2))
   q = rz(q, theta + py(math.pi))
   q = rx(q, py(math.pi / 2))
   return rz(q, py(math.pi))
```

Issue #330, opened 2024-07-25
Support tensor calls with inout arguments


mark-koch:

mark-koch:
Closed by #350

PR #331, opened 2024-07-25
feat: Hugr lowering for calls with inout arguments

### LINKED ISSUE #319: Update Hugr lowering for calls with inout arguments

mark-koch:
Closes #319.

* `CallableCompiledDef.compile_call` now additionally returns a list of inout ports that were passed through the function
* Similarly, there is now a `CustomInoutCallCompiler` that does the same. The original `CustomCallCompiler` still exists, but always returns an empty inout list
* `ExprCompiler` now updates the `Place -> Port` mapping for inout arguments after a call (see `ExprCompiler._update_inout_ports` helper method)
* Enabled validation for the inout integration tests

Notes:
* This PR doesn't handle tensor calls with inout arguments (see #330)
* `CustomInoutCallCompiler` isn't used yet since the prelude doesn't use `@inout` yet


tests/integration/test_inout.py
```diff
@@ -97,17 +89,14 @@ def test1(a: MyStruct) -> MyStruct:
         bar(a)
```
aborgna-q, line 89: Uncomment the `# bar(a)` a couple lines above this?

guppylang/definition/value.py
```diff
@@ -71,8 +71,12 @@ def compile_call(
         graph: Hugr,
         globals: "CompiledGlobals",
         node: AstNode,
-    ) -> list[OutPortV]:
-        """Compiles a call to the function."""
+    ) -> tuple[list[OutPortV], list[OutPortV]]:
```
aborgna-q, line 74: It'd be nice to return a Namedtuple or similar to avoid mixing up the two outputs.

Same with `CustomInoutCallCompiler.compile_with_inouts`

aborgna-q (approved):
Nice!

PR #342, opened 2024-07-29
fix: Use places in BB signatures for Hugr generation

### LINKED ISSUE #337: Struct linearity checking bug
The following program incorrectly fails the linearity check

```python
@guppy.struct(module)
class MyStruct:
    q: qubit

@guppy(module)
def test(s: MyStruct, b: bool) -> qubit:
    measure(s.q)
    if b:
        s.q = qubit()
    else:
        s.q = qubit()
    return s.q
```

mark-koch:
Fixes  #337.

The fix for the linearity checker is easy: When checking assignments, only complain if a *local* unused linear variable is shadowed.

However, correctly lowering the program in #337 to Hugr requires some additional changes to Hugr codegen. Namely, we need to consider if a BB uses a *whole* struct `s` or only some of its fields. In the latter case, we should only feed those used values into the BB. This requires specifying the signature of BBs in the form of `Place`s rather than `Variable`s.

Concretely, this PR does the following:

* Make `Signature` generic over the representation of program variables so it can capture both `Variable` and `Place`
* Similarly, make `CheckedBB` and `CheckedCFG` generic
* During checking, we first construct a `CheckedCFG[Variable]` which then gets turned into a `CheckedCFG[Place]` during linearity checking
* Add a test case

### DIFF
--- a/guppylang/checker/cfg_checker.py
+++ b/guppylang/checker/cfg_checker.py
@@ -63,7 +63,8 @@ def check_cfg(
     """Type checks a control-flow graph.
 
     Annotates the basic blocks with input and output type signatures and removes
-    unreachable blocks.
+    unreachable blocks. Note that the inputs/outputs are annotated in the form of
+    *places* rather than just variables.
     """
     # First, we need to run program analysis
     ass_before = {v.name for v in inputs}
--- a/guppylang/checker/cfg_checker.py
+++ b/guppylang/checker/cfg_checker.py
@@ -6,45 +6,48 @@ Operates on CFGs produced by the `CFGBuilder`. Produces a `CheckedCFG` consistin
 
 import collections
 from collections.abc import Iterator, Sequence
-from dataclasses import dataclass
-from typing import TypeVar
+from dataclasses import dataclass, field
+from typing import Generic, TypeVar
 
 from guppylang.ast_util import line_col
 from guppylang.cfg.bb import BB
 from guppylang.cfg.cfg import CFG, BaseCFG
-from guppylang.checker.core import Context, Globals, Locals, Variable
+from guppylang.checker.core import Context, Globals, Locals, Place, V, Variable
 from guppylang.checker.expr_checker import ExprSynthesizer, to_bool
-from guppylang.checker.linearity_checker import check_cfg_linearity
 from guppylang.checker.stmt_checker import StmtChecker
 from guppylang.error import GuppyError
 from guppylang.tys.ty import Type
 
-VarRow = Sequence[Variable]
+Row = Sequence[V]
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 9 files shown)


guppylang/checker/cfg_checker.py
```diff
@@ -55,8 +58,8 @@ def __init__(self, input_tys: list[Type], output_ty: Type) -> None:
 
 
 def check_cfg(
-    cfg: CFG, inputs: VarRow, return_ty: Type, globals: Globals
-) -> CheckedCFG:
+    cfg: CFG, inputs: Row[Variable], return_ty: Type, globals: Globals
+) -> CheckedCFG[Place]:
     """Type checks a control-flow graph.
 
     Annotates the basic blocks with input and output type signatures and removes
```
aborgna-q, line 65: And also mention the places instead of variables in the return type here.

guppylang/checker/cfg_checker.py
```diff
@@ -117,18 +120,20 @@ def check_cfg(
     }
 
     # Finally, run the linearity check
-    check_cfg_linearity(checked_cfg)
+    from guppylang.checker.linearity_checker import check_cfg_linearity
```
aborgna-q, line 123: Why the local import here, instead of having it at the top?

aborgna-q (commented):
Nice solution!

guppylang/compiler/cfg_compiler.py
```diff
@@ -149,20 +153,21 @@ def choose_vars_for_tuple_sum(
...
     linearity and name. The only exception are return vars which must be outputted in
     order.
     """
-    if is_return_var(x.name) and is_return_var(y.name):
-        return -1 if x.name < y.name else 1
-    return -1 if (x.ty.linear, x.name) < (y.ty.linear, y.name) else 1
+    x, y = str(p1), str(p2)
+    if is_return_var(x) and is_return_var(y):
+        return -1 if x < y else 1
+    return -1 if (p1.ty.linear, x) < (p2.ty.linear, y) else 1
```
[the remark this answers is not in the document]
mark-koch: Indeed, this `%ret` business is quite hacky. This currently works since return vars are only compared with each other, hower this changes with `@inout`.

This special handling of return vars is removed in #340, so I think it's not really worth putting effort into a nice solution here though

guppylang/checker/cfg_checker.py
mark-koch: Circuilarity, since `linearity_checker.py` already depends on `cfg_checker.py`.

I could move `Signature`, `CheckedBB`, and `CheckedCFG` to `checker/core.py` to break the cycle if you prefer?

aborgna-q (approved):

PR #345, opened 2024-07-29
chore: Downgrade mypy to 1.10.0

mark-koch:
For some reason, the recent mypy 1.11.0 release breaks type checking for AST related code. For example, see https://github.com/CQCL/guppylang/actions/runs/10146815891/job/28055775030?pr=341

We need to downgrade to 1.10 for now to fix CI runs


ss2165 (approved):
is there an associated mypy issue we could track?

mark-koch:
I couldn't find any issues mentioning this. It looks like typeshed got updated and subclassing ast nodes no longer works properly. Not sure if this can be considered a mypy/typeshed bug?

I created #346 to investigate further

PR #360, opened 2024-07-31
test: Add a T factory example program

croyzor:


examples/t_factory.py
```diff
@@ -0,0 +1,94 @@
...
+    qubit,
+    rx,
+    rz,
+)
+
+module = GuppyModule("tfactory")
+module.load(quantum)
+
+phi = np.arccos(1 / 3)
+pi = np.pi
```
acl-cqc, line 21: nit: something like `from numpy import pi`

examples/t_factory.py
```diff
@@ -0,0 +1,94 @@
...
+    discard,
+    h,
+    measure,
+    quantum,
+    qubit,
+    rx,
+    rz,
+)
+
+module = GuppyModule("tfactory")
```
acl-cqc, line 17: Maybe `t-factory` or `t_factory` ?

examples/t_factory.py
```diff
@@ -0,0 +1,94 @@
...
+
+@guppy(module)
+def ry(q: qubit, theta: float) -> qubit:
+    q = rx(q, py(pi / 2))
+    q = rz(q, theta + py(pi))
+    q = rx(q, py(pi / 2))
+    return rz(q, py(pi))
+
+
+# Preparation of approximate T state, from 2310.12106
```
acl-cqc, line 32: If an arxiv link, suggest you put in the full address (at least `arxiv`, preferably hostname and path, possibly protocol `https://`)

examples/t_factory.py
```diff
@@ -0,0 +1,94 @@
...
+@guppy(module)
+def prepare_approx(q: qubit) -> qubit:
+    phi_ = py(phi)
+    pi_ = py(pi)
+
+    q = ry(q, phi_)
+    return rz(q, pi_ / 4.0)
+
+
+# The inverse of the [[5,3,1]] encoder in figure 3 of 2208.01863
```
acl-cqc, line 42: ditto arxiv link

examples/t_factory.py
```diff
@@ -0,0 +1,94 @@
...
+    """
+    q0, q1 = cz(q0, q1)
+    q2, q3 = cz(q2, q3)
+    target, q0 = cz(target, q0)
+    q1, q2 = cz(q1, q2)
+    target, q3 = cz(target, q3)
+    # Measuring gives false for success, true for failure.
+    # We check for all falses to say whether distillation succeeded.
+    bits = [not (measure(h(q))) for q in [q0, q1, q2, q3]]
+    success = True
```
acl-cqc, line 59: Can this be just `success = all(bits)`? Hmmm, maybe that's not in guppy yet - if so, suggest commenting that, I bet it'll go in there at some point :-)

examples/t_factory.py
```diff
@@ -0,0 +1,94 @@
...
+def t_state(timeout: int) -> tuple[linst[qubit], bool]:
+    """Create a T state using magic state distillation with `timeout` attempts.
+
+    On success returns a singleton `linst` containing a qubit in a magic T state
+    and the boolean `True`.
+
+    If the number of attempts is exceeded, the empty `linst` will be returned
+    along with the boolean `False`.
+    """
+    if timeout > 0:
```
acl-cqc, line 75: Is this deliberately recursive? You could just make this a `while`...

examples/t_factory.py
```diff
@@ -0,0 +1,94 @@
...
+        q2 = prepare_approx(qubit())
+        q3 = prepare_approx(qubit())
+
+        q, success = distill(tgt, q0, q1, q2, q3)
+        if success:
+            return [q], True
+        else:
+            # Discard the qubit and start over
+            discard(q)
+            return t_state(timeout - 1)
```
acl-cqc, line 88: ...and change this to `timeout -= 1`. Optional, YMMV, you might explicitly want to test recursion!

acl-cqc (approved):
Thanks, Craig - looks very nice! A few nits but LGTM

examples/t_factory.py
croyzor: That's what I did on my first attempt, but it's not supported by guppy yet!

examples/t_factory.py
croyzor: Yeah I'm of two minds - obviously it could go either way. The `while` version is probably more realistic but I thought a recursive definition would be better for test coverage

acl-cqc: Yeah, testing recursion is no bad thing!! Just comment "recursive for test coverage" on the recursive call, say

PR #370, opened 2024-08-12
fix!: Use latest results extension spec

### LINKED ISSUE #365: result feature doesn't seem to generate correct HUGR for classical programs
Result function introduced in CQCL/guppylang#271 generates correct HUGR for quantum measurements, such as, `result(0, measure(q))`, but not for a program like the following:

```py
mod = GuppyModule("main")
mod.load(quantum)
@guppy(mod)
def main() -> None:
    a = 1 + 2  # noqa: F841
    result(0, a)
```

### LINKED ISSUE #369: Use latest results extension spec
See https://github.com/CQCL/tket2/pull/494

This should fix #365

mark-koch:
Fixes #369 and fixes #365.

Make Guppy conform with the latest results spec introduced in https://github.com/CQCL/tket2/pull/494

BREAKING CHANGE:
* Result tags are now strings instead of ints
* Only numeric values and arrays thereof are allowed as results


guppylang/compiler/expr_compiler.py
```diff
@@ -297,14 +302,51 @@ def visit_FieldAccessAndDrop(self, node: FieldAccessAndDrop) -> OutPortV:
...
+                    base_name = "f64"
+                case kind:
+                    assert_never(kind)
+        else:
+            # The only other valid base type is bool
+            assert is_bool_type(node.base_ty)
+            base_name = "bool"
+        if node.array_len is not None:
+            op_name = f"result_array_{base_name}"
+            assert node.array_len is not None
```
ss2165, line 328: redundant assert?

guppylang/prelude/_internal.py
```diff
@@ -277,22 +283,34 @@ def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
...
+            f"Expression of type `{ty}` is not a valid result. Only numeric values or "
+            "arrays thereof are allowed."
+        )
+        if isinstance(ty, NumericType) or is_bool_type(ty):
+            base_ty = ty
+            array_len: Const | None = None
+        elif is_array_type(ty):
+            [ty_arg, len_arg] = ty.args
+            assert isinstance(ty_arg, TypeArg)
+            assert isinstance(len_arg, ConstArg)
```
ss2165, line 305: can these asserts result in unhelpful errors?

guppylang/prelude/_internal.py
```diff
@@ -277,22 +283,34 @@ def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
...
+            if not isinstance(ty_arg.ty, NumericType) and not is_bool_type(ty_arg.ty):
```
ss2165, line 306: seems worth commoning up the "is numeric or bool" check somewhere nearby

ss2165 (commented):
Some minor queries

guppylang/prelude/_internal.py
mark-koch: No, failing asserts would be a compiler bug since you shouldn't be able to construct an array type with invalid args

ss2165 (approved):
LGTM

PR #374, opened 2024-08-12
fix: Fix struct definitions in notebooks

### LINKED ISSUE #373: Struct definitions don't work in Jupyter notebooks
Trying to do `inspect.getsourclines` for a class defined in a notebook yields a `OSError: source code not available`.

https://github.com/ipython/ipython/issues/11249
https://bugs.python.org/issue33826

mark-koch:
Fixes #373.

Trying to do `inspect.getsourclines` for a class defined in a notebook yields an `OSError: source code not available`. See
* https://github.com/ipython/ipython/issues/11249
* https://bugs.python.org/issue33826

The fix suggested in https://github.com/ipython/ipython/issues/11249 only works for classes with methods. Instead, I'm using a solution inspired by https://github.com/wandb/weave/pull/1864 that retrieves the cell sources of the active IPython session and searches them one by one.

### DIFF
--- a/guppylang/definition/struct.py
+++ b/guppylang/definition/struct.py
@@ -4,7 +4,7 @@ import textwrap
 from collections.abc import Sequence
 from dataclasses import dataclass
 from functools import cached_property
-from typing import Any
+from typing import Any, cast
 
 from guppylang.ast_util import AstNode, annotate_location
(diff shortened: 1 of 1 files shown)


aborgna-q (approved):
🫠

Issue #375, opened 2024-08-12
Check for example notebook regressions in CI


mark-koch:
We can use https://pytest-notebook.readthedocs.io/en/latest/

This requires making the notebook outputs deterministic (currently, error messages mention a temporary file as the error location). See #381

PR #376, opened 2024-08-12
fix: Fix and update demo notebook

### LINKED ISSUE #371: Fix demo notebook

mark-koch:
Closes #371. Depends on #374.

See https://github.com/CQCL/guppylang/blob/fd3b2fc43018ff1a9c51728453cfd09cd7e664a2/examples/demo.ipynb for the rendered notebook.

Drive-by: Add `bool.__xor__` method for the example.

We should check for notebook regressions in CI to avoid breaking the example in the future. See #375


guppylang/prelude/builtins.py
```diff
@@ -87,6 +87,9 @@ def __new__(x): ...
     @guppy.hugr_op(builtins, logic_op("Or", [tys.TypeArg(tys.BoundedNatArg(n=2))]))
     def __or__(self: bool, other: bool) -> bool: ...
 
+    @guppy.hugr_op(builtins, DummyOp("Xor"))
+    def __xor__(self: bool, other: bool) -> bool: ...
```
mark-koch, line 91: The logic extension doesn't define Xor yet. See https://github.com/CQCL/hugr/issues/1418

ss2165 (approved):
Really nice intro, just two notes:
1. Specify a version of guppy the notebook works with at the top, at least until we have CI checks.
2. `Qubit in Guppy` typo, should be "Qubits"

Issue #66, opened 2023-12-08
Enable pre-commit on github CI


qartik:
I added a pre-commit config in #64, it will be ideal to run the same setup on the CI as well.

PR #388, opened 2024-08-16
chore: Use local ruff and mypy for pre-commit checks

aborgna-q:
Ensures that the checks run by pre-commit are the same as the ones run in CI, and the `ruff` and `mypy` versions are always controlled by `pyproject.toml`.
Before, we used pre-commit actions with their own versions that needed to be updated separately and tended to get out of date.

drive-by: Fix ruff lints added in the latest version

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -45,17 +45,10 @@ jobs:
         run: poetry run mypy guppylang
 
       - name: Check formatting with ruff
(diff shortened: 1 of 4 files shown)


qartik:
Would it make sense to instead call pre-commit in the CI? See #66. We do it in [pytket-phir](https://github.com/CQCL/pytket-phir/blob/main/.github/workflows/python-app.yml#L31-L38)

PR #394, opened 2024-08-20
feat: Add equality test for booleans

### LINKED ISSUE #363: `Bool` does not implement `__eq__`

croyzor:
Closes #363

### DIFF
--- a/guppylang/prelude/builtins.py
+++ b/guppylang/prelude/builtins.py
@@ -75,7 +75,7 @@ class Bool:
     @guppy.custom(builtins, NoopCompiler())
     def __bool__(self: bool) -> bool: ...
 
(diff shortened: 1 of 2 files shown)


guppylang/prelude/builtins.py
```diff
@@ -75,6 +75,9 @@ def __and__(self: bool, other: bool) -> bool: ...
     @guppy.custom(builtins, NoopCompiler())
     def __bool__(self: bool) -> bool: ...
 
+    @guppy.hugr_op(builtins, logic_op("Eq", [tys.TypeArg(tys.BoundedNatArg(n=2))]))
```
ss2165, line 78: shouldn't n=1?

croyzor: The argument here is to the template for boolean ops, the 2 is specifying how many inputs the function has, rather than the bound of the input Ints

ss2165 (approved):

PR #398, opened 2024-08-23
chore: Replace `poetry` with `uv`

### LINKED ISSUE #395: Try using `uv` instead of `poetry` for project management
https://astral.sh/blog/uv-unified-python-packaging

aborgna-q:
Replaces the python project manager with `uv`.
For most commands that used poetry before, it just be as easy as replacing `poetry {things}` with `uv {things}`.

Some notable exeptions:

- Instead of `poetry install --with extra1,extra2`,
  use `uv sync --extra extra1 --extra extra2`.
  Note that will delete any extraneous packages/extras so the venv state is deterministic.
  Add `--inexact` if you want to replicate the old poetry behaviour.

- `poetry shell` is no longer available. The preferred workflow is to `uv run` / `uvx` things, to ensure you are always in sync with the project definition.
  If that's too annoying, it is always possible to `source .venv/bin/activate`.

Updates the justfile, so using that should be the same as before.

I updated the `devenv` definition, but someone that uses it should double-check it.

Closes #395.


aborgna-q:
This should be good to go.

@ss2165 Can you check if the devenv config works? The lockfile needs to be updated too.

.github/workflows/docs.yml
```diff
@@ -26,15 +29,21 @@ jobs:
       - uses: actions/checkout@v4
       - name: Run sccache-cache
         uses: mozilla-actions/sccache-action@v0.0.5
-      - name: Install poetry
-        run: pipx install poetry
-      - name: Set up Python '3.10'
+
+      - name: Set up uv
+        run: curl -LsSf https://astral.sh/uv/0.3.4/install.sh | sh
```
croyzor, line 34: It's a little bit scary to install via curl in our CI - I guess there's no actoin for installing uv yet?

croyzor (approved):
LGTM

.github/workflows/docs.yml
aborgna-q: It's the recommended method for CI runs :/
https://docs.astral.sh/uv/guides/integration/github/#installation

But yeah, I hope they create an action soon.

Issue #253, opened 2024-06-19
Tracking issue: Arrays


mark-koch:
- [x] #250
- [x] #228
- [x] #214
- [x] #251
- [x] #416
- [x] #417
- [x] #252
- [x] #418
- [x] #419
- [x] #396

PR #415, opened 2024-08-28
feat: Array indexing

mark-koch:
This is the feature branch for array indexing. Tracked by #253.

The goal is to allow array indexing in `@inout` positions that ignores linearity constraints in the indices:

```python
qs: array[qubit, 42]
cx(qs[0], qs[1])  # Ok
cx(qs[0], qs[0])  # Compiles, but will panic at runtime
q = qs[0]  # Error: Indexing only allowed in inout position
```

This is achieved by lowering `array[qubit]` to `array[qubit | None]` and making `array.__getitem__` and `array.__setitem__` swap in `None` whenever a qubit is taken out or put back in. The functions `array.__getitem__` and `array.__setitem__` take the array as `@inout`, so it suffices to apply the following desugaring:

```python
cx(qs[expr1], qs[expr2])

# Desugars to
idx1 = expr1
idx2 = expr2
tmp1 = qs.__getitem__(idx1)
tmp2 = qs.__getitem__(idx2)
cx(tmp1, tmp2)
qs.__setitem__(tmp1, idx1)
qs.__setitem__(tmp2, idx2)
```

This also means that arrays containing structs might behave slightly unexpectedly:

```python
@guppy.struct
class QubitPair:
    q1: qubit
    q2: qubit

ps: array[QubitPair, 42]
cx(ps[0].q1, ps[0].q2)  # Panics at runtime :(

# Since it desugars to
idx1 = 0
idx2 = 0
tmp1 = ps.__getitem__(idx1)
tmp2 = ps.__getitem__(idx2)  # Panic: Struct at index 0 has already been replaced with None
cx(tmp1.q1, tmp2.q2)
...
```

To solve this, we need to change the Hugr lowering of arrays to structually replace any occurrence of `qubit` with `qubit | None`, i.e. instead of doing `array[QubitPair | None]`, we would need to do `array[tuple[qubit | None, qubit | None]]`. I'll leave this to a future PR in case we are interested in that.

The following feature PRs target this branch:
* #420 
* #421
* #422
* #447


qartik:
> ```python
> h(tmp2)
> ```

This should be `cx(tmp1, tmp2)`

ss2165 (approved):
🎉 🎉

PR #420, opened 2024-08-28
feat: Type check array subscripts

### LINKED ISSUE #416: Emit place when checking array subscripts

mark-koch:
Closes #416. See #415 for context.

* Adds a new `SubscriptAccess` place that will be used to track array subscripts during linearity checking
* This place is emitted when checking a subscript AST node
* Ensures that subscripts in inout positions also implement a `__setitem__` method


tests/error/inout_errors/subscript_not_setable.err
```diff
@@ -0,0 +1,7 @@
+Guppy compilation failed. Error in file $FILE:24
+
+22:    @guppy(module)
+23:    def test(c: MyImmutableContainer) -> MyImmutableContainer:
+24:        foo(c[0])
+               ^^^^
+GuppyTypeError: Expression of type `MyImmutableContainer` is not assignable since it does not implement the `__setitem__` method
```
qartik, line 7: The error message seems confusing here as the effect of `c[0]` is not an _assignment_. To me it reads more like `__getitem__`. Isn't the error really because `foo` consumes an `inout` and doesn't return it back?

Assuming that is allowed if `__setitem__` is defined for the container, perhaps the error messaging needs to be improved here. Something about `foo` requiring an assignment that's not implemented by the container.

qartik (approved):
LGTM, I think an error message can be improved, however.

tests/error/inout_errors/subscript_not_setable.err
mark-koch: The problem is that once the `inout` value has been given back by `foo`, we cannot assign it back to `c[0]` since `c` doesn't implement `__setitem__`.

But you're right, the current message is misleading. I'll update it 👍

qartik: Ah, that explains the reassignment, certainly better error message will help here.

PR #423, opened 2024-08-28
chore: Update hugr llvm and test new op lowering

croyzor:
Update hugr_llvm so that we get lowering for multiply and divide. Test the new functionality

Drive-by: Fix inkwell to version 0.4.0 - the new release (0.5.0) has just come out and has api breaking changes


mark-koch (approved):
Nice!

PR #432, opened 2024-08-30
feat: Allow imports of function definitions and aliased imports

### LINKED ISSUE #134: Rethink the import system

### LINKED ISSUE #425: Implement imports w/o linking
Just lower everything into the same Hugr until we have implemented proper Hugr linking

### LINKED ISSUE #427: Add aliased imports

mark-koch:
Closes #425 and closes #427 and closes #134

* Allow importing of function definitions (before we could only import declarations)
* Until we have Hugr linking, we just lower everything into the same Hugr. This requires some changes to the compilation logic:
  - Differentiate between checking (`module.check()`) and full Hugr compilation (`module.compile()`) of modules
  - Importing modules only checks them
  - Hugr lowering is done at the end in one step over all imported modules
* Allow to import individual definitions instead of whole modules
* Allow aliased imports using kwargs (see test examples)

### DIFF
--- a/guppylang/module.py
+++ b/guppylang/module.py
@@ -1,24 +1,23 @@
 import inspect
-import itertools
 import sys
 from collections.abc import Callable, Mapping
 from types import ModuleType
-from typing import Any, Union
+from typing import Any
 
 from hugr import Hugr, ops
 from hugr.function import Module
 
 from guppylang.checker.core import Globals, PyScope
-from guppylang.compiler.core import CompiledGlobals
 from guppylang.definition.common import (
     CheckableDef,
     CheckedDef,
     CompilableDef,
+    CompiledDef,
     DefId,
+    Definition,
     ParsableDef,
     RawDef,
 )
-from guppylang.definition.custom import CustomFunctionDef
 from guppylang.definition.declaration import RawFunctionDecl
 from guppylang.definition.function import RawFunctionDef
 from guppylang.definition.parameter import ParamDef
@@ -35,6 +34,9 @@ class GuppyModule:
 
     name: str
 
+    # Whether the module has already been checked
+    _checked: bool
+
     # Whether the module has already been compiled
     _compiled: bool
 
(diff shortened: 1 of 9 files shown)


tests/error/misc_errors/return_flag_callable.py
```diff
@@ -11,7 +11,7 @@
 
 
 @guppy.declare(module)
-def foo(f: Callable[[], qubit @inout]) -> None: ...
+def foo(f: "Callable[[], qubit @inout]") -> None: ...
```
mark-koch, line 14: This is a bit annyoing. The `qubit` type now needs to be a `Definition` so we can import it, however Pythons `Callable` checks that the generic args are `type`s at runtime...

Fixing this with a string annotation for now, but should figure out something better. See #433

tests/integration/test_imports.py
```diff
@@ -0,0 +1,128 @@
...
+    module = GuppyModule("test")
+    module.load(g)  # `g` returns a type that was defined in `mod_a`
+
+    @guppy(module)
+    def test() -> None:
+        x = g()  # Type of x was defined in `mod_a`
+        # TODO: Should we be allowed to use a method that was defined in `mod_a` even
+        #  though we have not explicitly imported it? This would work in Python, but
+        #  not in Rust
+        # int(-x)
```
mark-koch, line 98: Any thoughts on this?

mark-koch: Nevermind, this also works in Rust. I'll implement it

guppylang/module.py
```diff
@@ -63,44 +70,95 @@ class GuppyModule:
...
+        names: dict[str, DefId] = {}
+
+        # Collect imports in reverse since we'll use it as a stack to push and pop from
+        imports: list[tuple[str, Definition | GuppyModule | ModuleType]] = [
+            *reversed(kwargs.items()),
+            *(("", arg) for arg in reversed(args)),
+        ]
+        while imports:
+            alias, imp = imports.pop()
+            if isinstance(imp, Definition):
```
croyzor, line 108: It'd be nice to check for name clashes. I.e. if `mod_a` and `mod_b` both define `foo`, and I write

```
mod_c = GuppyModule('')
mod_c.load(mod_a)
mod_c.load(mod_b)

@guppy(mod_c)
def bar() -> ():
    return foo()
```

Then I assume `mod_b`'s `foo` is shadowing the one from `mod_a`, but I'd rather this was made explicit

croyzor (approved):
Nice!

guppylang/module.py
mark-koch: Not sure, Python let's you shadow stuff if you do `from module import *`.

In the past we were for example checking that people don't define functions with clashing names, but this got annoying in the Jupyter setting. My guess is that the same applies here

croyzor: If that's the python semantics then I guess it's okay for this example. How does python deal with aliases? Because if we make an alias that could also be shadowed by something else, which seems a bit odd

mark-koch: > Because if we make an alias that could also be shadowed by something else, which seems a bit odd

That's Python for you 😅

`import` brings names into scope, override whatever was bound to the name before

croyzor: At least it's consistent!

PR #441, opened 2024-09-02
fix: Fix linearity checking bug

### LINKED ISSUE #438: Linearity checking bug
The following program incorrectly fails linearity checking:

```python
@guppy.struct(module)
class MyStruct:
    q: qubit


@guppy.declare(module)
def use(q: qubit) -> None: ...


@guppy(module)
def test(s: MyStruct) -> MyStruct:
    use(s.q)
    while True:
        s.q = qubit()
        return s
    # Guppy is not yet smart enough to detect that this code is unreachable
    s.q = qubit()
    return s
```

mark-koch:
Fixes #438.

When checking that all live linear places in scope are definitely used, we have to be more careful about determining whether variables are *actually* live rather than only being included in scope because of the less precise first pass of liveness analysis.

### DIFF
--- a/tests/integration/test_inout.py
+++ b/tests/integration/test_inout.py
@@ -266,7 +266,6 @@ def test_move_back(validate):
     validate(module.compile())
 
 
-@pytest.mark.skip("Fails due to https://github.com/CQCL/guppylang/issues/337")
 def test_move_back_branch(validate):
     module = GuppyModule("test")
     module.load(quantum)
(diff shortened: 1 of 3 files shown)


tests/integration/test_inout.py
```diff
@@ -266,7 +266,6 @@ def main(s: MyStruct, t: MyStruct) -> MyStruct:
     validate(module.compile())
 
 
-@pytest.mark.skip("Fails due to https://github.com/CQCL/guppylang/issues/337")
```
mark-koch, line 269: Drive-by: This test works now

ss2165 (approved):
reassuring fix!

PR #446, opened 2024-09-03
feat: Add array literals

### LINKED ISSUE #396: Add array literals
`xs: array[int, 4] = array(1, 2, 3, 4)`

mark-koch:
Closes #396.

* New arrays can be created using the constructor `array(x, y, z)`.
* Uses a `CustomCallChecker` to support varargs
* This broke some code that relied on the static signature of functions. Added special cases for `CustomFunctions` without known signature (such as the array constructor) that infers a concrete signature based on the context.


qartik (approved):
LGTM

Issue #429, opened 2024-08-30
Implement `range` in prelude


mark-koch:
Blocked by #425

ss2165:
Leaving more than single-argument ranges to future issues

Issue #277, opened 2024-06-28
Automatically annotate guppy definitions with `@no_type_check`


ss2165:
So mypy knows to ignore it https://docs.python.org/3/library/typing.html#typing.no_type_check

mark-koch:
The thing we want is https://docs.python.org/3/library/typing.html#typing.no_type_check_decorator to annotate `@guppy` as always being `no_type_check`.

Unfortunalty, no Python type checker has implemented it and it's going to be deprecated in Python 3.13: https://github.com/python/cpython/issues/106309

croyzor:
Relatedly, when we call tuples of functions in guppy code, mypy complains that tuples aren't callable 😅

PR #452, opened 2024-09-04
feat: range() with single-argument

acl-cqc:
Addresses #429 but maybe only partially.

* `xfail` some tests that look for single nodes in the compiled Hugr, as this currently includes the whole stdlib. See https://github.com/CQCL/guppylang/issues/470.
* `@typing.overload` the `@guppy` decorator to remove mypy error when using `@guppy(module)`
* `parse_py_class` attempts `inspect.getsourcelines` when running IPython, as this works for the stdlib


guppylang/decorator.py
```diff
@@ -61,6 +61,12 @@ class _Guppy:
     def __init__(self) -> None:
         self._modules = {}
 
+    @overload
+    def __call__(self, arg: PyFunc) -> RawFunctionDef: ...
```
acl-cqc, line 65: These got rid of a load of errors on `@guppy(builtins)` but we could say that typechecking `builtins.py` or any other guppy file is such a lost cause that we shouldn't bother....?

guppylang/prelude/builtins.py
```diff
@@ -882,8 +882,45 @@ def print(x): ...
...
 
-@guppy.custom(builtins, checker=UnsupportedChecker(), higher_order_value=False)
-def range(x): ...
+@guppy.struct(builtins)
+class Range:
+    stop: int
+
+    @guppy(builtins)
+    def __iter__(self: "Range") -> "RangeIter":
+        return RangeIter(0, self.stop)  # type: ignore[call-arg]
```
acl-cqc, line 891: Still lots of `type: ignore` here even with the `@overload`s in `decorator.py`. Maybe we should just disable mypy for the whole file??

tests/integration/test_basic.py
```diff
@@ -63,6 +64,7 @@ def foo(x: bool) -> bool:
     validate(foo)
 
 
+@pytest.mark.xfail(reason="hugr-includes-whole-stdlib")
```
acl-cqc, line 67: I did try fixing this one, see commit b663e8c5144a21c6e09aea2a7b1c144af7e3c4ab. But gave up when I realized how many other problems there were, so reverted and xfail this too to be consistent...

guppylang/decorator.py
mark-koch: Yeah, this is a difficult topic. I think for us it's acceptable to disable mypy for `builtins.py`, but actual users will run into the same problem when they run mypy on their code. Also see #277

Happy to keep these `__call__` overloads if they help 👍

guppylang/definition/struct.py
```diff
@@ -232,24 +232,29 @@ def parse_py_class(cls: type) -> ast.ClassDef:
...
-        annotate_location(defn.node, defn.cell_source, f"<{defn.cell_name}>", 1)
-        if not isinstance(defn.node, ast.ClassDef):
-            raise GuppyError("Expected a class definition", defn.node)
-        return defn.node
+        if defn is not None:
+            annotate_location(defn.node, defn.cell_source, f"<{defn.cell_name}>", 1)
+            if not isinstance(defn.node, ast.ClassDef):
+                raise GuppyError("Expected a class definition", defn.node)
+            return defn.node
+        # inspect.getsourcelines works for classes defined in the guppy stdlib/builtins
```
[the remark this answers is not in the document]
mark-koch: Yes, fall-through without else would be nicer. Throwing the `OSError` is fine 👍

guppylang/prelude/builtins.py
mark-koch: There is a `mypy: ignore` pragma at the start of the file. You can add `call-arg` to that.

Alternatively, you can add a `@no_type_check` decorator to the function

tests/integration/test_array.py
```diff
@@ -11,6 +11,7 @@
 import guppylang.prelude.quantum as quantum
 
 
+@pytest.mark.xfail(reason="hugr-includes-whole-stdlib")
```
mark-koch, line 14: Maybe we should just comment out the asserts below instead of marking the whole test as xfail? Then we'd at least still be testing the compilation code paths and validation

tests/integration/test_basic.py
mark-koch: Yeah, xfail is fine for this PR. Can you add an issue so we don't forget about it?

> The "right" solution - while we wait for proper Hugr linking - might well be to add metadata to every Hugr FuncDefn saying which source file it's from so we can filter better....but that's another PR? Or is there something like this already?

This doesn't exist yet, but sounds like good idea 👍

mark-koch (approved):
Nice work! Sorry you had to run into problems with my IPython hacks 😅

tests/integration/test_array.py
acl-cqc: Done with explicit call to `pytest.xfail` so we have a good marker, will document in issue

guppylang/prelude/builtins.py
acl-cqc: `@no_type_check` does work, but still has to be done in three places.

I've gone with whole-file `mypy: disable-error-code`, but the killer here is that this makes the `disable-error-code` line too long (so `ruff` complains), and I couldn't see any way to disable the line-length check for just that line (if you stick a `# noqa: E501` on the end, ruff doesn't recognize this because it's within an existing comment, and mypy takes that as part of an error-code to disable, doesn't recognize it and barfs everywhere...)

So, significant cost to disabling `call-arg` across the file - perhaps best to run with three separate `#type: ignore`s?

acl-cqc: > Oof that's annoying, I presume you can't do multiple `mypy: disable-error-code` lines?

Sadly not - see e.g. https://github.com/python/mypy/issues/14750

> I'm happy with `@no_type_check` or `#type: ignore`

Ok, I'll reinstate the type-ignores.

PR #453, opened 2024-09-04
feat: [FYI] RangeChecker/RangeCompiler to support 1/2/3 args

acl-cqc:


guppylang/prelude/_internal/compiler.py
```diff
@@ -438,3 +438,37 @@ def build_array_set(
...
+
+
+class RangeCompiler(CustomCallCompiler):
+    def compile(self, args: list[Wire]) -> list[Wire]:
+        from typing import cast
+
+        from guppylang.compiler.expr_compiler import python_value_to_hugr
+        from guppylang.definition.custom import CustomFunctionDef
+
+        # Find the builtins module. Brute-force search!
```
acl-cqc, line 450: This is the worst bit! Custom *Checkers* get access to the Func + full Context, *Compilers* only get the CompiledGlobals which is rather hard to use

mark-koch:
We can now do this without custom checkers! Closing in favour of #1157

PR #455, opened 2024-09-05
feat: Use `hugr-cli` for validation

### LINKED ISSUE #390: Drop the local validator
Use the `hugr` cli tool instead. If we wish to maintain feature-parity we'll need to export an extension definition for tket2 and include it here (via git subtrees?).

See https://github.com/CQCL/tket2/issues/556

aborgna-q:
Spinoff of #454 

Replaces the validator python library with a small binary that calls out to `hugr-cli validate`.

`cargo install` lets us download tools like `hugr-cli` and use them for local development, but the support for locally scoped tools is quite flaky. By default, tools are installed in a global directory using the latest version available.
It can be forced to use a local target directory via an env variable / `cargo` config (which didn't always work while I was testing it), and version selection can only be done by passing explicit arguments to the cmd. As there is no central `Cargo.toml` (or similar) config for it, this relies on every usage point always passing exactly the same arguments. Updating the tool version / patching in a git ref ends up being quite error prone.

The solution in this PR is inspired by cargo's own `xtask-` subcrates. These are internal crates meant to run some external tool, while centrally defining the dependency versions and artifact paths.

The new `validator` bin crate here checks if `hugr-cli` is installed, and calls `cargo install` with the appropriate parameters otherwise, before running the tool. The version to install matches the one configured in `Cargo.toml`. This ensures we control the validation in the same way as all the other hugr dependencies (e.g. the `hugr` dep used in `execute_llvm`).

Running `cargo run` now acts the same as executing a locally-versioned `hugr-cli`.

drive-by: Created a cargo workspace.

note: I temporarily disabled the tests that required the "collections" extension definition. This will get fixed once we merge #454, I just didn't want to wire in a hacky solution that'll get dropped in the next PR.

Closes #390


validator/src/lib.rs
```diff
@@ -1,38 +1,139 @@
...
+            if let Some(rev) = &self.dep.git_rev() {
+                cmd.arg("--rev").arg(rev);
+            }
+
+            if let Some(branch) = &self.dep.detail().and_then(|d| d.branch.clone()) {
+                cmd.arg("--branch").arg(branch);
+            }
+
+            if let Some(tag) = &self.dep.detail().and_then(|d| d.tag.clone()) {
+                cmd.arg("--branch").arg(tag);
```
croyzor, line 121:
```suggestion
                cmd.arg("--tag").arg(tag);
```

tests/integration/test_arithmetic.py
```diff
@@ -57,6 +58,7 @@ def main() -> int:
     run_int_fn(compiled, 6)
 
 
+@pytest.mark.skip(reason="trunc_u has an invalid signature")
```
croyzor, line 61: What's going on here?

tests/integration/conftest.py
```diff
@@ -12,26 +13,52 @@ def export_test_cases_dir(request):
...
 @pytest.fixture
 def validate(request, export_test_cases_dir: Path):
     def validate_json(hugr: str):
-        try:
-            import guppyval
-
-            guppyval.validate_json(hugr)
-        except ImportError:
-            pytest.skip("Skipping validation")
+        # Install and run the `hugr-cli` validator
```
croyzor, line 31: This comment seems inaccurate

tests/integration/test_arithmetic.py
aborgna-q: Not quite sure, I think guppy defined an invalid signature.
It's getting fixed in #454, so I didn't investigate much.

croyzor (approved):

Issue #264, opened 2024-06-24
Turn bool into a numeric type?


mark-koch:
Should we turn bool into a numeric type to match Python's behaviour? This means all of the following equalities would hold:

```python
True == 1
False == 0
True + True == 2
True - False == 1
2 * False == 0
...
```

See #263 for a PR that implements this.

Personally, I'd be in favour to increase consistency with Python. It also allows some nice patterns like

```python
results = [measure(q) for q in qs]
parity = sum(results) % 2
```

What do you think? @croyzor @qartik @ss2165

qartik:
> If we are considering guppy as a strongly typed language, I think `bool` should not be implicitly treated as a numeric type. We could certainly provide casting/conversion functions for explicit coercion, if they don't exist already.

_Originally posted by @qartik in https://github.com/CQCL/guppylang/issues/225#issuecomment-2139705909_

Type-theoretically speaking, I wouldn't want to allow the equalities in the OP.

The pattern shown is very easy and way more explicit with a coercion function:

> ```python
> - results = [measure(q) for q in qs]
> + results = [int(measure(q)) for q in qs]
>   parity = sum(results) % 2
> ```

PR #459, opened 2024-09-05
feat: Support `nat`/`int` ↔ `bool` cast operations

### LINKED ISSUE #281: Converting bool to int doesn't work
Adding an explicit `int` coercion to a boolean value currently produces an invalid hugr, with the error:
```
TypeArgMismatch(WrongNumberArgs(1, 0))
```
This is to do with the fact that the hugr "conversion" extension creates an int with width 1, whereas all ints in guppy have width 64. We need to widen the ints that come back!

### LINKED ISSUE #409: Support `ifrombool` / `itobool` for arbitrary ints
Hugr's `std.int.ifrombool` / `itobool` ops only work with bits. Guppy should emit `>0` and `iwiden` operations respectively when using `bool` / `int` with wider ints.

aborgna-q:
* Closes #281
* Closes #409

### DIFF
--- a/guppylang/prelude/_internal/compiler/arithmetic.py
+++ b/guppylang/prelude/_internal/compiler/arithmetic.py
@@ -2,7 +2,7 @@
 
 from collections.abc import Sequence
 
-import hugr
+import hugr.std.int
 from hugr import Wire, ops
 from hugr import tys as ht
 from hugr.std.float import FLOAT_T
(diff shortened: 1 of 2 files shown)


qartik:
This might help resolve #264 as well. 

Question: Why use `I` to refer to int when `Bool` is fully spelled out in `IToBool` etc?

aborgna-q:
> Looks good, but we need some tests. Happy to approve if those tests are somewhere else

We do have tests! Before they emitted a black box (`guppylang.unsupported`). When I started emiting prelude operations the validation started failing on the ones that had the wrong signature vs what's declared in hugr's extension.

> Question: Why use I to refer to int when Bool is fully spelled out in IToBool etc?

All the int-related operations start with `i`; `iadd`, `ieq`, `ifromusize`, ...

PR #461, opened 2024-09-05
feat!: Add implicit importing of modules

### LINKED ISSUE #133: Implicit module imports

### LINKED ISSUE #462: Allow importing of implicit modules

mark-koch:
Closes #133 and closes #462.

* When using implicit modules, automatically import definitions by inspecting the `__dict__` of the containing Python module.
* Implicit modules are no longer discarded once they are compiled for the first time.
* Updated examples to use implicit modules

BREAKING CHANGE: `guppy.take_module` renamed to `guppy.get_module` and no longer removes the module from the state.


guppylang/decorator.py
```diff
@@ -291,23 +291,32 @@ def load(self, m: ModuleType | GuppyModule) -> None:
         module = self._modules[caller]
         module.load_all(m)
 
-    def take_module(self, id: ModuleIdentifier | None = None) -> GuppyModule:
```
aborgna-q, line 294: This is a breaking change. I imagine no-one is calling this directly, but we may want to add a `@deprecated` alias `0.10.1` ?

aborgna-q (approved):
Nice!

PR #467, opened 2024-09-09
feat!: Use angle type in quantum operations

### LINKED ISSUE #240: Use tket2 angle type for quantum gate parameters

mark-koch:
Closes #240


examples/random_walk_qpe.py
```diff
@@ -68,10 +70,10 @@ def random_walk_phase_estimation(
 def example_controlled_oracle(q1: qubit, q2: qubit, t: float) -> tuple[qubit, qubit]:
```
ss2165, line 70: should this function take an angle?

mark-koch: Depends how you interpret things. I always think of the `t` in `e^(-itH)` as time so I kept float here. But happy to change the angle if you prefer (will require more explicit angle to float and vice versa conversions though)

ss2165: never mind

ss2165 (approved):

PR #468, opened 2024-09-09
test: cargo build validator in CI

acl-cqc:
Many hugr-building tests are showing up as `s` == skipped after #455. We need to `cargo build` the validator first.

Also "fix" (perhaps hack) `run_int_fn` to extract the first `module` Hugr from the Package coming back from compilation.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -63,6 +63,9 @@ jobs:
       - name: Install Guppy with llvm-based execution
         run: uv sync --extra execution
 
+      - name: Cargo build validator
+        run: cargo build -p validator --release
+
       - name: Run tests
         run: uv run pytest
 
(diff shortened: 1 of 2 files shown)


mark-koch (approved):
Thanks!

> I had to fix run_int_fn somehow to cope with that it now receives a Package not a Hugr. However, this is very much not a complete fix - perhaps it is even, dare I say it, a hack - but I guess one might argue it works fine for the kinds of function for which we expect run_int_fn to work anyway.

Yes, I think this is fine for now

> We should probably have an issue for this, nonetheless.

Feel free to create one 👍

acl-cqc:
> > We should probably have an issue for this, nonetheless.
> 
> Feel free to create one 👍

I think I reasoned that anyone wanting `run_int_fn` to do anything else....would encounter the same problem, and have to deal with it then. `run_int_fn` already has no docs of what it expects/contracts to do...

However, the bigger issue that playing around here reveals - is that the type annotations here (i.e. in `tests/`) are not checked; we never run `mypy`. So the only way a type annotation results in an error is at execution (e.g. pytest) time, when an annotation using a type that doesn't exist, fails to load the type. Do we plan to enable them? I guess not, since a lot of test code is `@guppy` ?
(If so, we could provide a custom Protocol for the return value of the `run_int_fn` fixture that would have detected this problem, say.)

PR #472, opened 2024-09-10
test: only skip validation tests when given explicit flag

acl-cqc:
pytest previously skipped the validation tests if the validator was not built, leading to the situation in #468 where many tests were accidentally skipped in CI.

Instead, fail the tests if the validator is not installed, but allow `pytest --no_validation` to skip them instead.

The possibility of doing similar for execution tests (somewhat-silently skipped if `execute-llvm` isn't installed) is left for a later PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,8 +96,10 @@ uv run pre-commit install
 Run tests using
 
 ```sh
-uv run pytest -v
+uv run pytest [-v]  # -v just enables verbose test output
 ```
+
+(If you have not build the validator, you can `uv run pytest --no_validation`)
 
 You have to install extra dependencies to test automatic circuit conversion from `pytket`.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -96,9 +96,11 @@ uv run pre-commit install
 Run tests using
 
 ```sh
-uv run pytest -v
+uv run pytest [-v]  # -v just enables verbose test output
 ```
 
+(If you have not build the validator, you can `uv run pytest --no_validation`)
```
qartik, line 102:
```suggestion
(If you have not built the validator, you can do `uv run pytest --no_validation`.)
```

qartik (approved):
LGTM, left a minor copy suggestion.

qartik:
> This raises the obvious question, what about execution tests, which are somewhat-silently skipped if `execute-llvm` isn't installed.

I see the CI does run execution tests with `Run uv sync --extra execution --extra pytket`, so perhaps not a major concern?

PR #485, opened 2024-09-11
fix: `angle` is now a struct and emitted as a rotation 

ss2165:
update extension definitions and lower angle type to rotation.


ss2165:
should be brought up to date with https://github.com/CQCL/tket2/pull/611 before merging

guppylang/prelude/angles.py
```diff
@@ -2,99 +2,70 @@
...
     @guppy(angles)
     @no_type_check
-    def log_denominator(self: "angle") -> nat:
-        _, log_denominator = self._parts()
-        return log_denominator
+    def __eq__(self: "angle", other: "angle") -> bool:
+        return self.halfturns == other.halfturns
+
+
+pi = guppy.constant(angles, "pi", ty="angle", value=hv.Tuple(FloatVal(math.pi)))
```
mark-koch, line 71:
```suggestion
pi = guppy.constant(angles, "pi", ty="angle", value=hv.Tuple(FloatVal(1.0)))
```

guppylang/prelude/_internal/json_defs/tket2/rotation.json
```diff
@@ -0,0 +1,126 @@
...
+          ],
+          "extension_reqs": []
+        }
+      },
+      "binary": false
+    },
+    "radd": {
+      "extension": "tket2.rotation",
+      "name": "radd",
+      "description": "Add two angles together (experimental).",
```
acl-cqc, line 61:
```suggestion
      "description": "Add two rotations together (experimental).",
```

guppylang/prelude/angles.py
```diff
@@ -1,96 +1,68 @@
...
-
-    @guppy.custom(AngleOpCompiler("aadd"))
-    def __add__(self: "angle", other: "angle") -> "angle": ...
-
-    @guppy.custom(AngleOpCompiler("asub"))
-    def __sub__(self: "angle", other: "angle") -> "angle": ...
-
-    @guppy.custom(AngleOpCompiler("aneg"))
-    def __neg__(self: "angle") -> "angle": ...
+    halfturns: float
```
acl-cqc, line 20: Doc-comment that this (is misnamed) is not an angle in the truest sense but a rotation, i.e. does not identify with itself modulo 2pi (/2halfturns)

tests/integration/test_arithmetic.py
```diff
@@ -112,7 +112,7 @@ def main(a1: angle, a2: angle) -> bool:
         a3 = -a1 + a2 * -3
         a3 -= a1
         a3 += 2 * a1
-        return a3 / 3 == -a2
+        return a3 == -a2
```
acl-cqc, line 115: Isn't the idea of this being a rotation rather than a true angle, that you *can* divide by 3? So we can keep the previous test?

tests/integration/conftest.py
```diff
@@ -67,17 +68,17 @@ class LLVMException(Exception):
...
-    def f(hugr: Package, expected: int, fn_name: str = "main"):
+def _run_fn(run_fn_name: str):
+    def f(hugr: Package, expected: Any, fn_name: str = "main"):
         try:
             import execute_llvm
 
-            if not hasattr(execute_llvm, "run_int_function"):
+            fn = getattr(execute_llvm, run_fn_name)
+            if not fn:
                 pytest.skip("Skipping llvm execution")
```
acl-cqc, line 78:
```suggestion
                pytest.skip(f"Skipping llvm execution, {run_fn_name} not found")
```

Note we've already checked for `execute_llvm` not being imported, so we don't really expect this to fail (do we?)

tests/integration/test_arithmetic.py
acl-cqc: after all, we have the new one below too

acl-cqc: this isn't an execution test, though. Might be better if it were!

tests/integration/test_arithmetic.py
```diff
@@ -230,3 +230,21 @@ def run_rem() -> int:
...
+        a1 = pi
+        a2 = pi * 2
+        a3 = -a1 + a2 * -3
+        a3 -= a1
+        a3 += 2 * a1
+        return float(a3)
+    hugr = module.compile()
+    validate(hugr)
+    import math
+    run_float_fn(hugr, expected=-6 * math.pi)
```
acl-cqc, line 250: Hmmmm, the worrying thing here is that `expected==`. We appear to be performing exact equality on floats, and where the expected value uses all the precision, to boot....

I think `pytest` has `assert_close`, maybe `run_float_fn` should just return the `float` (even if we keep `run_int_fn` as is to avoid a major refactor)

guppylang/prelude/_internal/json_defs/tket2/rotation.json
acl-cqc: Done in #506

guppylang/prelude/angles.py
acl-cqc: Done in #506

tests/integration/test_arithmetic.py
acl-cqc: Change reverted, have not made into an execution test or anything

tests/integration/test_arithmetic.py
acl-cqc: Added `pytest.approx` here

acl-cqc (approved):

PR #479, opened 2024-09-11
chore: Small error print improvements

aborgna-q:
Implement `__str__` for `DefId`, so it's easier to see the module name

before:
```py
DefId(id=298, module=<guppylang.module.GuppyModule object at 0x109e18110>)
```
after:
```py
DefId(298, quantum)
```

drive-by: Update uv.lock with the latest release version

PR #486, opened 2024-09-12
feat!: Make linear types @inout by default; add @owned annotation

### LINKED ISSUE #413: Make linear function arguments `@inout` by default

croyzor:
Resolves #413 and #414.

BREAKING CHANGE: Linear function arguments are now borrowed by default; removed the now redundant `@inout` annotation


guppylang/prelude/quantum_functional.py
```diff
@@ -0,0 +1,108 @@
...
+
+@guppy.hugr_op(quantum, quantum_op("ZZMax", ext=HSERIES_EXTENSION))
+def zz_max(q1: qubit @ owned, q2: qubit @ owned) -> tuple[qubit, qubit]: ...
+
+
+@guppy.hugr_op(quantum, quantum_op("Rz"))
+def rz(q: qubit @ owned, angle: angle) -> qubit: ...
+
+
+@guppy.hugr_op(quantum, quantum_op("Rx"))
```
croyzor, line 63: N.B. This is broken in main. Both `rz` and `rx` are annoted with `quantum_op("Rz")`

guppylang/checker/func_checker.py
```diff
@@ -163,7 +163,9 @@ def check_signature(func_def: ast.FunctionDef, globals: Globals) -> FunctionType
     )
 
 
-def parse_docstring(func_ast: ast.FunctionDef) -> tuple[ast.FunctionDef, str | None]:
+def parse_function_with_docstring(
```
croyzor, line 166: Drive-by: I updated the name of this function because I found it too confusing

guppylang/tys/parsing.py
```diff
@@ -82,11 +82,11 @@ def _try_parse_defn(node: AstNode, globals: Globals) -> Definition | None:
     match node:
         case ast.Name(id=x):
             if x not in globals:
-                raise GuppyError("Unknown identifier", node)
+                raise GuppyError(f"Unknown identifier: {x}", node)
```
croyzor, line 85: Drive-by: Printing the unknown identifiers here and on line 89

tests/error/linear_errors/copy_qubit.py
```diff
@@ -1,6 +1,7 @@
 import guppylang.prelude.quantum as quantum
 from guppylang.decorator import guppy
 from guppylang.module import GuppyModule
+from guppylang.prelude.builtins import owned
```
croyzor, line 4: I moved this file from `copy.py` to `copy_qubit.py` because of some weird case where it was being imported by the guppy prelude instead of the python `copy` library. Not 100% sure what was going on there but I don't want to touch it with a 10ft pole.

guppylang/tys/parsing.py
```diff
@@ -82,11 +82,11 @@ def _try_parse_defn(node: AstNode, globals: Globals) -> Definition | None:
...
             return globals[x]
         case ast.Attribute(value=ast.Name(id=module_name) as value, attr=x):
             if module_name not in globals:
-                raise GuppyError("Unknown identifier", value)
+                raise GuppyError(f"Unknown identifier: {module_name}.{x}", value)
```
mark-koch:
```suggestion
                raise GuppyError(f"Unknown identifier: `{module_name}`", value)
```

guppylang/prelude/builtins.py
```diff
@@ -74,7 +74,14 @@ def __rmatmul__(self, other: Any) -> Any:
         return other
 
 
-inout = _Inout()
```
mark-koch, line 77: We should also remove the `_Inout` class

guppylang/tys/parsing.py
aborgna-q: See #479. It also makes the printed id show the module name instead of the object repr.

guppylang/tys/parsing.py
```diff
@@ -220,18 +223,18 @@ def type_with_flags_from_ast(
...
                 if not ty.linear:
                     raise GuppyError(
-                        f"Non-linear type `{ty}` cannot be annotated as `@inout`",
+                        f"Non-linear type `{ty}` cannot be annotated as `@owned`",
                         node.right,
                     )
-                flags |= InputFlags.Inout
+                flags |= InputFlags.Owned
+            case ast.Name(name):
+                raise GuppyError(f"Invalid annotation: {name}", node.right)
```
mark-koch, line 237:
```suggestion
                raise GuppyError(f"Invalid annotation: `{name}`", node.right)
```

guppylang/tys/parsing.py
```diff
@@ -252,8 +256,11 @@ def type_from_ast(
 ) -> Type:
     """Turns an AST expression into a Guppy type."""
     ty, flags = type_with_flags_from_ast(node, globals, param_var_mapping)
-    if flags != InputFlags.NoFlags:
-        raise GuppyError("`@` type annotations are not allowed in this position", node)
+    if flags != InputFlags.NoFlags and flags != InputFlags.Inout:
```
mark-koch, line 259: Why exclude `InputFlags.Inout` here?

pyproject.toml
```diff
@@ -65,6 +65,8 @@ build-backend = "hatchling.build"
 plugins = ["pydantic.mypy"]
 strict = true
 allow_redefinition = true
+disable_error_code = ['valid-type' # Otherwise mypy will complain about @owned annotations
+                     ]
```
mark-koch, line 69: I think we shouldn't disable this globally, only for files that actually contain Guppy code

tests/error/comprehension_errors/used_twice2.err
```diff
@@ -1,7 +1,7 @@
 Guppy compilation failed. Error in file $FILE:18
 
 16:    @guppy(module)
-17:    def foo(qs: linst[qubit]) -> linst[qubit]:
+17:    def foo(qs: linst[qubit] @owned) -> linst[qubit]:
 18:        return [q for q in qs if bar(q)]
-                   ^
-GuppyError: Variable `q` with linear type `qubit` was already used (at 18:33)
+                                        ^
+GuppyTypeError: Variable `q` with linear type `qubit` is not used on all control-flow paths of the list comprehension
```
mark-koch, line 7: You need to annotate `bar` as `owned` as well to reproduce the old error

tests/error/linear_errors/__init__.err
mark-koch, line 1: Can you delete this file? Same for `__pycache__.err`

tests/error/poly_errors/non_linear2.err
```diff
@@ -3,5 +3,5 @@ Guppy compilation failed. Error in file $FILE:23
 21:    @guppy(module)
 22:    def main() -> None:
 23:        foo(h)
-           ^^^^^^
-GuppyTypeError: Cannot instantiate non-linear type variable `T` in type `forall T. (T -> T) -> None` with linear type `qubit`
+               ^
+GuppyTypeError: Expected argument of type `?T -> ?T`, got `qubit -> qubit`
```
mark-koch, line 7: Why did this error get worse?

croyzor: I think it's failing comparing `T` with `NoFlags` against `qubit` with `Inout` before it can get to the linearity check. I'll add a better error message for these flag mismatches

guppylang/tys/parsing.py
croyzor: If a linear variable has `InputFlags.Inout`, it's because we set it - it can't have been done via an annotation from the user

tests/error/poly_errors/non_linear2.err
croyzor: BUT the case I was thinking of in my first comment _also_ happens. I've just added `poly_errors/non_linear3.py` to demonstrate this. `@inout` flags are silent during printing because they're not user-facing, but they still mess up unification. I'll add an issue for this

croyzor: Added #487

croyzor:
> I think you also need to update the autogenerated constructors for structs (see `guppylang/definition/structs.py`) to take ownership of linear arguments. Surprising that this wasn't caught by the tests :/

I've added a test and the logic for this, but it seems like it was working by accident! Something's a bit fishy there...

qartik:
@croyzor could you also resolve the conflicts that github is pointing out in `tests/integration/test_array.py`?

guppylang/prelude/builtins.py
croyzor: Removed

guppylang/tys/parsing.py
croyzor: Updated to an assert

pyproject.toml
croyzor: Removed and readded locally to the definitions in quantum.py and quantum_functional.py

tests/error/comprehension_errors/used_twice2.err
croyzor: Good spot! Updated

tests/error/linear_errors/__init__.err
croyzor: Done

guppylang/prelude/quantum.py
```diff
@@ -29,12 +29,12 @@ def dirty_qubit() -> qubit: ...
 
 
 @guppy.hugr_op(quantum, quantum_op("Measure"))
-def measure_return(q: qubit @ owned) -> tuple[qubit, bool]: ...
+def measure_return(q: qubit @ owned) -> tuple[qubit, bool]: ...  # type: ignore[valid-type]
```
mark-koch, line 32: You could also add `valid-type` to the `mypy: disable-error-code` pragma at the start of the file.

Same in `quantum_functional.py`

guppylang/tys/ty.py
```diff
@@ -302,8 +302,7 @@ def transform(self, transformer: Transformer) -> "Type":
 class InputFlags(Flag):
     """Flags that can be set on inputs of function types.
 
-    Currently, the only supported flag is `Inout`. In the future, we could add
-    additional flags like `Frozen`, `Owned`, etc.
+    In the future, we could add  additional flags like `Frozen`, `Owned`, etc.
```
mark-koch, line 305:
```suggestion
    In the future, we could add  additional flags like `Frozen`.
```

guppylang/checker/linearity_checker.py
```diff
@@ -134,13 +134,13 @@ def new_scope(self) -> Generator[Scope, None, None]:
...
-        # them in another @inout position of a function call. In that case, our
+        # Usage of borrowed variables is generally forbidden. The only exception is
+        # letting them be borrowed by another  function call. In that case, our
         # `_visit_call_args` helper will set `is_inout_arg=True`.
         if is_inout_var(node.place) and not is_inout_arg:
             raise GuppyError(
-                f"{node.place.describe} may only be used in an `@inout` position since "
-                "it is annotated as `@inout`. Consider removing the annotation to get "
+                f"{node.place.describe} may not be used in an `@owned` position since "
+                "it lacks an `@owned` annotation. Consider adding `@owned` to get "
```
mark-koch, line 143:
```suggestion
                "it isn't owned. Consider adding a `@owned` annotation to get "
```
?

guppylang/tys/parsing.py
```diff
@@ -256,7 +256,8 @@ def type_from_ast(
 ) -> Type:
     """Turns an AST expression into a Guppy type."""
     ty, flags = type_with_flags_from_ast(node, globals, param_var_mapping)
-    if flags != InputFlags.NoFlags and flags != InputFlags.Inout:
+    if flags != InputFlags.NoFlags:
+        assert flags != InputFlags.Inout  # Users shouldn't be able to set this
```
mark-koch, line 260:
```suggestion
        assert InputFlags.Inout not in flags  # Users shouldn't be able to set this
```

mark-koch (approved):

PR #496, opened 2024-09-13
feat: Only lower definitions to Hugr if they are used

### LINKED ISSUE #434: Only lower definitions to Hugr that are actually called

### LINKED ISSUE #470: Tests that inspect Hugrs `xfail`ed while we compile stdlib into the same Hugr
Since #452, `builtins.py` compiles to functions and nodes, currently in the same Hugr as everything else. Tests that look for nodes in the Hugr are confused by this, so are `pytest.xfail`ed with the string `hugr-includes-whole-stdlib`. At time of writing there are four (this may change!):
* `test_array.py`: [test_len](https://github.com/CQCL/guppylang/blob/d05f369041edbb10772117161d3a74414769361d/tests/integration/test_array.py#L14)
* `test_basic.py`: [test_func_def_name](https://github.com/CQCL/guppylang/blob/d05f369041edbb10772117161d3a74414769361d/tests/integration/test_basic.py#L67)
* `test_extern.py`: [test_extern_float](https://github.com/CQCL/guppylang/blob/d05f369041edbb10772117161d3a74414769361d/tests/integration/test_extern.py#L8) and [test_extern_alt_symbol](https://github.com/CQCL/guppylang/blob/d05f369041edbb10772117161d3a74414769361d/tests/integration/test_extern.py#L29)

We could possibly rewrite the tests to look for *any* node with the desired property, although that slightly weakens the test. The ideal solution is to wait for Hugr linking, and then compile the stdlib into a different module (Hugr) in the same package. An intermediate solution would be to add metadata to every FuncDefn or even every node, identifying the source file from which it was produced; the tests could then filter out the stdlib restoring the test that there should be *only one* node produced of the specific type.

mark-koch:
Closes  #434 and closes #470.


guppylang/compiler/core.py
```diff
@@ -3,17 +3,56 @@
...
+from guppylang.definition.common import CheckedDef, CompilableDef, CompiledDef, DefId
 from guppylang.error import InternalGuppyError
 from guppylang.tys.ty import StructType
 
-CompiledGlobals = dict[DefId, CompiledDef]
+# CompiledGlobals = dict[DefId, CompiledDef]
 CompiledLocals = dict[PlaceId, Wire]
 
 
+class CompiledGlobals:
```
mark-koch, line 17: We should probably rename this, considering the class does way more than just holding the globals. See #497

guppylang/compiler/core.py
```diff
@@ -3,17 +3,55 @@
...
 
-CompiledGlobals = dict[DefId, CompiledDef]
 CompiledLocals = dict[PlaceId, Wire]
 
 
+class CompiledGlobals:
+    """Compilation context containing all available definitions.
+
+    This context drives the Hugr lowering by keeping track of which definitions are
+    used when lowering the definitions provided via the `required` keyword. The
```
acl-cqc: I dropped the `required` keyword, as we already have other ways to add to the worklist (and `compile_outer`) from outside (e.g. for nested functions) - `__getitem__` does basically the same thing

guppylang/compiler/core.py
```diff
@@ -3,17 +3,55 @@
...
+
+    def __getitem__(self, def_id: DefId) -> CompiledDef:
+        if def_id not in self.compiled:
+            defn = self.checked[def_id]
+            self.compiled[def_id] = self._compile(defn)
+            self.worklist.add(def_id)
+        return self.compiled[def_id]
+
+    def _compile(self, defn: CheckedDef) -> CompiledDef:
+        if isinstance(defn, CompilableDef):
```
acl-cqc, line 47: For another PR, I reckon we could probably get rid of `CompilableDef` by giving `CheckedDef`s an implementation of `compile_outer` that just returns self

guppylang/module.py
```diff
@@ -325,19 +326,19 @@ def compile(self) -> Package:
             return self._compiled_hugr
 
         self.check()
+        checked_defs = self._imported_checked_defs | self._checked_defs
```
acl-cqc, line 315: I removed `def _compile_defs` as the old code (removed here) was the only place calling it

acl-cqc:
Fixing #470 is a nice side-effect, not the right reason to do this, but this is good anyway :)

acl-cqc:
> Note I deleted the unused Definition.compile

This was actually mentioned in the PR description:
> Also adds a `compile` function to every definition that is useful when working with implicit modules: Instead of `guppy.compile_module()`, user can now do `main.compile()`.
....but given there are no uses in tests, I think that's fair.

acl-cqc:
> > Note I deleted the unused Definition.compile
> 
> This was actually mentioned in the PR description:
> ....but given there are no uses in tests, I think that's fair, can become another PR with usage examples easily enough

#504

PR #501, opened 2024-09-15
feat!: Hide lists and function tensors behind experimental flag

### LINKED ISSUE #437: Add switch to toggle experimental features
We should turn off by default:
* Lists
* Function tensors?

mark-koch:
Closes #437.

BREAKING CHANGE: Lists and function tensors are no longer available by default. `guppylang.enable_experimental_features()` must be called before compilation to enable them.

### DIFF
--- a/guppylang/checker/expr_checker.py
+++ b/guppylang/checker/expr_checker.py
@@ -27,6 +27,7 @@ from contextlib import suppress
 from dataclasses import replace
 from typing import Any, NoReturn, cast
 
+import guppylang
 from guppylang.ast_util import (
     AstNode,
     AstVisitor,
@@ -78,6 +79,7 @@ from guppylang.nodes import (
 from guppylang.tys.arg import TypeArg
 from guppylang.tys.builtin import (
     bool_type,
+    check_lists_enabled,
     get_element_type,
     is_bool_type,
     is_linst_type,
@@ -214,6 +216,7 @@ class ExprChecker(AstVisitor[tuple[ast.expr, Subst]]):
         return node, subst
 
     def visit_List(self, node: ast.List, ty: Type) -> tuple[ast.expr, Subst]:
+        check_lists_enabled(node)
         if not is_list_type(ty) and not is_linst_type(ty):
             return self._fail(ty, node)
         el_ty = get_element_type(ty)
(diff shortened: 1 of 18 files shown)


guppylang/checker/expr_checker.py
```diff
@@ -990,6 +996,15 @@ def instantiate_poly(node: ast.expr, ty: FunctionType, inst: Inst) -> ast.expr:
     return with_type(ty, node)
 
 
+def check_function_tensors_enabled(node: ast.expr | None = None) -> None:
```
acl-cqc, line 999: nit: I wonder if the two `check_xxx_enabled` should live together in `experimental.py`

guppylang/checker/expr_checker.py
```diff
@@ -990,6 +996,15 @@ def instantiate_poly(node: ast.expr, ty: FunctionType, inst: Inst) -> ast.expr:
...
+    if not guppylang.experimental.EXPERIMENTAL_FEATURES_ENABLED:
+        raise GuppyError(
+            "Function tensors are an experimental feature. Call "
```
acl-cqc, line 1002: Maybe `Use` would be better than `Call` ?

guppylang/tys/builtin.py
```diff
@@ -118,6 +120,30 @@ def check_instantiate(
...
+        self, args: Sequence[Argument], globals: "Globals", loc: AstNode | None = None
+    ) -> OpaqueType:
+        check_lists_enabled(loc)
+        return super().check_instantiate(args, globals, loc)
+
+
+def check_lists_enabled(loc: AstNode | None = None) -> None:
+    if not guppylang.experimental.EXPERIMENTAL_FEATURES_ENABLED:
+        raise GuppyError(
+            "Lists are an experimental feature and not fully supported yet. Call "
```
acl-cqc, line 141: ditto*2 (`Call` -> `Use`, move into `experimental.py` ?)

acl-cqc (approved):
Generally looks good to me, a couple of suggestions...

acl-cqc:
Changes enacted.

I think even "experimental features" would be good to have tests but this is better than simply not feature-flagging them ;) so merging...

PR #505, opened 2024-09-17
ci: optional tests fail in CI

doug-q:


tests/integration/conftest.py
```diff
@@ -32,9 +35,12 @@ def validate(request, export_test_cases_dir: Path):
...
         validator = get_validator()
         if validator is None:
-            pytest.fail("Run `cargo build -p release` to install the validator")
-    else:
-        pytest.skip("Skipping validation tests as requested")
+            if IN_CI:
+                pytest.fail("Validator not installed")
+            else:
+                pytest.skip(
+                    "Skipping validation: Run `cargo build` to install the validator"
```
acl-cqc, line 42: You can also just run `pytest --no-validation`, following #472. So this is a different take on the same idea. If we go with this, we might want to revert #472? Or, we might instead want to do something like that but for execution tests too?

tests/integration/conftest.py
```diff
@@ -74,15 +80,20 @@ def f(hugr: Package, expected: int, fn_name: str = "main"):
...
             if res != expected:
                 raise LLVMException(
                     f"Expected value ({expected}) doesn't match actual value ({res})"
                 )
-        except ImportError:
-            pytest.skip("Skipping llvm execution")
-
+        except ImportError as e:
+            if IN_CI:
+                pytest.fail(f"run_int_fn failed in CI: {e}")
```
croyzor, line 96: This should be something like "execute_llvm not installed in CI"

PR #506, opened 2024-09-17
fix: Proposed review changes to #485

acl-cqc:


tests/integration/conftest.py
```diff
@@ -92,7 +88,14 @@ def f(hugr: Package, expected: Any, fn_name: str = "main"):
 
 @pytest.fixture
 def run_int_fn():
-    return _run_fn("run_int_function")
+    run_to_res = _run_fn("run_int_function")
```
acl-cqc, line 91: The awkward thing is that this makes `run_int_fn` (the same as before but) different from `run_float_fn`, so perhaps they should be named differently. `check_int_fn` perhaps, although it'd be less change to rename the new `run_float_fn` instead....thoughts?

tests/integration/test_arithmetic.py
```diff
@@ -112,7 +112,7 @@ def main(a1: angle, a2: angle) -> bool:
         a3 = -a1 + a2 * -3
         a3 -= a1
         a3 += 2 * a1
-        return a3 == -a2
+        return a3 / 3 == -a2
```
acl-cqc, line 115: This is a straightforward reversion of an unnecessary change. (The code is not executed, the result does not have to be `true`, and division is valid.)

PR #509, opened 2024-09-18
chore: doug/add-doug-q-to-codeowners

doug-q:


cqc-alec (approved):

acl-cqc (approved):
So it's both ridiculous that a "CODEOWNER" is not required to add oneself to CODEOWNERS....and also that Alec's approval is not sufficient but mine is because (IIUC) review was requested from me not from Alec?!?!

But anyway, since all codeowners are simultaneously on holiday, desperate measures are required, let's go!

Issue #484, opened 2024-09-11
Add missing tket2.quantum operations to quantum module


ss2165:
see Tk2ops, we should add missing (I can see at least Ry, CRz, Toffoli missing)

_Originally posted by @ss2165 in https://github.com/CQCL/guppylang/pull/483#discussion_r1754795839_

jake-arkinstall:
Any chance of rushing this in on Monday? Could do with it for the notebooks (specifically crz, possibly others)

jake-arkinstall:
I'll give it a go now

PR #510, opened 2024-09-23
feat: Add missing tket2 quantum gates, generalise RotationCompile

jake-arkinstall:
Addresses #484

### DIFF
--- a/guppylang/prelude/_internal/compiler/quantum.py
+++ b/guppylang/prelude/_internal/compiler/quantum.py
@@ -83,7 +83,10 @@ class RotationCompiler(CustomInoutCallCompiler):
 
         qs = self.builder.add_op(
             quantum_op(self.opname)(
-                ht.FunctionType([ht.Qubit for _ in qs] + [ROTATION_T], [ht.Qubit for _ in qs]), []
+                ht.FunctionType(
+                    [ht.Qubit for _ in qs] + [ROTATION_T], [ht.Qubit for _ in qs]
+                ),
+                [],
             ),
             *qs,
             conditional,
--- a/guppylang/prelude/_internal/compiler/quantum.py
+++ b/guppylang/prelude/_internal/compiler/quantum.py
@@ -91,4 +91,4 @@ class RotationCompiler(CustomInoutCallCompiler):
             *qs,
             conditional,
         )
-        return CallReturnWires(regular_returns=[], inout_returns=list(*qs))
+        return CallReturnWires(regular_returns=[], inout_returns=list(qs))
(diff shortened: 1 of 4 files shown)


guppylang/prelude/_internal/compiler/quantum.py
```diff
@@ -81,11 +81,11 @@ def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
...
             quantum_op(self.opname)(
-                ht.FunctionType([ht.Qubit, ROTATION_T], [ht.Qubit]), []
+                ht.FunctionType([ht.Qubit for _ in qs] + [ROTATION_T], [ht.Qubit for _ in qs]), []
             ),
-            q,
+            *qs,
             conditional,
         )
-        return CallReturnWires(regular_returns=[], inout_returns=[q])
+        return CallReturnWires(regular_returns=[], inout_returns=qs)
```
mark-koch, line 91:
```suggestion
        return CallReturnWires(regular_returns=[], inout_returns=list(*qs))
```

Something like this should fix the mypy error

guppylang/prelude/_internal/compiler/quantum.py
```diff
@@ -70,7 +70,7 @@ def __init__(self, opname: str):
     def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
         from guppylang.prelude._internal.util import quantum_op
 
-        [q, angle] = args
+        qs, angle = args[:-1], args[-1]
```
mark-koch, line 73:
```suggestion
        [*qs, angle] = args
```

guppylang/prelude/quantum.py
```diff
@@ -35,6 +36,10 @@ def h(q: qubit) -> None: ...
 def cz(control: qubit, target: qubit) -> None: ...
 
 
+@guppy.hugr_op(quantum_op("CY"))
+def cy(control: qubit, target: qubit) -> None: ...
```
mark-koch, line 40: Can you also add these functions to the `quantum_functional` module?

guppylang/prelude/_internal/compiler/quantum.py
jake-arkinstall: This gets me:
```
>       return CallReturnWires(regular_returns=[], inout_returns=list(*qs))
E       TypeError: 'OutPort' object is not iterable
```

Note I am assigning qs to the return of self.builder.add_op - I probably need advice on that.

mark-koch: Oops I think it should be `list(qs)` then

PR #516, opened 2024-10-01
fix: Fix printing of generic function parameters

### LINKED ISSUE #482: bug: Printing functions with multiple parameters is wrong
In the printing code, it looks like we append the parameters for a function onto the end of `bound_vars`, but then when the visitor reaches a type param, it prints `bound_vars[-idx-1]`. This would be right if the param `idx` referred to a de Bruijn _index_, but IIUC it actually refers to a de Bruijn _level_.

This can be evidenced by printing off the signature of an array operation, such as `__setitem__`, which yields:
```
forall n, L: nat. (array[L, n] @inout, int, L) -> None
```
the `n` should have kind `nat`, not `L`!

mark-koch:
Fixes #482

### DIFF
--- a/guppylang/tys/param.py
+++ b/guppylang/tys/param.py
@@ -110,7 +110,7 @@ class TypeParam(ParameterBase):
         var = ExistentialTypeVar.fresh(self.name, self.can_be_linear)
         return TypeArg(var), var
 
-    def to_bound(self, idx: int | None = None) -> Argument:
+    def to_bound(self, idx: int | None = None) -> TypeArg:
         """Creates a bound variable with a given index that can be instantiated for this
         parameter.
         """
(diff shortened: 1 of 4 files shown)


guppylang/tys/param.py
```diff
@@ -110,7 +110,7 @@ def to_existential(self) -> tuple[Argument, ExistentialVar]:
         var = ExistentialTypeVar.fresh(self.name, self.can_be_linear)
         return TypeArg(var), var
 
-    def to_bound(self, idx: int | None = None) -> Argument:
+    def to_bound(self, idx: int | None = None) -> TypeArg:
```
mark-koch, line 113: Drive by: Use a more precise return type which is used in the test below

tests/error/array_errors/linear_len.err
```diff
@@ -4,4 +4,4 @@ Guppy compilation failed. Error in file $FILE:14
 13:    def main(qs: array[qubit, 42]) -> int:
 14:        return len(qs)
                   ^^^^^^^
-GuppyTypeError: Cannot instantiate non-linear type variable `T` in type `forall n, T: nat. array[T, n] -> int` with linear type `qubit`
+GuppyTypeError: Cannot instantiate non-linear type variable `T` in type `forall T, n: nat. array[T, n] -> int` with linear type `qubit`
```
mark-koch, line 7: This was actually wrong in the golden error and no one noticed 😅

croyzor: 🫠

Issue #528, opened 2024-10-02
Add missing list methods


mark-koch:
* `__add__`, `__radd__` (and `__iadd__` following #527)
* `__bool__`
* `__mul__`, `__rmul__` (and `__imul__` following #527)
* `__contains__`
* `clear`
* `copy`
* `count`
* `extend`
* `index`
* `remove`
* `reverse`
* `sort`

Most of these should be implementable via Guppy source following #522

nicolaassolini-qntm:
When solving this issue, we should remove 2 skips in `test_list`

Issue #487, opened 2024-09-12
Report errors where variables are instantiated as an `@inout` type


croyzor:
After #486, linear types get the `Inout` flag by default, but this is hidden to the user. This means the function:
```
def h(q: qubit) -> None: ...
```
has signature `qubit @inout -> None`, but will print `qubit -> None`. The problem arises when we try to unify this type. We can end up with an error message saying that unification failed comparing
```
qubit -> None
```
against
```
?T -> None
```
(see `tests/error/poly_errors/non_linear3.py`)
which is very opaque. We need a custom error message saying when unification fails by trying to instantiate a type variable to a borrowed type

mark-koch:
Fixed in #531

PR #531, opened 2024-10-02
feat!: Allow linear data inside lists

### LINKED ISSUE #524: Use inout for lists

### LINKED ISSUE #530: Allow linear types inside lists

mark-koch:
Closes #530 and closes #524.

* Changes the signature of list methods to allow linear list elements (implicitly using the inout feature).
* Removes methods that are not supported for now (see #528)
* Small change to the unification logic to be a bit laxer about requiring matching `@owned` flags if one of the arguments is not linear (otherwise we couldn't unify `L @owned` with any classical type). We'll need to revisit that once we have flags that can be put on classical types.

BREAKING CHANGE: Unsupported list methods have been removed.


tests/error/poly_errors/non_linear2.err
```diff
@@ -3,5 +3,5 @@ Guppy compilation failed. Error in file $FILE:23
 21:    @guppy(module)
 22:    def main() -> None:
 23:        foo(h)
-               ^
-GuppyTypeError: Expected argument of type `?T -> ?T`, got `qubit @owned -> qubit`
+           ^^^^^^
+GuppyTypeError: Cannot instantiate non-linear type variable `T` in type `forall T. (T -> T) -> None` with linear type `qubit`
```
mark-koch, line 7: A side effect of the unification change is that we get a better error message here.

However, I think #487 still applies

guppylang/prelude/builtins.py
```diff
@@ -505,77 +504,35 @@ def __trunc__(self: float) -> float: ...
...
 
-    @guppy.hugr_op(unsupported_op("IsNotEmpty"))
-    def __hasnext__(self: list[T]) -> tuple[bool, list[T]]: ...
-
-    @guppy.custom(NoopCompiler())
-    def __iter__(self: list[T]) -> list[T]: ...
-
-    @guppy.hugr_op(unsupported_op("Length"))
-    def __len__(self: list[T]) -> int: ...
+    @guppy.hugr_op(unsupported_op("pop"))  # TODO: check None and unwrap
```
ss2165, line 510: "pop" doesn't seem right here?

guppylang/prelude/builtins.py
```diff
@@ -505,77 +504,35 @@ def __trunc__(self: float) -> float: ...
...
+    def __setitem__(self: list[L], idx: int, value: L @ owned) -> None: ...
 
-    @guppy.hugr_op(unsupported_op("Repeat"))
-    def __mul__(self: list[T], other: int) -> list[T]: ...
-
-    @guppy.hugr_op(unsupported_op("Pop"))
-    def __next__(self: list[T]) -> tuple[T, list[T]]: ...
+    @guppy.hugr_op(unsupported_op("length"))  # TODO: inout return in wrong order
```
ss2165, line 513: is this ordering being resolved elsewhere?

guppylang/prelude/builtins.py
```diff
@@ -505,77 +504,35 @@ def __trunc__(self: float) -> float: ...
...
-
-    @guppy.hugr_op(unsupported_op("Find"))
-    def index(self: list[T], elt: T) -> int: ...
-
-    @guppy.custom(checker=FailingChecker("Guppy lists are immutable"))
-    def pop(self: list[T], idx: int) -> None: ...
-
-    @guppy.custom(checker=FailingChecker("Guppy lists are immutable"))
-    def remove(self: list[T], elt: T) -> None: ...
+    @guppy.hugr_op(unsupported_op("Pop"))
```
ss2165, line 528: case difference between "pop" and "Pop" ?

ss2165 (commented):
Looks good, few queries on the unsupporteds

guppylang/prelude/builtins.py
mark-koch: Actually turns out that the length op doesn't return the list at all! https://github.com/CQCL/hugr/issues/1543

ss2165 (approved):

PR #533, opened 2024-10-02
chore!: Remove linst type

### LINKED ISSUE #380: Get rid of linst
Now that we have inout arguments, we can use the same datastructure for classical and linear lists

mark-koch:
Closes #380.

BREAKING CHANGE: Removed the `linst` type, use `list` instead.


guppylang/tys/ty.py
```diff
@@ -667,10 +667,11 @@ def unify(s: Type | Const, t: Type | Const, subst: "Subst | None") -> "Subst | N
         case NoneType(), NoneType():
             return subst
         case FunctionType() as s, FunctionType() as t if s.params == t.params:
-            if len(s.inputs) != len(t.inputs) or any(
-                a.flags != b.flags for a, b in zip(s.inputs, t.inputs, strict=True)
-            ):
+            if len(s.inputs) != len(t.inputs):
                 return None
+            for a, b in zip(s.inputs, t.inputs, strict=True):
+                if a.ty.linear and b.ty.linear and a.flags != b.flags:
```
croyzor, line 673: I think we're going to end up with flags for non-linear types, so would it be better to just implicitly add `@owned` to non-linear types and hide it when printing (for now)?

mark-koch: This was introduced in #531

croyzor (approved):
RIP linst 🪦

PR #551, opened 2024-10-09
feat: Cache accessed source files

mark-koch:
We used to store the entire source code that was used to build an AST as a field within *every node* of the AST. This was a hack so that we can render the source line from an AST node when an error occurs. With our new diagnostics system, we should do better than that!

With this PR, we cache the source of every file accessed by the compiler in a new `SourceMap` class. Then, we can look up the source associated with a span when rendering a diagnostic.

Note that we can't rely on Pythons `linecache` since we also need to keep track of cells in jupyter notebooks as individual "files".

### DIFF
--- a/guppylang/decorator.py
+++ b/guppylang/decorator.py
@@ -38,6 +38,7 @@ from guppylang.module import (
     find_guppy_module_in_py_module,
     get_calling_frame,
 )
+from guppylang.span import SourceMap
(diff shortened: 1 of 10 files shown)


mark-koch:
> I think the various parse methods that take and never use a SourceMap are various constructs that don't track their source location yet?

Yes, exactly!

PR #552, opened 2024-10-09
feat: Render diagnostics

### LINKED ISSUE #537: Implement diagnostics rendering

mark-koch:
Closes #537

### DIFF
--- a/guppylang/diagnostic.py
+++ b/guppylang/diagnostic.py
@@ -1,11 +1,12 @@
-from dataclasses import dataclass, field
+import textwrap
+from dataclasses import dataclass, field, fields
 from enum import Enum, auto
-from typing import ClassVar, Literal, Protocol, runtime_checkable
+from typing import ClassVar, Final, Literal, Protocol, runtime_checkable
 
 from typing_extensions import Self
 
 from guppylang.error import InternalGuppyError
-from guppylang.span import ToSpan, to_span
+from guppylang.span import Loc, SourceMap, Span, ToSpan, to_span
 
 
 class DiagnosticLevel(Enum):
@@ -60,8 +61,32 @@ class Diagnostic(Protocol):
     children: list["SubDiagnostic"] = field(default_factory=list, init=False)
 
     def __post_init__(self) -> None:
-        if self.span_label and not self.span:
+        if self.span_label and self.span is None:
             raise InternalGuppyError("Diagnostic: Span label provided without span")
+
+    @property
+    def rendered_title(self) -> str:
+        """The title of this diagnostic with formatted placeholders."""
+        title = self._render(self.title)
+        assert title is not None
+        return title
+
+    @property
+    def rendered_long_message(self) -> str | None:
+        """The message of this diagnostic with formatted placeholders if provided."""
+        return self._render(self.long_message)
+
+    @property
+    def rendered_span_label(self) -> str | None:
+        """The span label of this diagnostic with formatted placeholders if provided."""
+        return self._render(self.span_label)
+
+    def _render(self, s: str | None) -> str | None:
+        """Helper method to fill in placeholder values in strings with fields of this
+        diagnostic.
+        """
+        values = {f.name: getattr(self, f.name) for f in fields(self)}
+        return s.format(**values) if s is not None else None
 
     def add_sub_diagnostic(self, sub: "SubDiagnostic") -> Self:
         """Adds a new sub-diagnostic."""
@@ -101,6 +126,23 @@ class SubDiagnostic(Protocol):
         if not self.span and not self.message:
             raise InternalGuppyError("SubDiagnostic: Empty diagnostic")
 
+    @property
+    def rendered_message(self) -> str | None:
+        """The message of this diagnostic with formatted placeholders if provided."""
+        return self._render(self.message)
+
+    @property
+    def rendered_span_label(self) -> str | None:
+        """The span label of this diagnostic with formatted placeholders if provided."""
+        return self._render(self.span_label)
+
+    def _render(self, s: str | None) -> str | None:
+        """Helper method to fill in placeholder values in strings with fields of this
+        diagnostic.
+        """
+        values = {f.name: getattr(self, f.name) for f in fields(self)}
+        return s.format(**values) if s is not None else None
+
 
 @runtime_checkable
 @dataclass(frozen=True)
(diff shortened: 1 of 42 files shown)


guppylang/diagnostic.py
```diff
@@ -109,3 +151,222 @@ class Error(Diagnostic, Protocol):
...
+            render_line(first, span.start.line)
+            # Compute the subspan that only covers the first line and render it's
+            # highlight banner
+            first_span = Span(span.start, Loc(span.file, span.start.line, len(first)))
+            first_highlight = " " * first_span.start.column + highlight_char * len(
+                first_span
+            )
+            render_line(first_highlight)
+            # Omit everything in the middle
+            if middle:
```
acl-cqc, line 330: `if len(middle)>1`, maybe. It'd be silly to omit a single line to display a line `...`

acl-cqc: Oh, but I guess then you'd need to highlight that line as well, hmmm. Ok. Not sure about this, maybe I should just shut up and delete the parent comment ;)

guppylang/diagnostic.py
```diff
@@ -109,3 +151,222 @@ class Error(Diagnostic, Protocol):
...
+
+        # Render the last span line and add highlights
+        render_line(last, span.end.line)
+        last_highlight = " " * last_span.start.column + highlight_char * len(last_span)
+
+        # Render the label next to the highlight
+        if label:
+            [label_first, *label_rest] = textwrap.wrap(
+                label,
+                self.MAX_LABEL_LINE_LEN,
```
acl-cqc, line 347: So this is the length of the label AFTER the highlight bar. Irrespective of how much space there is left on the line after said highlight bar. How about instead you ask textwap to layout `highlight_char * len(last_span) + " " + label` with initial indent `last_span.start.column` and subsequent indents, hmmm, you might want to go *backwards* ?
```
           ^^^^ there is |
    some problem here    |
    that won't fit on the|
    rest of the line
```
(`|` to indicate line end boundary, not printed)

Think if your `test_long_label` has very little space after the `==`...

guppylang/diagnostic.py
```diff
@@ -109,3 +151,222 @@ class Error(Diagnostic, Protocol):
...
+                subsequent_indent=" " * (len(last_highlight) + 1),
+            )
+            render_line(last_highlight + label_first)
+            for lbl in label_rest:
+                render_line(lbl)
+        else:
+            render_line(last_highlight)
+
+    @staticmethod
+    def level_str(level: DiagnosticLevel) -> str:
```
acl-cqc, line 360: This isn't `DiagnosticLevel.to_string` / format / something ?

guppylang/diagnostic.py
```diff
@@ -60,9 +61,33 @@ class Diagnostic(Protocol):
...
+    def rendered_span_label(self) -> str | None:
+        """The span label of this diagnostic with formatted placeholders if provided."""
+        return self._render(self.span_label)
+
+    def _render(self, s: str | None) -> str | None:
+        """Helper method to fill in placeholder values in strings with fields of this
+        diagnostic.
+        """
+        values = {f.name: getattr(self, f.name) for f in fields(self)}
+        return s.format(**values) if s is not None else None
```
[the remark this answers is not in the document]
mark-koch: I like the overloading idea. Mypy also seems to be happy with it 👍

guppylang/diagnostic.py
```diff
@@ -109,3 +151,222 @@ class Error(Diagnostic, Protocol):
...
+            remove = leading_whitespace - self.OPTIMAL_LEADING_WHITESPACE
+            all_lines = [line[remove:] for line in all_lines]
+            span = span.shift_left(remove)
+
+        # Render prefix lines
+        for i, line in enumerate(all_lines[:prefix_lines]):
+            render_line(line, span.start.line - prefix_lines + i)
+        span_lines = all_lines[prefix_lines:]
+
+        if span.is_multiline:
```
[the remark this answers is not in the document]
mark-koch: Yeah, not sure either. This replicates the current behaviour in Guppy, but easy enough to change in the future by updating `DiagnosticsRenderer.PREFIX_CONTEXT_LINES`

guppylang/diagnostic.py
mark-koch: I think the current alignment looks nicer, but I agree that the spacing might become problematic. 

In a previous version I implemented a layout algorithm that abbreviates the start/end of the line and/or the middle of the span with `...` ro ensure that the `^^^` doesn't go too far to the right and everything fits into a specified width

```
   |
42 | ...else foo(looo...ng_id).more_stuff(...
   |             ^^^^^^^^^^^^ Span label with
   |                          line break
```

I think that is the best solution but would have made this PR 150-200 lines longer 😅

I created #556 as a followup

guppylang/diagnostic.py
```diff
@@ -109,3 +151,222 @@ class Error(Diagnostic, Protocol):
...
+        ```
+        """
+        if diag.span is None:
+            # Omit the title if we don't have a span, but a long message. This case
+            # should be fairly rare.
+            msg = diag.rendered_long_message or diag.rendered_title
+            self.buffer += textwrap.wrap(
+                f"{self.level_str(diag.level)}: {msg}",
+                self.MAX_MESSAGE_LINE_LEN,
+            )
```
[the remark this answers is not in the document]
mark-koch: > Is the assumption that there won't be any if the parent doesn't have a span?

I think that's an ok assumption to make for now since basically 99% of diagnostics are going to come with a span. We can come back to that if it turns into an issue later.

I'll add a check though 👍

guppylang/diagnostic.py
```diff
@@ -109,3 +151,222 @@ class Error(Diagnostic, Protocol):
...
+                        to_span(sub_diag.span),
+                        sub_diag.rendered_span_label,
+                        is_primary=False,
+                    )
+            if diag.long_message:
+                self.buffer.append("")
+                self.buffer += textwrap.wrap(
+                    f"{diag.rendered_long_message}", self.MAX_MESSAGE_LINE_LEN
+                )
+        # Finally, render all sub-diagnostics that have a non-span message
```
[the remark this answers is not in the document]
mark-koch: The doc comment of `SubDiagnostic.message` says "Message that is printed if no span is provided."

PR #553, opened 2024-10-09
feat: Allow emission of diagnostics inside GuppyError

### LINKED ISSUE #538: Allow diagnostics inside `GuppyError`

mark-koch:
Closes #538.

* Allow to pass a `Diagnostic` object to `GuppyError` instead of an error message
* The old error reporting system remains untouched until we have migrated everything to the new diagnostics system
* Adds a `TypeMismatchError` diagnostic as a test to ensure that the pipeline works

### DIFF
--- a/guppylang/checker/expr_checker.py
+++ b/guppylang/checker/expr_checker.py
@@ -24,8 +24,8 @@ import ast
 import sys
 import traceback
 from contextlib import suppress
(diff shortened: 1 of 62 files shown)


ss2165 (approved):
Finally using python for what it's good at: pretty printing


This really feels like a standalone library to me 🤷

PR #560, opened 2024-10-15
feat: Emit infallible `float`→`rotation` conversions

### LINKED ISSUE #550: Only include the required tket2 extensions when compiling packages
Currently we bundle all the tket2 extensions in the packages all the time.
We should detect which extensions are actually being used, and only include those.

https://github.com/CQCL/guppylang/blob/473745fcf952fd7bc2b442cb257e6539b30d0e98/guppylang/module.py#L337-L343

aborgna-q:
Closes #550 

- Updates `tket2` dependency to `>0.4.1`, where we added the `tket2.rotation.from_halfturns_unchecked` op
- Updates the json extension definitions
- Removes the panicking conditional when converting floats into rotations for quantum op params.


qartik (dismissed):

aborgna-q:
We may need some more fixes before updating the tket2 dep -.-

aborgna-q:
I temporarily disabled the `guppy → pytket` tests, as they need extra support from tket2.
https://github.com/CQCL/tket2/issues/504

qartik (approved):

PR #568, opened 2024-10-17
feat!: return packages with pointers when compiling

### LINKED ISSUE #559: Compiling specific definitions should return pointers to compiled versions
e.g for a main function `main.compile()` returns a whole package. This should return a `FuncDefnPointer` once https://github.com/CQCL/hugr/issues/1561 is released.

ss2165:
Closes #559 
Is it valid to assume there is always a top-level funcdef or funcdecl for a Definition.compile()?

BREAKING CHANGE: decorator `compile_module` and `GuppyModule.compile` return `ModulePointer` rather than `Package` (package can be retrieved using `ModulePointer.package`. `Definition.compile` returns a `Func{Defn/Decl}Pointer`.

### DIFF
--- a/guppylang/definition/common.py
+++ b/guppylang/definition/common.py
@@ -6,8 +6,6 @@ from dataclasses import dataclass, field
 from typing import TYPE_CHECKING, ClassVar, TypeAlias
 
 from hugr.build.dfg import DefinitionBuilder, OpVar
-from hugr.ops import FuncDecl, FuncDefn
-from hugr.package import FuncDeclPointer, FuncDefnPointer
 
 if TYPE_CHECKING:
     from guppylang.checker.core import Globals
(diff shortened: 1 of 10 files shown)


guppylang/definition/common.py
```diff
@@ -73,9 +72,21 @@ def description(self) -> str:
...
+    #     for node in hugr:
+    #         op = hugr[node].op
+    #         match op:
+    #             case FuncDefn(f_name=self.name):
+    #                 return FuncDefnPointer(module.package, module.module_index, node)
+    #             case FuncDecl(f_name=self.name):
+    #                 return FuncDeclPointer(module.package, module.module_index, node)
+    #             case _:
+    #                 continue
+    #     raise ValueError(f"Unexpected op {op}")
```
mark-koch, line 89:
```suggestion
```

PR #572, opened 2024-10-17
fix: Fix list lowering bugs

mark-koch:
Fixes some lowering bugs for list subscripts. Also adds more tests for list subscripting (mostly copied from `test_array.py`)

### DIFF
--- a/guppylang/prelude/_internal/compiler/list.py
+++ b/guppylang/prelude/_internal/compiler/list.py
@@ -19,6 +19,7 @@ from guppylang.prelude._internal.compiler.prelude import (
     build_error,
     build_panic,
     build_unwrap,
+    build_unwrap_right,
 )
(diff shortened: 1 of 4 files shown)


guppylang/prelude/_internal/compiler/list.py
```diff
@@ -133,7 +134,7 @@ def build_linear_getitem(
         list_wire, result = self.builder.add_op(
             list_set(elem_opt_ty), list_wire, idx, none
         )
-        elem_opt = build_unwrap(self.builder, result, "List index out of bounds")
+        elem_opt = build_unwrap_right(self.builder, result, "List index out of bounds")
```
mark-koch, line 137: The list `set` operation returns an `Either` instead of an `Optional`

aborgna-q (approved):

PR #575, opened 2024-10-17
fix: Fix array lowering bugs

### LINKED ISSUE #573: bug: function returning array fails with type mismatch
Example:

```python
def test_return_linear_array(validate ):
    module = GuppyModule("test")
    module.load(qubit)
    @guppy(module)
    def foo() -> array[qubit, 2]:
        a = array(qubit(), qubit())
        return a

    package = module.compile()
    validate(package)
```

fails with:
```python
ValueError: The function has fixed output type [ExtType(type_def=TypeDef(name='array', description='array', params=[BoundedNatParam(upper_bound=None), TypeTypeParam(bound=<TypeBound.Any: 'A'>)], bound=FromParamsBound(indices=[1])), args=[BoundedNatArg(n=2), TypeTypeArg(ty=Sum([[Qubit], []]))])], but was given output wires with types [ExtType(type_def=TypeDef(name='array', description='array', params=[BoundedNatParam(upper_bound=None), TypeTypeParam(bound=<TypeBound.Any: 'A'>)], bound=FromParamsBound(indices=[1])), args=[BoundedNatArg(n=2), TypeTypeArg(ty=Qubit)])].
```

Looks like some wiring is not accounting for linear arrays being backed by hugr arrays over Option<T>

mark-koch:
Fixes #573 together with other list lowering issues.

There were some inconsistencies whether `None` is `Tag(0)` or `Tag(1)`. Now we consistently use `ht.Option` and the `build_unwrap` helper.

### DIFF
--- a/guppylang/prelude/_internal/compiler/array.py
+++ b/guppylang/prelude/_internal/compiler/array.py
@@ -26,11 +26,8 @@ from guppylang.tys.const import ConstValue
 def _instantiate_array_op(
     name: str, elem_ty: ht.Type, length: int, inp: list[ht.Type], out: list[ht.Type]
 ) -> ops.ExtOp:
-    op_def = hugr.std.PRELUDE.get_op(name)
-    return ops.ExtOp(
-        op_def,
-        ht.FunctionType(inp, out),
-        [ht.BoundedNatArg(length), ht.TypeTypeArg(elem_ty)],
+    return hugr.std.PRELUDE.get_op(name).instantiate(
+        [ht.BoundedNatArg(length), ht.TypeTypeArg(elem_ty)], ht.FunctionType(inp, out)
     )
 
 
--- a/guppylang/prelude/_internal/compiler/array.py
+++ b/guppylang/prelude/_internal/compiler/array.py
@@ -2,8 +2,6 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING
-
 import hugr.std
 from hugr import Wire, ops
 from hugr import tys as ht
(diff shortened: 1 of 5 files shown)


guppylang/prelude/_internal/compiler/array.py
```diff
@@ -12,229 +10,196 @@
...
+
+def _instantiate_array_op(
+    name: str, elem_ty: ht.Type, length: int, inp: list[ht.Type], out: list[ht.Type]
+) -> ops.ExtOp:
+    op_def = hugr.std.PRELUDE.get_op(name)
+    return ops.ExtOp(
+        op_def,
+        ht.FunctionType(inp, out),
+        [ht.BoundedNatArg(length), ht.TypeTypeArg(elem_ty)],
+    )
```
ss2165, line 34: optional: use `OpDef.instantiate` (equivalent in this case)

guppylang/prelude/_internal/compiler/array.py
```diff
@@ -12,229 +10,196 @@
...
-            self.builder,
-            array,
-            array_ty,
-            idx,
-            idx_ty,
-            none,
-            elem_opt_ty,
-            length,
+        elem_opt_ty = ht.Option(self.elem_ty)
+        none = self.builder.add_op(ops.Tag(0, elem_opt_ty))
```
ss2165, line 138: maybe we should add `build_some` `build_none` `build_left` `build_ok` `build_error` `build_right` etc. to hugr-py?

ss2165: or maybe just constructors for the `Tag` op

guppylang/prelude/_internal/compiler/prelude.py
```diff
@@ -94,6 +94,24 @@ def build_unwrap_right(
     return conditional.to_node()
 
 
+def build_unwrap_left(
```
ss2165, line 97: seems worth commoning up with `build_unwrap_right`, only the case numbers differ

tests/integration/test_array.py
```diff
@@ -61,6 +61,18 @@ def main(ys: array[int, 0]) -> array[array[int, 0], 2]:
     validate(main)
 
 
+def test_return_linear_array(validate ):
```
ss2165, line 64:
```suggestion
def test_return_linear_array(validate):
```
unformatted?

ss2165 (approved):
Much nicer, only some minor queries

guppylang/prelude/_internal/compiler/array.py
mark-koch: I think that's a good idea 👍 See https://github.com/CQCL/hugr/issues/1592

guppylang/prelude/_internal/compiler/prelude.py
mark-koch: We can do that once https://github.com/CQCL/hugr/issues/1596 is fixed

ss2165: leave a todo

PR #580, opened 2024-10-18
feat: add `qubit` discard/measure methods

ss2165:
Extends the fact that allocation is tied to the `qubit` type by implementing measure/discard methods.

Note this includes a drive-by `measure_reset` operation which avoids resetting the qubit and just flip, might not be something to add in this PR.

### DIFF
--- a/guppylang/prelude/quantum.py
+++ b/guppylang/prelude/quantum.py
@@ -28,21 +28,6 @@ class qubit:
 
     @guppy
     @no_type_check
-    def h(self: "qubit") -> None:
-        h(self)
-
-    @guppy
-    @no_type_check
-    def cz(self: "qubit", target: "qubit") -> None:
-        cz(self, target)
-
-    @guppy
-    @no_type_check
-    def cx(self: "qubit", target: "qubit") -> None:
-        cx(self, target)
-
-    @guppy
-    @no_type_check
     def measure(self: "qubit" @ owned) -> bool:
         return measure(self)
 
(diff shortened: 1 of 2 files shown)


tests/integration/test_inout.py
```diff
@@ -17,7 +17,7 @@ def test_basic(validate):
     def foo(q: qubit) -> None: ...
 
     @guppy(module)
-    def test(q: qubit @owned) -> qubit:
+    def test(q: qubit @ owned) -> qubit:
```
ss2165, line 20: it appears my formatter has been at all the decorators. I can undo this if necessary

guppylang/prelude/quantum.py
```diff
@@ -26,84 +26,209 @@ def __new__() -> "qubit":
...
+        h(self)
+
+    @guppy
+    @no_type_check
+    def cz(self: "qubit", target: "qubit") -> None:
+        cz(self, target)
+
+    @guppy
+    @no_type_check
+    def cx(self: "qubit", target: "qubit") -> None:
```
aborgna-q, line 41: I would suggest`target.cx(control)` instead,
so the n-controlled cases keep the same structure.
```py
target.toffoli(control0, control1)
target.ncx(*controls)
```

cqc-alec: Whichever way round it is, I see a lot of potential for bugs caused by people getting it the wrong way round.
In general I'm not so keen on forcing multi-qubit operations to have a "distinguished" qubit in this way: in general there's no natural way to select which qubit it should be.

ss2165: I think that's very valid. How about we restrict it a specific set of special single qubit operations: alloc (already the case), discard, measure (non destructive), reset, and any variants of those above (e.g. measure_reset, destructive measure). These special set can be implemented in terms of the existing inout functions as in this PR currently.

ss2165: Done

PR #588, opened 2024-10-23
fix: Take all spans into account when determining snippet alignment

mark-koch:


guppylang/diagnostic.py
```diff
@@ -208,10 +208,16 @@ def render_diagnostic(self, diag: Diagnostic) -> None:
         else:
             span = to_span(diag.span)
             level = self.level_str(diag.level)
+            all_spans = [
+                span,
+                *(to_span(child.span) for child in diag.children if child.span),
```
acl-cqc, line 213: nit: took me awhile to figure out what this `*` was about - it does seem like there's one more construction/destruction than I'd expect. It might be less efficient but `[span] + [to_span(child.span) for child in diag.children if child.span]` might be clearer. Up to you!

tests/diagnostics/snapshots/test_two_spans_different_lineno_lens.txt
```diff
@@ -0,0 +1,9 @@
+Error: Can't compare apples with oranges (at <unknown>:99:6)
+    | 
+ 97 | apple == orange
+ 98 | apple == orange
+ 99 | apple == orange
+    |       ^^ Comparison attempted here
+    | 
+100 | apple == orange
```
acl-cqc, line 8: This confused me a bit too - I realize this is purely formatting, but this looks like the subdiagnostic should *also* refer to line 99! (What happens if it does? Is that a separate test worth having?)

Changing the "This is an apple" to "This apple is on another line" might have avoided confusion ;)

acl-cqc (approved):
Looks good to me - basically, I am just feeling silly that I was confused. But a couple of suggestions if you like :)

PR #589, opened 2024-10-23
feat: Update CFG checker to use new diagnostics

mark-koch:


tests/error/errors_on_usage/and_not_defined.err
```diff
@@ -1,7 +1,11 @@
-Guppy compilation failed. Error in file $FILE:9
+Error: Variable not defined (at $FILE:9:15)
+  | 
+7 |         return z
+8 |     else:
+9 |         return z
+  |                ^ `z` might be undefined ...
+  | 
+6 |     if x and (z := y + 1):
+  |        - ... if this expression is `False`
```
ss2165, line 9: to my eye this looks like `if x False` rather than `if x and (z := y + 1) is False` ...

tests/error/errors_on_usage/for_new_var.err
```diff
@@ -1,7 +1,11 @@
-Guppy compilation failed. Error in file $FILE:8
+Error: Variable not defined (at $FILE:8:11)
+  | 
+6 |     for _ in xs:
+7 |         y = 5
+8 |     return y
+  |            ^ `y` might be undefined ...
+  | 
+6 |     for _ in xs:
+  |              -- ... if this expression is `False`
```
ss2165, line 9: this is a bit confusing, do we need to say "Falsey"...?

tests/error/errors_on_usage/or_not_defined.err
```diff
@@ -1,7 +1,11 @@
-Guppy compilation failed. Error in file $FILE:7
+Error: Variable not defined (at $FILE:7:15)
+  | 
+5 | def foo(x: bool, y: int) -> int:
+6 |     if x or (z := y + 1):
+7 |         return z
+  |                ^ `z` might be undefined ...
+  | 
+6 |     if x or (z := y + 1):
+  |        - ... if this expression is `True`
```
ss2165, line 9: same as previous query

ss2165 (commented):
These are some very cool errors!
Some queries about falsy loops and compound predicate expressions

tests/error/errors_on_usage/and_not_defined.err
mark-koch: I think the message is correct:

If `x` were true then `z` would be defined because of the walrus assignment even if the whole thing tends out to be false.

If `x` is false, then `z` is not defined because `and` short circuits

mark-koch: This is the test:
```python
@guppy
def foo(x: bool, y: int) -> int:
    if x and (z := y + 1):
        return z
    else:
        return z
```

tests/error/errors_on_usage/for_new_var.err
mark-koch: I like the idea of differentiating between bools and values that are truthy. The problem is that we can no longer make this distinction here since we already inserted an implicit call to `__bool__`. 

An additional problem is that for loops are desugared to
```python
iter = xs.__iter__()
while True:
   b, iter = iter.__hasnext__()
   if b:
      x, it = iter.__next__()
   else:
      break
iter.__end__()
```
and the message actually points to the `while True` bit which actually *is* a bool.

Created issues #591 and #592 to figure out how to solve this

tests/error/errors_on_usage/and_not_defined.err
ss2165: that makes sense! thanks

ss2165 (approved):

PR #590, opened 2024-10-23
refactor: Update function checker to use new diagnostics

mark-koch:


tests/error/misc_errors/return_not_annotated_none2.err
```diff
@@ -1,6 +1,14 @@
...
+  | ^^^^^^^^^^
+  | ...
+7 |         return x
+  | ^^^^^^^^^^^^^^^^ Return type must be annotated
 
-3:    @compile_guppy
-4:    def foo():
-      ^^^^^^^^^^
-GuppyError: Return type must be annotated. Try adding a `-> None` annotation.
+Help: Looks like `foo` doesn't return anything. Consider annotating it with `->
```
croyzor, line 11: 🔥

tests/error/misc_errors/arg_not_annotated.err
```diff
@@ -1,6 +1,8 @@
-Guppy compilation failed. Error in file $FILE:5
+Error: Missing type annotation (at $FILE:5:17)
+  | 
+3 | 
+4 | @compile_guppy
+5 | def foo(x: bool, y) -> int:
+  |                  ^ Argument type must be annotated
```
croyzor, line 6: Maybe "argument requires a type annotation"?

tests/diagnostics/snapshots/test_two_spans_different_lineno_lens.txt
```diff
@@ -0,0 +1,9 @@
+Error: Can't compare apples with oranges (at <unknown>:99:6)
+    | 
+ 97 | apple == orange
+ 98 | apple == orange
+ 99 | apple == orange
```
croyzor, line 5: 😁

examples/demo.ipynb
```diff
@@ -486,13 +486,18 @@
...
+      "Error: Illegal assignment (at <In [14]>:6:8)\n",
+      "  | \n",
+      "4 | def outer(x: int) -> int:\n",
+      "5 |     def nested() -> None:\n",
+      "6 |         x += 1  # Mutation of captured variable x is not allowed\n",
+      "  |         ^^^^^^ Variable `x` may not be assigned to since `x` is captured\n",
+      "  |                from an outer scope\n",
+      "  | \n",
+      "4 | def outer(x: int) -> int:\n",
+      "  |           ------ `x` defined here\n",
```
croyzor, line 498: It would be cool to detect overlapping spans and only print the `outer` once like
```
def outer(x: int) -> int
                ----- `x` defined here
    def nested() -> None
        x += 1
        ^^^^^^ Variable `x`...
```
but that seems tricky...

croyzor (approved):

tests/error/nested_errors/reassign_capture_3.err
```diff
@@ -1,7 +1,12 @@
...
+Error: Illegal assignment (at $FILE:11:12)
+   | 
+ 9 | 
+10 |         def baz() -> None:
+11 |             y += 2
+   |             ^^^^^^ Variable `y` may not be assigned to since `y` is captured
+   |                    from an outer scope
+   | 
+ 6 |     y = x + 1
+   |     - `y` defined here
```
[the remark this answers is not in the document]
mark-koch: Yes, that's a good idea. However, we'll need to be careful because the grammar might break in some cases when reordering:

```
Error: Linearity violation (at $FILE:21:8)
   | 
19 | def test(q: qubit @owned) -> None:
20 |     use(q)
21 |     foo(q)
   |         ^ Variable `q` with linear type `qubit` cannot be borrowed
   |           ...
   | 
20 |     use(q)
   |         - since it was already consumed here
```

We'd need two version of the span labels depending on which order they are outputted. I created #608 to collect these improvement ideas

tests/error/misc_errors/return_not_annotated.err
```diff
@@ -1,6 +1,10 @@
-Guppy compilation failed. Error in file $FILE:5
+Error: Missing type annotation (at $FILE:5:0)
+  | 
+3 | 
+4 | @compile_guppy
+5 | def foo(x: bool):
+  | ^^^^^^^^^^^^^^^^^
+6 |     return x
+  | ^^^^^^^^^^^^ Return type must be annotated
```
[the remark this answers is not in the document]
mark-koch: Unfortunately, the Python AST doesn't give us a span for this. Taking the first line is also not valid sice you could have weird stuff like

```python
def foo(
  # No arguments, so no argument spans
)     :   # Closing paren on separate line + extra spaces before the colon
```

examples/demo.ipynb
mark-koch: Yes, good idea, Rust also does this! Added to #608

PR #593, opened 2024-10-24
feat: conventional results post processing

ss2165:
Post processing of result streams, for converting to traditional distributions over bitstrings if some tagging convention is used, including conversion to a pytket BackendResult.

## Convention
Tags are assumed to be a name of a bit register unless they fit the regex pattern `^([a-z][\w_]*)\[(\d+)\]$` (like `reg[12]`) in which case they are assumed to refer to the nth element of a bit register.

For results of the form 
```
result("<register>", value)
```
`value` can be `{0, 1}`, wherein the register is assumed to be length 1, or lists over those values, wherein the list is taken to be the value of the entire register.

For results of the form
```
result("<register>[n]", value)
```
`value` can only be `{0, 1}`.
The register is assumed to be at least `n+1` in size and unset elements are assumed to be `0`.

Subsequent writes to the same register/element in the same shot will overwrite.

To convert to a `BackendResult` all registers must be present in all shots, and register sizes cannot change between shots.

## TODO 
Clear documentation of the above conventions and examples.


guppylang/hresult.py
```diff
@@ -0,0 +1,218 @@
...
+                reg_name, reg_index_str = match.groups()
+                reg_index = int(reg_index_str)
+
+                if reg_name not in reg_bits:
+                    # Initialize register counts to False
+                    reg_bits[reg_name] = [False] * (int(reg_index) + 1)
+                bitlst = reg_bits[reg_name]
+                if reg_index >= len(bitlst):
+                    # Extend register counts with False
+                    bitlst += [False] * (int(reg_index) - len(bitlst) + 1)
```
qartik, line 95:
```suggestion
                    bitlst += [False] * (reg_index - len(bitlst) + 1)
```

guppylang/hresult.py
```diff
@@ -0,0 +1,218 @@
...
+            >>> results = Results()
+            >>> results.append("tag1", 1)
+            >>> results.append("tag2", 2)
+            >>> results.append("tag2", 3)
+            >>> results.as_dict()
+            {"tag1": 1, "tag2": 3}
+        """
+        return dict(self.entries)
+
+    def to_register_bits(self) -> dict[str, list[bool]]:
```
qartik, line 78: I think, in its current state, this function should really be called `to_register_bools`. But I also wonder why you are converting everything to bools and then to bitstrings, it involves unnecessary back and forth conversion for the usual case where you are already given bits.

I'd think that `_cast_primitive_bool` should return a bit after doing necessary type checking and we stay away from bools for results.

guppylang/hresult.py
```diff
@@ -0,0 +1,218 @@
...
+Under this convention, tags are assumed to be a name of a bit register unless they fit
+the regex pattern `^([a-z][\w_]*)\[(\d+)\]$` (like `my_Reg[12]`) in which case they
+are assumed to refer to the nth element of a bit register.
+
+For results of the form ``` result("<register>", value) ``` `value` can be `{0, 1, True,
+False}`, wherein the register is assumed to be length 1, or lists over those values,
+wherein the list is taken to be the value of the entire register.
+
+For results of the form ``` result("<register>[n]", value) ``` `value` can only be `{0,
+1, True, False}`. The register is assumed to be at least `n+1` in size and unset
```
qartik, line 16: Is there a great benefit in supporting both boolean literals and bits 0 and 1? From the hardware perspective, we probably only want bits.

guppylang/hresult.py
```diff
@@ -0,0 +1,218 @@
...
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from pytket.backends.backendresult import BackendResult
+
+#: Primitive data types that can be returned by a result
+DataPrimitive = int | float | bool
+#: Data value that can be returned by a result: a primitive or a list of primitives
```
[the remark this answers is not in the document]
ss2165: yes, it lets sphinx pick up on docstrings for attributes/constants

tests/test_hresult.py
```diff
@@ -0,0 +1,110 @@
...
+        ("asdf3h32", None),
+        ("dsf[3]asdf", None),
+        ("_s34fd_fd[12]", None),
+        ("afsd3[34]sdf", None),
+        ("asdf[2]", ("asdf", 2)),
+        ("as3df[21234]", ("as3df", 21234)),
+        ("as3ABdfAB[2]", ("as3ABdfAB", 2)),
+    ],
+)
+def test_reg_index_pattern_match(identifier, match: tuple[str, int] | None):
```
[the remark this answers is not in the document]
ss2165: AFAIK this pattern matches the QASM2 allowed names

guppylang/hresult.py
ss2165: my initial handling of bools came about because a guppy `measure` returns a boolean, not an integer. This difference may be lost in the results handling (which is not too terrible because python booleans are integers anyway). I just wanted to avoid someone printing `result("c[0]", 3)` and accidentally getting that set to `1`, which we still can.

I agree with your simplification, and have made the suggested changes.

guppylang/hresult.py
```diff
@@ -0,0 +1,206 @@
...
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, Literal
+
+if TYPE_CHECKING:
+    from collections.abc import Iterable
+
+    from pytket.backends.backendresult import BackendResult
+
+#: Primitive data types that can be returned by a result
+DataPrimitive = int | float | bool
```
qartik, line 40: I am confused by the spec description above:
> ``` result("<register>", value) ``` `value` can be `{0, 1}`
and the code here.

Are floats supported or not? I'd imagine, if guppy says `result(x, pi)`, we can return an approx value of pi as part of the results.

Or is this PR limited to bitstrings? If so, why include `float`?

ss2165: float is allowed as a result type, it is not supported under the "register bitstring convention" adopted as part of this PR. The `HResult` and `Shots` classes faithfully capture all types, it is only the post-processing/conversions that are limited.

qartik (approved):
LGTM

guppylang/hresult.py
```diff
@@ -0,0 +1,206 @@
...
+            >>> results.append("tag1", 1)
+            >>> results.append("tag2", 2)
+            >>> results.append("tag2", 3)
+            >>> results.as_dict()
+            {"tag1": 1, "tag2": 3}
+        """
+        return dict(self.entries)
+
+    def to_register_bits(self) -> dict[str, str]:
+        """Convert results to a dictionary of register bit values."""
```
mark-koch, line 82: Let's say I do

```python
results = Results()
results.append("qs[0]", 1)
results.append("qs", 0)
results.append("qs[1]", 1)
results.to_register_bits()
```

The expected output should be `{"qs": "01"}`, right? I think this works since dicts are iterated in insertion order and `HResult.as_dict` keeps the order of `entries`. You could consider using `OrderedDict` to make this invariant explicit?

guppylang/hresult.py
```diff
@@ -0,0 +1,206 @@
...
+    if isinstance(data, int) and data in {0, 1}:
+        return str(data)  # type: ignore[return-value]
+    raise ValueError(f"Expected bit data for register value found {data}")
+
+
+@dataclass
+class HShots:
+    """Results accumulated over multiple shots."""
+
+    results: list[HResult] = field(default_factory=list)
```
mark-koch, line 121: No need for the `default_factory` since you have a custom `__init__`

guppylang/hresult.py
```diff
@@ -0,0 +1,206 @@
...
+            from pytket._tket.unit_id import Bit
+            from pytket.backends.backendresult import BackendResult
+            from pytket.utils.outcomearray import OutcomeArray
+        except ImportError as e:
+            raise ImportError(
+                "Pytket is an optional dependency, install with the `pytket` extra"
+            ) from e
+        counts = self.register_bitstrings(strict_lengths=True, strict_names=True)
+        reg_sizes: dict[str, int] = {
+            reg: len(next(iter(counts[reg]), "")) for reg in counts
```
mark-koch, line 198: Why is the `""` default in `next` needed? Shouldn't all registers be populated with at least one value?

mark-koch (approved):

guppylang/hresult.py
ss2165: yes, it is just there to avoid having to handle unexpected stopiteration errors induced by changes to other code. It is safe to assume unpopulated registers are size 0

guppylang/hresult.py
ss2165: yes it relies on insertion order preservation guaranteed from python 3.7 onwards.
`OrderedDict` is more useful for things that require reordering these days so I will
just add a comment.

PR #598, opened 2024-10-28
chore: PEP 735 dev dependencies

aborgna-q:
[PEP 735](https://peps.python.org/pep-0735/) specifies the syntax for listing development dependencies in `pyproject.toml` in a package manager-agnostic way.

`uv 0.4.27` includes support for this, so we can now cleanup our pyproject.

drive-by: Use `--frozen --python` for `uv sync` in CI, to ensure we are using the right dep versions.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -31,21 +31,30 @@ execution = ["execute-llvm"]
 homepage = "https://github.com/CQCL/guppylang"
 repository = "https://github.com/CQCL/guppylang"
 
-[tool.uv]
-dev-dependencies = [
-    "ipykernel >=6.29.5,<7",
+[dependency-groups]
+# Default dev dependency group
+dev = [
+    { include-group = "lint" },
+    { include-group = "test" },
+    { include-group = "llvm_integration" },
+    { include-group = "validation" },
+]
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -31,21 +31,29 @@ execution = ["execute-llvm"]
...
+]
+llvm_integration = [
+    { include-group = "test" },
+    # Required to run the llvm integration tests
     "maturin >=1.4.0,<2",
-    "pytket >=1.30.0,<2",
-    # Required to run `maturin develop`
     "pip >=24",
 ]
+validation = [{ include-group = "test" }, "pytket >=1.34.0,<2"]
```
ss2165, line 56: why is pytket in the validation group?

aborgna-q: Right, wrong name. Changed it to `pytket_integration`.

ss2165:
This would break local workflows relying on uv below this version, update should be easy for most people but needs to be flagged.

For nix, it will be broken until unstable does a version bump...

aborgna-q:
Nix now has `uv 0.4.28`, so I updated the devenv lock.
Could you check if it works?

ss2165 (approved):
Success!

PR #600, opened 2024-10-29
feat: Allow sub-diagnostic labels to refer to fields of their parent

mark-koch:
Sub-diagnostics now store a pointer to their parent main diagnostic that is used to look up field values when formatting labels and messages.

This is achieved by subclassing `string.Formatter` to implement the custom lookup

### DIFF
--- a/tests/diagnostics/snapshots/test_advanced_formatting.txt
+++ b/tests/diagnostics/snapshots/test_advanced_formatting.txt
@@ -1,0 +1,7 @@
+Error: Can't compare apples with oranges (at <unknown>:1:0)
+  | 
+1 | apple == orange
+  | ^^^^^ This is an apple
+  | 
+1 | apple == orange
+  |          ------ This is not an apple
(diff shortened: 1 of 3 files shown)


tests/diagnostics/snapshots/test_advanced_formatting.txt
```diff
@@ -0,0 +1,7 @@
+Error: Can't compare apples with oranges (at <unknown>:1:0)
+  | 
+1 | apple == orange
+  | ^^^^^ This is an apple
+  | 
+1 | apple == orange
+  |          ------ This is not an apple
```
aborgna-q, line 7: Shouldn't this have printed "aple", since the subdiagnostic overwrites the `pp` attribute ?

aborgna-q: Ah, nvm. The format string is "{p}{pp}".

aborgna-q (approved):

PR #605, opened 2024-10-29
feat: Update type parsing to use new diagnostics

mark-koch:


guppylang/decorator.py
```diff
@@ -492,11 +492,12 @@ def _parse_expr_string(ty_str: str, parse_err: str, sources: SourceMap) -> ast.e
             sources.add_file(info.filename)
             source_lines, _ = inspect.getsourcelines(caller_module)
             source = "".join(source_lines)
-            annotate_location(expr_ast, source, info.filename, 0)
+            annotate_location(expr_ast, source, info.filename, 1)
```
mark-koch, line 495: Drive-by: Our new spans start counting lines at 1

guppylang/tys/errors.py
```diff
@@ -0,0 +1,127 @@
...
+    title: ClassVar[str] = ""  # Custom implementation in `rendered_title`
+    expected: int
+    actual: int
+    type_name: str
+
+    @property
+    def rendered_title(self) -> str:
+        if self.expected > self.actual:
+            return "Missing type arguments"
+        elif 0 == self.expected < self.actual:
```
acl-cqc, line 22: nit: this case is a bit confusing - I mean, the assumption is that `expected>=0 and actual>=0 and expected!=actual` in which case the `< self.actual` here is redundant. I'd just be consistent with `rendered_span_label` i.e. check `self.expected == 0` *first* and then the other two.

acl-cqc (approved):
Looks good, ta. (Though I note there are quite a few TODOs!)

guppylang/decorator.py
mark-koch: Columns also start at 1, but that's the same in Python AST afaik, so it works out without having to modify it

tests/error/inout_errors/nonlinear.err
```diff
@@ -1,6 +1,8 @@
-Guppy compilation failed. Error in file $FILE:11
+Error: Invalid annotation (at $FILE:11:16)
```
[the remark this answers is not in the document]
mark-koch: This is only a placeholder in the golden value. The testing framework replaces it with the actual path (which will be different on every machine that cloned the repo)

PR #606, opened 2024-10-29
feat: Update remaining code to use new diagnostics

mark-koch:


guppylang/decorator.py
```diff
@@ -149,7 +149,7 @@ def _get_python_caller(self, fn: PyFunc | None = None) -> ModuleIdentifier:
                         break
                 frame = frame.f_back
             else:
-                raise GuppyError("Could not find a caller for the `@guppy` decorator")
+                raise RuntimeError("Could not find a caller for the `@guppy` decorator")
```
mark-koch, line 152: Guppy errors are now reserved for things that go wrong during compilation. This error is thrown immediately when the user calls the decorator, so we shouldn't use `GuppyError`

Same for other errors in this file

guppylang/decorator.py
```diff
@@ -310,11 +310,6 @@ def custom(
 
         def dec(f: PyFunc) -> RawCustomFunctionDef:
             func_ast, docstring = parse_py_func(f, self._sources)
-            if not has_empty_body(func_ast):
```
croyzor, line 313: What's going on here?

guppylang/experimental.py
```diff
@@ -55,19 +58,30 @@ def __exit__(
...
 
 
+@dataclass(frozen=True)
+class ExperimentalFeatureError(Error):
+    title: ClassVar[str] = "Experimental feature"
+    span_label: ClassVar[str] = "{things} are an experimental feature"
+    things: str
+
+    @dataclass(frozen=True)
+    class Suggestion(Help):
```
croyzor, line 68: Is this error ever going to be created without adding this Help?

guppylang/prelude/_internal/checker.py
```diff
@@ -75,37 +79,19 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
     """Call checker for Python builtin functions that are not available in Guppy.
 
     Gives the uses a nicer error message when they try to use an unsupported feature.
     """
 
     def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
-        raise GuppyError(
-            f"Builtin method `{self.func.name}` is not supported by Guppy", self.node
-        )
+        err = UnsupportedError(self.node, f"Builtin method `{self.func.name}`")
```
croyzor, line 89: I think this should be `singular=True`

croyzor (approved):
New errors are looking very nice!

I'm seeing a lot of a pattern where an error declares a subdiagnostic as a nested class, which is added by the caller. Can we not define classes with default subdiagnostics?

mark-koch:
> I'm seeing a lot of a pattern where an error declares a subdiagnostic as a nested class, which is added by the caller. Can we not define classes with default subdiagnostics?

I like the idea 👍  I created issue #620 to implement this

PR #611, opened 2024-11-04
feat: Add `SizedIter` wrapper type

### LINKED ISSUE #610: Statically sized iterators
Required to reason about the length or array comprehensions

mark-koch:
Adds a new wrapper type `SizedIter` to the standard library that annotates an iterator with a static size hint. This will be used for example in array comprehensions to infer the size of the resulting array. The type gets erased when lowering to Hugr.

Updates the `range` function to emit such a size hint when the range stop is given by a static number. For example the expression `range(10)` is typed as `SizedIter[Range, 10]` whereas `range(n + 1)` is typed as `Range`. Currently, this is implemented via a `CustomCallChecker`, but in the future we could use function overloading and `Literal` types to implement this in Guppy source:

```python
@guppy.overloaded
def range[n: nat](stop: Literal[n]) -> SizedIter[Range, n]:
   return SizedIter(Range(0, stop))

@guppy.overloaded
def range(stop: int) -> Range:
   return Range(0, stop)
```

Closes #610.

### DIFF
--- a/guppylang/definition/custom.py
+++ b/guppylang/definition/custom.py
@@ -261,12 +261,16 @@ class CustomCallChecker(ABC):
         self.node = node
         self.func = func
 
-    @abstractmethod
     def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
         """Checks the return value against a given type.
 
         Returns a (possibly) transformed and annotated AST node for the call.
         """
+        from guppylang.checker.expr_checker import check_type_against
+
+        expr, res_ty = self.synthesize(args)
+        subst, _ = check_type_against(res_ty, ty, self.node)
+        return expr, subst
 
     @abstractmethod
     def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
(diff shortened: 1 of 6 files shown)


guppylang/definition/custom.py
```diff
@@ -261,12 +261,16 @@ def _setup(self, ctx: Context, node: AstNode, func: CustomFunctionDef) -> None:
...
     def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
         """Checks the return value against a given type.
 
         Returns a (possibly) transformed and annotated AST node for the call.
         """
+        from guppylang.checker.expr_checker import check_type_against
+
+        expr, res_ty = self.synthesize(args)
+        subst, _ = check_type_against(res_ty, ty, self.node)
+        return expr, subst
```
mark-koch, line 273: Drive-by: Add a default implementation for `CustomCallChecker.check` that falls back to the synthesizing case if we can't make use of the incoming type information

qartik (commented):
Overall LGTM, it might be nice to add some tests as well.

mark-koch:
`range` is already tested in `tests/integration/test_range.py`, but I added new tests for non-static range lengths and one that checks that the inferred size annotation is correct

qartik (approved):

PR #614, opened 2024-11-04
feat: Type check array comprehensions

mark-koch:


guppylang/prelude/builtins.py
```diff
@@ -92,7 +92,7 @@ class nat:
 class array(Generic[_T, _n]):
     """Class to import in order to use arrays."""
 
-    def __init__(self, *args: _T):
+    def __init__(self, *args: Any):
```
mark-koch, line 95: To prevent mypy from complaining

guppylang/prelude/_internal/checker.py
```diff
@@ -195,47 +206,107 @@ class NewArrayChecker(CustomCallChecker):
...
+                for i in range(len(rest)):
+                    rest[i], subst = checker.check(rest[i], ty)
+                    assert len(subst) == 0, "Array element type is closed"
+                result_ty = array_type(ty, len(args))
+                call = GlobalCall(
+                    def_id=self.func.id, args=[fst, *rest], type_args=result_ty.args
+                )
+                return with_loc(self.node, call), result_ty
+            case args:
+                return assert_never(args)  # type: ignore[arg-type]
```
mark-koch, line 232: Mypy isn't smart enough to infer that this code is unreachable, therefore `type: ignore[arg-type]`

aborgna-q (approved):

Issue #497, opened 2024-09-13
Rename `CompiledGlobals`


mark-koch:
A better name would be something like `CompilationContext` following the changes in #496

tatiana-s:
Done in #799

PR #616, opened 2024-11-04
feat: Compile array comprehensions to Hugr

mark-koch:
* Refactor compilation of list comprehensions so we can use the same logic for array comprehensions. The only difference is that lists call `push` to add the next element to the accumulator whereas arrays use `set` and update a counter. This is now implemented via a `build_update` hook provided to `_compile_generators`
* Store checked `Globals` inside the compilation context so we can invoke Guppy methods by name while building Hugr

### DIFF
--- a/guppylang/compiler/core.py
+++ b/guppylang/compiler/core.py
@@ -5,10 +5,12 @@ from typing import cast
 from hugr import Wire, ops
 from hugr.build.dfg import DP, DefinitionBuilder, DfBase
 
-from guppylang.checker.core import FieldAccess, Place, PlaceId, Variable
+from guppylang.checker.core import FieldAccess, Globals, Place, PlaceId, Variable
 from guppylang.definition.common import CheckedDef, CompilableDef, CompiledDef, DefId
+from guppylang.definition.ty import TypeDef
+from guppylang.definition.value import CompiledCallableDef
 from guppylang.error import InternalGuppyError
-from guppylang.tys.ty import StructType
+from guppylang.tys.ty import StructType, Type
 
 CompiledLocals = dict[PlaceId, Wire]
 
@@ -26,15 +28,19 @@ class CompiledGlobals:
     compiled: dict[DefId, CompiledDef]
     worklist: set[DefId]
 
+    checked_globals: Globals
+
     def __init__(
         self,
         checked: dict[DefId, CheckedDef],
         module: DefinitionBuilder[ops.Module],
+        checked_globals: Globals,
     ) -> None:
         self.module = module
         self.checked = checked
         self.worklist = set()
         self.compiled = {}
+        self.checked_globals = checked_globals
 
     def __getitem__(self, def_id: DefId) -> CompiledDef:
         if def_id not in self.compiled:
@@ -47,6 +53,16 @@ class CompiledGlobals:
         if isinstance(defn, CompilableDef):
             return defn.compile_outer(self.module)
         return defn
+
+    def get_instance_func(
+        self, ty: Type | TypeDef, name: str
+    ) -> CompiledCallableDef | None:
+        checked_func = self.checked_globals.get_instance_func(ty, name)
+        if checked_func is None:
+            return None
+        compiled_func = self[checked_func.id]
+        assert isinstance(compiled_func, CompiledCallableDef)
+        return compiled_func
 
 
 @dataclass
--- a/guppylang/compiler/expr_compiler.py
+++ b/guppylang/compiler/expr_compiler.py
@@ -1,5 +1,5 @@
 import ast
-from collections.abc import Iterable, Iterator, Sequence
+from collections.abc import Callable, Iterable, Iterator, Sequence
 from contextlib import contextmanager
 from typing import Any, TypeGuard, TypeVar
 
@@ -15,16 +15,21 @@ from hugr.build.dfg import DP, DfBase
 from hugr.std.collections import ListVal
 from typing_extensions import assert_never
 
-from guppylang.ast_util import AstVisitor, get_type, with_loc, with_type
+from guppylang.ast_util import AstNode, AstVisitor, get_type
 from guppylang.cfg.builder import tmp_vars
 from guppylang.checker.core import Variable
 from guppylang.checker.linearity_checker import contains_subscript
 from guppylang.compiler.core import CompilerBase, DFContainer
 from guppylang.compiler.hugr_extension import PartialOp
 from guppylang.definition.custom import CustomFunctionDef
-from guppylang.definition.value import CompiledCallableDef, CompiledValueDef
+from guppylang.definition.value import (
+    CallReturnWires,
+    CompiledCallableDef,
+    CompiledValueDef,
+)
 from guppylang.error import GuppyError, InternalGuppyError
 from guppylang.nodes import (
+    DesugaredArrayComp,
     DesugaredGenerator,
     DesugaredListComp,
     FieldAccessAndDrop,
@@ -39,12 +44,16 @@ from guppylang.nodes import (
     TensorCall,
     TypeApply,
 )
+from guppylang.prelude._internal.compiler.array import (
+    array_new_uninitialized,
+)
 from guppylang.prelude._internal.compiler.list import (
     list_new,
-    list_push,
 )
+from guppylang.tys.arg import Argument
 from guppylang.tys.builtin import (
     get_element_type,
+    int_type,
     is_bool_type,
     is_list_type,
 )
@@ -57,6 +66,7 @@ from guppylang.tys.ty import (
     InputFlags,
     NoneType,
     NumericType,
+    OpaqueType,
     TupleType,
     Type,
     type_to_row,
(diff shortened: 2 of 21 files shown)


guppylang/prelude/_internal/compiler/array.py
```diff
@@ -67,6 +68,14 @@ def array_set(elem_ty: ht.Type, length: int) -> ops.ExtOp:
     )
 
 
+def array_new_uninitialized(elem_ty: ht.Type, length: int) -> ops.ExtOp:
+    """Returns an array `uninitialized` operation."""
+    # TODO
+    return UnsupportedOp(
+        op_name="array.uninitialized", inputs=[], outputs=[array_type(elem_ty, length)]
+    ).ext_op
```
mark-koch, line 76: See https://github.com/CQCL/hugr/issues/1627

guppylang/compiler/expr_compiler.py
```diff
@@ -452,68 +462,117 @@ def visit_ResultExpr(self, node: ResultExpr) -> Wire:
...
+        self.dfg[count_var] = self.builder.load(
+            hugr.std.int.IntVal(0, width=NumericType.INT_WIDTH)
+        )
+
+        def build_update(elt: Wire) -> None:
+            """Callback for putting an element into the array."""
+            array, count = self.dfg[array_var], self.dfg[count_var]
+            (self.dfg[array_var],) = self._build_method_call(
+                array_ty, "__setitem__", node, [array, count, elt], array_ty.args
+            ).inout_returns
```
acl-cqc, line 505: very-optional, probably for another PR - but how about `all_returns` that gives back a tuple of both inout and regular, so you can check the other is empty in your (non-exhaustive) patterm-match, and so you can't accidentally forget which one to call?
(I slightly dislike that which of `inout_returns` and `regular_returns` you want, depends on the *sugar*/style with which the function was declared, rather than the actual functionality, right?)

guppylang/compiler/core.py
```diff
@@ -48,6 +54,16 @@ def _compile(self, defn: CheckedDef) -> CompiledDef:
             return defn.compile_outer(self.module)
         return defn
 
+    def get_instance_func(
+        self, ty: Type | TypeDef, name: str
+    ) -> CompiledCallableDef | None:
+        checked_func = self.checked_globals.get_instance_func(ty, name)
```
acl-cqc, line 60: This does feel like some refactoring could help, perhaps after the array stuff is all merged in; all we actually want from the checked-globals is the mapping from name to id...

acl-cqc (commented):
Some initial thoughts, am still thinking about the interface for `build_update`.

guppylang/compiler/core.py
```diff
@@ -26,15 +28,19 @@ class CompiledGlobals:
     compiled: dict[DefId, CompiledDef]
     worklist: set[DefId]
 
+    checked_globals: Globals
```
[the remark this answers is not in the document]
acl-cqc: And/or see [other comment](https://github.com/CQCL/guppylang/pull/616#discussion_r1829671780)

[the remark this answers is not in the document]
mark-koch: Agreed, we should do some refactoring here. Also see #497

guppylang/compiler/expr_compiler.py
mark-koch: It's a `NamedTuple`, so we can actually unpack it directly

guppylang/compiler/core.py
mark-koch: We also need the mapping from class definitions to their methods

guppylang/compiler/expr_compiler.py
```diff
@@ -452,68 +462,117 @@ def visit_ResultExpr(self, node: ResultExpr) -> Wire:
...
+        assert isinstance(gen.iter, PlaceNode)
+        assert isinstance(gen.hasnext, PlaceNode)
+        inputs = [gen.iter] + [PlaceNode(place=var) for var in loop_vars]
+        with self._new_loop(inputs, gen.hasnext):
+            # If there is a next element, compile it and continue with the next
+            # generator
+            compiler.compile_stmts([gen.hasnext_assign], self.dfg)
+            with self._if_true(gen.hasnext, inputs):
+
+                def compile_ifs(ifs: list[ast.expr]) -> None:
```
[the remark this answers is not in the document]
mark-koch: Very nice, I have never used that. Makes this function a lot cleaner 👍

guppylang/compiler/expr_compiler.py
mark-koch: > I slightly dislike that which of inout_returns and regular_returns you want, depends on the sugar/style with which the function was declared, rather than the actual functionality, right?

Yes, but I don't see a way around that...

guppylang/compiler/expr_compiler.py
```diff
@@ -452,68 +463,104 @@ def visit_ResultExpr(self, node: ResultExpr) -> Wire:
...
+        method: str,
+        node: AstNode,
+        args: list[Wire],
+        type_args: Sequence[Argument] | None = None,
+    ) -> CallReturnWires:
+        func = self.globals.get_instance_func(ty, method)
+        assert func is not None
+        return func.compile_call(args, type_args or [], self.dfg, self.globals, node)
+
+    def _compile_generators(
```
acl-cqc, line 530: Ok, I was just wondering whether you could use `ExitStack` to make the whole `_compile_generators` nonrecursive too....and you'd gone and done it :-D :-) Good job! 👍 

Next...I am wondering if you can turn the whole `_compile_generators` into a context manager. The `build_update` callback is a bit restrictive (lots of hidden state because `build_update` has a fixed single-Wire interface, etc.), so rather than `self._compile_generators(node.elt, node.generators, [list_place], build_update)`, you could call it by
```
with _compile_generators(self, node.generators, [list_place]) as _:
  build_update(node.elt)
```
....which would allow inlining `build_update` at the use site.

For the implementation of `_compile_generators`, see the [`@contextmanager`](https://docs.python.org/3/library/contextlib.html#contextlib.contextmanager) decorator - you'd put your `yield` in place of the existing call to the `build_update` Callable.

guppylang/compiler/expr_compiler.py
```diff
@@ -452,68 +463,104 @@ def visit_ResultExpr(self, node: ResultExpr) -> Wire:
...
+        compiler = StmtCompiler(self.globals)
+        with ExitStack() as stack:
+            for gen in gens:
+                # Build the generator
+                compiler.compile_stmts([gen.iter_assign], self.dfg)
+                assert isinstance(gen.iter, PlaceNode)
+                assert isinstance(gen.hasnext, PlaceNode)
+                inputs = [gen.iter] + [PlaceNode(place=var) for var in loop_vars]
+                # Remember to finalize the iterator once we are done with it. Note that
+                # we need to use partial in the callback, so that we bind the *current*
```
acl-cqc, line 549: This is the seriously awkward bit, which of course I hadn't spotted when I suggested it ;-), well done!

tests/integration/test_array_comprehension.py
```diff
@@ -0,0 +1,79 @@
...
+    def test() -> array[qubit, 42]:
+        return array(qubit() for _ in range(42))
+
+    validate(module.compile())
+
+
+def test_zero_length(validate):
+    @compile_guppy
+    def test() -> array[int, 0]:
+        return array(i for i in range(0))
```
acl-cqc, line 35:
```suggestion
        return array(i/0 for i in range(0))
```
I.e. something that would runtime panic if it were ever actually evaluated

tests/integration/test_array_comprehension.py
```diff
@@ -0,0 +1,79 @@
...
+
+    validate(test)
+
+
+def test_capture(validate):
+    @compile_guppy
+    def test(x: int) -> array[int, 42]:
+        return array(i + x for i in range(42))
```
acl-cqc, line 43: Would be good to make this one an execution test to check it does actually capture the correct value of `x`....

tests/integration/test_array_comprehension.py
```diff
@@ -0,0 +1,79 @@
...
+
+    validate(test)
+
+
+def test_nested_left(validate):
+    @compile_guppy
+    def test() -> array[array[int, 10], 20]:
+        return array(array(x + y for y in range(10)) for x in range(20))
+
+    validate(test)
```
acl-cqc, line 79: This might be good to sum the whole lot too

acl-cqc (commented):
I'd approve now, but I'm just wondering whether the contextmanager trick works (i.e. I might have other ideas if it doesn't)....

Looks good, thanks @mark-koch !

tests/integration/test_array_comprehension.py
mark-koch: Agreed, but unfortunately it doesn't run yet until we have https://github.com/CQCL/hugr/issues/1627 and corresponding lowering in hugr-llvm.

tests/integration/test_array_comprehension.py
mark-koch: Array comprehensions with multiple generators aren't supported yet since we can't express arrays of size `n * m` if `n` or `m` are generic arguments.

guppylang/compiler/expr_compiler.py
mark-koch: Love the idea! 👍

guppylang/compiler/expr_compiler.py
```diff
@@ -452,68 +463,96 @@ def visit_ResultExpr(self, node: ResultExpr) -> Wire:
...
+        return self.dfg[list_place]
+
+    def visit_DesugaredArrayComp(self, node: DesugaredArrayComp) -> Wire:
+        # Allocate an uninitialised array of the desired size and a counter variable
+        array_ty = get_type(node)
+        assert isinstance(array_ty, OpaqueType)
+        array_var = Variable(next(tmp_vars), array_ty, node)
+        count_var = Variable(next(tmp_vars), int_type(), node)
+        hugr_elt_ty = (
+            ht.Option(node.elt_ty.to_hugr())
```
acl-cqc, line 487: It might possibly be worth making this into a method that you call `node.elt_ty.array_element_type()`

tests/integration/test_array_comprehension.py
acl-cqc: can't we do
```
@compile_guppy
def test2() -> int:
  arr = test()
  total = 0
  for inner in arr:
    for elem in inner:
      total += elem
  return total

...
run_int_fn(test2, ....)
```

can we only iterate over arrays in a comprehension?

If we have 1d `sum` only (of type `array[int,_] -> int`) then you can lift to 2d via a comprehension `sum(array(sum(inner) for inner in test()))`

modulo there are no execution tests!

guppylang/compiler/expr_compiler.py
mark-koch: I think this will be redundant following #630

tests/integration/test_array_comprehension.py
mark-koch: Array iteration is added in PR #632, but not yet available here

PR #619, opened 2024-11-04
refactor: Refactor `GuppyModule.compile` to get rid of side effects

### LINKED ISSUE #574: Refactor `GuppyModule.compile` and `CompiledGlobals` to get rid of evil side effect

mark-koch:
Closes #574

### DIFF
--- a/guppylang/compiler/core.py
+++ b/guppylang/compiler/core.py
@@ -48,6 +48,19 @@ class CompiledGlobals:
             return defn.compile_outer(self.module)
         return defn
 
+    def compile(self, defn: CheckedDef) -> None:
+        """Compiles the given definition and all of its dependencies into the current
+        Hugr."""
+        if defn.id in self.compiled:
+            return
+
+        self.compiled[defn.id] = self._compile(defn)
+        self.worklist.add(defn.id)
+        while self.worklist:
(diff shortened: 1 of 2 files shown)


guppylang/compiler/core.py
```diff
@@ -48,6 +48,19 @@ def _compile(self, defn: CheckedDef) -> CompiledDef:
...
+        """Compiles the given definition and all of its dependencies into the current
+        Hugr."""
+        if defn.id in self.compiled:
+            return
+
+        self.compiled[defn.id] = self._compile(defn)
+        self.worklist.add(defn.id)
+        while self.worklist:
+            next_id = self.worklist.pop()
+            next_def = self[next_id]
```
ss2165, line 61: is there still compilation happening inside `__getitem__`? I don't see any change to that behaviour here

mark-koch: Kind of. The `__getitem__` call only builds the empty function def and registers it in the worklist. Imo that's fine, but I can turn it into an explicit method call if you prefer?

ss2165: yes please, mutating __getitem__ is best avoided if possible

ss2165 (approved):
thanks!

PR #621, opened 2024-11-05
feat: Update statement checker to use new diagnostics

### LINKED ISSUE #540: Update statement checker to use new diagnostics

mark-koch:
Closes #540

### DIFF
--- a/tests/error/struct_errors/invalid_attribute_assign1.err
+++ b/tests/error/struct_errors/invalid_attribute_assign1.err
@@ -3,6 +3,6 @@ Error: Attribute not found (at $FILE:15:6)
 13 | @guppy(module)
 14 | def foo(s: MyStruct) -> None:
 15 |     s.z = 2
-   |       ^ Attribute `z` not found on type `int`
+   |       ^ `MyStruct` has no attribute `z`
 
 Guppy compilation failed due to 1 previous error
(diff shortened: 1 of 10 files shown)


tests/error/struct_errors/invalid_attribute_assign1.err
```diff
@@ -1,7 +1,8 @@
-Guppy compilation failed. Error in file $FILE:15
+Error: Attribute not found (at $FILE:15:6)
+   | 
+13 | @guppy(module)
+14 | def foo(s: MyStruct) -> None:
+15 |     s.z = 2
+   |       ^ Attribute `z` not found on type `int`
```
qartik, line 6: Would it be an easier read if this error was:
```suggestion
   |       ^ Attribute `z` of type `int` not found
```
or even better
```suggestion
   |       ^ `MyStruct` has no attribute `z` of type `int`
```

I actually don't see the relevance of the inferred type here, the error message on the left is quite descriptive as is. Perhaps I am missing some context here.

qartik (commented):
The error messages look way more friendly now! Had a question about one of them. Unsure if it's even part of this PR.

tests/error/struct_errors/invalid_attribute_assign1.err
mark-koch: Oops yes it's supposed to say "MyStruct" instead of "int" 👍

qartik (approved):

PR #622, opened 2024-11-05
fix: Mock guppy decorator during sphinx builds

### LINKED ISSUE #583: Sphinx builds are failing
Maybe an OOM issue?

mark-koch:
Fixes #583

The problem was that sphinx autodoc got confused by `@guppy` decorated functions. Now, we mock the decorator when we detect that sphinx is running.


ss2165 (approved):
😬

PR #613, opened 2024-11-04
feat: Array comprehension

mark-koch:
Adds array comprehensions of the form `array(i for i in range(10))`.

This is the base PR that adds new AST nodes for array comprehensions and general generator expressions. Checking and lowering logic follows in subsequent PRs targetting this branch:
* #614
* #615 
* #616

Closes #612

PR #632, opened 2024-11-07
feat: Make arrays iterable

### LINKED ISSUE #617: Make arrays iterable

mark-koch:
* Add `ArrayIter` type to the standard library
* Add `array.__iter__` method

Drive by: Fix array method compilers to handle generic array lengths

Note: The `ArrayIter.__end__` method depends on a new `array_map` Hugr op that doesn't exist yet to discard linear arrays

Closes #617


guppylang/checker/expr_checker.py
```diff
@@ -544,7 +545,8 @@ def synthesize_instance_func(
                 f"expected `{exp_sig}`",
                 node,
             )
-        return func.synthesize_call([node, *args], node, self.ctx)
+        result, result_ty = func.synthesize_call([node, *args], node, self.ctx)
+        return with_type(result_ty, result), result_ty
```
mark-koch, line 549: Drive-by: Annotate the AST node with the inferred type to avoid repeated checking

guppylang/decorator.py
```diff
@@ -277,14 +277,16 @@ def type_var(
...
         """Creates a new const nat variable in a module."""
         module = module or self.get_module()
         defn = ConstVarDef(
             DefId.fresh(module), name, None, NumericType(NumericType.Kind.Nat)
         )
         module.register_def(defn)
-        return defn
+        # Return an actual Python `TypeVar` so it can be used as an actual type in code
+        # that is executed by interpreter before handing it to Guppy.
+        return TypeVar(name)
```
mark-koch, line 289: Otherwise, we couldn't write `Generic[T, n]`

guppylang/prelude/_internal/compiler/array.py
```diff
@@ -207,3 +225,30 @@ def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
...
+        # been taken out (i.e. replaced by `None`)
+        if self.elem_ty.type_bound() == ht.TypeBound.Any:
+            elem_opt_ty = ht.Option(self.elem_ty)
+            none_ty = ht.UnitSum(1)
+            # Define `unwrap_none` function
+            func = self.builder.define_function("unwrap_none", [elem_opt_ty], [none_ty])
+            err_msg = "Linear array element has not been used in iterator"
+            build_expect_none(
+                cast(DfBase[ops.DfParentOp], func), func.inputs()[0], err_msg
+            )
```
mark-koch, line 244: Annoying that we have this `cast` here.

`func` is a `Function` which inherits `DfBase[ops.FuncDefn]` and `ops.FuncDefn` inherits `ops.DfParentOp`. The problem is that `DfBase` is not covariant, so this cast is technically not safe. But I also don't want to define `build_expect_none` only for `DfBase[ops.FuncDefn]`. Any ideas?

guppylang/tys/arg.py
```diff
@@ -81,7 +81,7 @@ def to_hugr(self) -> ht.TypeArg:
             case ConstValue(value=v, ty=NumericType(kind=NumericType.Kind.Nat)):
                 assert isinstance(v, int)
                 return ht.BoundedNatArg(n=v)
-            case BoundConstVar(idx=idx):
+            case BoundConstVar(idx=idx, ty=NumericType(kind=NumericType.Kind.Nat)):
```
mark-koch, line 84: Drive-by: This case should only trigger for `nat` type args

guppylang/prelude/_internal/compiler/array.py
```diff
@@ -207,3 +225,30 @@ def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
...
         raise InternalGuppyError("Call compile_with_inouts instead")
+
+
+class ArrayIterEndCompiler(ArrayCompiler):
+    """Compiler for the `ArrayIter.__end__` method."""
+
+    def compile(self, args: list[Wire]) -> list[Wire]:
+        # For linear array iterators, we need to make sure that all array elements have
+        # been taken out (i.e. replaced by `None`)
+        if self.elem_ty.type_bound() == ht.TypeBound.Any:
```
acl-cqc, line 236: Isn't there something like `self.elem_ty.is_linear()` you can use here?

guppylang/prelude/_internal/compiler/prelude.py
```diff
@@ -123,3 +123,12 @@ def build_unwrap(
     result is an error.
     """
     return build_unwrap_right(builder, result, error_msg, error_signal)
+
+
+def build_expect_none(
+    builder: DfBase[ops.DfParentOp], result: Wire, error_msg: str, error_signal: int = 1
```
acl-cqc, line 129: Suggest rename `result` -> `input` (or input_option if you like)

tests/integration/test_array.py
```diff
@@ -247,6 +247,19 @@ def main() -> tuple[array[int, 3], array[qubit, 2]]:
...
+    module = GuppyModule("test")
+    module.load(qubit, discard)
+
+    @guppy(module)
+    def main() -> None:
+        qs = array(qubit(), qubit(), qubit())
+        for q in qs:
+            discard(q)
+
+    validate(module.compile())
```
acl-cqc, line 260: I am just wondering here if you could check that the module *does* include the expect_none stuff, and then have a counterexample
```
qs = array(qubit(), qubit(), qubit())
[discard(q) for q in qs]
```
that does not need expect none (because the array returned by the comprehension is copyable)....I think??

acl-cqc (commented):
Looks good, a few suggestions, possibly some answers, and a few questions...

guppylang/prelude/_internal/compiler/array.py
```diff
@@ -2,51 +2,59 @@
...
 # ------------------------------------------------------
 # --------------- std.array operations -----------------
 # ------------------------------------------------------
 
 
 def _instantiate_array_op(
-    name: str, elem_ty: ht.Type, length: int, inp: list[ht.Type], out: list[ht.Type]
+    name: str,
+    elem_ty: ht.Type,
+    length: ht.TypeArg,
```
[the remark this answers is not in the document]
mark-koch: Exactly 👍

guppylang/prelude/_internal/compiler/array.py
```diff
@@ -85,11 +103,11 @@ def elem_ty(self) -> ht.Type:
                 raise InternalGuppyError("Invalid array type args")
 
     @property
-    def length(self) -> int:
+    def length(self) -> ht.TypeArg:
         """The length for the array op that is being compiled."""
         match self.type_args:
-            case [_, ConstArg(ConstValue(value=int(length)))]:
-                return length
+            case [_, ConstArg(const=const)]:
```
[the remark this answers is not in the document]
mark-koch: `ConstArg` here means "compile-time constant", so including literals and variables. Terminology borrowed from rustc: https://doc.rust-lang.org/nightly/nightly-rustc/rustc_middle/ty/consts/type.ConstKind.html

guppylang/prelude/_internal/compiler/array.py
mark-koch: I don't think that exists: https://cqcl.github.io/hugr/generated/hugr.tys.Type.html

guppylang/prelude/builtins.py
```diff
@@ -560,6 +561,35 @@ def __len__(self: array[L, n]) -> int: ...
...
+
+@guppy.struct
+class ArrayIter(Generic[L, n]):
+    xs: array[L, n]
+    i: int
+
+    @guppy
+    @no_type_check
+    def __hasnext__(self: "ArrayIter[L, n]" @ owned) -> tuple[bool, "ArrayIter[L, n]"]:
+        return self.i < int(n), self
```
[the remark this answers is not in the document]
mark-koch: This was added in #618

guppylang/prelude/builtins.py
```diff
@@ -560,6 +561,35 @@ def __len__(self: array[L, n]) -> int: ...
...
+    def __next__(self: "ArrayIter[L, n]" @ owned) -> tuple[L, "ArrayIter[L, n]"]:
+        elem = _array_unsafe_getitem(self.xs, self.i)
+        return elem, ArrayIter(self.xs, self.i + 1)
+
+    @guppy.custom(ArrayIterEndCompiler())
+    def __end__(self: "ArrayIter[L, n]" @ owned) -> None: ...
+
+
+@guppy.custom(ArrayGetitemCompiler())
+def _array_unsafe_getitem(xs: array[L, n], idx: int) -> L: ...
```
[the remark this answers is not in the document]
mark-koch: Not on guppy level since we don't have an option type yet

tests/integration/test_array.py
mark-koch: >  (because the array returned by the comprehension is copyable)....I think??

Yes, indeed! Unfortunately we can't add this test since #613 isn't merged yet

guppylang/prelude/_internal/compiler/array.py
acl-cqc: Oh this is not guppy-specific anymore, right

guppylang/prelude/_internal/compiler/array.py
acl-cqc: See #676

guppylang/prelude/_internal/compiler/array.py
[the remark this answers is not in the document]
acl-cqc: Might be worth a little comment `# Inc. literals and variables` - it's not super-obvious that a Const could include a variable!!

acl-cqc (approved):
Sorry for dropping the ball on this one! Looks good to me, modulo I'm hoping #676 fixes that `cast` issue

PR #647, opened 2024-11-14
refactor: use `OutcomeArray.from_ints` in `HShots.to_pytket`

ss2165:
just a drive by I happened to do while debugging something else. Performance gains are not a bottleneck but it is more readable so might as well.


aborgna-q (approved):
LGTM

PR #653, opened 2024-11-15
feat: mem_swap function for swapping two inout values

### LINKED ISSUE #652: builtin function for swapping two inout arguments
`def swap(x: L, y: L) -> None  where L = guppy.type_var("L", linear=True)`

ss2165:
not sure if prelude is best place for the custom compiler

also not sure about name `mem_swap` but don't want to confuse with quantum logic swap

...And not sure about the test location either.

Closes #652

// to coerce release-please: 

Release-As: 0.13.1


mark-koch (approved):
I think putting it in `prelude.py` is ok, the name also sounds good 👍

PR #662, opened 2024-11-25
fix: Fix implicit modules in IPython shells

### LINKED ISSUE #661: Compiled module's `main` function misnamed in ipython shell.
When compiling guppy in an ipython shell, instead of the main function being called `main`, it is called something like `<ipython-input-3-3e9b5833de21>` in the compiled hugr.

For example:
```
alec@QuantUK-6R64C14:~/r/guppy-integration$ uv run ipython
Python 3.12.3 (main, Nov  6 2024, 18:32:19) [GCC 13.2.0]
Type 'copyright', 'credits' or 'license' for more information
IPython 8.29.0 -- An enhanced Interactive Python. Type '?' for help.

In [1]: from guppylang import guppy
   ...: from guppylang.std.builtins import result
   ...: from guppylang.std.quantum import h, measure, qubit

In [2]: @guppy
   ...: def main() -> None:
   ...:     q0 = qubit()
   ...:     h(q0)
   ...:     result("m", measure(q0))
   ...: 

In [3]: out = guppy.compile_module()

In [4]: "main" in str(out)
Out[4]: False
```

mark-koch:
Fixes #661


guppylang/ipython_inspect.py
```diff
@@ -12,6 +12,18 @@ def is_running_ipython() -> bool:
...
+def is_ipython_dummy_file(filename: str) -> bool:
+    """Checks whether a given filename is a dummy name generated for an IPython cell."""
+    # TODO: The approach below is false-positive prone. Figure out if there is a better
+    #  way to do this.
+    return (
+        # IPython cells have filenames like "<ipython-input-3-3e9b5833de21>"
+        filename.startswith("<ipython-input-")
+        # Jupyter cells have filenames like "/var/{...}/ipykernel_82076/61218616.py"
+        or "ipykernel_" in filename
+    )
```
mark-koch, line 24: I'm not really happy with this, but I don't know a better way to solve this problem :/

cqc-alec:
(The only difference between the generated hugr when run in ipython with this fix and when run as a python script is that the Module metadata is `{'name': '<jupyter-notebook>'}` versus `{'name': '__main__'}`.)

ss2165 (approved):
:/

Issue #671, opened 2024-11-29
Support arrays in `pytket` function stubs


tatiana-s:
See #657 for initial issue.

Instead of having to write 
```
@guppy.pytket(circ)
def guppy_circ(q1: qubit, q2: qubit) -> None: ...
```
we want to be able to write
```
@guppy.pytket(circ)
def guppy_circ(array[qubit, 2]) -> None: ...
```

and maybe more complicated combinations of both of the above.

NathanCQC:
Big dog feature

Issue #670, opened 2024-11-29
Load `pytket` functions without explicit function stub


tatiana-s:
See #657 for initial issue.

Add a `load_pytket` method. One complication here is finding the call location for error reporting (see `_parse_expr_string`  for `extern` method which does something similar).

Issue #669, opened 2024-11-29
Support circuits with classical bits and measurements in `pytket` circuit functions


tatiana-s:
See #657 for initial issue.

The current implementation leads to port errors as the guppy function inputs and outputs differ from the circuit hugr inputs and outputs. This is complicated by the ownership system, as measurements require the qubits to be owned rather than borrowed, and writing back results to bits should lead to those bools being returned from the function.

As a start, we could allow measurements if they happen in the very end but finding the output node and then looking if there is a measure node directly before it - then modify the function to own those qubits and return bools for the results.

PR #672, opened 2024-11-29
feat: Load `pytket` circuit as a function definition

### LINKED ISSUE #657: Load circuit as a function definition
So that function value execution isn't required

Use uuids for circuits - can use circuit names or custom names in future once we have #563 and use that for circuits

tatiana-s:
Closes #657, however the following limitations still exist which are covered by further issues:
-  #669: ~~Classical bits and measurements aren't currently supported as this requires some checks related to ownership and return values.~~ They are supported but without taking ownership.
- #670: It would be nice to not have to provide an explicit function stub for the circuit and also have  `load_pytket` method which infers it.
- #671: It would also be nice to be able to provide arrays of qubits in the function stub.


guppylang/definition/pytket_circuits.py
```diff
@@ -0,0 +1,262 @@
...
+            return CallReturnWires(
+                regular_returns=[],
+                inout_returns=list(call[0:]),
+            )
+        else:
+            # Circuit function returns are the other way round to Guppy returns.
+            return CallReturnWires(
+                regular_returns=list(call[-num_returns:]),
+                inout_returns=list(call[:-num_returns]),
+            )
```
mark-koch, line 262: Consider doing
```suggestion
        num_inouts = len(list(call)) - num_returns
        return CallReturnWires(
            regular_returns=list(call[inout_returns:]),
            inout_returns=list(call[:inout_returns]),
        )
```
to avoid the 0 edge case

tests/integration/test_pytket_circuits.py
```diff
@@ -0,0 +1,137 @@
...
+
+    module = GuppyModule("test")
+    module.load_all(quantum)
+
+    @guppy.pytket(circ, module)
+    def guppy_circ(q: qubit) -> bool: ...
+
+    @guppy(module)
+    def foo(q: qubit) -> None:
+        result = guppy_circ(q)
```
mark-koch, line 72:
```suggestion
    def foo(q: qubit) -> bool:
        return guppy_circ(q)
```

This way, validation checks that the bool is wired correctly

mark-koch (commented):
Looks good! Mostly just some nits and suggestions.

It looks like the coverage report doesn't include the test you've written for some reason. It should be running with tket2 installed, so no idea what's going on there.

Speaking of tests, could you also add some error tests for every `GuppyError` that is thrown. See the `tests/error` directory

guppylang/definition/pytket_circuits.py
```diff
@@ -0,0 +1,262 @@
...
+            if isinstance(self.input_circuit, pytket.circuit.Circuit):
+                from tket2.circuit import (  # type: ignore[import-untyped, import-not-found, unused-ignore]
+                    Tk2Circuit,
+                )
+
+                circ = Hugr.load_json(Tk2Circuit(self.input_circuit).to_hugr_json())  # type: ignore[attr-defined, unused-ignore]
+                mapping = module.hugr.insert_hugr(circ)
+                hugr_func = mapping[circ.root]
+
+                # We need to remove input bits from both signature and input node.
```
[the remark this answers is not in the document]
tatiana-s: If I don't remove them, there is an runtime error while validating hugr along the lines of `The node Node(...) has an unconnected port Port(Incoming, 1) of type []+[].` where the node in question is `Call` (if I counted correctly) - I think this could maybe be fixed if we also explicitly add `bool`s to the function stub inputs, so it depends onwards whether we want that.

tatiana-s:
Perhaps related to the question of why the coverage report didn't run for the tests which require tket2, how do I make a error test run with tket2 (since it seems looking at the `tket2_not_installed` test not having it installed is the default)?

tatiana-s:
The reason coverage is not taking the tests into account is because `-- extra pytket` isn't actually installing tket2 - there is some issue with clashing versions when trying to install tket2 which is why it was commented out I assume.

guppylang/definition/pytket_circuits.py
```diff
@@ -154,26 +151,29 @@ def compile_outer(self, module: DefinitionBuilder[OpVar]) -> "CompiledPytketDef"
...
                     node_data = module.hugr.get(child)
-                    if node_data and isinstance(node_data.op, Input):
+                    if node_data and isinstance(node_data.op, ops.Input):
                         input_types = node_data.op.types
-                        num_bools = input_types.count(Bool)
+                        num_bools = input_types.count(tys.Bool)
                         for _ in range(num_bools):
-                            input_types.remove(Bool)
+                            bool_node = outer_func.load(val.FALSE)
+                            bool_wires.append(*bool_node.outputs())
```
mark-koch:
```suggestion
                bool_wires = [outer_func.load(val.FALSE) for _ in range(self.input_circuit.n_bits)]
```

guppylang/definition/pytket_circuits.py
```diff
@@ -154,26 +151,29 @@ def compile_outer(self, module: DefinitionBuilder[OpVar]) -> "CompiledPytketDef"
...
+
+                call_node = outer_func.call(
+                    hugr_func, *(list(outer_func.inputs()) + bool_wires)
+                )
+                # Pytket circuit hugr has qubit and bool wires in the opposite order.
+                outer_func.set_outputs(*reversed(list(call_node.outputs())))
```
mark-koch, line 175: This also reverses the order inside the qubit and bit registers themselves

guppylang/checker/errors/py_errors.py
```diff
@@ -58,13 +58,5 @@ class InstallInstruction(Help):
 @dataclass(frozen=True)
 class PytketSignatureMismatch(Error):
     title: ClassVar[str] = "Signature mismatch"
-    span_label: ClassVar[str] = (
-        "Function signature {name} doesn't match provided pytket circuit"
-    )
+    span_label: ClassVar[str] = "Signature {name} doesn't match provided pytket circuit"
```
mark-koch, line 61:
```suggestion
    span_label: ClassVar[str] = "Signature `{name}` doesn't match provided pytket circuit"
```

mark-koch (approved):
Looks good, thank you!

PR #676, opened 2024-12-02
Generalize build_{expect_none,unwrap_{left,right}}, remove cast

acl-cqc:


mark-koch (approved):
Thank you!

The CI problem with hugr-llvm is fixed on main, don't worry about it

PR #678, opened 2024-12-02
ci: Open an issue when building docs fails on `main`

aborgna-q:
Based on hugr's [unsoundness check](https://github.com/CQCL/hugr/blob/main/.github/workflows/unsoundness.yml#L43-L56) notification.


mark-koch (approved):
Thank you, very useful!

PR #679, opened 2024-12-02
feat!: qsystem std functions with updated primitives

### LINKED ISSUE #625: Get rid of `measure_return`
Replace it with `measure_reset`

### LINKED ISSUE #645: Add `project_z` and `hseries.measure_and_reset` functions
To precisely capture the semantics of the tket2 measurement operations and the hseries specific MeasureReset operation:

| Guppy function            | semantics                                                                                                       | Guppy signature      | Hugr implementation    |
| ------------------------- | --------------------------------------------------------------------------------------------------------------- | -------------------- | ---------------------- |
| measure                   | Measure qubit and lose it                                                                                       | qubit -> bool        | T2::DestructiveMeasure |
| project_z                 | Projective z basis measurement                                                                                  | qubit -> qubit, bool | T2::Measure            |
| hseries.measure_and_reset | Measure a qubit and immediately reset it to 0 | qubit -> qubit, bool | HS::MeasureReset       |

Blocked by: 
https://github.com/CQCL/tket2/issues/691
https://github.com/CQCL/tket2/issues/692

Tracked by:
https://github.com/CQCL-DEV/hugrverse/issues/77

ss2165:
Closes #625 Closes #645
- hseries renamed to qsystem

BREAKING CHANGE: hseries primitive functions moved to std.qsystem
BREAKING CHANGE: measure_return renamed to `project_z`

### DIFF
--- a/guppylang/std/quantum.py
+++ b/guppylang/std/quantum.py
@@ -159,46 +159,6 @@ def measure(q: qubit @ owned) -> bool:
     return res
 
 
-@guppy
-@no_type_check
-def phased_x(q: qubit, angle1: angle, angle2: angle) -> None:
-    f1 = float(angle1)
-    f2 = float(angle2)
-    _phased_x(q, f1, f2)
-
-
-@guppy
-@no_type_check
-def zz_phase(q1: qubit, q2: qubit, angle: angle) -> None:
-    f = float(angle)
-    _zz_phase(q1, q2, f)
-
-
 @guppy.hugr_op(quantum_op("Reset"))
 @no_type_check
 def reset(q: qubit) -> None: ...
-
-
-# ------------------------------------------------------
-# --------- Internal definitions -----------------------
-# ------------------------------------------------------
-
-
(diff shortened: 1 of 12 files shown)


guppylang/std/quantum.py
```diff
@@ -34,7 +33,7 @@ def measure(self: "qubit" @ owned) -> bool:
     @guppy
     @no_type_check
     def measure_return(self: "qubit") -> bool:
-        return measure_return(self)
+        return project_z(self)
```
qartik, line 36: Just a comment, no change needed. From this wrapper, it looks like `project_z` is not exposed directly to the user, so I think it's a nice rename. I wouldn't want users to have to use `project_z` call for measurements.

qartik (commented):
Overall LGTM.

Question: Is qsystem broadly agreed upon as a term to refer to quantinuum-specific hardware? Why not just call it quantinuum, let's say? I must be missing some context here.

qartik (approved):

PR #692, opened 2024-12-04
fix: Stop throwing NotImplementedError for unsupported expressions

### LINKED ISSUE #691: Stop throwing `NotImplementedError` for unsupported expressions
For example, compiling `return *xs` panics with a `NotImplementedError`. It should be a compiler error instead

mark-koch:
Fixes #691


acl-cqc (approved):
I'd call this "Properly report errors for unsupported expressions" but LGTM

PR #696, opened 2024-12-09
feat: Add Option type to standard library

### LINKED ISSUE #667: Add Option type
To map to HUGR option (sum of None, Some[T])

Ideally should be an interface we can in future replace with a guppy algebraic data type.

Functions:
- new None
- new Some(T)
- unwrap()
- is_none()
- is_some()

mark-koch:
Closes #667.

Also generalises the `@guppy.type` decorator to handle generic types.

### DIFF
--- a/guppylang/std/option.py
+++ b/guppylang/std/option.py
@@ -1,0 +1,61 @@
+from collections.abc import Sequence
+from typing import Generic, no_type_check
+
+import hugr.tys as ht
+
+from guppylang.decorator import guppy
+from guppylang.error import InternalGuppyError
+from guppylang.std._internal.compiler.option import (
+    OptionConstructor,
+    OptionTestCompiler,
+    OptionUnwrapCompiler,
+)
+from guppylang.std.builtins import owned
+from guppylang.tys.arg import Argument, TypeArg
+from guppylang.tys.param import TypeParam
+
+
+def _option_to_hugr(args: Sequence[Argument]) -> ht.Type:
+    match args:
+        case [TypeArg(ty)]:
+            return ht.Option(ty.to_hugr())
+        case _:
+            raise InternalGuppyError("Invalid type args for Option")
+
+
+T = guppy.type_var("T", linear=True)
+
+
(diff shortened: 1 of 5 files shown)


guppylang/std/option.py
```diff
@@ -0,0 +1,61 @@
...
+    def unwrap(self: "Option[T]" @ owned) -> T:
+        """Returns the contained `some` value, consuming `self`.
+
+        Panics if the option is a `none` value.
+        """
+
+
+@guppy.custom(OptionConstructor(0))
+@no_type_check
+def none() -> Option[T]:
```
ss2165, line 54: I thought you wanted to go with "Nothing" to future-proof? Or is the plan to overload the `None` keyword for this value? (And keep it for the unit type for type signatures?)

tests/integration/test_option.py
```diff
@@ -0,0 +1,36 @@
...
+from guppylang.std.option import Option, none, some
+
+
+def test_none(validate, run_int_fn):
+    module = GuppyModule("test_range")
+    module.load(Option, none)
+
+    @guppy(module)
+    def main() -> int:
+        x: Option[int] = none()
```
ss2165, line 12: is this annotation required or here for readability?

tests/integration/test_option.py
```diff
@@ -0,0 +1,36 @@
+from guppylang.decorator import guppy
+from guppylang.module import GuppyModule
+from guppylang.std.option import Option, none, some
+
+
+def test_none(validate, run_int_fn):
+    module = GuppyModule("test_range")
+    module.load(Option, none)
```
ss2165, line 8: bit inconvenient to have to load these separately

ss2165 (approved):
the inout stuff is very nice, thanks!

tests/integration/test_option.py
mark-koch: It's required, we can't infer the type

tests/integration/test_option.py
mark-koch: Yeah, but this is only an issue for explicit modules

PR #697, opened 2024-12-09
feat!: Turn py expression lists into arrays

### LINKED ISSUE #545: Lists in py expressions should be turned into arrays
We statically know their size. This information shouldn't be thrown away!

mark-koch:
Closes #545.

BREAKING CHANGE: Lists in `py(...)` expressions are now turned into Guppy arrays instead of lists.


ss2165 (approved):
thanks!

PR #699, opened 2024-12-10
fix: remove use of deprecated Ellipsis

ss2165:
Closes https://github.com/quantinuum-dev/guppy-integration/issues/33

### DIFF
--- a/guppylang/ast_util.py
+++ b/guppylang/ast_util.py
@@ -343,4 +343,8 @@ def has_empty_body(func_ast: ast.FunctionDef) -> bool:
     if len(func_ast.body) > 1:
         return False
     [n] = func_ast.body
-    return isinstance(n, ast.Expr) and isinstance(n.value, ast.Ellipsis)
+    return (
+        isinstance(n, ast.Expr)
+        and isinstance(n.value, ast.Constant)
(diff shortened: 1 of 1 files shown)


guppylang/ast_util.py
```diff
@@ -335,12 +335,12 @@ def get_type(node: AstNode) -> "Type":
...
+    This is the case if the body only contains a single `pass` statement or a constant
     `...` expression.
     """
     if len(func_ast.body) == 0:
         return True
     if len(func_ast.body) > 1:
         return False
     [n] = func_ast.body
-    return isinstance(n, ast.Expr) and isinstance(n.value, ast.Ellipsis)
+    return isinstance(n, ast.Expr) and isinstance(n.value, ast.Constant)
```
cqc-alec, line 346:
```suggestion
    return isinstance(n, ast.Expr) and isinstance(n.value, ast.Constant) and n.value.value is Ellipsis
```
I am not certain about this, but it looks to me from https://docs.python.org/3/library/ast.html#ast.Constant that it should be thus.

cqc-alec (approved):

aborgna-q (approved):

PR #712, opened 2024-12-12
feat: Add method to load pytket circuit without function stub

### LINKED ISSUE #670: Load `pytket` functions without explicit function stub

tatiana-s:
Closes #670 

Both finding the call location and then having either a node or a span for errors feels a bit hacky but I haven't been able to think of something better so far and it seems to work.

### DIFF
--- a/guppylang/definition/pytket_circuits.py
+++ b/guppylang/definition/pytket_circuits.py
@@ -28,9 +28,9 @@ from guppylang.definition.function import (
 )
 from guppylang.definition.ty import TypeDef
 from guppylang.definition.value import CallableDef, CallReturnWires, CompiledCallableDef
-from guppylang.error import GuppyError
+from guppylang.error import GuppyError, InternalGuppyError
 from guppylang.nodes import GlobalCall
-from guppylang.span import SourceMap
+from guppylang.span import SourceMap, Span, ToSpan
 from guppylang.tys.builtin import bool_type
 from guppylang.tys.subst import Inst, Subst
 from guppylang.tys.ty import (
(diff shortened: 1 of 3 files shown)


tatiana-s:
Python 3.10 doesn't have `Traceback.positions` hence why the check fails, will look into it.

guppylang/definition/pytket_circuits.py
```diff
@@ -74,65 +74,73 @@ def parse(self, globals: Globals, sources: SourceMap) -> "ParsedPytketDef":
...
-                    raise GuppyError(err) from None
-        except ImportError:
-            pass
+        circuit_signature = _signature_from_circuit(
+            self.input_circuit, globals, self.defined_at
+        )
+        if not (
+            circuit_signature.inputs == stub_signature.inputs
+            and circuit_signature.output == stub_signature.output
+        ):
```
mark-koch, line 85: drive-by:
```suggestion
        if circuit_signature != stub_signature:
```

mark-koch (commented):
Nice!

I also had a look at `Traceback.positions` before, but as you said, we can't use it for Python 3.10. You could have a look at `_parse_expr_string` in `decorator.py`, there we only use the line number to construct a span. We could do the same here.

Also, could you add a test that shows an error with this span?

guppylang/definition/pytket_circuits.py
tatiana-s: This doesn't work because they have different `input_names`.

mark-koch (approved):
Thanks, looks good to me 👍

PR #713, opened 2024-12-12
feat: result appending post-processing

ss2165:
To enable e.g. all the elements of an array to be collated in to the same bitstrings, see tests

Flattens lists so can be used once entire arrays can be printed too.

So some guppy that looks like for a given shot:

```python
vals = (1, 0, 1, 0)
for v in vals:
    result("c", v)
```

results in post-processed output, over 100 shots
```python
Counter({(("c", 1010"),): 100})
```

### DIFF
--- a/tests/test_hresult.py
+++ b/tests/test_hresult.py
@@ -113,3 +113,41 @@ def test_pytket():
     )
 
     assert pytket_result == expected
+
+
+def test_collate_tag():
+    # test use of same tag for all entries of array
+
+    shotlist = []
+    for _ in range(10):
+        shot = HResult()
+        _ = [
+            shot.append(reg, 1)
+            for reg, size in (("c", 3), ("d", 5))
+            for _ in range(size)
+        ]
+        shotlist.append(shot)
+
+    weird_shot = HResult((("c", 1), ("d", 1), ("d", 0), ("e", 1)))
+    assert weird_shot.collate_tags() == {"c": [1], "d": [1, 0], "e": [1]}
+
+    lst_shot = HResult([("lst", [1, 0, 1]), ("lst", [1, 0, 1])])
+    shots = HShots([*shotlist, weird_shot, lst_shot])
+
+    counter = shots.collated_counts()
+
+    assert counter == {
+        "c": Counter({"111": 10, "1": 1}),
+        "d": Counter({"11111": 10, "10": 1}),
+        "e": Counter({"1": 1}),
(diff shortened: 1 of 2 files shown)


tests/test_hresult.py
```diff
@@ -113,3 +113,41 @@ def test_pytket():
...
+    shots = HShots([*shotlist, weird_shot, lst_shot])
+
+    counter = shots.collated_counts()
+
+    assert counter == {
+        "c": Counter({"111": 10, "1": 1}),
+        "d": Counter({"11111": 10, "10": 1}),
+        "e": Counter({"1": 1}),
+        "lst": Counter({"101101": 1}),
+    }
```
cqc-alec, line 144: In this counter, are we losing information about the correlation between different tags across shots?

ss2165: yes, but it is still preserved in the full results object. The conversion to pytket preserves it for example.

cqc-alec: I see. I think those correlations would be important in most cases. No harm in offering this format, but I think we should also offer a counter-like result format that doesn't lose any information _except_ for the shot ordering. I guess this would have to be something like `Counter({(("c", "111"), ("d", "10")): 1, ...})`?

ss2165: yeah maybe that makes more sense as a default

cqc-alec: Hmm, although even that format isn't sufficient to capture the order of results within a shot (e.g. whether the "d" came between the first and second "c" or not), representing different control-flow paths. For that, we may need a format like `Counter({("c", 1), ("c", 1), ("d", 1), ("c", 1), ("d", 0)) : 1, ...})`.

ss2165: that's too noisy to be useful in this context I think

cqc-alec: We may need to offer several different "views" into results to serve different use cases.

ss2165: yep, we already have `register_counts` for serving the traditional "classical register" style use, this RFC is effectively view number 2

cqc-alec: Sounds good to me.

ss2165: > representing different control-flow paths
I think `Counter(shots.results)` where `shots` is an instance of `HShots` already gets you this

tatiana-s (approved):
Looks good to me

cqc-alec (approved):

PR #723, opened 2024-12-16
fix: Allow borrowing inside comprehensions

### LINKED ISSUE #719: Borrowing comprehension bodies report linearity violation
For example:

```python

@guppy
def foo(qs: array[qubit, N]) -> bool:
    return True

@guppy
def main() -> None:
    outcomes = array(foo(qs) for _ in range(10))
_ = main.compile()

```

```
Error: Linearity violation (at <In [4]>:8:25)
  | 
6 | def main() -> None:
7 |     qs = array(qubit() for _ in range(10))
8 |     outcomes = array(foo(qs) for _ in range(10))
  |                          ^^ Variable `qs` with linear type `array[qubit, 10]` is leaked

```

mark-koch:
Fixes #719


guppylang/checker/linearity_checker.py
```diff
@@ -317,13 +317,13 @@ def _reassign_inout_args(self, func_ty: FunctionType, call: AnyCall) -> None:
             if InputFlags.Inout in inp.flags:
                 match arg:
                     case PlaceNode(place=place):
-                        self._reassign_single_inout_arg(place, arg)
+                        self._reassign_single_inout_arg(place, place.defined_at or arg)
```
mark-koch, line 320: drive-by: Improve leak errors by keeping track of the original place where a variable was defined if possible

tests/error/comprehension_errors/borrow_leaked2.err
```diff
@@ -0,0 +1,8 @@
+Error: Linearity violation (at $FILE:17:23)
+   | 
+15 | @guppy(module)
+16 | def foo(qs: list[qubit] @owned) -> list[int]:
+17 |     return [bar(q) for q in qs]
```
ss2165, line 5: obvs not for this PR but it would be super sweet if a hint said "try removing owned"

tests/integration/test_array_comprehension.py
```diff
@@ -88,3 +88,76 @@ def test(xs: array[int, n]) -> array[int, n]:
...
+
+    @guppy.declare(module)
+    def foo(q: qubit) -> int: ...
+
+    @guppy.declare(module)
+    def bar(q: qubit @ owned) -> int: ...
+
+    @guppy(module)
+    def test(qs: array[qubit, n] @ owned) -> array[int, n]:
+        return array(foo(q) + bar(q) for q in qs)
```
ss2165, line 160: flex

ss2165 (approved):
thanks!

tests/error/comprehension_errors/borrow_leaked2.err
mark-koch: Created #726

PR #737, opened 2024-12-19
chore: Attempt at making execute_llvm editable in local `uv` development

aborgna-q:
This is an attempt at avoiding `uv` from caching `execute_llvm` and reflecting new changes to the package.

As part of this I updated various dependencies:

- `maturin 1.4 -> 1.7.7`, since [this issue](https://www.github.com/PyO3/maturin/issues/2346) seemed related. 
- `pyo3 0.19 -> 0.23`, and updated the bindings syntax.

The change that actually worked was removing the unversioned "execute_llvm" entry in the main package's dependencies. Since we manually install it using `maturin develop`, it now ensures that changes get re-compiled and reflected on the python runtime.

drive-by: Update `execute_llvm` to read packages instead of hugrs, so we can pass it guppys with quantum primitives.


pyproject.toml
```diff
@@ -70,9 +70,10 @@ test = [
 execution = [
     { include-group = "test" },
     # Required to run the llvm integration tests
-    "maturin >=1.4.0,<2",
+    # The `execute_llvm` package itself should be installed directly
+    # by running `uv run maturin develop -m execute_llvm/Cargo.toml`.
+    "maturin >=1.7.7,<2",
     "pip >=24",
-    "execute-llvm",
```
aborgna-q, line 75: ^ Removing the unversioned dependency from here seems to prevent `uv` from using an old cached version even if we've ran `maturin develop`.

Ideally `uv` would trigger `maturin` on its own, but there seems to be missing some rebuild triggers :/

execute_llvm/src/lib.rs
```diff
@@ -105,27 +105,39 @@ fn run_function<T>(
...
+        Ok(module.print_to_string().to_str().unwrap().to_string())
+    }
+
+    #[pyfunction]
+    fn run_int_function(hugr_json: &str, fn_name: &str) -> PyResult<i64> {
+        run_function::<i64>(hugr_json, fn_name, |_, llvm_val| {
+            // GenericVal is 64 bits wide
+            let int_with_sign = llvm_val.as_int(true);
+            let signed_int = int_with_sign as i64;
+            panic!("test 6");
```
ss2165, line 132: !

tests/integration/test_array.py
```diff
@@ -310,3 +312,27 @@ def main() -> array[qubit, 2]:
 
     package = module.compile()
     validate(package)
+
```
ss2165, line 315: i dont think this test should be in here

ss2165: (or any diff to this file)

ss2165 (changes requested):

execute_llvm/src/lib.rs
aborgna-q: -.-
Should this have triggered a CI fail?

ss2165 (approved):

Issue #739, opened 2024-12-20
Emit warning instead of error for unreachable code


mark-koch:
Guppy currently errors if it produces unreachable basic blocks. It would be nicer if this were only a warning

PR #740, opened 2024-12-20
feat: Make `True` and `False` branches unconditional

mark-koch:
For example, when building a CFG for `if True: ...`, build an unconditional branch. This improves linearity checking for functions involving `while True` loops, since dataflow analysis now understands that such loops can only exit on an explicit `break` or `return`.

It also means that Guppy is now better at detecting unreachable code. ~For now, we error out in this case~. In the future it would be nice to emit a warning when we detect this (see #739).

One complication introduced by this is that Guppy can now generate CFGs where the exit block is unreachable. This lead to some problems with borrowed variables. This edge case now needs to be specially handled.

I recommend reviewing each of the three commits separately


tests/error/linear_errors/unused_non_terminating.err
```diff
@@ -0,0 +1,10 @@
+Error: Linearity violation (at $FILE:13:8)
+   | 
+11 | 
+12 | @guppy(module)
+13 | def foo(q: qubit @owned) -> None:
+   |         ^^^^^^^^^^^^^^^ Variable `q` with linear type `qubit` is leaked
```
mark-koch, line 6: Arguably, this shouldn't be an error since non-terminating functions technically cannot leak. However, it would be difficult to generate valid Hugr for cases like this, e.g.

```python
while True:
    qubit()
```

Therefore, I think it's ok to still complain about leaks, even if the function provably doesn't terminate

aborgna-q (commented):
Uhm, I don't think erroring out on unreachable code is a good option.
Is there no alternative asides from implementing warnings? Even completely ignoring the unreachable code may be better.

Adding `if False and ...`, mid-point `return`s, and unused branches is something to be expected while coding and testing things. This errors would make development a lot more annoying.

mark-koch:
> Uhm, I don't think erroring out on unreachable code is a good option.
Is there no alternative asides from implementing warnings? Even completely ignoring the unreachable code may be better.

Ok, fair enough, I updated it so the compiler now accepts unreachable code. Unreachable code is still type checked, but I'm skipping linearity checking (see `tests/integration/test_unreachable.py` for code that would normally produce an error but is now accepted because the compiler inferred that it's unreachable).

I had to introduce a notion of "dummy CFG edges" to mark branches that are never taken. I use those to propagate type information into and between unreachable code. For the linearity checking dataflow analysis, these dummy edges are simply ignored.

PR #753, opened 2025-01-06
feat: support integer exponentiation in guppy source

ss2165:


guppylang/std/builtins.py
```diff
@@ -355,8 +355,16 @@ def __or__(self: int, other: int) -> int: ...
     @guppy.custom(NoopCompiler())
     def __pos__(self: int) -> int: ...
 
-    @guppy.hugr_op(int_op("ipow"))
-    def __pow__(self: int, other: int) -> int: ...
+    # TODO replace with @guppy.hugr_op(int_op("ipow"))
+    # once lowering available
+    @guppy
+    @no_type_check
+    def __pow__(self: int, other: int) -> int:
```
croyzor, line 362: Can we have `other: nat` with typechecking enabled?

ss2165: guppy can't infer literals in to different types so you'd have to use a nat constructor to use exponentiation `3 ** nat(2)` etc. 

I don't think that's worth it, but I agree that it might lead to surprising bugs. Maybe we should panic on negative "other"?

croyzor: I think panicking on negative exponent seems a good solution :+1:

ss2165: blocked by #756 then

croyzor (approved):

PR #754, opened 2025-01-06
fix: Fix pytest hanging

### LINKED ISSUE #569: Pytest hangs if when `GuppyError` is thrown
E.g. when trying to run the test

```python
def test_guppy_error(validate):
    module = GuppyModule("test")
    
    @guppy(module)
    def bad(x: int) -> int:
        return y  # y not defined, so this will error

    validate(module.compile())
```

pytest will just hang. This seems to happen whenever the compiler throws a `GuppyError`. However, `GuppyTypeError` works fine?!

This seems to be the reason for #565

mark-koch:
Fixes #569.

Ensure termination of `repr` calls on objects in the compiler by excluding `python_scope` fields due to cyclic references.

See https://github.com/CQCL/guppylang/issues/569#issuecomment-2573157994 for a more detailed description of how this fixes the pytest issue.

### DIFF
--- a/guppylang/checker/core.py
+++ b/guppylang/checker/core.py
@@ -2,7 +2,7 @@ import ast
 import copy
 import itertools
 from collections.abc import Iterable, Iterator
-from dataclasses import dataclass, replace
+from dataclasses import dataclass, field, replace
 from functools import cached_property
 from typing import (
     TYPE_CHECKING,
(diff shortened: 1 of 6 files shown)


ss2165 (approved):
💯

ss2165 (commented):
Might be nice to have a test that now works without hanging

PR #758, opened 2025-01-08
fix: panic on negative exponent in ipow

ss2165:


guppylang/std/builtins.py
```diff
@@ -362,7 +362,8 @@ def __pos__(self: int) -> int: ...
     @no_type_check
     def __pow__(self: int, other: int) -> int:
         # only positive exponents are supported
-        # TODO add panic for negative exponents once #756 is resolved
+        if other < 0:
+            panic("Negative exponent not supported in __pow__.")
```
croyzor, line 366: Not supported for now? Or does guppy disallow it? I think using negative exponents to take roots is the kind of freaky thing physicists would do so I'm inclined to say we should support it. It is very awkward however, the right behaviour is like:
```
def __pow(self: int, other: int) -> int | float:
    if other < 0:
        return float(self) ** float(other)
    else:
        # do the ipow thing, return an int
```
but I think the `float | int` return type is problematic? I tried to do it quickly and I'm not quite sure what goes wrong

ss2165: guppy does not support unions and probably never will, we will support algebraic data types at some point but it is unclear whether we can make them behave like unions (especially with regards to type inference).

ss2165: Could add a message saying "cast to float first" though?

croyzor: good!

croyzor: That would work - make sure to say it's the base that needs cast to a float

mark-koch: Casting the exponent should also work due to `__rpow__`

croyzor (approved):

guppylang/std/builtins.py
croyzor: > Casting the exponent should also work due to `__rpow__`

Nice, I didn't realise!

croyzor: Does that work as written here, or does it require `__pow__` to use `ReversingChecker`?

mark-koch: Should work as written. The `ReversingChecker` is on `__rpow__`

Issue #749, opened 2025-01-03
Array subscript assignments


mark-koch:
This will also require either making arrays linear, or breaking Python style sematics:

```python
xs = array(1, 2, 3)
ys = xs
ys[0] = 0  # What about xs[0]?
```

ss2165:
make arrays linear and add an explicit array copy function?

ss2165:
can this approach be used for structs too?

mark-koch:
Yes it should work for arrays containing structs as well as structs containing arrays (where the whole struct becomes linear).

An assignment would need to be desugared as follows:

```python
struct.arr_field[idx].field = value

# Desugars to
tmp = struct.arr_field.__getitem__(idx)
tmp.field = value
struct.arr_field.__setitem__(idx, tmp)
```

Note that this probably requires copying the `tmp` value out of the array (at least using our current lowering strategy for arrays), but LLVM might be able to optimise the copy away. If not, this could become a performance problem when working with arrays containing arrays...

acl-cqc:
Yeah I can see LLVM optimising that away for a struct of known size but whether it would for an array...if the array is also of statically known size then yeah it might.

A proposed "array::apply" op (taking an index and a child graph) would avoid these issues, right? There'd be no copying?

Also presumably `xs[0] = 0` is only plausible for non-linear element types, otherwise the only operation you can do on an element is `swap` ?

PR #768, opened 2025-01-15
feat!: Refactor to support affine arrays

tatiana-s:
Needed for #749

Summary:
- types now have `copyable` and `droppable` as property (with the linear property being inferred from those)
- `can_be_linear` is now `must_be_copyable` and `must_be_droppable` (where it was previously `True` it is now `False` and `False` and vice versa)
- `always_linear` is now `never_copyable` and `never_droppable`
- to check whether parametrised types are copyable / droppable, all type arguments must be copyable / droppable as well as the type being intrinsically copyable / droppable
- HUGR bounds are left as they are for now
- an attempt at changing linear to just not copyable or not droppable has been made where possible, but there are probably still some cases that have been missed
- arrays are the only type for now that is affine, there are no tests for relevant types

BREAKING CHANGE: classical arrays can no longer be implicitly copied


mark-koch (approved):
Nice!

As a follow up we could consider to stop using terms like linearity or affineness in user-facing messages and say "cannot be implicitly copied/dropped" instead?

tests/integration/test_poly.py
```diff
@@ -232,7 +232,7 @@ def main() -> None:
 def test_linear(validate):
     module = GuppyModule("test")
     module.load_all(quantum)
-    T = guppy.type_var("T", linear=True, module=module)
+    T = guppy.type_var("T", copyable=False, droppable=False, module=module)
```
mark-koch, line 235: It would also be good to add tests for `copyable and not droppable` and `not copyable and droppable` type vars

PR #773, opened 2025-01-17
refactor: try conditionally applying no type check

ss2165:
Following a suggestion from https://github.com/python/cpython/issues/106309

It seems to _partly_ work, not sure what's limiting it.


acl-cqc:
I'm never quite sure how much type-checkers actually run the code, but can you do:
```
if TYPE_CHECKING:
  guppy = cast(_Guppy, no_type_check)
else:
  ....
```

PR #776, opened 2025-01-21
feat: Array subscript assignment for classical arrays

### LINKED ISSUE #749: Array subscript assignments

tatiana-s:
Closes #749


tests/integration/test_array.py
```diff
@@ -318,3 +320,34 @@ def main(xs: array[int, 2] @ owned) -> None:
         ys = xs
 
     validate(main)
+
+
+def test_subscript_assign(validate):
+    @compile_guppy
+    def main(xs: array[int, 3] @ owned) -> array[int, 3]:
+        xs[0] = 1
+        return xs
```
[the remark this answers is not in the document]
mark-koch: Also an unpacking assignment involving arrays: 

```python
xs[i], y = 1, 2
a, *zs[j] = array(1, 2, 3)
```

guppylang/checker/stmt_checker.py
```diff
@@ -152,8 +171,47 @@ def _check_field_assign(
...
         self, lhs: ast.Subscript, rhs: ast.expr, rhs_ty: Type
-    ) -> AnyUnpack:
-        raise GuppyError(UnsupportedError(lhs, "Subscript assignments"))
+    ) -> PlaceNode:
+        # Check subscript is array subscript.
+        value, container_ty = self._synth_expr(lhs.value)
+        if not is_array_type(container_ty):
+            raise GuppyError(
+                UnsupportedError(lhs, "Subscript assignments to non-arrays")
+            )
```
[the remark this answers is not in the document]
mark-koch: Also a test that the program below produces the expected linearity violation error:

```python
qs[0] = qubit()  # Leaks the old qubit at index 0
```

guppylang/checker/stmt_checker.py
```diff
@@ -283,6 +341,36 @@ def visit_Assign(self, node: ast.Assign) -> ast.Assign:
...
+            setitem_args = [
+                with_type(parent.ty, with_loc(node, PlaceNode(parent))),
+                with_type(item.ty, with_loc(node, PlaceNode(item))),
+                node.value,
+            ]
+            setitem_call, _ = self._synth_instance_fun(
+                setitem_args[0],
+                setitem_args[1:],
+                "__setitem__",
+                "able to borrow subscripted elements",
```
mark-koch, line 367: This message needs to be updated since it doesn't involve borrowing

guppylang/compiler/stmt_compiler.py
```diff
@@ -72,6 +73,13 @@ def _assign(self, lhs: ast.expr, port: Wire) -> None:
     @_assign.register
     def _assign_place(self, lhs: PlaceNode, port: Wire) -> None:
         self.dfg[lhs.place] = port
+        if subscript := contains_subscript(lhs.place):
+            assert subscript.setitem_call is not None
+            if subscript.item not in self.dfg:
+                self.dfg[subscript.item] = self.expr_compiler.compile(
+                    subscript.item_expr, self.dfg
+                )
+            self.expr_compiler.visit(subscript.setitem_call)
```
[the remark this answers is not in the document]
tatiana-s: I don't think this would work because it requires calls to the expression compiler, but the expression compiler has `DFContainer` so you'd have a circular import?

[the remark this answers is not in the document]
mark-koch: It would only be a circular type checking import, right? Those are fine if you do

```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from guppylang.compiler.expr_compiler import ExprCompiler
```

tatiana-s:
(I will do making `rhs` optional during checking as a separate PR)

guppylang/checker/core.py
```diff
@@ -175,7 +175,8 @@ class SubscriptAccess:
     ty: Type
     item_expr: ast.expr
     getitem_call: ast.expr | None = None
-    setitem_call: ast.expr | None = None
+    # Store a temp variable for the RHS of an assignment so it can be assigned a port.
+    setitem_call: tuple[ast.expr, ast.expr] | None = None
```
mark-koch, line 179: I think it would be better to store the temporary as a `Variable` instead of an `ast.expr`

mark-koch: Also we could consider creating a

```python
class SetitemCall(NamedTuple):
    """Docstring ..."""

    #: Expression corresponding to the `__setitem__` call
    call: ast.expr

    #: Variable holding the value that should be written to the subscript.
    #: This variable *must be* assigned before compiling th call! 
    value_var: Variable

    def compile(value: ast.expr, ...):
        """Writes the provided value to the specified location by compiling the `__setitem__` call."""
```

guppylang/checker/expr_checker.py
```diff
@@ -944,7 +945,7 @@ def check_inout_arg_place(place: Place, ctx: Context, node: PlaceNode) -> Place:
                 exp_sig,
                 True,
             )
-            return replace(place, setitem_call=setitem_call)
+            return replace(place, setitem_call=(setitem_call, rhs))
```
mark-koch, line 948: RHS is not a variable. We should create a new temporary to satisfy the contract with `setitem_call`

guppylang/checker/linearity_checker.py
```diff
@@ -332,7 +332,7 @@ def _reassign_single_inout_arg(self, place: Place, node: AstNode) -> None:
         # Places involving subscripts are given back by visiting the `__setitem__` call
         if subscript := contains_subscript(place):
             assert subscript.setitem_call is not None
-            self.visit(subscript.setitem_call)
+            self.visit(subscript.setitem_call[0])
```
mark-koch, line 343: We should assign the temporary to `self.dfg[place]` before compiling the call

guppylang/compiler/expr_compiler.py
```diff
@@ -282,7 +282,8 @@ def _update_inout_ports(
                 # `arg.place.parent` occurs as an arg of this call, so will also
                 # be recursively reassigned.
                 if subscript := contains_subscript(arg.place):
-                    self.visit(subscript.setitem_call)
+                    assert subscript.setitem_call is not None
+                    self.visit(subscript.setitem_call[0])
```
mark-koch, line 290: Same here

guppylang/checker/linearity_checker.py
tatiana-s: The visit here is not compilation - how can I assign it when this is inside the linearity checker?

PR #783, opened 2025-01-29
Add CopyNode, update array copy compiler

croyzor:


guppylang/std/_internal/checker.py
```diff
@@ -198,7 +203,10 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
                 ArrayCopyChecker.NonCopyableElementsError.Explanation(None)
             )
             raise GuppyTypeError(err)
-        return array_arg, array_ty
+        if isinstance(array_arg, PlaceNode):
```
croyzor, line 206: I don't like this. My thinking was that `copy` makes sense when called on a `Place`: `copy(xs); copy(xs[0:4]); copy(xs.a)` and in other cases like constants it's a noop e.g. `copy(array(1,2,3))`. But surely there's more cases I'm not thinking of?

guppylang/compiler/expr_compiler.py
```diff
@@ -186,6 +187,12 @@ def _if_true(self, cond: ast.expr, inputs: list[PlaceNode]) -> Iterator[None]:
         for node, wire in zip(inputs, conditional, strict=True):
             self.dfg[node.place] = wire
 
+    def visit_CopyNode(self, node: CopyNode) -> Wire:
+        if node.value.place not in self.dfg:
```
acl-cqc, line 191: nit: you could do something like `return self.dfg.get(node.value.place) or self.visit(node.value)`.  (`.get` returns None if not found rather than raising an exception like `[...]`)

guppylang/std/_internal/checker.py
acl-cqc: I was wondering this too when I saw `value: PlaceNode` in `class CopyNode` above. It may be true for the moment though while arrays are owned, though - it'll change when we have dynamically-allocated arrays (i.e. the expression you're copying is just a reference).

guppylang/std/_internal/checker.py
```diff
@@ -198,7 +203,10 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+            return CopyNode(array_arg), array_ty
+        else:
+            return array_arg, array_ty
```
acl-cqc, line 209: However, you could try raising an error in this case (so `copy(array(1,2,3))` fails to compile) and then seeing if anyone can break it ;)

guppylang/compiler/expr_compiler.py
acl-cqc: But hang on, given `node: CopyNode` we have `node.value: PlaceNode` and so `self.visit(node.value)` will call `visit_PlaceNode` which....ignoring the subscript checks....does exactly the `return self.dfg[...place]` that you want. So this can just be `return self.visit(node.value)` ?

guppylang/std/_internal/checker.py
croyzor: It's trivial to break by writing silly code, but that code shouldn't affect runtime behaviour so I don't think it's worth raising an error over 🙂

guppylang/compiler/expr_compiler.py
croyzor: good point :+1:

guppylang/std/_internal/checker.py
croyzor: As in we'll have to work harder to copy the actual array when we're given a reference instead of duplicating the reference as would happen here :+1:

tatiana-s: I can't think of other cases at the moment, and there are other places in code where we only allow certain things to just happen to places (e.g. field and subscript access). Given we don't have references at the moment (and we made arrays affine exactly to avoid confusion with implicit copies) I think this works fine for now 👍

tatiana-s (approved):

PR #799, opened 2025-02-06
refactor!: Stop inlining array.__getitem__ and arrary.__setitem__

### LINKED ISSUE #786: Stop inlining `array.__getitem__` and `arrary.__setitem__`

tatiana-s:
Closes #786

BREAKING CHANGE: `CompiledGlobals` renamed to `CompilerContext`


guppylang/compiler/core.py
```diff
@@ -1,21 +1,43 @@
...
+    """
+    Abstract base class for functions in custom compilers that should only be
+    constructed once to avoid inlining.
+    """
+
+    @abstractmethod
+    def call(self, *args: Any) -> CallReturnWires: ...
+
+
+class CompiledContext:
```
mark-koch, line 40: I'd prefer `CompilerContext`

guppylang/compiler/core.py
```diff
@@ -27,6 +49,7 @@ class CompiledGlobals:
     checked: dict[DefId, CheckedDef]
     compiled: dict[DefId, CompiledDef]
     worklist: set[DefId]
+    compiler_functions: dict[str, CustomCompilerFunction]
```
mark-koch, line 52: I'd prefer to just store a dict of `hugr.build.function.Function`s and leave the call logic to whoever wants to use this function.

Maybe even just store `Wire`s corresponding to the node, then we could use the same system to store global constants and stuff like that?

guppylang/std/_internal/compiler/array.py
```diff
@@ -180,18 +183,103 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
     def build_classical_getitem(self, array: Wire, idx: Wire) -> CallReturnWires:
         """Lowers a call to `array.__getitem__` for classical arrays."""
-        # See https://github.com/CQCL/guppylang/issues/629
-        elem_opt_ty = ht.Option(self.elem_ty)
-        idx = self.builder.add_op(convert_itousize(), idx)
-        result = self.builder.add_op(array_get(elem_opt_ty, self.length), array, idx)
-        elem_opt = build_unwrap(self.builder, result, "Array index out of bounds")
-        elem = build_unwrap(self.builder, elem_opt, "array.__getitem__: Internal error")
-        return CallReturnWires(regular_returns=[elem], inout_returns=[array])
+        func_name = "classical_array.__getitem__"
```
mark-koch, line 270: Specifying these by name feels a bit unsafe. We could consider using a global supply of names to guarantee uniqueness.

In `core.py`:
```python
_fresh_ids = itertools.count()

@dataclass(frozen=True)
class GlobalConstId:
    id: int

    @staticmethod
    def fresh() -> HugrConstId:
        return HugrConstId(next(_fresh_ids))
    
    def unique_name(str: base) -> str:
        return f"{self.id}.{base}"
```

In `array.py`:
```python
# On module level:
ARRAY_GETITEM: Final[GlobalConstId] = GlobalConstId.fresh()

# In compile_call:
if ARRAY_GETITEM not in self.globals.compiler_functions:
    parent_op = ops.FuncDefn(name=ARRAY_GETITEM.unique_name, ...)
    ...
```

tatiana-s:
Benchmarks:
Screenshot 2025-02-06 at 17 23 39

On the `big_array` example this seems to reduce number of nodes from 44064 to 4868!

guppylang/std/_internal/compiler/array.py
```diff
@@ -180,44 +186,158 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
+
+
 class ArrayGetitemCompiler(ArrayCompiler):
     """Compiler for the `array.__getitem__` function."""
 
-    def build_classical_getitem(self, array: Wire, idx: Wire) -> CallReturnWires:
-        """Lowers a call to `array.__getitem__` for classical arrays."""
+    def _construct_getitem_ty(self, bound: ht.TypeBound) -> ht.PolyFuncType:
+        """Constructs a polymorphic function type for `__getitem__`"""
+        # a(T, N), int -> T, a(T, N)
```
mark-koch, line 200:
```suggestion
        # a(Option(T), N), int -> T, a(Option(T), N)
```

guppylang/std/_internal/compiler/array.py
```diff
@@ -180,44 +186,158 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
+        length_param = ht.BoundedNatParam()
+        return ht.PolyFuncType(
+            params=[elem_ty_param, length_param],
+            body=ht.FunctionType(
+                input=[
+                    array_type(
+                        ht.Option(ht.Variable(0, bound)),
+                        ht.VariableArg(1, length_param),
+                    ),
+                    INT_T,
```
mark-koch, line 213: Use `int_type().to_hugr()` (imported from `guppylang.tys.builtin`) here to make sure it's using the same bit-width as the rest of the compiler

guppylang/std/_internal/compiler/array.py
```diff
@@ -180,44 +186,158 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
+            ),
+        )
+
+    def _construct_classical_getitem(self, name: str) -> Wire:
+        """Constructs a generic function for `__getitem__` for classical arrays."""
+        func_ty = self._construct_getitem_ty(ht.TypeBound.Copyable)
+        parent_op = ops.FuncDefn(
+            name, func_ty.body.input, func_ty.params, func_ty.body.output
+        )
+        func = hf.Function.new_nested(parent_op, self.builder.hugr)
```
mark-koch, line 231: This places the function into the current builder, but it should be on module level. This should fail if you try to use arrays in two different functions (maybe also add a test for that!)

Also no need to define the op by hand. You can try something like this:

```suggestion
        func = self.ctx.module.define_function(
           name, func_type.body.input, func_type.body.output, func_type.params
        )
```

guppylang/compiler/cfg_compiler.py
```diff
@@ -24,7 +24,7 @@ def compile_cfg(
     cfg: CheckedCFG[Place],
     container: DfBase[DP],
     inputs: Sequence[Wire],
-    globals: CompiledGlobals,
+    globals: CompilerContext,
```
mark-koch, line 27: We should also rename all of these arguments to `ctx`

guppylang/std/_internal/compiler/array.py
```diff
@@ -180,44 +186,158 @@ def compile(self, args: list[Wire]) -> list[Wire]:
             return [self.build_classical_array(args)]
 
 
+ARRAY_GETITEM_CLASSICAL: Final[GlobalConstId] = GlobalConstId.fresh()
```
mark-koch, line 189: On second thought, maybe we should store the user-facing name also inside `GlobalConstId`?

guppylang/std/_internal/compiler/array.py
```diff
@@ -180,44 +186,158 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
+    def compile_classical_getitem(self, array: Wire, idx: Wire) -> CallReturnWires:
+        """Lowers a call to `array.__getitem__` for classical arrays."""
+        if ARRAY_GETITEM_CLASSICAL not in self.globals.global_consts:
+            self.globals.global_consts[ARRAY_GETITEM_CLASSICAL] = (
+                self._construct_classical_getitem(
+                    name=ARRAY_GETITEM_CLASSICAL.unique_name(
+                        "array.__getitem__.classical"
+                    )
+                )
+            )
```
mark-koch: I think it would be nicer to factor this into the `CompilerContext`. What do you think about something like this:

```python
f, already_exists = self.ctx.define_global_function(ARRAY_GETITEM_CLASSICAL, params, input_tys)
# f is the function if it already exists, otherwise it's a fresh builder
if not already_exists:
    # Build the function
    ...
    f.set_outputs(...)  # Hugr-py automatically infers the return type

array, item = f.call(array, idx, instantiation=..., type_args=...)
return CallReturnWires(regular_returns=[item], inout_returns=[array])
```

tatiana-s: Not sure I like this that much because of the passing of params / types even when you don't need them if the function already exists, but I haven't managed to come up with anything better so I did this for now. Also means I am storing functions instead of wires but I think I prefer that anyway.

mark-koch (approved):

Issue #798, opened 2025-02-06
Add `get_current_shot` to qsystem module


ss2165:
```python

def get_current_shot() -> int: ...
```

Lowers to tket2 extension: https://github.com/CQCL/tket2/issues/767

PR #806, opened 2025-02-12
feat: add `get_current_shot()` to qsystem module

qartik:
Closes: #798

It's not clear to me whether the type in tket2-hseries for `get_current_shot()` should be `usize` or `int(6)`. I used usize initially but got the following runtime error during json validation (`[] -> [tket2.qsystem.utils][usize] but stored signature was [] -> [tket2.qsystem.utils][int(6)]`):
```py
tests/integration/conftest.py:64: in validate_impl
    validate_json(js)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

hugr = '{"modules":[{"version":"live","nodes":[{"parent":0,"op":"Module"},{"parent":0,"op":"FuncDefn","name":"test","signatur..."i":1,"b":"A"}],"output":[{"t":"R","i":2,"b":"A"}],"runtime_reqs":["guppylang"]}},"binary":false,"lower_funcs":[]}}}]}'

    def validate_json(hugr: str):
        # Executes `cargo run -p validator -- validate -`
        # passing the hugr JSON as stdin
        p = subprocess.run(  # noqa: S603
            [validator, "validate", "-"],
            text=True,
            input=hugr,
            capture_output=True,
        )

        if p.returncode != 0:
>           raise RuntimeError(f"{p.stderr}")
E           RuntimeError: Error parsing package: Error resolving opaque operation: Conflicting signature: resolved GetCurrentShot in extension tket2.qsystem.utils to a concrete implementation which computed [] -> [tket2.qsystem.utils][usize] but stored signature was [] -> [tket2.qsystem.utils][int(6)]
```

I [changed the signature](https://github.com/CQCL/tket2/pull/772/commits/a173d44814a853327683ecc821658eb4ee942fc6) to use `int(6)`, which then made the types compatible.

### DIFF
--- a/guppylang/std/qsystem/__init__.py
+++ b/guppylang/std/qsystem/__init__.py
@@ -5,10 +5,9 @@ from guppylang.module import GuppyModule
 from guppylang.std import angles
 from guppylang.std._internal.compiler.quantum import (
     QSYSTEM_EXTENSION,
-    QSYSTEM_UTILS_EXTENSION,
     InoutMeasureCompiler,
 )
-from guppylang.std._internal.util import external_op, quantum_op
+from guppylang.std._internal.util import quantum_op
 from guppylang.std.angles import angle
 from guppylang.std.builtins import owned
 from guppylang.std.quantum import qubit
@@ -72,13 +71,6 @@ def reset(q: qubit) -> None: ...
 @guppy.hugr_op(quantum_op("QFree", ext=QSYSTEM_EXTENSION), module=qsystem)
 @no_type_check
 def qfree(q: qubit @ owned) -> None: ...
-
-
-@guppy.hugr_op(
-    external_op("GetCurrentShot", [], ext=QSYSTEM_UTILS_EXTENSION), module=qsystem
-)
-@no_type_check
-def get_current_shot() -> int: ...
 
 
 # ------------------------------------------------------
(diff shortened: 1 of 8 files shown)


guppylang/std/qsystem/__init__.py
```diff
@@ -73,6 +74,13 @@ def reset(q: qubit) -> None: ...
 def qfree(q: qubit @ owned) -> None: ...
 
 
+@guppy.hugr_op(
+    external_op("GetCurrentShot", [], ext=QSYSTEM_UTILS_EXTENSION), module=qsystem
```
ss2165, line 78: I'm fine with this but just checking we don't want a `qsystem.utils` guppy module to reflect the extension structure?

ss2165 (approved):
Looks good, but good to merge after next tket2-exts release

guppylang/std/qsystem/__init__.py
qartik: Yeah, I was wondering about the same, what's your preference? Happy to create a submodule if that's better.

ss2165: i think yes - good to keep in line with the extensions as far as is reasonable to keep mental models simpler

qartik:
> Looks good, but good to merge after next tket2-exts release

Should be ready to merge now.

PR #809, opened 2025-02-18
feat: add `Option.take()` for swapping with None

ss2165:
Example use case: measure without changing array indexing

```python

@guppy
def measure_option(ar: array[Option[qubit], N], index: int) -> bool:
    return measure(ar[index].take().unwrap())
```

### DIFF
--- a/guppylang/std/option.py
+++ b/guppylang/std/option.py
@@ -53,13 +53,11 @@ class Option(Generic[T]):  # type: ignore[misc]
 
     @guppy
     @no_type_check
-    def take(self: "Option[T]") -> T:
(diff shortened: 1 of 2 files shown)


qartik (approved):
Just a side comment, what was the rationale behind using `nothing` in Guppy's option type as opposed to `None`?

ss2165:
> what was the rationale behind using nothing in Guppy's option type as opposed to None?

to avoid confusion with python's `None` which works quite differently due to how python `Union`s work dynamically. 

Guppy uses `None` in type signatures to indicate void return

PR #815, opened 2025-02-25
fix: Fix error message for conditional shadowing of global variables

### LINKED ISSUE #772: Problems with conditional shadowing of global variables
```python
@guppy
def x() -> None:
    pass


@guppy
def bad(b: bool) -> int:
    if b:
        x = 4
    return x
```

This crashes instead of complaining that the type of `x` is not fixed

tatiana-s:
Closes #772


guppylang/checker/cfg_checker.py
```diff
@@ -260,13 +266,43 @@ def check_bb(
...
+                    # since  it could lead to cross-file diagnostics (which are not
+                    # supported) or refer to long function definitions.
+                    err.add_sub_diagnostic(BranchTypeError.GlobalHint(None, g.ty))
+                    raise GuppyError(err)
+                # Remove variables from comparison below if types match.
+                map1.pop(x, None)
+                map2.pop(x, None)
+
     assert map1.keys() == map2.keys()
     for x in map1:
```
mark-koch, line 307: It might be nicer to move this logic into the existing loop to avoid code duplication? Maybe something along the lines of

```python
for x in map1.keys() | map2.keys():
    v1 = map1.get(x) or globals[x]
    v2 = map2.get(x) or globals[x]
    if v1.ty != v2.ty:
        # ...
        err = BranchTypeError(use, ident)
        sub1 = (
            BranchTypeError.TypeHint(v1.defined_at, v1.ty) if isinstance(v1, Variable)
            else BranchTypeError.GlobalHint(None, v1.ty)
        )
        sub2 = (
            BranchTypeError.TypeHint(v2.defined_at, v2.ty) if isinstance(v2, Variable)
            else BranchTypeError.GlobalHint(None, v2.ty)
        )
        err.add_sub_diagnostic(sub1)
        err.add_sub_diagnostic(sub2)
        raise GuppyError(err)
```

guppylang/checker/cfg_checker.py
```diff
@@ -191,6 +190,13 @@ class TypeHint(Note):
         span_label: ClassVar[str] = "This is of type `{ty}`"
         ty: Type
 
+    @dataclass(frozen=True)
+    class GlobalHint(Note):
+        message: ClassVar[str] = (
+            "{ident} may be shadowing a global name with type `{ty}` on some branches"
```
mark-koch, line 196: If you also strore the definition in the note, you can use `defn.describe` instead of "name". Then it will explicitly say "function", "const", etc.

guppylang/checker/cfg_checker.py
```diff
@@ -260,13 +266,43 @@ def check_bb(
...
+                        "Expression"
+                        if v.name.startswith("%")
+                        else f"Variable `{v.name}`"
+                    )
+                    use = bb.containing_cfg.live_before[bb][v.name].vars.used[v.name]
+                    err = BranchTypeError(use, ident)
+                    err.add_sub_diagnostic(BranchTypeError.TypeHint(v.defined_at, v.ty))
+                    # We don't add a location to the type hint for the global variable,
+                    # since  it could lead to cross-file diagnostics (which are not
+                    # supported) or refer to long function definitions.
```
mark-koch, line 299: Imo this would be a good reason to support cross-file diagnostics, but that is a separate issue 😅

tests/error/misc_errors/conditional_global_shadow.err
```diff
@@ -0,0 +1,14 @@
...
+12 |     if b:
+13 |         x = 4
+14 |     return x
+   |            ^ Variable `x` may refer to different types
+   | 
+13 |         x = 4
+   |         - This is of type `int`
+
+Note: Variable `x` may be shadowing a global name with type `() -> None` on some
+branches
```
mark-koch, line 12: This message reads a bit akward, but hard to improve without cross-file diagnostics... :/

mark-koch (commented):
Looks good, just some refactoring suggestions.

Also, could you try a case where the global actually has a matching type. I have feeling that this might break:

```python
x = guppy.extern("x", ty="int", module=module)

@guppy(module)
def foo(b: bool) -> int:
    if b:
        x = 4
    return x
```

guppylang/checker/cfg_checker.py
tatiana-s: As far as I can see only places have a describe method?

mark-koch: Yes it should be `Definition.description`, sorry

mark-koch (approved):

PR #821, opened 2025-02-25
feat: Allow explicit application of type arguments

### LINKED ISSUE #770: Allow explicit specification of type args
Proposed syntax: `func[T1, T2, n, ...](arg1, arg2, ...)`

Example:

```python
n = guppy.nat_var("n")

@guppy
def foo() -> array[qubit, n]: ...

@guppy
def main():
    qs = foo[42]()
```

mark-koch:
Closes #770

### DIFF
--- a/guppylang/tys/parsing.py
+++ b/guppylang/tys/parsing.py
@@ -9,7 +9,6 @@ from guppylang.ast_util import (
 from guppylang.cfg.builder import is_py_expression
 from guppylang.checker.core import Context, Globals, Locals
 from guppylang.checker.errors.generic import ExpectedError
-from guppylang.checker.expr_checker import eval_py_expr
 from guppylang.definition.common import Definition
 from guppylang.definition.module import ModuleDef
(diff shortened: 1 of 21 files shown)


tests/error/poly_errors/define.err
mark-koch, line 1: drive-by: Deleted unused file

guppylang/tys/parsing.py
```diff
@@ -92,6 +91,8 @@ def arg_from_ast(
 
     # Py-expressions can also be used to specify static numbers
     if py_expr := is_py_expression(node):
+        from guppylang.checker.expr_checker import eval_py_expr
```
mark-koch, line 94: To prevent cyclic import

aborgna-q (approved):
That's a short diff!

I guess that we are restricted by python's syntax, but is there no way to declare which params a function uses / in which order?

I didn't dive into how `FunctionType.params` gets computed, does it only consider the signature? Can I bind and use a variable only in the body?

Issue #769, opened 2025-01-15
Add PRNG qsystem lib


ss2165:
Module name: `guppylang.std.random`

Functions:
Note we should not use names from the python [standard library random](https://docs.python.org/3/library/random.html) because they have different signatures and behaviours

```python
@guppy.struct
class RNGContext:
  # can only be initialised once per program. Second init fails at rutime.
  def __init__(seed: int) -> "RNGContext": ...
  # init safely, returning None if already acquired.
  def init_safe(seed: int) -> Option["RNGContext"]: ...
  def random_nat(self) -> nat: ...
  def random_int(self) -> int: ...
  def random_float(self) -> float: ...
  def random_int_bounded(self, bound: int) -> int: ...
```

qartik:
> Module name: `guppylang.std.random`

Should `random` be namespaced under `qsystem` module or `std`?

ss2165:
probably safer to go with qystem for now

qartik:
> def random_nat(self) -> nat: ...

is this just an alias for `random_int` with some type conversions?

ss2165:
> is this just an alias for random_int with some type conversions?
yup

Issue #810, opened 2025-02-18
Can't discard `None` of an optional qubit


ss2165:
If I have an `Option` qubit and I want to discard it, I could try and unwrap then discard the qubit, but I can't get rid of a `nothing` variant in any way. Do we need pattern matching to enable the `nothing` to be treated non-linearly?

mark-koch:
Adding a consuming `unwrap_none` or `expect_none` method would solve this

ss2165:
So something like 

```python
def expect_nothing[T](v: Option[T] @ owned) -> None: ...
```
?

mark-koch:
Yes 👍

Not sure if we should stick to the Rust naming scheme where `expect` usually also takes a string message?

PR #822, opened 2025-02-26
feat(qsystem): add Random number generation module

qartik:
Closes: #769

Depends on:
- https://github.com/CQCL/tket2/pull/786
  - https://github.com/CQCL/tket2/pull/790
- https://github.com/CQCL/guppylang/issues/810

Some changes from the issue:
- Introduces `RNG` guppy type, instead of the `RNGContext` guppy struct.
- We have `def __new__(seed: int) -> "RNG":` instead of `__init__(seed: int) -> "RNGContext"`
- We have `maybe_rng` instead of `init_safe(seed: int) -> Option["RNGContext"]`

Other thoughts:

Unsure if this interface is the best we can do:
```py
def test() -> tuple[int, nat, float, int]:
    rng = RNG(42)
    rint = rng.random_int()
    rnat = rng.random_nat()
    rfloat = rng.random_float()
    rint_bnd = rng.random_int_bounded(100)
    rng.discard()
    return rint, rnat, rfloat, rint_bnd
```
It might be nice to provide a context manager interface such as:
```py
def test() -> tuple[int, nat, float, int]:
    with RNG(42) as rng:
        rint = rng.random_int()
        rnat = rng.random_nat()
        rfloat = rng.random_float()
        rint_bnd = rng.random_int_bounded(100)
    return rint, rnat, rfloat, rint_bnd
```
which would enforce the linear discipline on `RNG`. Mark confirms it's currently not possible within guppy.

---

Release-As: 0.16.1


qartik:
Locally getting a test failure:
```py
FAILED tests/integration/test_qsystem.py::test_qsystem_random
    - AttributeError: type object 'int' has no attribute '_to_serial_root'
```

guppylang/std/qsystem/random.py
```diff
@@ -0,0 +1,69 @@
...
+    @guppy.hugr_op(
+        external_op("RandomIntBounded", [], ext=QSYSTEM_RANDOM_EXTENSION),
+        module=qsystem_random,
+    )
+    @no_type_check
+    def _random_int_bounded(self: "RNG", bound: int) -> int: ...
+
+    @guppy(qsystem_random)
+    def random_int_bounded(self: "RNG", bound: int) -> int:
+        return int(self._random_int_bounded(bound))
```
qartik, line 69: Getting `Conflicting signature: resolved RandomIntBounded in extension tket2.qsystem.random to a concrete implementation which computed [context, int(5)] -> [tket2.qsystem.random][int(5), context] but stored signature was [context, int(6)] -> [tket2.qsystem.random][int(6), context]` unsure if I can somehow coerce `bound` to act like an `int(5)` or if there is a way in Guppy to ensure that `bound` is 32-bits.

qartik: @mark-koch @ss2165 might need some help here. Do I need to write a custom compiler here?

qartik: After chatting with Alex, I decided to simplify this problem by requiring 64-bits in the tket2 signatures. The truncation can happen automatically.

I'd like to see how to document the truncation and correct signatures for guppy users. Any suggestions on the right place in this PR to do that? Docstrings don't seem to work for placeholder functions.

mark-koch: > Do I need to write a custom compiler here?

Yes, you would need a `CustomCallCompiler` to do that

> Docstrings don't seem to work for placeholder functions.

Created issue #826 to fix this

pyproject.toml
```diff
@@ -86,7 +86,7 @@ execute-llvm = { workspace = true }
 
 # Uncomment these to test the latest dependency version during development
 #  hugr = { git = "https://github.com/CQCL/hugr", subdirectory = "hugr-py", rev = "e40b6c7" }
-# tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", rev = "175a02d"}
+tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", branch = "fix-rngcontext"}
```
mark-koch, line 89: I presume this branch will be deleted once you merge the tket2 PR?

So better wait until the tket2 PR is merged before you merge this

tests/integration/test_qsystem.py
```diff
@@ -13,6 +14,7 @@
     qsystem_functional,
     measure_and_reset,
     zz_max,
+    reset,
```
mark-koch, line 17: Unused import?

mark-koch (approved):
Nice, thank you!

pyproject.toml
qartik: Yes, I will wait indeed, thanks!

ss2165:
> It's unclear whether init_safe(seed: int) -> Option["RNGContext"] can be written without maintaining any state and whether it's actually useful given the explicit linear typing currently required to use RNG


I don't follow why this is sorry - the `__new__` method is unwrapping the option, __init_safe__ would presumably just return it as-is?

qartik:
> I don't follow why this is sorry - the `__new__` method is unwrapping the option, `__init_safe__` would presumably just return it as-is?

`__new__` is a special dunder that's called for memory allocation even before `__init__`. It returns a new object and hence doesn't require passing in `self` as its first argument unlike `__init__`. So apart from `__new__`, I can't write another method that doesn't require `self`.

I also tried using `@staticmethod` decorator for `__init_safe__`, but then I get:
```py
tests/integration/test_qsystem.py:8: in <module>
    from guppylang.std.qsystem.random import RNG
guppylang/std/qsystem/random.py:26: in <module>
    class RNG:
guppylang/std/qsystem/random.py:34: in RNG
    @guppy(qsystem_random)
guppylang/decorator.py:141: in <lambda>
    return lambda s: dec(s, arg)
guppylang/decorator.py:118: in dec
    return module.register_func_def(f)
guppylang/module.py:229: in register_func_def
    defn = RawFunctionDef(DefId.fresh(self), f.__name__, None, f, get_py_scope(f))
guppylang/module.py:404: in get_py_scope
    code = f.__code__
E   AttributeError: 'staticmethod' object has no attribute '__code__'. Did you mean: '__call__'?
```
which perhaps suggests that static methods are not yet supported in Guppy?

ss2165:
ah that's fine (indeed static methods are not supported) - just have it available as a standalone function (`new_rng_checked` or something like that). I think this just means making your `_new_rng_context` function public effectively with a docstring and and a less ambiguous name.

qartik:
Sounds good. Blocked by https://github.com/CQCL/guppylang/issues/810 Also I will need to write a wrapper `maybe_rng` instead of directly using `_new_rng_context` as docstrings don't yet work on placeholder functions, to be fixed in https://github.com/CQCL/guppylang/issues/826

ss2165:
#810 has been closed: https://github.com/CQCL/guppylang/commit/abb1aa1707e94cbbce82a74b3d0c388c252483ef

ss2165:
I think that's right - I didn't realise they weren't being emitted apologies

guppylang/std/_internal/compiler/arithmetic.py
```diff
@@ -55,6 +55,28 @@ def iwiden_s(from_width: int, to_width: int) -> ops.ExtOp:
...
+    )
+
+
+def inarrow_s(from_width: int, to_width: int) -> ops.ExtOp:
+    """Returns a signed `std.arithmetic.int.narrow_s` operation."""
+    return _instantiate_int_op(
+        "inarrow_s",
+        [from_width, to_width],
+        [int_t(from_width)],
+        # TODO: add type for error
```
qartik, line 75:
```py
RuntimeError: Error parsing package: Error resolving opaque operation: Conflicting signature: resolved inarrow_s in extension arithmetic.int to a concrete implementation which computed [int(6)] -> [arithmetic.int][[error]+[int(5)]] but stored signature was [int(6)] -> [[]+[int(5)]]
```
Unsure if there is a type in hugr-py for "error".

mark-koch: Not in `hugr-py` unfortunately, but you can use `guppylang.std._internal.compiler.prelude.error_type` and other methods in that module

guppylang/std/qsystem/random.py
```diff
@@ -0,0 +1,105 @@
...
+    RNGCONTEXT_T,
+)
+from guppylang.std._internal.util import external_op
+from guppylang.std.builtins import nat, owned
+from guppylang.std.option import Option
+
+qsystem_random = GuppyModule("qsystem.random")
+
+
+class RandomIntCompiler(CustomInoutCallCompiler):
```
mark-koch, line 24: Could you move those `CustomInoutCallCompiler`s into the `_internal` module.

Maybe into `guppylang.std._internal.compiler.quantum` or a new `guppylang.std._internal.compiler.qsystem`?

mark-koch (approved):
Nice, looks good!

qartik:
I have a patch to take this to merge, but it might require #846 
```diff
diff --git a/pyproject.toml b/pyproject.toml
index b5b31b2..5a826c0 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -39,7 +39,7 @@ dependencies = [
     "pydantic >=2.7.0b1,<3",
     "typing-extensions >=4.9.0,<5",
     "hugr >=0.10.3,<0.11",
-    "tket2-exts>=0.4.0,<0.5",
+    "tket2-exts>=0.5.0,<0.6",
 ]

 [project.optional-dependencies]
@@ -65,7 +65,7 @@ test = [
     "pytest-notebook >=0.10.0,<0.11",
     "pytest-snapshot >=0.9.0,<1",
     "ipykernel >=6.29.5,<7",
-    "tket2>=0.6.1",
+    "tket2>=0.7.0",
     "pytest-benchmark>=5.1.0",
 ]
 execution = [
@@ -86,7 +86,7 @@ execute-llvm = { workspace = true }

 # Uncomment these to test the latest dependency version during development
 #  hugr = { git = "https://github.com/CQCL/hugr", subdirectory = "hugr-py", rev = "e40b6c7" }
-tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", rev = "633ebd7"}
+# tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", rev = "633ebd7"}
 # tket2 = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-py", rev = "259cf88"}
```

qartik:
> ah that's fine (indeed static methods are not supported) - just have it available as a standalone function (`new_rng_checked` or something like that). I think this just means making your `_new_rng_context` function public effectively with a docstring and and a less ambiguous name.

Hi @ss2165, I think `maybe_rng` or `init_safe` are useless, given there is no failure at runtime on seeding the RNG twice. I am removing them in https://github.com/CQCL/guppylang/pull/873

ss2165:
The issue is not seeding the rng twice, it is a lack of guarantee that you have a unique handle to rng, so reproducibility isn't guaranteed because rng calls can be reordered. The failure at runtime is from calling initialisation twice.

qartik:
> @qartik The issue is not seeding the rng twice, it is a lack of guarantee that you have a unique handle to rng state, so reproducibility isn't guaranteed because rng calls could be reordered. The failure at runtime should be from calling initialisation twice without discarding first.

I am unsure which runtime we are referring to. If the platform call doesn't fail, how can we detect this ordering violation?

Since `RNG` state is a linear type, it's tracked statically by the guppy type checker. The other way I was thinking of implementing maybe_rng was storing some state in RNG type itself, but that also will not be a runtime failure.

PR #831, opened 2025-03-04
feat: Implement `float` to `int` and `nat` casts

### LINKED ISSUE #794: Implement `float` to `int` and `nat` cats
They currently emit an unsupported op

tatiana-s:
Closes #794


guppylang/std/_internal/compiler/prelude.py
```diff
@@ -151,3 +158,36 @@ def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
...
+        assert len(self.ty.output) == 1
+        # To instantiate the op we need a function signature that wraps the output of
+        # the function that is being compiled into a sum type.
+        opt_func_type = ht.FunctionType(
+            input=self.ty.input,
+            output=[ht.Either([error_type()], self.ty.output)],
+        )
+        op = self.op(opt_func_type, self.type_args)
+        option = self.builder.add_op(op, *args)
+        result = build_unwrap(self.builder, option, self.err_msg)
```
mark-koch, line 189: If the op alreday returns an `error_type()`, imo it would be better to panic with that value instead of making up a new message

mark-koch, line 189: Also note that this inserts the `unwrap` logic in-place. Maybe we should use the same system we have for `array.__setitem__` etc. to only define this once and emit a call here?

mark-koch (commented):
Have you considered to define

guppylang/std/_internal/compiler/prelude.py
tatiana-s: Agree that would be better but also should probably be a separate PR since it affects other usages of this as well?

tatiana-s: Is there a good way in general to get the value of a wire - so I know how to check if this is an `error_type` but not how to get the `ErrorVal` which has the message?

mark-koch: `Option` is a wire with type `Either<error, T>`. Just build a conditional on that wire:
* In the left case, you'll get the error value and can panic with it
* In the right case, you get the unwrapped value and can return it

You won't be able to rely on `build_unwrap` for that since that was only designed for `Option`, not `Either<error, T>`.

mark-koch: Indeed, probably all uses should be refactored. However, to adress the other comment you would need to stop using `build_unwrap` anyways

mark-koch: Maybe we should call it `unwrap_result`?

guppylang/std/_internal/compiler/prelude.py
```diff
@@ -151,3 +160,97 @@ def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
...
+    either = func.inputs()[0]
+    conditional = func.add_conditional(either)
+    with conditional.add_case(0) as case:
+        [error] = list(case.inputs())
+        case.set_outputs(
+            *build_panic(case, [error_type()], [result_type_var], error, *case.inputs())
+        )
+    with conditional.add_case(1) as case:
+        case.set_outputs(*case.inputs())
+    func.set_outputs(conditional.to_node())
```
mark-koch, line 182:
```suggestion
    func.set_outputs(*conditional.outputs())
```

guppylang/std/_internal/compiler/prelude.py
```diff
@@ -151,3 +160,97 @@ def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
...
+    )
+    # Build global unwrap result function if it doesn't already exist.
+    if result_bound == ht.TypeBound.Any:
+        func, already_exists = ctx.declare_global_func(UNWRAP_RESULT_LINEAR, func_ty)
+        if not already_exists:
+            _build_unwrap_result(func, ht.Variable(0, result_bound))
+    else:
+        func, already_exists = ctx.declare_global_func(UNWRAP_RESULT_CLASSICAL, func_ty)
+        if not already_exists:
+            _build_unwrap_result(func, ht.Variable(0, result_bound))
```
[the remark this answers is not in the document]
mark-koch: It should be fine to define the function for `TypeBound.Any`. You can still call it with an instantiation that is `TypeBound.Copyable`

mark-koch (approved):

Issue #732, opened 2024-12-17
Break builtins.py in to multiple files


ss2165:
It is quite hard to navigate

CalMacCQ:
Just a note that having all of the unsupported builtins (e.g. `print`, `bin`, `isinstance`) in a single file should make it much cleaner to exclude them from the sphinx docs.

PR #833, opened 2025-03-05
feat!: Switch to improved iterator protocol

### LINKED ISSUE #96: Improve iterator protocol
It would be better to only have a single function `__next__: Iter -> Sum((Elem, Iter), ())`, compared to the current protocol with `__hasnext__`, `__next__`, and `__iterend__`.

But this requires a notion of sum types first (see #95)

mark-koch:
Closes #96

BREAKING CHANGE: Iterator protocol updated to now use a single method `__next__: Iterator -> Option[tuple[Elem, Iterator]]` instead of the previous `__hasnext__`, `__next__`, `__end__` methods.

BREAKING CHANGE: `Option` is now a builtin type.


guppylang/std/option.py
```diff
@@ -1,81 +1,68 @@
...
-        case _:
-            raise InternalGuppyError("Invalid type args for Option")
+from guppylang.std.builtins import L, mem_swap, owned
+from guppylang.tys.builtin import option_type_def
 
-
-T = guppy.type_var("T", copyable=False, droppable=False)
+builtins_module = guppy.get_module(
+    ModuleIdentifier(Path(builtins.__file__), "builtins", builtins)
+)
```
mark-koch, line 17: This is hack to make `Option` a builtin type without moving it into the `builtins` module to cause less breakage.

We want to break up `builtins.py` anyways (see #732), so we can find a better solution once we tackle #732/

PR #839, opened 2025-03-06
fix: Fix mutation of nested arrays

### LINKED ISSUE #835: Guppy arrays can be indexed into after a move
I expect the following code to give a compiler error. (Guppy v0.16.0)

```python
from guppylang import guppy

@guppy
def make_array() -> array[int, 3]:
    return array(1, 2, 3)

@guppy
def main() -> None:
    arr = make_array()
    arr2 = arr
    arr[1] = 17

guppy.compile_module() # Compilation succeeds.
```

I expect this to fail as I am accessing the `arr` variable after the move has ocurred. If I replace ` arr[1] = 17` with `arr3 = arr` then I get a compiler error as I would expect.

### LINKED ISSUE #836: Mutation of nested arrays in structs is broken
```python
@guppy.struct
class S:
    xs: array[int, 1]

@guppy
def main() -> int:
    arr = array(S(array(0)), S(array(1)))
    arr[0].xs = array(3)
    return arr[0].xs[0]
```

This should output 3, but it actually outputs 0

mark-koch:
Fixes #836 and as a drive-by also fixes #835

### DIFF
--- a/guppylang/checker/core.py
+++ b/guppylang/checker/core.py
@@ -225,6 +225,15 @@ class SubscriptAccess:
         return f"{self.parent}[...]"
 
 
+def contains_subscript(place: Place) -> SubscriptAccess | None:
+    """Checks if a place contains a subscript access and returns the rightmost one."""
+    while not isinstance(place, Variable):
+        if isinstance(place, SubscriptAccess):
+            return place
+        place = place.parent
+    return None
+
+
 PyScope = dict[str, Any]
 
 
--- a/guppylang/checker/expr_checker.py
+++ b/guppylang/checker/expr_checker.py
@@ -930,7 +930,9 @@ def type_check_args(
     for inp, func_inp in zip(inputs, func_ty.inputs, strict=True):
         a, s = ExprChecker(ctx).check(inp, func_inp.ty.substitute(subst), "argument")
         if InputFlags.Inout in func_inp.flags and isinstance(a, PlaceNode):
-            a.place = check_inout_arg_place(a.place, ctx, a)
+            a.place = check_place_assignable(
+                a.place, ctx, a, "able to borrow subscripted elements"
+            )
         new_args.append(a)
         subst |= s
 
@@ -949,8 +951,11 @@ def type_check_args(
     return new_args, subst
 
 
-def check_inout_arg_place(place: Place, ctx: Context, node: PlaceNode) -> Place:
-    """Performs additional checks for borrowed place arguments.
+def check_place_assignable(
+    place: Place, ctx: Context, node: ast.expr, reason: str
+) -> Place:
+    """Performs additional checks for assignments to places, for example for borrowed
+    place arguments after function returns.
 
     In particular, we need to check that places involving `place[item]` subscripts
     implement the corresponding `__setitem__` method.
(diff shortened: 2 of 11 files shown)


guppylang/checker/core.py
```diff
@@ -225,6 +225,15 @@ def __str__(self) -> str:
         return f"{self.parent}[...]"
 
 
+def contains_subscript(place: Place) -> SubscriptAccess | None:
+    """Checks if a place contains a subscript access and returns the rightmost one."""
+    while not isinstance(place, Variable):
+        if isinstance(place, SubscriptAccess):
+            return place
+        place = place.parent
+    return None
```
mark-koch, line 234: Moved from `linearity_checker.py` so we can also use it during type checking

guppylang/checker/linearity_checker.py
```diff
@@ -463,17 +464,25 @@ def _check_assign_targets(self, targets: list[ast.expr]) -> None:
...
-                        err.add_sub_diagnostic(PlaceNotUsedError.Fix(None))
-                        raise GuppyError(err)
-                self.scope.assign(tgt_place)
+            # Subscript assignments also require checking the `__setitem__` call
+            if subscript := contains_subscript(tgt.place):
+                assert subscript.setitem_call is not None
+                self.visit(subscript.item_expr)
+                self.scope.assign(subscript.item)
+                self.scope.assign(subscript.setitem_call.value_var)
+                self.visit(subscript.setitem_call.call)
```
mark-koch, line 473: This fixes #835

guppylang/checker/expr_checker.py
```diff
@@ -949,8 +951,11 @@ def type_check_args(
     return new_args, subst
 
 
-def check_inout_arg_place(place: Place, ctx: Context, node: PlaceNode) -> Place:
-    """Performs additional checks for borrowed place arguments.
+def check_place_assignable(
+    place: Place, ctx: Context, node: ast.expr, reason: str
+) -> Place:
```
mark-koch, line 956: I generalised this function since this code was duplicated in the statement checker

guppylang/compiler/stmt_compiler.py
```diff
@@ -72,15 +71,27 @@ def _assign(self, lhs: ast.expr, port: Wire) -> None:
...
+                # Instead of `tmp` just use `xs[i]` as a "name", the dfg tracker doesn't
+                # care about this
+                self.dfg[subscript] = self.expr_compiler.compile(
+                    subscript.getitem_call, self.dfg
+                )
+            # Assign to the name `xs[i].y`
+            self.dfg[lhs.place] = port
+            # Look up `xs[i]` again since it was mutated by the assignment above, then
+            # compile a call to `__setitem__` to actually mutate
+            self.dfg[subscript.setitem_call.value_var] = self.dfg[subscript]
```
mark-koch, line 94: This is the main change that fixes #839

guppylang/compiler/stmt_compiler.py
```diff
@@ -72,15 +71,27 @@ def _assign(self, lhs: ast.expr, port: Wire) -> None:
 
     @_assign.register
     def _assign_place(self, lhs: PlaceNode, port: Wire) -> None:
-        if (subscript := contains_subscript(lhs.place)) and isinstance(
-            lhs.place, SubscriptAccess
-        ):
+        if subscript := contains_subscript(lhs.place):
```
mark-koch, line 74: We previously didn't handle the case where the subscript is nested in the place

tatiana-s (approved):

PR #841, opened 2025-03-06
fix: Fix linearity checking for array copies

### LINKED ISSUE #838: Guppy arrays can be explicitly copied after a move
Similar to #835 

It seems I can explicitly copy a guppy array after an assignment (move) Which should be a linearity violation. This compiles with guppy 0.16.0.

```python
from guppylang import guppy

@guppy
def make_big_array() -> array[int, 96]:
    return array(x*x for x in range(96))


def main() -> None:
    arr1 = make_big_array()
    arr2 = arr1 # Move the value arr1 to arr2
    arr3 = arr1.copy() # copy a moved value?

guppy.compile_module()
```

mark-koch:
Fixes #838

### DIFF
--- a/guppylang/nodes.py
+++ b/guppylang/nodes.py
@@ -387,12 +387,3 @@ class CheckedNestedFunctionDef(ast.FunctionDef):
         self.cfg = cfg
         self.ty = ty
         self.captured = captured
-
-
-class CopyNode(ast.expr):
-    value: "PlaceNode"
-
-    def __init__(self, val: "PlaceNode"):
-        self.value = val
-
-    fields = ("value",)
--- a/guppylang/std/_internal/checker.py
+++ b/guppylang/std/_internal/checker.py
@@ -4,7 +4,7 @@ from typing import ClassVar, cast
 
 from typing_extensions import assert_never
 
-from guppylang.ast_util import with_loc, with_type
+from guppylang.ast_util import get_type, with_loc, with_type
 from guppylang.checker.core import Context
 from guppylang.checker.errors.generic import ExpectedError, UnsupportedError
 from guppylang.checker.errors.type_errors import (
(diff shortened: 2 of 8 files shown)


guppylang/std/_internal/checker.py
```diff
@@ -179,34 +178,22 @@ class NonCopyableElementsError(Error):
...
-        ty: Type
-
     def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
-        check_num_args(1, len(args), self.node)
-        [array_arg] = args
-        array_arg, array_ty = ExprSynthesizer(self.ctx).synthesize(array_arg)
-        if not is_array_type(array_ty):
-            raise GuppyTypeError(
-                ArrayCopyChecker.ExpectedArrayError(self.node, array_ty)
-            )
```
mark-koch, line 197: We shouldn't be hand-rolling our own checking logic here, just call `synthesize_call`. This makes sure we're doing all the required inout checks and also makes error messages more consistent.

guppylang/std/builtins.py
```diff
@@ -597,7 +597,7 @@ def __iter__(self: array[L, n] @ owned) -> "SizedIter[ArrayIter[L, n], n]":
         return SizedIter(ArrayIter(self, 0))
 
     @guppy.custom(CopyInoutCompiler(), ArrayCopyChecker())
-    def copy(self: array[L, n]) -> array[L, n]: ...
+    def copy(self: array[T, n]) -> array[T, n]: ...
```
mark-koch, line 600: The type variable `T` has a `copyable=True` requirement which better matches the actual behaviour enforcced by `ArrayCopyChecker`.

In a sense, the only purpose of `ArrayCopyChecker` now is to give a better error message if the array elements are not implicitly copyable

guppylang/nodes.py
```diff
@@ -387,12 +387,3 @@ def __init__(
...
         self.captured = captured
-
-
-class CopyNode(ast.expr):
-    value: "PlaceNode"
-
-    def __init__(self, val: "PlaceNode"):
-        self.value = val
-
-    fields = ("value",)
```
mark-koch, line 398: Don't treat `copy` as its own special AST node. Just use a regular call and linearity checking will magically work on its own

PR #849, opened 2025-03-10
feat: add barrier operation to builtins

ss2165:
Required a hack custom expression to support inout varargs

Shouldn't be merged until next hugr python release

Couple of small fixed to do with hugr updates


tests/integration/test_quantum.py
```diff
@@ -169,3 +169,30 @@ def test() -> None:
         panic("I panicked!", q)
 
     validate(test)
+
+
+
+def test_barrier(validate):
```
mark-koch, line 175: Could you add a test for barriers with structs/array indexing:

```python
barrier(qs[42], struct.q)
```

mark-koch: Also barrier on classical data and whole arrays. Stuff like

```python
x = False
qs = array(qubit(), qubit())
barrier(42, x, array(1, 2, 3), qs)
```

mark-koch: Also empty `barrier()`

mark-koch (commented):
Could you add some error tests to check that linearity violations are caught correctly? E.g.

```python
barrier(q, q)
```

```python
q = qubit()
barrier(q)
# Drop q without use
```

```python
barrier(qubit())
```

Cargo.toml
```diff
@@ -26,13 +26,13 @@ pyo3 = "0.23.3"
...
 
 [patch.crates-io]
 
 # Uncomment these to test the latest dependency version during development
- hugr = { git = "https://github.com/CQCL/hugr", rev = "5e7c81e" }
- hugr-cli = { git = "https://github.com/CQCL/hugr", rev = "5e7c81e" }
- hugr-llvm = { git = "https://github.com/CQCL/hugr", rev = "5e7c81e" }
+ hugr = { git = "https://github.com/CQCL/hugr", rev = "4e6b6d8" }
+ hugr-cli = { git = "https://github.com/CQCL/hugr", rev = "4e6b6d8" }
+ hugr-llvm = { git = "https://github.com/CQCL/hugr", rev = "4e6b6d8" }
```
mark-koch, line 38: Should we wait to merge this until those are released?

guppylang/compiler/expr_compiler.py
```diff
@@ -514,6 +514,20 @@ def visit_PanicExpr(self, node: PanicExpr) -> Wire:
...
+        op = hugr.std.prelude.PRELUDE_EXTENSION.get_op("Barrier").instantiate(
+            [ht.SequenceArg([ht.TypeTypeArg(ty) for ty in hugr_tys])],
+            ht.FunctionType.endo(hugr_tys),
+        )
+
+        barrier_n = self.builder.add_op(op, *(self.visit(e) for e in node.args))
+
+        self._update_inout_ports(
+            node.args, (barrier_n[i] for i in range(len(hugr_tys))), node.func_ty
+        )
```
mark-koch, line 528:
```suggestion
        self._update_inout_ports(node.args, list(barrier_n), node.func_ty)
```

Surprising that mypy doesn't complain about passing a generator instead of a list?

guppylang/std/_internal/checker.py
```diff
@@ -474,3 +477,17 @@ def to_sized_iter(
...
+class BarrierChecker(CustomCallChecker):
+    """Call checker for the `barrier` function."""
+
+    def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
+        tys = [ExprSynthesizer(self.ctx).synthesize(val)[1] for val in args]
+        func_ty = FunctionType(
+            [FuncInput(t, InputFlags.Inout) for t in tys],
+            NoneType(),
+        )
+        args, ret_ty, _inst = synthesize_call(func_ty, args, self.node, self.ctx)
```
mark-koch, line 491:
```suggestion
        args, ret_ty, inst = synthesize_call(func_ty, args, self.node, self.ctx)
        assert len(inst) == 0, "func_ty is not generic"
```

tests/error/misc_errors/barrier_drop.err
mark-koch, line 1: Imo better to move these tests to `tests/error/linear_errors` / `tests/error/inout_errors`

mark-koch (approved):

Cargo.toml
ss2165: I want to wait until hugr-py release but not too fussed about validation dependencies

Issue #778, opened 2025-01-21
Track functions with side-effects


mark-koch:
We need to ensure the ordering of side-effectful functions like `result`, `panic`, `rng_seed`, `rng_next`, etc.

Since there are no dataflow dependencies between those functions, we need to emit extra order-edges between them within a basic block. This requires tracking of these effects by the compiler and is not trivial:

* A function calling another side-effectful function becomes side-effectful itself
* A call of a higher-order function could also be side-effectful, so we need to track this in the function type, not only the function definition
* There are different kinds of side effects: E.g. it's not required to specify the order between `result` and `rng_next` since their side-effects don't interfer

cqc-alec:
See also https://github.com/CQCL/tket2/issues/748 . Having a linear "state" type passed through these calls seems like a good solution. (We would need several such types, one per type of state).

mark-koch:
But should we expose this to the user? Having to pass the state around seems very cumbersome

cqc-alec:
Agree it would be nice to avoid exposing to the user if we can.

mark-koch:
Some thoughts on how to do this:

* Add an extendable type of `SideEffect` that captures different kinds of side effects (e.g. `PRNG`, `IO`, ...)
* Each builtin function specifies its set of side effects
* During type checking, construct a call-graph
* Run a fixed point iteration over the call graph to infer the side effects of user-defined functions
* During Hugr generation, insert order edges in each basic block to ensure odering of conflicting side-effects
* We also need to ensure ordering of basic blocks when unrolling control-flow. To do this, we need to:
  * Infer the side-effects of each basic block
  * For each BB, compute the set of BBs that can reach it and that are reachable from it
  * If any of those interfer with the side-effects of the current BB, we need to somehow mark that they may not be reordered during unrolling
  * How would this look like in Hugr?

aborgna-q:
MLIR has something similar with their ["categorization"](https://mlir.llvm.org/docs/Rationale/SideEffectsAndSpeculation/#categorization) of side-effect types, but they use numeric "stages" to order execution inside a single basic block.
In our case using order edges should work correctly.

Mind that we may want separate _shared_ vs _unique_ side-effect consumers.
We don't want to enforce order between quantum gates with no data-dependency, but we want _those_ to always come before some other operation that depends on the global quantum state.
The order edges can be added only for _unique_ operations, connecting them to the subset of shared side-effect operations with no data dependencies between them.

[...]

PR #855, opened 2025-03-13
fix: Prevent reordering of operations with side-effects

### LINKED ISSUE #853: Conservatively insert order edges
Users have already run into issues with ordering (see #778). Until we have a proper solution, we should just conservatively insert order edges between the following ops:
* Result
* Panic
* Qubit allocation
* Qubit discard and measurement
* Call
* IndirectCall

mark-koch:
Closes #853
 
This is a temporary solution until we properly solve #778. Therefore, the implementation is somewhat hacky as we're going to remove this in the future anyways. 

The approach is to monkey-patch the `Hugr.add_node` method to implicitly insert state-order edges between nodes that could have side-effects. The list of side-effectful nodes is hard-coded in the compiler.

Also relies on the fix from https://github.com/CQCL/hugr/pull/1971, so we should wait to merge until `hugr-py` is released

### DIFF
--- a/guppylang/compiler/core.py
+++ b/guppylang/compiler/core.py
@@ -1,12 +1,19 @@
 import itertools
 from abc import ABC
+from collections import defaultdict
+from collections.abc import Iterator
+from contextlib import contextmanager
 from dataclasses import dataclass, field
-from typing import cast
+from typing import Any, cast
 
-from hugr import Wire, ops
+import tket2_exts
+from hugr import Hugr, Node, Wire, ops
 from hugr import tys as ht
 from hugr.build import function as hf
 from hugr.build.dfg import DP, DefinitionBuilder, DfBase
+from hugr.hugr.base import OpVarCov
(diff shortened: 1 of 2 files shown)


guppylang/compiler/core.py
```diff
@@ -216,3 +224,93 @@ def return_var(n: int) -> str:
...
+        parent = hugr[node].parent
+        if parent is not None:
+            if prev := prev_node_with_side_effect[parent]:
+                hugr.add_order_link(prev, node)
+            else:
+                # If this is the first side-effectful op in this DFG, make a recursive
+                # call with the parent since the parent is also considered side-
+                # effectful now. We shouldn't walk up through function definitions
+                # or basic blocks though
+                if not isinstance(hugr[parent].op, ops.FuncDefn | ops.DataflowBlock):
```
ss2165, line 307: I think this is ok through Case nodes but just noting it

PR #858, opened 2025-03-13
feat!: Allow array arguments to `load_pytket`

### LINKED ISSUE #671: Support arrays in `pytket` function stubs

tatiana-s:
Closes #671 

Allows you to load pytket circuits as function definitions that use arrays as both inputs and outputs:
```
circ = Circuit(2)
circ.measure_all()

guppy.load_pytket("guppy_circ", circ)

@guppy
def foo(q: array[qubit, 2]) -> array[qubit, 2]:
    return guppy_circ(q)

```

Note that:
- If using more registers than just the default register, this relies on the user sticking to lexicographic order for registers based on their names, with the default qubit and bit registers being `q` and `c` respectively. This is done because the tket2 conversion to HUGR relies on this order for qubit / bit ordering. It may be worth treating default registers differently in the future but that depends on user preferences
- I think the original idea was to use arrays when the user specifies them as such in the explicit function signature - I've decided to go with the implicit version as this a) seems to be used more often by users and b) would have been more complicated to implement with more questions with regards to how much flexibility you want to give / would be expected in the signature definition, but this should be a follow up issue (#859)

BREAKING CHANGE: `load_pytket` takes arrays by default (pass `use_arrays=False` for qubit arguments)


ss2165:
is it feasible to just have one `guppy.load` method? If the behaviour can't be inferred automatically I suggest `guppy.load_pytket` does arrays by default and you add a kwarg to have individual wires.

guppylang/decorator.py
```diff
@@ -512,7 +512,11 @@ def func(f: PyFunc) -> RawPytketDef:
 
     @pretty_errors
     def load_pytket(
-        self, name: str, input_circuit: Any, module: GuppyModule | None = None
+        self,
+        name: str,
+        input_circuit: Any,
+        use_arrays: bool = True,
+        module: GuppyModule | None = None,
     ) -> RawLoadPytketDef:
```
aborgna-q, line 520: Positional bool arguments are confusing. Could we have `use_arrays` be kwargs-only?
That would also avoid having a breaking change.

guppylang/definition/pytket_circuits.py
```diff
@@ -170,15 +178,84 @@ def compile_outer(self, module: DefinitionBuilder[OpVar]) -> "CompiledPytketDef"
...
+                        input_list.extend(unpack(wire, ht.Option(ht.Qubit), q_reg.size))
+
+                else:
+                    # Otherwise pass inputs directly.
+                    input_list = list(outer_func.inputs())
+
+                call_node = outer_func.call(hugr_func, *(input_list + bool_wires))
+
+                # Pytket circuit hugr has qubit and bool wires in the opposite
+                # order.
```
aborgna-q, line 218: Opposite order to what?

The `tket2` conversion should respect [pytket's input qubit order](https://github.com/CQCL/tket2/blob/5abef159c16d0db0149377af92b3a4187ba9b44c/tket2/src/serialize/pytket/decoder.rs#L103-L112), and have [qubits before bits](https://github.com/CQCL/tket2/blob/5abef159c16d0db0149377af92b3a4187ba9b44c/tket2/src/serialize/pytket/decoder.rs#L56-L58).

tatiana-s: The guppy function has regular bit returns first before inout qubit returns

aborgna-q (approved):

PR #863, opened 2025-03-13
fix: Define `len` of arrays using Guppy

### LINKED ISSUE #804: `len` of generic array causing `InternalGuppyError`
example:

```python
@guppy
@no_type_check
def array_any(arr: array[bool, n]) -> bool:
    for i in range(len(arr)):
        if arr[i]:
            return True
    return False
```

throws 
```

    @staticmethod
    def _get_const_len(inst: Inst) -> ast.expr:
        """Helper function to extract the static length from the inferred type args."""
        # TODO: This will stop working once we allow generic function defs. Then, the
        #  argument could also just be variable instead of a concrete number.
        match inst:
            case [_, ConstArg(const=ConstValue(value=int(n)))]:
                return ast.Constant(value=n)
>       raise InternalGuppyError(f"array.__len__: Invalid instantiation: {inst}")
E       guppylang.error.InternalGuppyError: array.__len__: Invalid instantiation: [TypeArg(ty=OpaqueType(args=[], defn=OpaqueTypeDef(id=DefId(id=3, module=None), name='bool', defined_at=None, description='type', params=[], never_copyable=False, never_droppable=False, to_hugr=<function <lambda> at 0x106ccae80>, bound=None))), ConstArg(const=BoundConstVar(ty=NumericType(kind=<Kind.Nat: 1>), display_name='n', idx=0))]
```

using `range(n)` instead works, but I believe this is a regression.

tatiana-s:
Closes #804


tests/integration/test_array.py
```diff
@@ -5,28 +5,34 @@
...
+    @guppy(module)
+    def main(qs: array[float, 42]) -> int:
+        return len(qs)
 
-    hg = package.module
-    [val] = [data.op for node, data in hg.nodes() if isinstance(data.op, ops.Const)]
-    assert isinstance(val, ops.Const)
-    assert isinstance(val.val, IntVal)
-    assert val.val.v == 42
+    validate(module.compile())
```
acl-cqc, line 35: Could you look for an appropriate Hugr "load nat typearg" op? Only if it's easy, really

tests/integration/test_array.py
```diff
@@ -37,14 +43,22 @@ def test_len_linear(validate):
...
-    assert isinstance(val.val, IntVal)
-    assert val.val.v == 42
+    @guppy(module)
+    def main(qs: array[bool, n]) -> bool:
+        for i in range(len(qs)):
+            if qs[i]:
+                return True
+        return False
+
+    module.compile()
```
acl-cqc, line 61: no validate?

guppylang/std/builtins.py
```diff
@@ -579,8 +578,10 @@ def __getitem__(self: array[L, n], idx: int) -> L: ...
     @guppy.custom(ArraySetitemCompiler())
     def __setitem__(self: array[L, n], idx: int, value: L @ owned) -> None: ...
 
-    @guppy.custom(checker=ArrayLenChecker())
-    def __len__(self: array[L, n]) -> int: ...
+    @guppy
+    @no_type_check
+    def __len__(self: array[L, n]) -> int:
+        return n
```
acl-cqc, line 584: I guess the magic here has been done in another PR!

tests/integration/test_array.py
```diff
@@ -5,28 +5,34 @@
...
     module = GuppyModule("test")
 
     @guppy(module)
     def main(xs: array[float, 42]) -> int:
         return len(xs)
 
-    package = module.compile()
-    validate(package)
+    compiled = module.compile()
+    validate(compiled)
```
acl-cqc, line 24: I was about to say, can we refactor this so the `validate` is done even if the `run_int_fn` isn't....but then I realize that, so up to this point, this is exactly equivalent to `test_len` without the run_int_fn (right?). How about changing this one to do
```
if run_int_fn is not None:
  run_int_fn(compiled, 42)
```

and then making `test_len(validate)` just call `test_len_execute(validate, None)`

acl-cqc (approved):
Looks good to me on the basis that the tests pass :). Just a couple of thoughts on those tests...

PR #865, opened 2025-03-17
fix: Fix array comprehensions with generic element type

### LINKED ISSUE #864: Generic array comprehensions produce invalid Hugrs
```python
T = guppy.type_var("T")

@guppy
def foo(x: T) -> None:
    array(x for _ in range(10))
```

fails validation. To fix, we need to make the `init_none` function that is used when compiling array comprehensions generic

mark-koch:
Fixes #864

### DIFF
--- a/guppylang/compiler/expr_compiler.py
+++ b/guppylang/compiler/expr_compiler.py
@@ -1,7 +1,7 @@
 import ast
 from collections.abc import Iterable, Iterator, Sequence
 from contextlib import AbstractContextManager, ExitStack, contextmanager
-from typing import Any, TypeGuard, TypeVar
+from typing import Any, Final, TypeGuard, TypeVar
 
 import hugr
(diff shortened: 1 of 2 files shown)


guppylang/compiler/expr_compiler.py
```diff
@@ -538,9 +544,12 @@ def visit_DesugaredArrayComp(self, node: DesugaredArrayComp) -> Wire:
         # See https://github.com/CQCL/guppylang/issues/629
         hugr_elt_ty = ht.Option(node.elt_ty.to_hugr())
         # Initialise array with `None`s
-        make_none = self.builder.define_function("init_none", [], [hugr_elt_ty])
```
mark-koch, line 541: This definiton was invalid if `hugr_elt_ty` contains a type variable

tatiana-s (approved):

PR #870, opened 2025-03-18
ci: Update uv.lock on release-please PRs

aborgna-q:
`uv 0.6.6` adds a hard error when the version of the package in `uv.lock` does not match the one declared in the editable wheel: https://github.com/astral-sh/uv/pull/12235.

`release-please` doesn't update the lock files automatically, so release PRs needed a manual lock update.
This PR adds the lockfile as an `extra-file`, to avoid this manual step.

Note that the _jsonpath_ in the config is not technically correct, as release-please's json array filters don't quite work for toml files. See https://github.com/googleapis/release-please/issues/2455, which includes this hacky workaround.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -7,7 +7,14 @@
       "bump-patch-for-minor-pre-major": false,
       "draft": false,
       "prerelease": false,
-      "draft-pull-request": true
+      "draft-pull-request": true,
+      "extra-files": [
+        {
+          "type": "toml",
+          "path": "uv.lock",
+          "jsonpath": "$.package[?(@.name.value=='guppylang')].version"
+        }
+      ]
     }
   },
   "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json"


ss2165 (approved):
pls wait for release before merging (I know it shouldn't matter but paranoia)

konstin:
I wasn't expecting https://github.com/astral-sh/uv/pull/12235 to break any existing workflows or require manual edits to `uv.lock`, do you mind sharing what your release workflow is / how this PR broke it?

aborgna-q:
Hi! Sorry for the noisy back-links.

The workflow here is running [googleapis/release-please](https://github.com/googleapis/release-please) to automatically create release PRs with including changelog entries and bumped package versions.

The version update logic is quite simple, and manually edits the version in a couple pre-determined places (`pyproject.toml`, `package.__version__`, etc.) but does not include `uv.lock`. This seems similar to the issue with `dependabot` mentioned in your PR.

I rolled back this change in a fork and updated uv to show the error.
See this automatically-generated release PR: https://github.com/aborgna-q/guppylang/pull/4
And [failing CI check](https://github.com/aborgna-q/guppylang/actions/runs/13973211117/job/39120191415?pr=4#step:6:22).

Note that [the check workflow](https://github.com/aborgna-q/guppylang/blob/6b10dd1f7252dbfb96920052de8833b486fbb7bc/.github/workflows/pull-request.yaml#L43) only fails if we run `uv` with `--locked`. We instead use `--frozen` in some places, and that doesn't raise issues.

PR #873, opened 2025-03-20
fix(qsystem)!: remove `random_nat` and `maybe_rng`

qartik:
For random_nat, see: quantinuum-dev/hugrverse#138

Discussion on maybe_rng at https://github.com/CQCL/guppylang/pull/822#issuecomment-2741571177

BREAKING CHANGE: removed `random.RNG.random_nat()` and `random.maybe_rng()`


ss2165:
This PR removes previously unsupported (downstream) functions so should not need a breaking release - I will override the commit message

PR #884, opened 2025-03-25
chore: Remove duplicate classes in builtins

mark-koch:
This is to improve API docs for the Guppy standard library.

Also added some docstrings as a drive-by.

### DIFF
--- a/guppylang/std/builtins.py
+++ b/guppylang/std/builtins.py
@@ -1,7 +1,11 @@
 """Guppy module for builtin types and operations."""
 
-# mypy: disable-error-code="empty-body, misc, override, valid-type, no-untyped-def"
+# ruff: noqa: E501
+# mypy: disable-error-code="empty-body, misc, override, valid-type, no-untyped-def, has-type"
 
+from __future__ import annotations
+
+import builtins
 from typing import Any, Generic, TypeVar, no_type_check
 
 import hugr.std.int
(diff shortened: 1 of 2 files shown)


guppylang/std/builtins.py
```diff
@@ -565,7 +572,9 @@ def __trunc__(self: float) -> float: ...
 
 
 @guppy.extend_type(list_type_def)
-class List:
+class list(Generic[_T]):
+    """Mutable sequence items with homogeneous types."""
```
CalMacCQ, line 576: Just checking that a `list` can't have `hetrogeneous` types? Is it just the fact that the length can change at compile time that distinguishes a list from a mutable `array`?

guppylang/std/builtins.py
```diff
@@ -612,31 +623,38 @@ def __new__(): ...
...
-    def __iter__(self: array[L, n] @ owned) -> "SizedIter[ArrayIter[L, n], n]":
+    def __iter__(self: array[L, n] @ owned) -> SizedIter[ArrayIter[L, n], n]:
         return SizedIter(ArrayIter(self, 0))
 
     @guppy.custom(CopyInoutCompiler(), ArrayCopyChecker())
     def copy(self: array[T, n]) -> array[T, n]: ...
 
+    def __new__(cls, *args: _T) -> builtins.list[_T]:  # type: ignore[no-redef]  # noqa: F811
+        # Runtime array constructor that is used for comptime. We return an actual list
+        # in line with the comptime unpacking logic that turns arrays into lists.
```
CalMacCQ, line 634: Just checking it shouldn't be "lists into arrays" rather than "arrays into lists"

CalMacCQ (approved):
Looks good to me, just a few questions

guppylang/std/builtins.py
mark-koch: >  Is it just the fact that the length can change at runtime that distinguishes a list from a mutable array?

Yes, that's the only difference

guppylang/std/builtins.py
mark-koch: The point is that if a user writes ``array(x, y, z)``, they will actually get back a list in comptime mode. Similarly, if a function takes an array, the comptime implementation will actually get a list

CalMacCQ: Ah I think I was mixing things up with `comptime` python lists being treated as static arrays, thanks

tatiana-s (approved):

PR #900, opened 2025-03-28
refactor!: Change bool lowering to an opaque type

### LINKED ISSUE #862: Change bool lowering to an opaque type
Stop using sum

tatiana-s:
Closes #862

BREAKING CHANGE: Bools in HUGRs generated by Guppy are represented using a `tket2.bool` instead of the HUGR sum type


tatiana-s:
(draft because this depends on both a tket2 release and also for the lowering to `Sum( Sum( () | () ), ResultId)` in hugr)

tatiana-s:
main things I am unsure about is where to put all the bool definitions (currently in `tket2_bool.py` in the std compiler), and also I was surprised I didn't seem to have to change anything with tail loops to make all the tests pass

guppylang/compiler/expr_compiler.py
```diff
@@ -185,8 +191,12 @@ def _if_else(
         """Builds a `Conditional`, returning context managers to build the `True` and
         `False` branch.
         """
+        cond_wire = self.visit(cond)
+        cond_ty = self.builder.hugr.port_type(cond_wire.out_port())
+        if cond_ty == OpaqueBool:
+            cond_wire = self.builder.add_op(bool_to_sum(), cond_wire)
```
mark-koch, line 197: I would also expect a similar change in `_new_loop`. Not sure why tests are passing?

guppylang/std/_internal/compiler/tket2_bool.py
mark-koch, line 1: This location for the bool ops looks good to me for now 👍

mark-koch (commented):
Looks good! My main comment is that I'd like to get rid of as many `if ty == OpaqueBool:` lines as possible. I feel like in most cases we already know because of type checking what the type of a wire will be.

I'm also concerned that the tests pass even though `ExprCompiler._new_loop` was not updated. Feels like something must be off...

guppylang/compiler/cfg_compiler.py
```diff
@@ -100,6 +101,10 @@ def compile_bb(
     if len(bb.successors) > 1:
         assert bb.branch_pred is not None
         branch_port = ExprCompiler(ctx).compile(bb.branch_pred, dfg)
+        # Convert the bool predicate into a sum for branching, unless it already is one.
+        pred_ty = builder.hugr.port_type(branch_port.out_port())
+        if pred_ty == OpaqueBool:
+            branch_port = dfg.builder.add_op(bool_to_sum(), branch_port)
```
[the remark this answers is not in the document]
tatiana-s: It seems to be in quite a lot of cases, all the tests in `test_for.py` for example, I am not sure why

guppylang/compiler/expr_compiler.py
```diff
@@ -185,8 +191,12 @@ def _if_else(
         """Builds a `Conditional`, returning context managers to build the `True` and
         `False` branch.
         """
+        cond_wire = self.visit(cond)
+        cond_ty = self.builder.hugr.port_type(cond_wire.out_port())
+        if cond_ty == OpaqueBool:
```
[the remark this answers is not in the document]
tatiana-s: Yes, all the comprehensions fail otherwise as the `cond_wire` seems to be a non-empty option of a tuple in those cases

guppylang/compiler/expr_compiler.py
tatiana-s: The break predicate in `_new_loop` always seems to take an `Either(Tuple(some type))`

guppylang/compiler/expr_compiler.py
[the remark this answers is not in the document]
mark-koch: I see, this makes sense following the changes in #833 👍

guppylang/compiler/expr_compiler.py
mark-koch: Also makes sense following the changes in #833 👍

mark-koch (approved):

PR #905, opened 2025-04-01
feat: Add debug module with `state_result` function

### LINKED ISSUE #899: Add `output_state` function
See https://github.com/quantinuum-dev/hugrverse/issues/65

tatiana-s:
Closes #899


guppylang/compiler/expr_compiler.py
```diff
@@ -533,6 +539,33 @@ def visit_BarrierExpr(self, node: BarrierExpr) -> Wire:
...
+        sig = ht.FunctionType(
+            [array_type(ht.Qubit, num_qubits_arg)],
+            [array_type(ht.Qubit, num_qubits_arg)],
+        )
+        op = ops.Custom(
+            op_name="StateResult", signature=sig, args=args, extension="tket2.debug"
+        )
+
+        qubits_in = [self.visit(e) for e in node.args[1:]]
+        # TODO: This should be caught during checking.
```
tatiana-s, line 555: Importing `qubit` for the `StateResultChecker` in `checker.py` leads to circular import errors - not sure if there is an obvious other place to move it to

guppylang/compiler/expr_compiler.py
```diff
@@ -533,6 +539,33 @@ def visit_BarrierExpr(self, node: BarrierExpr) -> Wire:
...
+        for q in qubits_in:
+            t = self.builder.hugr.port_type(q.out_port())
+            if t != ht.Qubit:
+                raise GuppyError(
+                    UnsupportedError(node, "Non-qubit inputs to state_result")
+                )
+        qubits_arr = self.builder.add_op(array_new(ht.Qubit, num_qubits), *qubits_in)
+        qubits_arr = self.builder.add_op(op, qubits_arr)
+        qubits_out = unpack_array(self.builder, qubits_arr)
```
tatiana-s, line 564: Not sure if using arrays in the hugr op is worth this overhead or there is a better way of representing the list of qubits in the extension

tests/error/misc_errors/state_result_missing_tag.err
```diff
@@ -0,0 +1,8 @@
+Error: Type mismatch (at $FILE:12:17)
+   | 
+10 | def main() -> None:
+11 |     q1 = qubit()
+12 |     state_result(q1)
+   |                  ^^ Expected expression of type `str`, got `qubit`
```
tatiana-s, line 6: (While it would be nice to be more specific about the missing tag this is the same as result at the moment)

guppylang/compiler/expr_compiler.py
mark-koch: I think this is fine for now. Would be a good idea to add an overload of `result_state` that takes an array anyways as a follow-up

guppylang/std/_internal/checker.py
```diff
@@ -536,3 +538,27 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+    def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
+        tag, _ = ExprChecker(self.ctx).check(args[0], string_type())
+        if not isinstance(tag, ast.Constant) or not isinstance(tag.value, str):
+            raise GuppyTypeError(ExpectedError(tag, "a string literal"))
+        if len(tag.value.encode("utf-8")) > TAG_MAX_LEN:
+            err: Error = TooLongError(tag)
+            err.add_sub_diagnostic(TooLongError.Hint(None))
+            raise GuppyTypeError(err)
+
+        tys = [ExprSynthesizer(self.ctx).synthesize(val)[1] for val in args]
```
mark-koch, line 555: You shouldn't ignore the expression returned by `synthesize`. E.g. instead

```suggestion
        args, tys = zip(*(ExprSynthesizer(self.ctx).synthesize(val) for val in args), strict=True)
```

mark-koch: Also no need to resynthesize the type for `args[0]` since you already checked it above

guppylang/std/_internal/checker.py
```diff
@@ -536,3 +538,27 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        func_ty = FunctionType(
+            [FuncInput(tys[0], InputFlags.NoFlags)]
+            + [FuncInput(t, InputFlags.Inout) for t in tys[1:]],
```
mark-koch, line 558: Then you could just do

```suggestion
            [FuncInput(string_type(), InputFlags.NoFlags)]
            + [FuncInput(t, InputFlags.Inout) for t in tys],
```

here

tests/error/misc_errors/state_result_not_qubit.err
```diff
@@ -0,0 +1,8 @@
+Error: Unsupported (at $FILE:9:4)
+  | 
+7 | @guppy(module)
+8 | def main(x: int) -> None:
+9 |     state_result("tag", x)
+  |     ^^^^^^^^^^^^^^^^^^^^^^ Non-qubit inputs to state_result are not supported
```
mark-koch, line 6: Would be nicer if the span only highlighted `x`

tests/integration/test_state_result.py
```diff
@@ -0,0 +1,98 @@
...
+        discard(q1)
+        discard(q2)
+
+    validate(main)
+
+
+def test_empty(validate):
+    @compile_quantum_guppy
+    def main() -> None:
+        state_result("tag")
```
mark-koch, line 27: Not sure if we should allow this. I could see users writing this with the assumption that the *whole state* will be outputted because they are not aware that you have to pass the qubits.

Since empty state results are useless anyways, maybe it would be more helpful to add a diagnostic along the lines of

```
2 |     state_result("tag")
  |                       ^ Qubits whose state should be reported must be passed explicitly
```

mark-koch (commented):
Looks good! Just a couple of nits and suggestions

tests/error/misc_errors/result_tag_too_big.py
```diff
@@ -1,11 +1,8 @@
...
-# from guppylang.prelude._internal.checker import TAG_MAX_LEN
-# BIG_TAG = "a" * (TAG_MAX_LEN + 1)
+TAG_MAX_LEN = 200
 
 @compile_guppy
-def foo(y: bool) -> None:
-    # each tick or cross is 3 bytes. The cross sends the tag over the limit.
-    result("✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅❌", y)
+def foo(y: bool) -> None:    
+    result(comptime("a" * (TAG_MAX_LEN + 1)), y)
```
mark-koch, line 8: I think we shouldn't change this since it's testing that we consider the utf8 encoding of the tag instead of its Python `len`

mark-koch (approved):

guppylang/std/_internal/debug.py
```diff
@@ -43,18 +54,38 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
-            [FuncInput(string_type(), InputFlags.NoFlags)]
-            + [FuncInput(qubit_ty, InputFlags.Inout)] * len(args[1:]),
-            NoneType(),
-        )
+
+        array_len = None
+        arg, ty = ExprSynthesizer(self.ctx).synthesize(args[1])
+        if is_array_type(ty):
+            if len(args) > 2:
+                raise GuppyTypeError(self.MoreThanOneArrayError(self.node))
```
mark-koch: The diagnostic wording "Only one array is allowed to be passed to `state_result`" might be slightly confusing if the second thing that is passed is not an array

mark-koch: Alternatively, you could consider to just throw a generic `WrongNumberOfArgsError`

mark-koch: (Also would prefer to use `args[2]` as the span instead of the whole `self.node`)

guppylang/std/_internal/debug.py
```diff
@@ -43,18 +54,38 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+            element_ty = get_element_type(ty)
+            if not element_ty == qubit_ty:
+                raise GuppyTypeError(ExpectedError(arg, "an array of qubits"))
```
mark-koch, line 70: Maybe a `TypeMismatchError` would be more consistent here? Your call

guppylang/std/_internal/debug.py
```diff
@@ -43,18 +54,38 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+                    FuncInput(ty, InputFlags.Inout),
+                ],
+                NoneType(),
+            )
+            array_len = get_array_length(ty)
+        else:
+            for arg in args[1:]:
+                qbt, _ = ExprChecker(self.ctx).check(arg, qubit_ty)
+                if not ExprChecker(self.ctx).check(arg, qubit_ty):
+                    raise GuppyTypeError(ExpectedError(arg, "a qubit"))
```
mark-koch, line 84:
```suggestion
```
You're already checking this in the line above. Also `ExprChecker.check` never returns a falsy value

PR #907, opened 2025-04-07
fix: Fix diagnostics spans for dynamically defined notebook functions

### LINKED ISSUE #906: Confusing undefined variable error message
Consider the following example

```python
from guppylang import GuppyModule, guppy
from guppylang.std.builtins import array, comptime
from guppylang.std import quantum
from guppylang.std.quantum import qubit


def build_phase_gadget_prog(n_qb: int, theta: float):
    my_module = GuppyModule(name="phase_gadget")
    my_module.load_all(quantum)
    
    @guppy(my_module)
    def phase_gadget(qs: array[qubit, comptime(n_qb)]) -> None:    
        for i in range(n_qb - 1):
            cx(qs[i], qs[i + 1])

        rz(qs[-1], angle(comptime(theta)))

        for i in range(n_qb - 1):
            cx(qs[i], qs[i + 1])


    return phase_gadget

phase_gadget1 = build_phase_gadget_prog(5, 0.5)
phase_gadget2 = build_phase_gadget_prog(10, 0.25)


@guppy
def main() -> None:
    qs1 = array(qubit() for _ in range(5))
    qs2 = array(qubit() for _ in range(10))

    phase_gadget1(qs1)
    phase_gadget2(qs2)

guppy.compile_module();
```

Running this in a notebook, I get the following error...

```
Error: Variable not defined (at <In [?]>:4:23)
  | 
2 |     def phase_gadget(qs: array[qubit, comptime(n_qb)]) -> None:    
3 |         for i in range(n_qb - 1):
4 |             cx(qs[i], qs[i + 1])
  |                        ^^^^ `n_qb` is not defined

```

The error is a bit misleading as it points to the array indexing as being the problematic portion of the code. The actual issue is that we need to use `comptime` when specifying the range. I.e. 

```python
for i in range(comptime(n_qb) - 1):
    cx(qs[i], qs[i + 1])
```

mark-koch:
Fixes #906

### DIFF
--- a/guppylang/definition/function.py
+++ b/guppylang/definition/function.py
@@ -265,7 +265,7 @@ def parse_py_func(f: PyFunc, sources: SourceMap) -> tuple[ast.FunctionDef, str |
                 # got from inspect. Line numbers will be wrong, but that's the best we
                 # can do.
                 sources.add_file(file, source)
-                line_offset = 1
+                line_offset = 0
     else:
         file = inspect.getsourcefile(f)
         if file is None:
--- a/tests/integration/notebooks/misc_notebook_tests.ipynb
+++ b/tests/integration/notebooks/misc_notebook_tests.ipynb
@@ -1,0 +1,98 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 1,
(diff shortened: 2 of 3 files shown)


aborgna-q (approved):
LGTM

PR #910, opened 2025-04-08
feat: Array results

### LINKED ISSUE #631: Reenable array results
They got disabled in #630 to fix #629

Blocked by #628

mark-koch:
Closes #631

### DIFF
--- a/guppylang/checker/linearity_checker.py
+++ b/guppylang/checker/linearity_checker.py
@@ -54,6 +54,7 @@ from guppylang.nodes import (
     LocalCall,
     PartialApply,
     PlaceNode,
+    ResultExpr,
     SubscriptAccessAndDrop,
     TensorCall,
 )
@@ -61,6 +62,7 @@ from guppylang.tys.ty import (
     FuncInput,
     FunctionType,
     InputFlags,
+    NoneType,
     StructType,
 )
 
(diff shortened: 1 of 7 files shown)


guppylang/compiler/expr_compiler.py
```diff
@@ -753,6 +759,21 @@ def array_comprehension_init_func(ctx: CompilerContext) -> hf.Function:
...
+def array_unwrap_elem(ctx: CompilerContext) -> hf.Function:
+    """Returns the Hugr function that is used to unwrap the elements in an option array
+    to turn it into a regular array."""
+    v = ht.Variable(0, ht.TypeBound(ht.TypeBound.Any))
+    sig = ht.PolyFuncType(
+        params=[ht.TypeTypeParam(ht.TypeBound.Any)],
+        body=ht.FunctionType([ht.Option(v)], [v]),
+    )
+    func, already_defined = ctx.declare_global_func(ARRAY_UNWRAP_ELEM, sig)
+    if not already_defined:
```
ss2165, line 771: guess not relevant to this PR but I would have expected the declare function to return `function | None` for this kind of pattern and to condition off that.

I assume there is some valid usage for non-null function with `already_defined` being False?

guppylang/checker/linearity_checker.py
```diff
@@ -419,6 +421,13 @@ def visit_BarrierExpr(self, node: BarrierExpr) -> None:
         self._visit_call_args(node.func_ty, node)
         self._reassign_inout_args(node.func_ty, node)
 
+    def visit_ResultExpr(self, node: ResultExpr) -> None:
```
ss2165, line 424: this is now necessary because arrays are affine and so we need the linearity checking to get involved correct?

ss2165: should we have some error test that such linearity errors are being caught?

guppylang/compiler/expr_compiler.py
mark-koch: If `already_defined = False`, the returned `func` is a builder to construct the function. This way, we don't need to separately put the defined function back into the context once we're done building

guppylang/checker/linearity_checker.py
mark-koch: > this is now necessary because arrays are affine and so we need the linearity checking to get involved correct?

Exactly!

> should we have some error test that such linearity errors are being caught?

Yes, I'll add a "result after move" test 👍

ss2165 (approved):
thanks for the tests

PR #911, opened 2025-04-08
feat: add optional progress_bar for QsysResult

jake-arkinstall:
QsysResult is given an Iterable over shots. In the case where this iterable is a generator, a progress bar is handy for seeing the number of shots produced so far, as well as an estimate of the amount of time remaining for the next shot. The typical use case would be a simulation over many shots, or long shots, or both.

To do this, I add tqdm as an optional dependency, and add an optional `progress_bar` argument to `QsysResult::__init__`. If this is given True, tqdm is imported (upon failure, an `ImportError` is raised telling the user that they need to install tqdm or add the `progress` optional dependency group), and the results are wrapped in a tqdm wrapper. This displays a progress bar to users.

Usage:
```python
shots = QsysResult(..., progress_bar=True)
```

Note: in notebooks that are subject to pytest, progress bars are not recommended - their output is nondeterministic in terms of width (which depends on the available output width) and in terms of time taken to run. Pytest checks for an exact string match in terms of the output cell. It'd be great if there is a way to avoid this, but I don't know of one.

### DIFF
--- a/guppylang/qsys_result.py
+++ b/guppylang/qsys_result.py
@@ -139,8 +139,19 @@ class QsysResult:
     results: list[QsysShot]
 
     def __init__(
-        self, results: Iterable[QsysShot | Iterable[TaggedResult]] | None = None
+        self,
+        results: Iterable[QsysShot | Iterable[TaggedResult]] | None = None,
+        progress_bar: bool = False
     ):
+        if progress_bar:
+            # verify tqdm is available
+            try:
+                from tqdm import tqdm
+            except ImportError as e:
+                raise ImportError(
+                    "tqdm is required for progress bar, install with the `progress` extra"
+                ) from e
+            results = tqdm(results, unit="shots")
         self.results = [
             res if isinstance(res, QsysShot) else QsysShot(res) for res in results or []
         ]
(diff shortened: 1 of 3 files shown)


qartik (approved):
Seems useful to me.

CalMacCQ:
We can close this I think as `QsysResult` now lives in the hugr-py package

acl-cqc:
@jake-arkinstall messaged you on slack but not sure if you've seen. Sorry that I'd ignored this for ages - I'd seen it was “approved” so left it, but now I see Kartik’s comment and hence you might still be looking for review! Sorry 😞….

As Callum says, QsysResult is now in the hugr repo - so I don’t know where’s best to go with this, but if you want to either update it, or move PR to Hugr, then let me know and I’ll review ASAP?

jake-arkinstall:
Sorry @acl-cqc , I remember you asking and I decided I'd look into it, and never did!

Guppy is gaining its own Selene interface for common guppy user patterns and, while I still recommend using tqdm for simulations, I would still like to see this in the run_shots wrapper that emerges.

Can I keep this around and rebase when the time comes?

PR #912, opened 2025-04-08
docs: add qsystem gate definitions

CalMacCQ:


guppylang/std/qsystem/__init__.py
```diff
@@ -21,33 +21,90 @@
...
+
     f1 = float(angle)
     _rz(q, f1)
 
 
 @guppy.hugr_op(quantum_op("Measure", ext=QSYSTEM_EXTENSION), module=qsystem)
 @no_type_check
-def measure(q: qubit @ owned) -> bool: ...
+def measure(q: qubit @ owned) -> bool:
+    """Measure a qubit destructively."""
```
qartik, line 107: It's worth noting that this version of `measure` is blocking unlike `quantum.measure` that is lazy. I am unsure if that should be documented here or now.

qartik: Never mind, I am wrong. Both of the variants of measure are lazy.

ss2165:
1. Increasingly it looks like it would be handy to have a utility that defines and compiles a single function. To be used in exmamples, doc tests, prototyping, etc. @mark-koch are you opposed to adding a dedicated decorator for this? (prompted by having to call `compile()` at the end of the doctest)
2. Don't duplicate docs, cross-link to the non-functional one in each case. e.g. "Functional variant of zz_phase, see [`guppylang.std.qsystem.zz_phase`]()"

mark-koch:
> are you opposed to adding a dedicated decorator for this? 

If we use this in all examples then that's what people will copy and use themselves. So I think we have to be careful that this doesn't become an anti-pattern. In particular, it would be bad if people slapped that decorator on every function.

Also, where would you get the compiled Hugr from? Returning it from the decorator would make it hard to use those functions for comptime

ss2165:
>  we have to be careful that this doesn't become an anti-pattern

yes good point, what about a kwarg to `@guppy` that makes it clear that this is a thing you shouldn't do normally? Or a `check` kwarg that only does the checking stage and not compilation to hugr?

guppylang/std/qsystem/__init__.py
```diff
@@ -17,6 +17,19 @@
 @guppy
 @no_type_check
 def phased_x(q: qubit, angle1: angle, angle2: angle) -> None:
+    r"""phased_x gate command.
+
+    .. math::
+
+        \mathrm{phased\_x}(q, \theta_1, \theta_2)=
+          \mathrm{Rz(\theta_2)Rx(\theta_1)Rz(\theta_2)}&=
```
cqc-alec:
```suggestion
          \mathrm{Rz(\theta_2)Rx(\theta_1)Rz(-\theta_2)}&=
```

guppylang/std/qsystem/__init__.py
```diff
@@ -25,30 +38,72 @@ def phased_x(q: qubit, angle1: angle, angle2: angle) -> None:
 @guppy
 @no_type_check
 def zz_max(q1: qubit, q2: qubit) -> None:
-    """Maximally entangling ZZPhase.
-
-    Equivalent to `zz_phase(q1, q2, pi / 2)`.
+    r"""zz_max gate command. A maximally entangling zz_phase gate.
+
+    This is a special case of the ZZPhase operation with angle = pi/2
```
cqc-alec: Let's be consistent about capitalization in the op names (here we have both `zz_phase` and `ZZPhase`).

cqc-alec:
```suggestion
    r"""zz_max gate command. A maximally entangling zz_phase gate.

    This is a special case of the ZZPhase operation with angle :math:`\pi/2`.
```

guppylang/std/qsystem/__init__.py
```diff
@@ -25,30 +38,72 @@ def phased_x(q: qubit, angle1: angle, angle2: angle) -> None:
...
 
 
 @guppy
 @no_type_check
 def rz(q: qubit, angle: angle) -> None:
+    r"""rz gate command.
+
+    .. math::
+        \mathrm{rz}(q,\theta)=
+        \exp(\frac{- i \pi \theta}{2}\big(Z \big))=
```
cqc-alec:
```suggestion
        \exp(\frac{- i \pi \theta}{2} Z ) =
```

guppylang/std/qsystem/functional.py
```diff
@@ -1,3 +1,9 @@
+"""Guppy standard module for functional qsystem native operations. For the mathematical
+definitions of these gates, see the guppylang.std.qsystem documentation.
+
+These gates are the same as those in the std.qsystem but use functional syntax.
```
cqc-alec:
```suggestion
These gates are the same as those in std.qsystem but use functional syntax.
```

guppylang/std/quantum/__init__.py
```diff
@@ -49,102 +49,274 @@ def maybe_qubit() -> Option[qubit]:
...
 @guppy.custom(RotationCompiler("Ry"))
 @no_type_check
-def ry(q: qubit, angle: angle) -> None: ...
+def ry(q: qubit, angle: angle) -> None:
+    r"""Ry gate.
+
+    .. math::
+        \mathrm{Ry}(q, \theta)=
+          \begin{pmatrix}
+            \cos(\frac{\pi \theta}{2}) & \sin(\frac{\pi \theta}{2}) \\
```
cqc-alec:
```suggestion
            \cos(\frac{\pi \theta}{2}) & -\sin(\frac{\pi \theta}{2}) \\
```

guppylang/std/qsystem/__init__.py
```diff
@@ -17,6 +17,19 @@
 @guppy
 @no_type_check
 def phased_x(q: qubit, angle1: angle, angle2: angle) -> None:
+    r"""phased_x gate command.
+
+    .. math::
+
+        \mathrm{phased\_x}(q, \theta_1, \theta_2)=
```
cqc-alec: I'm not sure we need the qubits in the argument lists here. They don't appear on the RHS so seems a bit odd to have them on the LHS.

I suggest using pytket capitalizations everywhere, e.g. `\mathrm{PhasedX}(\theta_1, \theta_2)`.

cqc-alec (changes requested):

guppylang/std/qsystem/__init__.py
CalMacCQ: The qubit arguments were a deliberate choice here (not sure if it was the optimal choice)

Unlike pytket, Guppy has no notion of a `CX` gate as an `Op` independent of qubits. Guppy implements gates as functions which take qubits as arguements (more like a `Command` in pytket. It seemed more correct to add them.

This is potentially hairy and unsatisfying though as the matrix defintion could vary in cases where the qubits passed as arguments are not adjacent to one another. Many multi-qubit gates are also not invariant under qubit reordering.

Do you have a preference @ss2165?

ss2165: I think you have to separate the matrix notation from the function call notation. So no qubits here, and state somewhere that qubit ordering is implicitly always in the order they appear in the function arg list, or introduce some extra notation 

```markdown
rzz(q1, q2, theta_2)
Qubit ordering: [q1, q2]
.. math::
    \mathrm{RZZ}(\theta_2} = 
```

guppylang/std/quantum/__init__.py
```diff
@@ -49,102 +49,294 @@ def maybe_qubit() -> Option[qubit]:
...
 
 @guppy.hugr_op(quantum_op("CZ"))
 @no_type_check
-def cz(control: qubit, target: qubit) -> None: ...
+def cz(control: qubit, target: qubit) -> None:
+    r"""Controlled-Z gate command.
+
+    cz(control, target)
+
+    Qubit ordering: [control, target]
```
CalMacCQ, line 71: This gate is symmetirc but added this line for consistency

guppylang/std/qsystem/__init__.py
CalMacCQ: Added in [109f294](https://github.com/CQCL/guppylang/pull/912/commits/109f294d1b19df84d27101ff3472d4bd1c02e681) and [7c730be](https://github.com/CQCL/guppylang/pull/912/commits/7c730be9f4a3764a7e0b25c2d3d2c0f09a375dad)

guppylang/std/qsystem/__init__.py
```diff
@@ -17,6 +17,19 @@
 @guppy
 @no_type_check
 def phased_x(q: qubit, angle1: angle, angle2: angle) -> None:
+    r"""phased_x gate command.
+
+    .. math::
+
+        \mathrm{PhasedX}(\theta_1, \theta_2)=
+          \mathrm{Rz(\theta_2)Rx(\theta_1)Rz(\theta_2)}&=
```
cqc-alec:
```suggestion
          \mathrm{Rz(\theta_2)Rx(\theta_1)Rz(-\theta_2)}&=
```
This was marked as resolved but the change was not made.

guppylang/std/qsystem/__init__.py
cqc-alec: Not addressed.

cqc-alec: Marked as resolved but not addressed.

cqc-alec (changes requested):
I started going through the changes but found that my comments had been marked as resolved but not addressed. Is this due to git issues?

CalMacCQ:
> I started going through the changes but found that my comments had been marked as resolved but not addressed. Is this due to git issues?

Hmm strange. I don't know why the changes were not applied. Will fix, sorry.

guppylang/std/qsystem/__init__.py
```diff
@@ -25,30 +38,80 @@ def phased_x(q: qubit, angle1: angle, angle2: angle) -> None:
 @guppy
 @no_type_check
 def zz_max(q1: qubit, q2: qubit) -> None:
-    """Maximally entangling ZZPhase.
+    r"""zz_max gate command. A maximally entangling zz_phase gate.
+
+    This is a special case of the zz_max command with theta = pi/2
```
cqc-alec:
```suggestion
    This is a special case of the zz_phase command with :math:`\theta = \pi/2`
```

guppylang/std/qsystem/__init__.py
cqc-alec: No need for the parens around Z.

guppylang/std/quantum/__init__.py
cqc-alec: Not done.

guppylang/std/qsystem/__init__.py
CalMacCQ: Something got messed up here with git, will fix again.

CalMacCQ: fixed [e683ab0](https://github.com/CQCL/guppylang/pull/912/commits/e683ab0dcbf67c17634489d08c2eaab76f6f74a2)

guppylang/std/quantum/functional.py
```diff
@@ -1,4 +1,8 @@
-"""Guppy standard module for quantum operations."""
+"""Guppy standard module for functional quantum operations. For the mathematical
+definitions of these gates, see the guppylang.std.quantum documentation.
+
+These gates are the same as those in the std.quantum but use functional syntax.
```
cqc-alec:
```suggestion
These gates are the same as those in std.quantum but use functional syntax.
```

cqc-alec (commented):
Just one correction missed I think.

guppylang/std/qsystem/__init__.py
```diff
@@ -25,30 +38,80 @@ def phased_x(q: qubit, angle1: angle, angle2: angle) -> None:
...
+    zz_max(q1, q2)
+
+    Qubit ordering: [q1, q2]
 
-    Equivalent to `zz_phase(q1, q2, pi / 2)`.
+    .. math::
+        \mathrm{ZZMax}=
+        \exp(\frac{- i}{4}\big(Z \otimes Z \big))=
+          \begin{pmatrix}
+            e^{\frac{-i}{4}} & 0 & 0 & 0 \\
```
cqc-alec: Missing pis here.

guppylang/std/quantum/__init__.py
```diff
@@ -49,102 +49,295 @@ def maybe_qubit() -> Option[qubit]:
...
 @no_type_check
-def s(q: qubit) -> None: ...
+def s(q: qubit) -> None:
+    r"""S gate.
+
+    .. math::
+        \mathrm{S}=
+          \begin{pmatrix}
+            1 & 0 \\
+            0 & e^{i \frac{\pi}{2}}
```
cqc-alec:
```suggestion
            0 & i
```
Simpler.

guppylang/std/quantum/__init__.py
```diff
@@ -49,102 +49,295 @@ def maybe_qubit() -> Option[qubit]:
...
 @no_type_check
-def sdg(q: qubit) -> None: ...
+def sdg(q: qubit) -> None:
+    r"""Sdg gate.
+
+    .. math::
+        \mathrm{S}^\dagger=
+          \begin{pmatrix}
+            1 & 0 \\
+            0 & e^{-i \frac{\pi}{2}}
```
cqc-alec:
```suggestion
            0 & -i
```

guppylang/std/qsystem/__init__.py
CalMacCQ: Good catch! thanks. Fixed it in [1188e13](https://github.com/CQCL/guppylang/pull/912/commits/1188e1386b595e01920afe108cb763869b6476d3).

cqc-alec (approved):

PR #913, opened 2025-04-09
feat: Add bytecast ops to builtins

### LINKED ISSUE #782: Byte casting between types
E.g. `int` to `float`

croyzor:
Resolves #782

### DIFF
--- a/guppylang/std/builtins.py
+++ b/guppylang/std/builtins.py
@@ -45,6 +45,7 @@ from guppylang.std._internal.compiler.prelude import (
     UnwrapOpCompiler,
 )
 from guppylang.std._internal.util import (
+    external_op,
     float_op,
     int_op,
     logic_op,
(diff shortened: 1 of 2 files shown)


qartik:
Thanks @croyzor, I think if we are resolving #782, then `int` should also be included. Further, we may also need a conversion between `int` and `nat`, especially when a negative integer is encoded in a `nat`.

Also, is it possible to use shorter names for these coercion functions? Here are some ideas:
```py
def nat_to_float(n: nat) -> float:
def float_to_nat(f: float) -> nat:

def nat2float(n: nat) -> float:
def float2nat(f: float) -> nat:
```

I am unsure if it's possible but the original suggestion from @mark-koch was to just provide a single `byte_cast` that could act based on the inferred types:
```
theta: float = byte_cast(nat_angle)
```
This will be ideal if possible.

croyzor:
My understanding is that this is a very niche use case for enabling RPC, so it's not a bad thing to have a verbose (descriptive) name. I'd hate for somebody to find these functions and use them thinking that they do the job of `Float.__nat__` and `Nat.__float__`

qartik (approved):

qartik:
Marking as approved as it's worth it if it's released sooner.

PR #920, opened 2025-04-17
feat: add wasm module

### LINKED ISSUE #755: Add support for wasm
See https://github.com/CQCL/tket2/pull/737 for proposed extension spec.

- Model stateful WASM context with affine type.
- WASM functions can take have form:
```python
E = int | bool | float
T = E | array[E, N]

def foo(*args: T) -> E: ...
def void(*args: T) -> None: ...
```

Require standard library `guppy.std.wasm` with functions:

```python
@guppy.struct
class WASMContext:
  """Affine type representing state of environment."""
  
def new_wasm_checked() -> Option[WASMContext]:
  """Returns a new WASM environment if available else None."""

def new_wasm() -> WASMContext:
"""Panicking version of `new_wasm_checked`""".
  new = new_wasm_checked()
  if new.is_some():
    return new.unwrap()
  panic("No more WASM environments available")
```

Example:
```python

# all declared wasm signatures are treated as belonging to single wasm module

@guppy.wasm
def add_one(x: int) -> int: ...

@guppy.wasm
def add_syndrome(syndrome: array[bool, 3]) -> None: ...

@guppy.wasm
def decode() -> int: ...



@guppy
def main() -> None:
wasm1 = new_wasm()
wasm2 = new_wasm()
y = wasm1.add_one(x) # all methods take WASMContext as inout
wasm2.add_syndrome(array(True, False, False))
decoded = wasm2.decode()
```

qartik:
Closes #755

Needs a lot more work, especially to add the `guppy.wasm` decorator.


qartik:
Closing in favor of #942

Issue #791, opened 2025-02-03
Stdlib: More range functions


croyzor:
It'd be nice to have the full spectrum of `range` functions available in python, e.g. `range(min, max)`.
This is tricky because in guppy we don't support function overloading, so range always has one argument.

If we implement [default arguments](https://github.com/CQCL/guppylang/discussions/790), we could consider `range` to have the signature: 
```python
range(max: int, min: int = 0, step: int = 1)
```
to do this without overloading.

----

It may be easier for guppy to reason about indices that come directly from a `range` iterator than the alternative, i.e.:
```python
for i in range(3, len(xs)):
    f(xs[i])
```
vs
```python
for i in range(len(xs)):
    f(xs[i + 3])
```
but maybe at this point it doesn't make a difference.

PR #925, opened 2025-04-25
feat(std): add array shuffling to rng module

ss2165:
https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle?utm_source=chatgpt.com#The_modern_algorithm

### DIFF
--- a/guppylang/std/qsystem/random.py
+++ b/guppylang/std/qsystem/random.py
@@ -12,10 +12,16 @@ from guppylang.std._internal.compiler.quantum import (
     RNGCONTEXT_T,
 )
 from guppylang.std._internal.util import external_op
-from guppylang.std.builtins import owned
+from guppylang.std.builtins import array, mem_swap, owned
 from guppylang.std.option import Option
 
 qsystem_random = GuppyModule("qsystem.random")
+
+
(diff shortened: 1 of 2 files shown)


guppylang/std/qsystem/random.py
```diff
@@ -80,3 +86,16 @@ def _random_float(self: "RNG") -> float: ...
...
     @no_type_check
     def _random_int_bounded(self: "RNG", bound: int) -> int: ...
+
+    @guppy(qsystem_random)
+    @no_type_check
+    def shuffle(self: "RNG", array: array[SHUFFLE_T, SHUFFLE_N]) -> None:
+        """Randomly shuffle the elements of a possibly linear array in place.
+        Uses the Fisher-Yates algorithm."""
+        for k in range(SHUFFLE_N):
+            i = SHUFFLE_N - 1 - k
```
qartik, line 96:
```suggestion
        for i in range(SHUFFLE_N - 1, 0, -1):
```
Would it be more transparent to write the iteration this way?

qartik (commented):
Generally, LGTM.

guppylang/std/qsystem/random.py
ss2165: It would but sadly range doesn't work that way in guppy 

See #791

qartik (approved):

PR #930, opened 2025-04-29
feat!: Bump to Hugr 0.20 and switch to value arrays for lowering

### LINKED ISSUE #927: Use new `value_array` extension

mark-koch:
Closes #927

BREAKING CHANGE: Generated HUGRs now refer to entrypoints and are always modules in envelopes.


guppylang/std/_internal/compiler/array.py
```diff
@@ -332,7 +370,9 @@ def _build_call_getitem(
 
     def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
         [array, idx] = args
-        if self.elem_ty.type_bound() == ht.TypeBound.Any:
+        [elem_ty_arg, _] = self.type_args
+        assert isinstance(elem_ty_arg, TypeArg)
+        if not elem_ty_arg.ty.copyable:
```
mark-koch, line 375: We need to relax this constraint to non-copyable since we have to avoid `get`s on nested arrays: Those only work for copyable element types

guppylang/compiler/cfg_compiler.py
```diff
@@ -130,12 +130,12 @@ def compile_bb(
             branch_port = choose_vars_for_tuple_sum(
                 unit_sum=branch_port,
                 output_vars=[
-                    [v for v in sort_vars(row) if not v.ty.linear]
+                    [v for v in sort_vars(row) if v.ty.droppable]
```
mark-koch, line 133: drive-by: We should generally output all non-droppable types here instead of just linear ones

guppylang/compiler/expr_compiler.py
```diff
@@ -501,6 +505,10 @@ def visit_ResultExpr(self, node: ResultExpr) -> Wire:
             )
             map_op = array_map(ht.Option(base_ty), size_arg, base_ty)
             value_wire = self.builder.add_op(map_op, value_wire, unwrap)
+            # Turn `value_array` into regular linear `array`
+            value_wire = self.builder.add_op(
+                array_convert_to_std_array(base_ty, size_arg), value_wire
```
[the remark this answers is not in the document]
mark-koch: Because the `result` op still takes a regular `array` (as it should!). The linearisation pass turns these convert ops into noops

PR #935, opened 2025-05-01
fix: Use correct copyable/droppable terminology for type arg errors

### LINKED ISSUE #867: Type argument checking still refers to "non-linear" instead of "copyable"

mark-koch:
Fixes #867


ss2165 (approved):
👍

PR #941, opened 2025-05-06
feat: Add some explicit copy diagnostic hints

### LINKED ISSUE #901: Diagnostic hint to copy values
E.g. when writing

```python
@guppy
def identity(xs: array[int, n]) -> array[int, n]:
    return xs
```

the compiler tells you

```
Error: Not owned (at <In [39]>:8:11)
  | 
6 | @guppy
7 | def identity(xs: array[int, n]) -> array[int, n]:
8 |     return xs
  |            ^^ Cannot return `xs` since `identity` doesn't own it
  | 
7 | def identity(xs: array[int, n]) -> array[int, n]:
  |              ----------------- Argument `xs` is only borrowed. Consider taking ownership:
  |                                `xs: array[int, n] @owned`
```

It would be nice if the compiler also suggested:

```
8 |     return xs
  |            ^^ Or consider copying this argument: `xs.copy()`
```

tatiana-s:
Closes #901 + hints for attempts at implicit copies


guppylang/checker/linearity_checker.py
```diff
@@ -603,6 +610,16 @@ def is_inout_var(place: Place) -> TypeGuard[Variable]:
...
 
 
+def has_explicit_copy(ty: Type) -> bool:
+    """Checks whether a type has an explicit copy function.
+
+    Currently, this is only the case for arrays with copyable elements."""
+    if not is_array_type(ty):
+        return False
+    elem_ty = get_element_type(ty)
+    return ty.droppable and elem_ty.copyable
```
mark-koch, line 620: Why is the `ty.droppable` restriction needed?

mark-koch (commented):
Very nice!

I think the `place.root` thing needs to be fixed. Apart from that only some questions and wording suggestions

guppylang/checker/linearity_checker.py
```diff
@@ -270,6 +275,8 @@ def visit_PlaceNode(
                     err.add_sub_diagnostic(
                         AlreadyUsedError.PrevUse(prev_use.node, prev_use.kind)
                     )
+                    if has_explicit_copy(place.ty):
+                        err.add_sub_diagnostic(AlreadyUsedError.MakeCopy(None))
```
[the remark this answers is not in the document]
tatiana-s: I assume you mean something like this which I tried, but in my opinion it looks a bit strange in terms of capitalisation and repeated span:
Screenshot 2025-05-08 at 11 35 20

guppylang/checker/errors/linearity.py
```diff
@@ -151,10 +156,38 @@ def rendered_span_label(self) -> str:
 
     @dataclass(frozen=True)
     class MakeOwned(Help):
-        span_label: ClassVar[str] = (
-            "Argument `{place.root.name}` is only borrowed. Consider taking ownership: "
-            "`{place.root.name}: {place.root.ty} @owned`"
-        )
+        place: Place
+        added_first: bool
```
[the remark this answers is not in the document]
tatiana-s: The idea was that it is a bit strange to have a hint that starts with "Or..." that you could in theory add by itself but yeah I guess it is unnecessary

guppylang/checker/linearity_checker.py
mark-koch: ping

mark-koch (approved):
Thanks, looks good 👍

Only wondering about the droppable check in `has_explicit_copy`?

guppylang/checker/linearity_checker.py
tatiana-s: I removed it now - it was accidentally left over from before I realised I just cared about the elements being copyable and the array being affine is automatically inferred from that

PR #942, opened 2025-05-07
feat: Declare WASM modules in guppy

croyzor:


guppylang/tys/builtin.py
```diff
@@ -101,6 +105,28 @@ def check_instantiate(
         return self.ty
 
 
+class WasmModuleTypeDef(OpaqueTypeDef):
```
mark-koch, line 106: Turn this into a `dataclass`, then you don't need to define `__init__`

mark-koch: Also, maybe we should move this into a `definitions/wasm.py` file, doesn't really feel like a builtin?

guppylang/tys/builtin.py
```diff
@@ -347,3 +379,17 @@ def get_iter_size(ty: Type) -> Const:
             return const
         case _:
             raise InternalGuppyError("Unexpected type args")
+
+
+class ConstStringArg(ConstArg):
```
mark-koch: This is a duplicate of the same thing in `constarg.py`

guppylang/tys/constarg.py
```diff
@@ -0,0 +1,19 @@
+from hugr import tys as ht
+
+from guppylang.tys.arg import ConstArg
+from guppylang.tys.builtin import is_string_type, string_type
+from guppylang.tys.const import ConstValue
+
+
+class ConstStringArg(ConstArg):
```
mark-koch: As far as I can tell, this is only used in `WasmModuleCallCompiler.compile_with_inouts`, i.e. not in any place that actually concerns the Guppy type system, so I'm not sure why this is needed?

guppylang/tys/constarg.py
```diff
@@ -0,0 +1,19 @@
...
+    def to_hugr(self) -> ht.TypeArg:
+        match self.const:
+            case ConstValue(value=v, ty=ty) if is_string_type(ty):
+                assert isinstance(v, str)
+                return ht.StringArg(v)
+            case _:
+                return super().to_hugr()
+
+
+def const_string_arg(s: str) -> ConstStringArg:
```
mark-koch: This seems to be unused

guppylang/checker/errors/wasm.py
```diff
@@ -0,0 +1,42 @@
...
+@dataclass(frozen=True)
+class WasmError(Error):
+    title: ClassVar[str] = "WASM signature error"
+
+
+@dataclass(frozen=True)
+class FirstArgNotModule(WasmError):
+    span_label: ClassVar[str] = (
+        "First argument to WASM function should be a reference to a WASM module."
+        " Found {ty} instead"
```
mark-koch:
```suggestion
        " Found `{ty}` instead"
```

guppylang/checker/errors/wasm.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    ty: Type
+
+
+@dataclass(frozen=True)
+class UnWasmableType(WasmError):
+    span_label: ClassVar[str] = (
+        "WASM function signature contained an unsupported type: {ty}"
```
mark-koch:
```suggestion
        "WASM function signature contained an unsupported type: `{ty}`"
```

guppylang/checker/errors/wasm.py
```diff
@@ -0,0 +1,42 @@
...
+    )
+    ty: Type
+
+
+@dataclass(frozen=True)
+class NonFunctionWasmType(WasmError):
+    span_label: ClassVar[str] = (
+        "WASM function didn't have a function type, instead found {ty}"
```
mark-koch:
```suggestion
        "WASM function didn't have a function type, instead found `{ty}`"
```

guppylang/decorator.py
```diff
@@ -420,6 +438,78 @@ def load_pytket(
...
+                filehash,
+            )
+
+            wasm_module_ty = wasm_module.check_instantiate([], None)
+
+            DEF_STORE.register_def(wasm_module, get_calling_frame())
+            for val in cls.__dict__.values():
+                if isinstance(val, GuppyDefinition):
+                    DEF_STORE.register_impl(wasm_module.id, val.wrapped.name, val.id)
+            # Add a __call__ to the class
```
mark-koch:
```suggestion
            # Add a constructor to the class
```

guppylang/definition/custom.py
```diff
@@ -15,6 +15,8 @@
 from guppylang.compiler.core import CompilerContext, DFContainer, GlobalConstId
 from guppylang.definition.common import ParsableDef
 from guppylang.definition.value import CallReturnWires, CompiledCallableDef
+
+# from guppylang.definition.wasm import WasmModule
```
mark-koch:
```suggestion
```

guppylang/definition/custom.py
```diff
@@ -179,7 +181,7 @@ class CustomFunctionDef(CompiledCallableDef):
         has_signature: Whether the function has a declared signature.
     """
 
-    defined_at: AstNode
+    defined_at: AstNode | None
```
mark-koch, line 182: Is this change still needed?

guppylang/std/_internal/checker.py
```diff
@@ -557,3 +568,71 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+                    pass
+                case FuncInput(ty=ty):
+                    raise GuppyError(FirstArgNotModule(loc, ty))
+            for inp in self.func.ty.inputs[1:]:
+                if not self.is_type_wasmable(inp.ty):
+                    raise GuppyError(UnWasmableType(loc, inp.ty))
+            if not self.is_type_wasmable(self.func.ty.output):
+                raise GuppyError(UnWasmableType(loc, self.func.ty.output))
+        else:
+            raise GuppyError(NonFunctionWasmType(loc, self.func.ty))
```
mark-koch: These last two cases don't have coverage, corresponding error tests would be good

guppylang/std/_internal/checker.py
```diff
@@ -557,3 +568,71 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+                return True
+            case NoneType():
+                return True
+            case TupleType(element_types=tys):
+                return all(self.is_type_wasmable(ty) for ty in tys)
+            case FunctionType() as f:
+                return self.is_type_wasmable(f.output) and all(
+                    self.is_type_wasmable(inp.ty) and inp.flags != InputFlags.Inout
+                    for inp in f.inputs
+                )
```
mark-koch: Surely we can't pass function pointers to wasm?!

guppylang/std/_internal/checker.py
```diff
@@ -557,3 +568,71 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+            raise GuppyTypeError(WasmTypeConversionError(self.node, ty))
+
+        # Use default implementation from the expression checker
+        args, subst, inst = check_call(self.func.ty, args, ty, self.node, self.ctx)
+
+        return GlobalCall(def_id=self.func.id, args=args, type_args=inst), subst
+
+    def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
+        if not self.type_sanitised:
+            self.sanitise_type()
```
mark-koch: Imo it's still not great that the signature is only checked when the function is called. It would be cleaner to define your own `ParseableDef` subsclass that performs these sanitise checks on `parse`.

guppylang/std/_internal/checker.py
```diff
@@ -557,3 +568,71 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+                    return all(
+                        self.is_type_wasmable(arg.ty)
+                        for arg in ty.args
+                        if isinstance(arg, TypeArg)
+                    )
+                return is_bool_type(ty)
+            case _:
+                return False
+
+    def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
```
mark-koch, line 564: You're skipping the `sanitise_type` call here

guppylang/std/_internal/checker.py
```diff
@@ -557,3 +568,71 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        if not self.is_type_wasmable(ty):
+            raise GuppyTypeError(WasmTypeConversionError(self.node, ty))
```
mark-koch: Why is this check needed? `ty` is the type we're checking against, if it's not wasmable then we'll just get a type mismatch error?

guppylang/std/_internal/compiler/wasm.py
```diff
@@ -0,0 +1,129 @@
...
+)
+from guppylang.tys.const import ConstValue
+from guppylang.tys.constarg import ConstStringArg
+from guppylang.tys.ty import (
+    FunctionType,
+)
+
+
+# Compiler for initialising WASM modules
+class WasmModuleInitCompiler(CustomInoutCallCompiler):
```
mark-koch: Can you turn this into a docstring? (Same for the othert classes below)

guppylang/tys/builtin.py
```diff
@@ -13,8 +14,11 @@
 from guppylang.error import GuppyError, InternalGuppyError
 from guppylang.experimental import check_lists_enabled
 from guppylang.std._internal.compiler.tket2_bool import OpaqueBool
+from guppylang.std._internal.compiler.tket2_exts import WASM_EXTENSION
 from guppylang.tys.arg import Argument, ConstArg, TypeArg
 from guppylang.tys.const import Const, ConstValue
+
+# from guppylang.tys.constarg import const_string_arg
```
mark-koch:
```suggestion
```

tests/integration/test_wasm.py
mark-koch, line 1: Can you add a test for calling wasm methods and spawning contexts in comptime functions

uv.lock
mark-koch, line 1: Not sure where this diff is coming from, you didn't change `pyproject.toml`?

tests/error/misc_errors/wasm_func_without_module.err
```diff
@@ -0,0 +1,9 @@
+Error: WASM signature error (at $FILE:7:4)
+  | 
+5 | class Foo:
+6 |     @guppy.wasm
+7 |     def foo(x: int) -> None: ...
+  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^ First argument to WASM function should be a reference to a
+  |                                  WASM module. Found int instead
```
mark-koch: Can you change the error location to point to the first argument?

guppylang/tys/builtin.py
croyzor: The dataclass magic doesn't work here because it inherits from `OpaqueTypeDef` which has a default argument

guppylang/tys/builtin.py
croyzor: Oops!

guppylang/tys/constarg.py
croyzor: Yes it's just for passing into the `wasm.lookup` op - I put it in a general place in case it was useful in the future 🤷

guppylang/tys/builtin.py
mark-koch: Oh I see, one workaround is setting your new fields to keyword-only:

```python
@dataclass(frozen=True)
class WasmModuleTypeDef(OpaqueTypeDef):
    wasm_file: str = field(kw_only=True)
    wasm_hash: int = field(kw_only=True)
```

or remove the default from `OpaqueTypeDef.bound`,

or leave it as is :D - your call!

guppylang/tys/constarg.py
mark-koch: But it shouldn't be a subclass of `ConstArg`, just an instace with the `ty` field set to string. If it were needed for this PR it would be ok as a hack, but if it's not needed I'd prefer to get rid of it and do it properly in the future

tests/error/misc_errors/wasm_func_without_module.err
croyzor: I'm not really sure how - the `FunctionType` that comes out of parsing `CustomFunctionDef`s doesn't have location info...
(We'd want to highlight the signature of the function as written, rather than the call site, which would be easier)

guppylang/definition/custom.py
croyzor: This is for the `__call__` and `discard` methods of WASM modules, which don't really have a place. I think we could get rid of it if we actually parsed WASM module definitions like we do structs, but I think that's a refactor for down the line...
If you have any other ideas I'm very happy to hear them!

guppylang/std/_internal/checker.py
croyzor: Oh the latter was meant to be covered by the `tests/error/misc/wasm_non_function` test case, but a syntax error gets there first! The former I can definitely do

croyzor: oh i didn't commit that file because the test suite doesn't like the parse error, so I'm not quite sure how to trigger this case

guppylang/definition/wasm.py
```diff
@@ -0,0 +1,63 @@
...
+                return all(self.is_type_wasmable(ty) for ty in tys)
+
+        return False
+
+    def parse(self, globals: "Globals", sources: SourceMap) -> "CustomFunctionDef":
+        parsed = super().parse(globals, sources)
+        if isinstance(parsed.ty, FunctionType):
+            self.sanitise_type(parsed.defined_at, parsed.ty)
+        else:
+            raise GuppyError(NonFunctionWasmType(parsed.defined_at, self.func.ty))
```
mark-koch: Isn't `CustomFunctionDef.ty` already annotated as `FunctionType`? The `isinstance` check should be unnecessary

guppylang/std/_internal/compiler/wasm.py
```diff
@@ -0,0 +1,136 @@
...
+
+        func_ty = WASM_EXTENSION.get_type("func").instantiate(
+            [inputs_row_arg, output_row_arg]
+        )
+        future_ty = FUTURES_EXTENSION.get_type("Future").instantiate(
+            [ht.Tuple(*wasm_sig.output).type_arg()]
+        )
+
+        # Get the WASM module information from the type
+        assert isinstance(self.node, GlobalCall | LocalCall)
```
mark-koch: I don't think this assert is needed anymore?

tests/error/misc_errors/wasm_func_without_module.err
mark-koch: Ah I see, yeah this hard without writing your own signature parsing - happy to leave it as is then

uv.lock
mark-koch, line 1: There's still a big diff here?

guppylang/checker/errors/wasm.py
mark-koch: Backticks still missing here

PR #949, opened 2025-05-13
fix: Fix `result` / `panic` /  `exit` calls in comptime functions

### LINKED ISSUE #919: comptime strings not allowed as `result` keys
```python
from guppylang import guppy
from guppylang.std.builtins import result, comptime

EXAMPLE_RESULTS = [
    ("boolean", False),
    ("int", 123),
]


@guppy.comptime
def main() -> None:
    for key, value in EXAMPLE_RESULTS:
        result(key, value)
main.compile()
```

fails with

```sh
GuppyComptimeError: Expected a string literal: Expected a string literal
```

@mark-koch believes this is to do with the checker for `result` not detecting that a comptime expression can be treated as a static value (?)

tatiana-s:
Closes #919

There are two issues with these functions:

1) The checker not knowing it can treat comptime arguments as static values - this fix is sort of a halfway between a general solution (variables passed to Guppy function calls inside comptime functions are now `ComptimeVariables` which just inherits from `Variable` with an additional field to pass argument values directly) and hack (it is only used for this particular case so isn't thought out completely)

2) As they are built-ins defined with a custom checker and no type annotations (as the inputs could have different types), `func.ty` is empty so you can't iterate through the function input types to update inouts - since as far as I can see any function where this is the case also won't need to update inouts, the current fix is to just not do that part of the input types are empty, but this seems a bit dodgy. Maybe there could be a way of allowing or forcing custom checker to provide a function signature, but not sure if this is more of a follow up and this fix is fine since it does make these functions work for now.


aborgna-q (approved):

guppylang/tracing/function.py
```diff
@@ -152,7 +152,10 @@ def trace_call(func: CompiledCallableDef, *args: Any) -> Any:
     ]
 
     # Create dummy variables and bind the objects to them
-    arg_vars = [Variable(next(tmp_vars), obj._ty, None) for obj in args_objs]
+    arg_vars: list[Variable] = [
+        ComptimeVariable(next(tmp_vars), obj._ty, None, static_value=arg)
```
mark-koch, line 156: Why make `static_value` optional if you always set it?

guppylang/tracing/function.py
```diff
@@ -152,7 +152,10 @@ def trace_call(func: CompiledCallableDef, *args: Any) -> Any:
...
+        for (obj, arg) in zip(args_objs, args, strict=False)
```
mark-koch, line 157: Why `strict=False`?

guppylang/std/_internal/checker.py
```diff
@@ -337,9 +337,18 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
         check_num_args(2, len(args), self.node)
         [tag, value] = args
         tag, _ = ExprChecker(self.ctx).check(tag, string_type())
+        tag_value = ""
         if not isinstance(tag, ast.Constant) or not isinstance(tag.value, str):
-            raise GuppyTypeError(ExpectedError(tag, "a string literal"))
-        if len(tag.value.encode("utf-8")) > TAG_MAX_LEN:
+            # Besides a constant it could be a comptime variable,
+            # which we can assume to be static
+            if hasattr(tag, "place") and isinstance(tag.place, ComptimeVariable):
```
mark-koch, line 344: Why `hasattr(tag, "place")` instead of `isinstance(tag, PlaceNode)`?

mark-koch: You could even refactor to a match

```python
match tag:
    case ast.Constant(value=str(v)):
        tag_value = v
    case PlaceNode(place=ComptimeVariable(static_value=str(v))):
        tag_value = v
```

guppylang/checker/core.py
```diff
@@ -109,6 +109,14 @@ def replace_defined_at(self, node: AstNode | None) -> "Variable":
         return replace(self, defined_at=node)
 
 
+@dataclass(frozen=True, kw_only=True)
+class ComptimeVariable(Variable):
```
mark-koch, line 113: Following https://github.com/CQCL/guppylang/issues/898 we can probably merge this functionality into the regular `Variable`, but this seems ok for a short-term fix 👍

guppylang/tracing/function.py
tatiana-s: I think I didn't add it at all and the check automatically added, changed it to True

guppylang/tracing/function.py
```diff
@@ -169,10 +172,13 @@ def trace_call(func: CompiledCallableDef, *args: Any) -> Any:
     ret_wire = ExprCompiler(state.ctx).compile(call_node, state.dfg)
 
     # Update inouts
-    for inp, arg, var in zip(func.ty.inputs, args, arg_vars, strict=True):
-        if InputFlags.Inout in inp.flags:
-            inout_wire = state.dfg[var]
-            update_packed_value(arg, GuppyObject(inp.ty, inout_wire), state.dfg.builder)
+    if len(func.ty.inputs) != 0:
```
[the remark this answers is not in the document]
tatiana-s: See point 2 in the description

PR #950, opened 2025-05-13
fix: Fix comptime angle arithmetic

### LINKED ISSUE #946: bug(comptime): const float multiplying `pi` fails
```python
from guppylang.std.angles import pi


@guppy
def foo() -> int:
    _ = 1.5 * pi
    return 1


@guppy.comptime
def main() -> None:
    x = 2 * foo()
    a = 1.5 * pi


guppy.compile_module()

```

```
    a = 1.5 * pi
        ~~~~^~~~
TypeError: unsupported operand type(s) for *: 'float' and 'GuppyDefinition'
```

The example shows that:
1. it works for integers returned from another function
2. float * angle works in normal function

So I would expect the comptime multiplication to work

mark-koch:
Fixes #946

The problem was that `pi` is a `@guppy.constant` (i.e. a `GuppyDefinition` at comptime) which didn't implement the corresponding `__add__` etc dunder methods.

I fixed this by making `GuppyDefinition` inherit from `DunderMixin` as well as refactoring the `binary_operation` and `unray_operation` wrappers to first load definitions as values (via `guppy_object_from_py`) before calling the methods.

### DIFF
--- a/guppylang/tracing/object.py
+++ b/guppylang/tracing/object.py
@@ -97,7 +97,9 @@ def binary_operation(f: BinaryDunderMethod) -> BinaryDunderMethod:
         with suppress(Exception):
             return f(self, other)
 
-        # If that failed, try the reverse method on `other`
+        # If that failed, try the reverse method on `other`.
+        # NB: We know that `f.__name__` is in one of the tables since we make sure to
+        # only put this decorator on the correct dunder methods below
         if f.__name__ in binary_table:
             reverse_method, display_name = binary_table[f.__name__]
             left_ty, right_ty = self._ty, other._ty
--- a/guppylang/tracing/object.py
+++ b/guppylang/tracing/object.py
@@ -115,168 +115,174 @@ def binary_operation(f: BinaryDunderMethod) -> BinaryDunderMethod:
 
 
 class DunderMixin:
-    """Mixin class to allow `GuppyObject`s to be used in arithmetic expressions etc.
-    via providing the corresponding dunder methods delegating to the objects impls.
+    """Mixin class to allow `GuppyObject`s and `GuppyDefinition`s to be used in
+    arithmetic expressions etc. via providing the corresponding dunder methods
+    delegating to the objects impls.
     """
 
-    def __getattr__(self, item: Any) -> Any:
-        return super().__getattribute__(item)
+    def _get_method(self, name: str) -> Any:
+        from guppylang.tracing.state import get_tracing_state
+        from guppylang.tracing.unpacking import guppy_object_from_py
+
+        state = get_tracing_state()
+        self = guppy_object_from_py(self, state.dfg.builder, state.node)
+        return self.__getattr__(name)
 
     def __abs__(self) -> Any:
-        return self.__getattr__("__abs__")()
+        return self._get_method("__abs__")()
 
     @binary_operation
     def __add__(self, other: Any) -> Any:
-        return self.__getattr__("__add__")(other)
+        return self._get_method("__add__")(other)
 
     @binary_operation
     def __and__(self, other: Any) -> Any:
-        return self.__getattr__("__and__")(other)
+        return self._get_method("__and__")(other)
 
     def __bool__(self: Any) -> Any:
-        return self.__getattr__("__bool__")()
+        return self._get_method("__bool__")()
 
     def __ceil__(self: Any) -> Any:
-        return self.__getattr__("__ceil__")()
+        return self._get_method("__ceil__")()
 
     def __divmod__(self, other: Any) -> Any:
-        return self.__getattr__("__divmod__")(other)
+        return self._get_method("__divmod__")(other)
 
     @binary_operation
     def __eq__(self, other: object) -> Any:
-        return self.__getattr__("__eq__")(other)
+        return self._get_method("__eq__")(other)
 
     def __float__(self) -> Any:
-        return self.__getattr__("__float__")()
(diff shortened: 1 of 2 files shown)


guppylang/tracing/object.py
```diff
@@ -85,38 +86,41 @@ def binary_operation(f: BinaryDunderMethod) -> BinaryDunderMethod:
...
-            else:
-                reverse_method, display_name = reverse_binary_table[f.__name__]
-                left_ty, right_ty = obj._ty, self._ty
-            return obj.__getattr__(reverse_method)(self)
+        # If that failed, try the reverse method on `other`
+        if f.__name__ in binary_table:
+            reverse_method, display_name = binary_table[f.__name__]
+            left_ty, right_ty = self._ty, other._ty
+        else:
+            reverse_method, display_name = reverse_binary_table[f.__name__]
```
acl-cqc, line 105: Does `reverse_binary_table` contain every name that might occur here, or should you check (or use `get`) here?

guppylang/tracing/object.py
```diff
@@ -85,38 +86,41 @@ def binary_operation(f: BinaryDunderMethod) -> BinaryDunderMethod:
...
-class DunderMixin(ABC):
+class DunderMixin:
     """Mixin class to allow `GuppyObject`s to be used in arithmetic expressions etc.
     via providing the corresponding dunder methods delegating to the objects impls.
     """
 
-    @abstractmethod
-    def __getattr__(self, item: Any) -> Any: ...
+    def __getattr__(self, item: Any) -> Any:
+        return super().__getattribute__(item)
```
acl-cqc, line 123: This looks a little odd - what's `getattribute` as opposed to `getattr`, and `super()` in a class with no superclasses? My understanding is that `getattribute` falls back to `getattr` if it doesn't find e.g. a field, so a little doc explaining what this is doing (and why you need a `getattr` at all) would be good

acl-cqc: e.g. Is the idea that `super()` is resolved dynamically to some *other* superclass of a subclass inheriting this?

guppylang/tracing/object.py
```diff
@@ -85,38 +86,41 @@ def binary_operation(f: BinaryDunderMethod) -> BinaryDunderMethod:
...
-        assert isinstance(self, GuppyObject)
         raise GuppyTypeError(
             BinaryOperatorNotDefinedError(state.node, left_ty, right_ty, display_name)
         )
 
     return wrapped
 
 
-class DunderMixin(ABC):
+class DunderMixin:
```
acl-cqc, line 117: Noting that DunderMixin's are converted into GuppyObjects by `guppy_object_from_py`, is the doc still accurate? Seems it provides dunder methods by delegating to its own superclass?

acl-cqc (commented):
converting via `guppy_object_from_py` seems plausible :-) but I haven't grokked how the rest works. Can you offer any pointers (as per comments!)?

guppylang/tracing/object.py
mark-koch: It should contain every name since the `@binary_operation` decorator is only put on correct dunder methods. Could add an additional assert though if you prefer?

guppylang/tracing/object.py
mark-koch: Yeah this looks bogus since the methods are invoked with `self` converted to a `GuppyObject`.

I'll refactor a bit to get rid of this `__getattr__` stuff 👍

guppylang/tracing/object.py
```diff
@@ -115,168 +115,174 @@ def wrapped(self: "DunderMixin", other: Any) -> Any:
...
     """
 
-    def __getattr__(self, item: Any) -> Any:
-        return super().__getattribute__(item)
+    def _get_method(self, name: str) -> Any:
+        from guppylang.tracing.state import get_tracing_state
+        from guppylang.tracing.unpacking import guppy_object_from_py
+
+        state = get_tracing_state()
+        self = guppy_object_from_py(self, state.dfg.builder, state.node)
```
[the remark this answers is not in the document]
acl-cqc: This looks very reasonable now! (Did I miss that we'd reassigned `self` before? Sorry if so, whatever.)

guppylang/tracing/object.py
```diff
@@ -85,194 +86,203 @@ def binary_operation(f: BinaryDunderMethod) -> BinaryDunderMethod:
...
 
-    @abstractmethod
-    def __getattr__(self, item: Any) -> Any: ...
+    def _get_method(self, name: str) -> Any:
+        from guppylang.tracing.state import get_tracing_state
+        from guppylang.tracing.unpacking import guppy_object_from_py
+
+        state = get_tracing_state()
+        self = guppy_object_from_py(self, state.dfg.builder, state.node)
+        return self.__getattr__(name)
```
acl-cqc, line 129: Should this in fact be `self._get_method(name)` ?

acl-cqc (approved):
Looks good to me, thanks @mark-koch ! :)

If I wanted to uber-nit, I could say that your tests only check each operator one way around (e.g. on const ints, we check `1.5 - x` and `x * 2` but not `x - 1.5` and `2 * x`). But that's fine :)

guppylang/tracing/object.py
mark-koch: No I think this would infinitely recurse? `GuppyObject.__getattr__` gives us the the method implementation (which is also called if users write `obj.method()` in a comptime section)

PR #952, opened 2025-05-13
fix: Capture entire Python scope when defining comptime functions

### LINKED ISSUE #945: `AssertionError` when trying to compile `comptime` functions
I'd expect to be able to compile this comptime function with `function.compile()` rather than having to rely on `guppy.compile_module()`

```python
from hugr.package import FuncDefnPointer
from guppylang.std.quantum import cx

def get_ladder_func(n: int) -> FuncDefnPointer:
    @guppy.comptime
    def ladder(qs: array[qubit, n]) -> None:
        for q1, q2 in zip(qs[1:], qs[:-1]):
            cx(q1, q2)
    return ladder.compile()

get_ladder_func(3)
```

Instead I get an `AssertionError` 

```
   564 def compile(self) -> Any:
    565     from guppylang.decorator import guppy
--> 567     assert isinstance(
    568         self.wrapped, RawFunctionDef | RawLoadPytketDef | RawPytketDef
    569     )
    570     return guppy.compile_function(self.wrapped)

AssertionError: 
```

mark-koch:
Fixes #945

### DIFF
--- a/tests/integration/tracing/test_basic.py
+++ b/tests/integration/tracing/test_basic.py
@@ -2,6 +2,7 @@ from collections.abc import Callable
 
 from guppylang.decorator import guppy
 from guppylang.module import GuppyModule
+from guppylang.std.builtins import array, comptime
 
 from hugr import ops
 from hugr.std.int import IntVal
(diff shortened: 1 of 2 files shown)


tests/integration/tracing/test_basic.py
```diff
@@ -82,3 +83,19 @@ def test() -> Callable[[int], int]:
...
+    def make(n: int):
+        @guppy.comptime(module)
+        def foo() -> int:
+            return n
+
+        @guppy.comptime(module)
+        def bar(xs: array[int, comptime(n)]) -> None:
+            pass
+
+    make(42)
```
qartik, line 99: It was not obvious to me what the test is trying to achieve. Perhaps a comment here will be helpful.

From the parent issue, I do see that 42 should now be available to guppy as the size of the array `xs`.

qartik (approved):

tests/integration/tracing/test_basic.py
mark-koch: Exactlty, it's checking that the variable `n` is in scope when evaluating the `comptime` expression. This was broken before if `n` is only defined inside a function scope instead of globally. Will add a comment

Issue #959, opened 2025-05-15
Library of high-level RNG functions


cqc-alec:
Create a library of high-level RNG functions in guppy, making use of the platform's RNG (which only provides methods for random 32-bit integers and random floats in the range `[0, 1)`).

There is already a "shuffle" function. A general discrete-distribution sampler has been requested. Need to determine what else is desired.

PR #961, opened 2025-05-15
feat: Implement sampling from a generic discrete distribution.

cqc-alec:
Towards #959 .

### DIFF
--- a/guppylang/std/qsystem/random.py
+++ b/guppylang/std/qsystem/random.py
@@ -12,7 +12,7 @@ from guppylang.std._internal.compiler.quantum import (
     RNGCONTEXT_T,
 )
 from guppylang.std._internal.util import external_op
-from guppylang.std.builtins import array, mem_swap, owned
+from guppylang.std.builtins import array, mem_swap, owned, panic
 from guppylang.std.option import Option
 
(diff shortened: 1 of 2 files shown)


guppylang/std/qsystem/random.py
```diff
@@ -99,3 +101,27 @@ def shuffle(self: "RNG", array: array[SHUFFLE_T, SHUFFLE_N]) -> None:
             # https://github.com/CQCL/guppylang/issues/924
             if i != j:
                 mem_swap(array[i], array[j])
+
+    @guppy(qsystem_random)
+    @no_type_check
+    def discrete_rv(self: "RNG", weights: array[float, DISCRETE_N]) -> int:
```
cqc-alec, line 107: Note, this is a "one-shot" version of the method. If the user wishes to sample multiple times from the same distribution, it would be slightly more efficient to provide a class that they can initialize with the array and then call a "sample" method on. Then we'd only have to sanity-check the weights once, and we could cache the partial sums. On reflection, since there is little downside even if they only want one sample, I think that would be preferable and we can dispense with the one-shot version (assuming all this is possible in guppy).

PR #971, opened 2025-05-20
fix: Specify bash shell in justfile.

cqc-alec:
Otherwise, it uses `sh` which doesn't recognize the `[[` ... `]]` syntax:
```
[[ -n "${JUST_INHIBIT_GIT_HOOKS:-}" ]] || uv run pre-commit install -t pre-commit
sh: 1: [[: not found
```

### DIFF
--- a/justfile
+++ b/justfile
@@ -1,3 +1,5 @@
+set shell := ["bash", "-uc"]
+
 # List the available commands
 help:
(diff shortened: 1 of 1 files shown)


aborgna-q (approved):
It seems that `just` uses the current shell by default.
Standardizing on `bash` should be harmless.

PR #982, opened 2025-05-27
fix: Swap back copyable elements in `array.__getitem__`

### LINKED ISSUE #980: `array.__getitem__ ` on copyable element should immediately swap back the element
Following #900, we stopped emitting `array.get` ops for classical `array.__getitem__` calls and switched to an `array.swap` with `None`. However, we never write something back into the array! This leads to panics, for example for the program

```python
@guppy
def main() -> int:
    xs = array(1)
    return xs[0] + xs[0]
```

After getting the copyable element out, we should immediately write something back!

tatiana-s:
Closes #980

### DIFF
--- a/tests/integration/test_array.py
+++ b/tests/integration/test_array.py
@@ -726,7 +726,7 @@ def test_multiple_functions(validate, run_int_fn):
 
 
 # Verifies that the same index in a classical array can be accessed twice.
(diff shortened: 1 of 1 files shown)


mark-koch (approved):
Thanks, looks good 👍

PR #983, opened 2025-05-28
feat!: Remove explicit Guppy modules

### LINKED ISSUE #956: Replace `GuppyModule` with global collection of definitions

### LINKED ISSUE #957: Track dependencies of definitions and look them up in corresponding stack frame

mark-koch:
Closes #956 and closes #957. See #936 for more context.

Most of the diff is coming from changing all the tests to no longer use modules. The significant changes are the following:

* Instead of tracking definitions by module, we now have a global `DEF_STORE` where every `@guppy` decorated definition is registered.
* For each defininition, this store also remembers the stack frame in which it was defined. When encountering a name while checking a function, we look it up in this stack frame and see if it corresponds to another Guppy definition. The old "name -> definition" mapping is gone, it's all based on what's in scope in the defining frame. See `guppylang.checker.core.Globals` for the new lookup mechanism.
* This also means that parsing, checking, and compilation of definitions is now on demand: We begin by checking the user-proveded entry point (via a new `guppy.compile` method) and only check other definitions if they are used by the entry point.
* This is driven via a new `CompilationEngine` class (defined in `guppylang/engine.py`). It maintains a worklist of which definitions still need to be checked/compiled and replaces the old `GuppyModule.check` and `GuppyModule.compile` logic.

BREAKING CHANGE: Explicit `GuppyModule` declarations are no longer possible. Instead, use the regular `@guppy` decorator everywhere without passing an explicit module. Compilation is now triggered via the `guppy.compile` function, passing the to-be-compiled function as an argument.


guppylang/definition/pytket_circuits.py
```diff
@@ -356,11 +355,14 @@ def _signature_from_circuit(
             try:
                 import tket2  # type: ignore[import-untyped, import-not-found, unused-ignore]  # noqa: F401
 
-                qubit = cast(TypeDef, globals["qubit"]).check_instantiate([], globals)
+                from guppylang.std.quantum import qubit
+
+                assert isinstance(qubit, GuppyDefinition)
+                qubit_ty = cast(TypeDef, qubit.wrapped).check_instantiate([])
```
mark-koch, line 361: Imo this is also way better: Before we tried to look up the qubit type by name in the globals which would break if the user had shadowed the name, or the name was not imported.

Now, we can import the type ourselves from the standard library within our own checking code!

guppylang/definition/struct.py
```diff
@@ -90,16 +91,6 @@ class RawStructDef(TypeDef, ParsableDef):
     """A raw struct type definition that has not been parsed yet."""
 
     python_class: type
-    python_scope: PyScope = field(repr=False)
-
-    def __getitem__(self, item: Any) -> "RawStructDef":
```
mark-koch, line 95: drive-by: This `__getitem__` should have been removed before when we added the comptime feature...

PR #985, opened 2025-05-28
refactor: move unsupported builtins into a new file

CalMacCQ:
The `builtins.py` file contains a bunch of Python std lib functions which are unsupported. My understanding is that having these unsupported functions here improves error messages when users try and make use of unsupported std lib functions from python.

I've refactoring these unsupported functions into their own file so that its easier to excude them from the sphinx docs. This avoids confusing users.

related to https://github.com/CQCL/guppylang/issues/732


guppylang/std/unsupported.py
```diff
@@ -0,0 +1,231 @@
+# Python builtins that are not supported yet:
+from guppylang import guppy
+from guppylang.std._internal.checker import UnsupportedChecker
+
+
+@guppy.custom(checker=UnsupportedChecker(), higher_order_value=False)
```
mark-koch, line 6: These are now placed into their own Guppy module but they should still be part of the builtins Guppy module (so they are always implicitly imported).

Have look at top of `std/options.py` for how to get a handle to the builtins guppy module

guppylang/std/builtins.py
```diff
@@ -1004,58 +825,6 @@ def range(stop: int) -> Range:
...
-
-
-@guppy.custom(checker=UnsupportedChecker(), higher_order_value=False)
-def zip(x): ...
-
-
-@guppy.custom(checker=UnsupportedChecker(), higher_order_value=False)
-def __import__(x): ...
-
-
```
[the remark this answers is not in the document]
CalMacCQ: Sorry I'm not completely sure what you mean by this.

Interestingly the `__import__` dunder isn't shown in the API docs to begin with so I could have probably left this in the same file.

CalMacCQ:
I've tested the sphinx build with the latest commit and the unsupported std members are being excluded as I intended even with the `import *` :)

CalMacCQ:
Not quite sure whats gone wrong with the C.I. will take a look shortly

guppylang/std/unsupported.py
CalMacCQ: Ah thanks, I've added that now. There still seems to be some sort of issue with C.I.

CalMacCQ:
> > Not quite sure whats gone wrong with the C.I. will take a look shortly. Some sort of recursion issue?
> 
> Try adding this to the top of the file
> 
> ```python
> # mypy: disable-error-code="empty-body, no-untyped-def"
> ```

Thanks, this resolved the mypy complaints. 

The recursion issue remains though. If I comment out the `import *` line locally this error doesn't show up.

guppylang/std/unsupported.py
```diff
@@ -0,0 +1,354 @@
+# mypy: disable-error-code="empty-body, no-untyped-def"
+"""Python builtins that are not supported yet"""
+
+from pathlib import Path
+
+from guppylang.decorator import ModuleIdentifier, guppy
+from guppylang.std import builtins
+from guppylang.std._internal.checker import UnsupportedChecker
+
+builtins_module = guppy.get_module(
```
mark-koch, line 10: So can you try making this private: `_builtins_module`?

CalMacCQ: Seems like I still get the same error as I did locally :(

CalMacCQ:
Managed to get the tests to pass by importing `builtins` again after the `import *`

mark-koch (approved):

PR #999, opened 2025-06-02
refactor: Split up builtins module into multiple files

### LINKED ISSUE #732: Break builtins.py in to multiple files

mark-koch:
Closes #732


guppylang/std/mem.py
```diff
@@ -1,14 +1,31 @@
...
 
 T = guppy.type_var("T", copyable=False, droppable=False)
 Out = guppy.type_var("Out", copyable=False, droppable=False)
 
 
+@guppy.custom(checker=BarrierChecker(), higher_order_value=False)
+@no_type_check
+def barrier(*args) -> None:
+    """Barrier to guarantee that all operations before the barrier are completed before
+    operations after the barrier are started."""
```
mark-koch: Not sure where we should put `barrier`?

guppylang/std/reflection.py
```diff
@@ -0,0 +1,10 @@
+"""Utilities for compile-time reflection."""
+
+# mypy: disable-error-code="empty-body, misc, override, valid-type, no-untyped-def"
+
+from guppylang.decorator import guppy
+from guppylang.std._internal.checker import CallableChecker
+
+
+@guppy.custom(checker=CallableChecker(), higher_order_value=False)
+def callable(x): ...
```
mark-koch, line 10: We could also consider to just turn this into unsupported, this function is pretty useless...

guppylang/std/mem.py
doug-q: yeah it's not clear. Here is fine for now I guess

doug-q: mem is not great. Maybe platform?

guppylang/std/reflection.py
doug-q: I don't understand well enough to form an opinion

mark-koch: It checks if the passed value is callable which is useful in Python, but very unnecessary in Guppy since wr already know this from the type system.

Still, I think I'll leave it for now to keep this non-breaking

PR #1000, opened 2025-06-02
feat: Function overloading via static dispatch

### LINKED ISSUE #995: Function overloading
See #466

mark-koch:
Closes #995

Allows to collect multiple function definitions into one overloaded function. Consider the following example:

```python
@guppy.declare
def variant1(x: int, y: int) -> int: ...

@guppy.declare
def variant2(x: float) -> int: ...

@guppy.overload(variant1, variant2)
def combined(): ...
```

Now, `combined` may be called with either one `float` or two `int` arguments, delegating to the implementation with the matching signature:

```python
combined(4.2)  # Calls `variant1`
combined(42, 43)  # Calls `variant2`
```

The compiler will pick the *first* implementation with matching signature and ignore all following ones, even if they would also match. This is in line with the typing semantics of the Python `@overload` decorator: https://mypy.readthedocs.io/en/stable/more_types.html#type-checking-calls-to-overloads

For example, if we added a third variant

```python
@guppy.declare
def variant3(x: int) -> int: ...

@guppy.overload(variant1, variant2, variant3)
def combined_new(): ...
```

then a call `combined_new(42)` will still select the `variant1` implementation since `42` is a valid argument for `variant1` and `variant1` comes before `variant3` in the `@guppy.overload` annotation.

### DIFF
--- a/guppylang/definition/overloaded.py
+++ b/guppylang/definition/overloaded.py
@@ -13,11 +13,12 @@ from guppylang.definition.common import (
     DefId,
 )
 from guppylang.definition.value import CallableDef, CallReturnWires, CompiledCallableDef
-from guppylang.diagnostic import Error
+from guppylang.diagnostic import Error, Note
 from guppylang.error import GuppyError, InternalGuppyError
 from guppylang.span import Span, to_span
+from guppylang.tys.printing import signature_to_str
 from guppylang.tys.subst import Inst, Subst
-from guppylang.tys.ty import Type
+from guppylang.tys.ty import FunctionType, Type
 
 
 @dataclass(frozen=True)
@@ -44,6 +45,18 @@ class OverloadNoMatchError(Error):
 
 
 @dataclass(frozen=True)
+class AvailableOverloadsHint(Note):
+    func_name: str
+    variants: list[FunctionType]
+
+    @property
+    def rendered_message(self) -> str:
+        return "Available overloads are:\n" + "\n".join(
+            f"  {signature_to_str(self.func_name, ty)}" for ty in self.variants
+        )
+
+
+@dataclass(frozen=True)
 class OverloadHigherOrderError(Error):
     title: ClassVar[str] = "Higher-order overloaded function"
     span_label: ClassVar[str] = (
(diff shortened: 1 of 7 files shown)


guppylang/tracing/function.py
```diff
@@ -144,7 +144,6 @@ def trace_call(func: CompiledCallableDef, *args: Any) -> Any:
     handles inout arguments.
     """
     state = get_tracing_state()
-    assert func.defined_at is not None
```
mark-koch, line 147: Overloaded functions don't have an AST definition, so we can no longer use this

guppylang/tracing/function.py
```diff
@@ -162,10 +161,10 @@ def trace_call(func: CompiledCallableDef, *args: Any) -> Any:
 
     # Check call
     arg_exprs: list[ast.expr] = [
-        with_loc(func.defined_at, with_type(var.ty, PlaceNode(var))) for var in arg_vars
+        with_loc(state.node, with_type(var.ty, PlaceNode(var))) for var in arg_vars
```
mark-koch, line 164: Instead, switch to the node we have stored in the tracing state. It doesn't really matter since we're not using the span-renderer for comptime errors anyways

guppylang/decorator.py
```diff
@@ -231,6 +241,72 @@ def declare(self, f: F) -> F:
...
+        def combined_new(): ...
+        ```
+
+        then a call `combined_new(42)` will still select the `variant1` implementation
+        `42` is a valid argument for `variant1` and `variant1` comes before `variant3`
+        in the `@guppy.overload` annotation.
+        """
+        funcs = list(funcs)
+        if len(funcs) < 2:
+            raise ValueError("Overload requires at least two functions")
```
aborgna-q, line 286: Any reason to not allow a single function?

Shouldn't be needed normally, but there's no logical reason not to.

guppylang/definition/overloaded.py
```diff
@@ -0,0 +1,127 @@
...
+from guppylang.definition.value import CallableDef, CallReturnWires, CompiledCallableDef
+from guppylang.diagnostic import Error
+from guppylang.error import GuppyError, InternalGuppyError
+from guppylang.span import Span, to_span
+from guppylang.tys.subst import Inst, Subst
+from guppylang.tys.ty import Type
+
+
+@dataclass(frozen=True)
+class OverloadNoMatchError(Error):
```
aborgna-q, line 25: Could we list the available options as hints?

aborgna-q (approved):

guppylang/decorator.py
mark-koch: It catches people trying to use it like the Python `@overload` decorator (i.e. not passing in functions):

```python
@guppy.overload
def foo(): ...
```

PR #1005, opened 2025-06-06
Add optional signature argument to RawCustomFunctionDef

### LINKED ISSUE #1003: Allow `RawCustomFunctionDef` to take an optional signature instead of parsing it from the function AST

hsemenenko:
Fixes #1003

### DIFF
--- a/guppylang/definition/custom.py
+++ b/guppylang/definition/custom.py
@@ -88,6 +88,7 @@ class RawCustomFunctionDef(ParsableDef):
         call_checker: The custom call checker.
         call_compiler: The custom call compiler.
         higher_order_value: Whether the function may be used as a higher-order value.
+        signature: User-provided signature.
     """
 
     python_func: "PyFunc"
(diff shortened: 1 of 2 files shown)


guppylang/definition/custom.py
```diff
@@ -98,10 +99,12 @@ class RawCustomFunctionDef(ParsableDef):
     # if a static type for the function is provided.
     higher_order_value: bool
 
+    signature: FunctionType | None
+
     description: str = field(default="function", init=False)
 
     def parse(self, globals: "Globals", sources: SourceMap) -> "CustomFunctionDef":
-        """Parses and checks the user-provided signature of the custom function.
+        """Parses and checks the user-annotated signature of the custom function.
```
acl-cqc: Grammar super-nit: I think provided is probably the right word here. "Annotated", as an adjective, means that something *has* annotations, not that it is an annotation itself.
The function-def is annotated *with* the signature; the signature itself *is* an annotation (as per a couple of lines down).

mark-koch (commented):
Looks good 👍 Could you add a test to make sure it works as expected?

guppylang/definition/custom.py
hsemenenko: What is meant by the docs is that the user has added type annotations to the decorated function. I changed the docs to try to make this clearer that the user was not providing the signature, but instead it was being parse from the type annotations. As providing a signature is now an option, I will update the docs to reflect this.

hsemenenko:
@mark-koch @acl-cqc I have added tests and update the docs to make it clearer how the signature is provided.

PR #1006, opened 2025-06-07
feat: Add PriorityQueue to standard library

### LINKED ISSUE #974: Add priority queue data structure to the standard library
Extend Guppy's standard library by adding a [priority queue](https://en.wikipedia.org/wiki/Priority_queue) data structure.

# Details

Read our development guide to set up your environment: [DEVELOPMENT.md](https://github.com/CQCL/guppylang/blob/main/DEVELOPMENT.md)

## Requirements

The priority queue implementation should be added to the [`guppylang/std/collections.py`](https://github.com/CQCL/guppylang/blob/main/guppylang/std/collections.py) file. It should offer the following methods:

* `__len__`: Return the number of elements currently in the queue
* `push`: Given an element and an integer priority, add it to the queue
* `pop`: Return the element with the highest priority and remove it from the queue
* `peek`: Return the element with the highest priority without removing it from the queue

## Implementation

The priority queue should be implemented via a [binary heap](https://en.wikipedia.org/wiki/Binary_heap) that is backed by an array (see for example [here](https://www.geeksforgeeks.org/array-representation-of-binary-heap/) for how to represent binary trees using arrays).

You can follow the [existing stack implementation](https://github.com/CQCL/guppylang/blob/f82ee429eb9703dc4b4014eb8763e6b6907e9bcc/guppylang/std/collections.py#L15) as an example for how to define new types in Guppy and how to work with arrays. In particular, the maximum size of the queue will need to be specified as a type parameter to ensure that everything can be statically allocated. Again, the stack implementation should serve as guidance here.

A good starting point would the following struct declaration:

```python
@guppy.struct
class PriorityQueue(Generic[T, MAX_SIZE]):
    buf: array[Option[tuple[int, T]], MAX_SIZE]  # For each node, store the element and its weight
    size: int
    
    def __len__(self: PriorityQueue[T, MAX_SIZE]) -> int: ...

    def push(self: PriorityQueue[T, MAX_SIZE] @ owned, value: T @ owned, priority: int) -> PriorityQueue[T, MAX_SIZE]: ...

    def pop(self: PriorityQueue[T, MAX_SIZE] @ owned) -> tuple[int, T, PriorityQueue[T, MAX_SIZE]]: ...

    def peek(self: PriorityQueue[TCopyable, MAX_SIZE] @ owned) -> tuple[int, T, PriorityQueue[TCopyable, MAX_SIZE]]: ...
```

## Tests

Add some tests to [`tests/integration/std/test_collections.py`](https://github.com/CQCL/guppylang/blob/main/tests/integration/std/test_collections.py) to verify correctness of your implementation.

andre-a-alves:
Closes #974

Adds a PriorityQueue that implements a min heap under the hood.

### DIFF
--- a/guppylang/std/collections.py
+++ b/guppylang/std/collections.py
@@ -206,8 +206,8 @@ class PriorityQueue(Generic[T, MAX_SIZE]):  # type: ignore[misc]
         """
         if self.size <= 0:
             panic("PriorityQueue.peek: priority queue is empty")
-        elem = self.buf[0].unwrap()
-        return elem[0], elem[1], PriorityQueue(self.buf, self.size)
+        prio, val = self.buf[0].unwrap()
+        return prio, val, PriorityQueue(self.buf, self.size)
 
 
 @guppy
--- a/guppylang/std/collections.py
+++ b/guppylang/std/collections.py
@@ -89,6 +89,17 @@ def empty_stack() -> Stack[T, MAX_SIZE]:
 
 @guppy.struct
 class PriorityQueue(Generic[T, MAX_SIZE]):  # type: ignore[misc]
+    """A queue of values ordered by priority.
+
+    Values with the lowest priority value will be popped from the queue first.
+    (e.g. Priority 1 will be returned before priority 2.) To ensure static
+    allocation, the maximum queue size must be specified in advance and
+    is tracked in the type. For example, the `PriorityQueue[int, 10]`
+    is a queue that can hold at most 10 prioritized integers.
+
+    Use `empty_priority_queue` to construct a new priority_queue.
+    """
+
     #: Underlying buffer holding the priority queue elements.
     #:
     #: INVARIANT: All array elements up to and including index `self.size - 1` are
--- a/guppylang/std/collections.py
+++ b/guppylang/std/collections.py
@@ -74,7 +74,7 @@ class Stack(Generic[T, MAX_SIZE]):  # type: ignore[misc]
         Note that this operation is only allowed if the stack elements are copyable.
         """
         if self.end <= 0:
-            panic("Stack.peel: stack is empty")
+            panic("Stack.peek: stack is empty")
         elem = self.buf[self.end - 1].unwrap()
         return elem, Stack(self.buf, self.end)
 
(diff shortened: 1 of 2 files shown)


guppylang/std/collections.py
```diff
@@ -85,3 +85,134 @@ def empty_stack() -> Stack[T, MAX_SIZE]:
...
+
+        Panics if the priority queue is empty.
+
+        Note that this operation is only allowed if the priority queue elements are
+        copyable.
+        """
+        if self.size <= 0:
+            panic("PriorityQueue.peek: priority queue is empty")
+        elem = self.buf[0].unwrap()
+        return elem[0], elem[1], PriorityQueue(self.buf, self.size)
```
tatiana-s: We don't have tuple indexing in Guppy (yet), so you have to use unpacking instead (it is generally worth testing all methods you've added to catch cases like this!):
```suggestion
        prio, val = self.buf[0].unwrap()
        return prio, val, PriorityQueue(self.buf, self.size)
```

tests/integration/std/test_collections.py
```diff
@@ -21,3 +21,20 @@ def main() -> int:
     run_int_fn(compiled, sum((i + 1) * x for i, x in enumerate(reversed(list(range(10))))))
 
 
+def test_priority_queue(validate, run_int_fn) -> None:
```
tatiana-s, line 24: This test fails on execution because the sum isn't quite correct - it should work with the priority set to `i` instead of `9 - i`.

tatiana-s (changes requested):
Hi, @andre-a-alves, thank you for your contribution! I have added some comments but generally this looks good :)

guppylang/std/collections.py
gringioca: Thanks for pointing this out. It passed all the tests I was able to run locally, but I couldn't get llvm14 to install (nix didn't work, either).
I've made the change.

tests/integration/std/test_collections.py
gringioca: I definitely did the math wrong! I'm leaving the priority opposite the value to ensure elements come out of the queue in the  correct order, but I fixed the math. Ironically, it looks even more similar to the stack test, but I added some comments explaining why it's put together the way it is.

tatiana-s (approved):
Nice, thank you!

Issue #1008, opened 2025-06-09
Extend comptime arguments to arbitrary non-linear types


mark-koch:
Via monomorphisation in Guppy

PR #1015, opened 2025-06-09
feat: Comptime `nat` arguments

### LINKED ISSUE #1007: Comptime `nat` arguments
Should lowering to Hugr bounded nat type args

mark-koch:
Closes #1007 

Adds a new `@comptime` annotation (analogous to `@owned`) to mark that a function argument must be known at compile-time. Currently, only `nat` arguments are allowed to annotated as `@comptime`. This restriction will be lifted as a follow-up (see #1008).

# Example

```python
@guppy
def foo(x: nat @comptime) -> None: ...

@guppy
def main(a: nat, b: nat @comptime) -> None:
    foo(42)  # Ok
    foo(42 + 1)  # Error: Addition not performed at compile-time 
    foo(comptime(42 + 1))  # Ok
    foo(b)  # Ok (since b is another comptime arg)
    c = 42
    foo(c)  # Error: c is a dynamic variable, so not tracked at compile-time
```

See #987 for follow-up ideas to improve tracking of comptime values.

# Implementation details

Comptime arguments are turned into implicit generic type arguments. E.g.

```python
def foo(n: nat @comptime): ...
```

is turned into

```python
def foo[n: nat](_: nat): ...
```

To achieve this, I made the following changes:

* While parsing a function signature  (`parse_function_io_types` in `tys/parsing.py`), when we encouter an argument that is annotated as `@comptime`, we generate a new `ConstParam` and bind it to the argument name in `param_var_mapping`. This way, the user can refer to the parameter by name at a later point in the signature.
* In the generated `ConstParam`, we set a new field `from_comptime_arg=True`. Later, when we collect the parsed signature into a `FunctionType`, we use this information tell these implicit type parameters apart from the regular ones.
* Concretely, `FunctionType` has a new field `comptime_args` that holds the instantiated values for all of its comptime arguments. These are either:
    * `BoundConstVar`s when the values are still bound to their parameters (i.e. they're still quantified)
    * `ExistentialConstVar`s after calling the `FunctionType.unquantified` method: This is to infer the type arguments when the user calls the functions
    * `ConstValue`s when we have picked a concrete number-instantiation for the comptime arguments
* During Hugr lowering, we remove comptime args from the function signature and only keep them as bounded nat type args (`FunctionType.to_hugr`). When compiling a call, we make sure that comptime args are not lowered to Hugr (`ExprCompiler._compile_call_args`)


guppylang/checker/core.py
```diff
@@ -80,6 +80,7 @@ class Variable:
     ty: Type
     defined_at: AstNode | None
     flags: InputFlags = InputFlags.NoFlags
+    is_func_input: bool = field(default=False, kw_only=True)
```
mark-koch, line 86: Remember if a variable is a function input. This way, we can suggest to add a `@comptime` annotation to users when appropriate

guppylang/checker/cfg_checker.py
```diff
@@ -103,10 +103,7 @@ def check_cfg(
     while len(queue) > 0:
         pred, num_output, bb = queue.popleft()
         pred_outputs = [*pred.sig.output_rows, *pred.sig.dummy_output_rows]
-        input_row = [
-            Variable(v.name, v.ty, v.defined_at, v.flags)
-            for v in pred_outputs[num_output]
-        ]
+        input_row = pred_outputs[num_output]
```
mark-koch, line 106: `Variable` is immutable, so no need to make a copy here. This also ensures that the new `Variable.is_func_input` field is carried through all basic blocks

guppylang/checker/cfg_checker.py
```diff
@@ -222,7 +219,11 @@ def check_bb(
     if bb == cfg.entry_bb:
         assert len(bb.predecessors) == 0
         for x, use in bb.vars.used.items():
-            if x not in cfg.ass_before[bb] and x not in globals:
+            if (
+                x not in cfg.ass_before[bb]
+                and x not in globals
+                and x not in generic_params
+            ):
```
mark-koch, line 226: There are a bunch of places where we forgot to check if variables are generic parameters. This used to be ok since generic params were also globals (since you declare them globally via `guppy.type_var`), but now we also have locally scoped ones

guppylang/definition/custom.py
```diff
@@ -131,28 +131,6 @@ def parse(self, globals: "Globals", sources: SourceMap) -> "CustomFunctionDef":
...
 
-    def compile_call(
-        self,
-        args: list[Wire],
-        type_args: Inst,
-        dfg: DFContainer,
-        ctx: CompilerContext,
-        node: AstNode,
-        function_ty: ht.FunctionType,
-    ) -> Sequence[Wire]:
```
mark-koch, line 142: drive-by: This function is actually unused. We should never call `compile_call` on a `RawCustomFunctionDef`, only on `CustomFunctionDef`s

guppylang/std/builtins.py
```diff
@@ -71,13 +71,20 @@
...
-def comptime(*args: Any) -> Any:
-    """Function to tag compile-time evaluated Python expressions in a Guppy context.
+class _Comptime:
+    """Dummy class to support `@comptime` annotations and `comptime(...)` expressions."""
+
+    def __call__(self, *args: Any) -> Any:
+        return tuple(args)
+
+    def __rmatmul__(self, other: Any) -> Any:
+        return other
```
mark-koch: This is to make the Python interpreter happy with `@comptime` at runtime

guppylang/tys/parsing.py
```diff
@@ -117,10 +128,10 @@ def _try_parse_defn(node: AstNode, globals: Globals) -> Definition | None:
     match node:
         case ast.Name(id=x):
             if x not in globals:
-                raise GuppyError(VarNotDefinedError(node, x))
+                return None
```
mark-koch, line 131: Small refactor: If `x` is not a global, we shouldn't error immediately because it could also refer to a generic parameter. This is now checked above

tests/error/comptime_arg_errors/unknown_arg.err
```diff
@@ -0,0 +1,17 @@
...
+   |             ^ Value of this argument must be known at compile-time
+   | 
+10 | def main(b: bool, m: nat) -> None:
+   |                   ------ `m` is a function argument, so its value is not statically
+   |                          known. Consider turning `m` into a comptime argument: `m:
+   |                          @comptime`
+
+Note: We are currently investigating ways to make Guppy's compile-time reasoning
+smarter. Please leave your feedback at
+https://github.com/CQCL/guppylang/discussions/987
```
mark-koch, line 15: Not sure if this is a good idea?

guppylang/std/builtins.py
croyzor: this would be better as a comment in the code

guppylang/checker/core.py
croyzor: this would be better as a comment in the code

guppylang/checker/expr_checker.py
```diff
@@ -951,14 +957,19 @@ def type_check_args(
     check_num_args(len(func_ty.inputs), len(inputs), node, func_ty)
 
     new_args: list[ast.expr] = []
+    comptime_args = iter(func_ty.comptime_args)
     for inp, func_inp in zip(inputs, func_ty.inputs, strict=True):
         a, s = ExprChecker(ctx).check(inp, func_inp.ty.substitute(subst), "argument")
```
croyzor, line 962: it seems like we'd want to skip this if `InputFlags.Comptime in func_inp.flags`?

guppylang/checker/expr_checker.py
```diff
@@ -1019,6 +1030,44 @@ def check_place_assignable(
...
+        case GenericParamValue(param=const_param):
+            const = const_param.to_bound().const
+        case arg:
+            # Anything else is considered unknown at comptime, but we can give some
+            # nicer error hints by inspecting in more detail
+            err = ComptimeUnknownError(arg, "argument")
+            s: SubDiagnostic
+            match arg:
+                case PlaceNode(place=place) if place.root.is_func_input:
+                    s = ComptimeUnknownError.InputHint(arg.place.defined_at, arg.place)
```
croyzor: nit: these could be using `place` instead of `arg.place`

guppylang/checker/func_checker.py
```diff
@@ -120,7 +122,7 @@ def check_nested_func_def(
 
     # Construct inputs for checking the body CFG
     inputs = [v for v, _ in captured.values()] + [
-        Variable(x, inp.ty, func_def.args.args[i], inp.flags)
+        Variable(x, inp.ty, func_def.args.args[i], inp.flags, is_func_input=True)
         for i, (x, inp) in enumerate(
             zip(func_ty.input_names, func_ty.inputs, strict=True)
```
croyzor, line 127: I think this should just be the non `CompTime` inputs (nested funcs can still take `@comptime` args right?)

mark-koch: > nested funcs can still take @comptime args right?

No, nested functions are not allowed to be generic for now. But you're right, we should filter them out to make our lives easier if we enable this at some point later 👍

guppylang/checker/expr_checker.py
mark-koch: We still need to type check the argument and make sure it has the correct type, otherwise we could produce an incorrect instantiation

guppylang/checker/errors/comptime_errors.py
```diff
@@ -65,3 +66,48 @@ class PytketSignatureMismatch(Error):
     class TypeHint(Note):
         message: ClassVar[str] = "Expected circuit signature is `{circ_sig}`"
         circ_sig: FunctionType
+
+
+@dataclass(frozen=True)
+class ComptimeCallableArgError(Error):
+    title: ClassVar[str] = "Invalid comptime an"
```
croyzor: "an"?

tests/error/comptime_arg_errors/unknown_arg.err
croyzor: 🤷

croyzor: For _this_ example at least, it doesn't seem super relevant

croyzor (approved):

tests/error/comptime_arg_errors/unknown_arg.err
mark-koch: Yeah, the problem is that detecting where the hint would be relevant is almost as hard as implementing #987 😅

PR #1017, opened 2025-06-10
feat: Allow users to wrap `guppy` in their own decorator

### LINKED ISSUE #1016: Allow users to wrap guppy decorator in their own decorator
Currently, if a user tries to define their own decorator that wraps `guppy`, everything will break since we do some stack-frame walking to look up variables in scope

mark-koch:
Closes #1016

Adds a new `@custom_guppy_decorator` decorator to annotate functions that wrap a call to the builtin `guppy` decorator. This makes the compiler aware that the function acts as a decorator which is taken into account during name resolution.

Examples:

```python
@custom_guppy_decorator
def my_guppy(f):
    # Some custom logic here ...
    return guppy(f)

@my_guppy
def main() -> int: ...
```

If the `@custom_guppy_decorator` were missing, then compilation of `main` would not work correctly.

### DIFF
--- a/guppylang/decorator.py
+++ b/guppylang/decorator.py
@@ -471,10 +471,40 @@ def _find_load_call(sources: SourceMap) -> Span | None:
     return None
 
 
+def custom_guppy_decorator(f: F) -> F:
+    """Decorator to mark user-defined decorators thag wrap builtin `guppy` decorators.
+
+    Example:
+
+    ```
+    @custom_guppy_decorator
+    def my_guppy(f):
+        # Some custom logic here ...
+        return guppy(f)
+
+    @my_guppy
+    def main() -> int: ...
+    ```
+
+    If the `custom_guppy_decorator` were missing, then the `@my_guppy` annotation would
+    not produce a valid guppy definition.
+    """
+    object.__setattr__(f, "__custom_guppy_decorator__", True)
+    return f
+
+
 def get_calling_frame() -> FrameType:
     """Finds the first frame that called this function outside the compiler modules."""
     frame = inspect.currentframe()
     while frame:
+        # Skip frame if we're inside a user-defined decorator that wraps the `guppy`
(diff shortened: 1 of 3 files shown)


mark-koch:
CC @hsemenenko

tests/integration/test_custom_decorator.py
```diff
@@ -0,0 +1,27 @@
...
+
+@custom_guppy_decorator
+def my_guppy(f):
+    return guppy(f)
+
+
+def test_custom_decorator(validate):
+    def make():
+        @my_guppy
+        def foooo() -> int:
```
ss2165, line 12: 👻

guppylang/decorator.py
```diff
@@ -471,10 +471,40 @@ def _find_load_call(sources: SourceMap) -> Span | None:
     return None
 
 
+def custom_guppy_decorator(f: F) -> F:
+    """Decorator to mark user-defined decorators thag wrap builtin `guppy` decorators.
```
ss2165:
```suggestion
    """Decorator to mark user-defined decorators that wrap builtin `guppy` decorators.
```

guppylang/decorator.py
```diff
@@ -471,10 +471,40 @@ def _find_load_call(sources: SourceMap) -> Span | None:
...
 def get_calling_frame() -> FrameType:
     """Finds the first frame that called this function outside the compiler modules."""
     frame = inspect.currentframe()
     while frame:
+        # Skip frame if we're inside a user-defined decorator that wraps the `guppy`
+        # decorator. Those are functions that have a `__custom_guppy_decorator__` field.
+        # We can get a reference to the calling function by looking up the code function
+        # name in the frame globals:
+        func = frame.f_globals.get(frame.f_code.co_name)
+        if hasattr(func, "__custom_guppy_decorator__"):
```
ss2165: to be safe I think worth checking the value of the attribute is True in case some future code elsewhere messes with it

PR #1020, opened 2025-06-10
fix: Make array results borrow array instead of copying implicitly

### LINKED ISSUE #981: Array results should borrow the passed array
Currently, programs like

```python
xs = array(1, 2, 3)
result("foo", xs)
result("bar", xs)
```

create a Hugr where the `xs` port is used twice. This violates our no-implicit-copy guarantee!

Note: Fixing this requires chaning the array result Hugr op to return back the result

tatiana-s:
Closes #981 

Requires a breaking tket2 release with https://github.com/CQCL/tket2/pull/888

### DIFF
--- a/guppylang/compiler/expr_compiler.py
+++ b/guppylang/compiler/expr_compiler.py
@@ -56,6 +56,7 @@ from guppylang.nodes import (
 )
 from guppylang.std._internal.compiler.arithmetic import convert_ifromusize
 from guppylang.std._internal.compiler.array import (
+    array_convert_from_std_array,
     array_convert_to_std_array,
     array_map,
     array_repeat,
@@ -72,11 +73,13 @@ from guppylang.std._internal.compiler.prelude import (
 from guppylang.std._internal.compiler.tket2_bool import (
     OpaqueBool,
     OpaqueBoolVal,
+    make_opaque,
     not_op,
     read_bool,
 )
 from guppylang.tys.arg import Argument
 from guppylang.tys.builtin import (
+    array_type,
     bool_type,
     get_element_type,
     int_type,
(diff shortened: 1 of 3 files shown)


guppylang/compiler/expr_compiler.py
```diff
@@ -815,9 +842,82 @@ def array_read_bool(ctx: CompilerContext) -> hf.Function:
...
+
 T = TypeVar("T")
 
 
 def doesnt_contain_none(xs: list[T | None]) -> TypeGuard[list[T]]:
     """Checks if a list contains `None`."""
     return all(x is not None for x in xs)
+
+
+def apply_array_op_with_conversions(
```
tatiana-s, line 882: Extracted into a separate function mainly because I am doing the same things in the state result compilation and I reckon it might show up again whenever there are Hugr array ops - but maybe there is a better way to abstract this. I did think about making this a global function as well but with some of it being optional that would get complicated quickly, could be a follow up maybe

pyproject.toml
```diff
@@ -86,7 +86,7 @@ execute-llvm = { workspace = true }
 
 # Uncomment these to test the latest dependency version during development
 # hugr = { git = "https://github.com/CQCL/hugr", subdirectory = "hugr-py", rev = "7f233fa" }
-# tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", rev = "652a7d0" }
+tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", branch = "ts/array-result-sig" }
```
doug-q: This is released now right?

guppylang/compiler/expr_compiler.py
doug-q: It seems like a good outlining. I don't quite understand the comment though. It looks like there are two modes, which you specify via the `/`s. But then which mode applies when?

tatiana-s: It's not two modes, it what's done to inputs and then consequently undone again on the output, tried to be a bit more descriptive in the comment

Issue #1018, opened 2025-06-10
Support arbitrary const generic params via monomorphisation


mark-koch:
So far, we only allow `nat` since they lower to Hugr bounded nats. Support all other copyable+droppable types via monomorphisation in Guppy.

In particular, this is a precursor to #1008

mark-koch:
Closed by #1033

PR #1022, opened 2025-06-11
chore!: Require a context when converting types to Hugr

mark-koch:
This is required as a precusor to #1018 to handle monomorphisation: When compiling to Hugr, some bound type variables will need to be replaced with a monotype. We will need access to the compiler context to look this information up.

Instead of requiring the whole compiler context, I added a new `ToHugrContext` protocol in which we can put everything that is needed during the Hugr translation of types (empty for now but will be filled in a subsequent PR). The main motivation for this is that the context needs to change when we go under a binder.

BREAKING CHANGE: All `to_hugr` methods on types, arguments, and parameters now require a `ToHugrContext`
BREAKING CHAGNE: `CompileableDef.compile_outer` now requires a `ToHugrContext`
BREAKING CHANGE: `guppy.hugr_op` now passes the compiler context to the function generating the op

### DIFF
--- a/guppylang/compiler/cfg_compiler.py
+++ b/guppylang/compiler/cfg_compiler.py
@@ -45,7 +45,7 @@ def compile_cfg(
 
     # Explicitly annotate the output types since Hugr can't infer them if the exit is
     # unreachable
-    out_tys = [place.ty.to_hugr() for place in cfg.exit_bb.sig.input_row]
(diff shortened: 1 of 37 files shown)


mark-koch:
Actually, I'm not sure if this is really needed. @acl-cqc maybe hold off on reviewing this until I've thought a bit more about this

mark-koch:
@acl-cqc This is ready for review now

PR #1025, opened 2025-06-11
fix: deterministic worklist iteration order

### LINKED ISSUE #1024: HUGR node ordering can be non-deterministic
Example was generating a CFG node at a different point depending on test execution order

ss2165:
Closes #1024


mark-koch (approved):
If we find other places where we need this, it might be worth defining our own `OrderedSet` class that wraps this logic

PR #1031, opened 2025-06-16
docs: use the sphinx furo theme for guppy compiler docs

CalMacCQ:
Here I've used the base furo theme just like the HUGR-py docs. As these docs pages I think its not worth using the quantinuum-sphinx theme with the navbar at present.

Driveby: add some useful docs commands to the justfile (see [f9e4f35](https://github.com/CQCL/guppylang/pull/1031/commits/f9e4f357db254b671ea2e83a60aa65268d365f84))

Screenshot 2025-06-16 at 12 57 56


docs/api-docs/conf.py
```diff
@@ -19,22 +19,13 @@
...
-    "logo": {
-        "image_light": "_static/Quantinuum_logo_black.png",
-        "image_dark": "_static/Quantinuum_logo_white.png",
-    },
-}
+html_theme_options = {}
 
-html_static_path = ["../_static"]
-html_css_files = ["custom.css"]
+html_static_path = []
```
CalMacCQ, line 28: No static files are needed at the moment. Could add the guppy mascot but that seems like it belongs in the user facing docs rather than the compiler docs.

mark-koch (approved):
Thanks!

PR #1035, opened 2025-06-17
feat!: Lower const generics via monomorphization

mark-koch:
Lowers the const generic introduced in #1034 to Hugr via monomorphization.

* In `definition.common`, we add new `MonomorphizableDef` abstract base classes to mark definitions that need to be lowered via monomorphisation. `CheckedFunctionDef` is changed to implement `MonomorphizableDef` instead of `CompileableDef`. All other definitions are unchanged.

* The `checker.core.CompilerContext` is the main driver for monomorphisation:
  * When calling `CompilerContext.build_compiled_def`, we now also need to pass the type arguments for the definition we want to compile.
  * If the to be compiled function is a `MonomorphizableDef`, the context decides which of its parameters will need to be monomorphised. This is done by the `requires_monomorphization` function. Currently, those are exactly the non-nat const parameters of the functio (not that nat params can be lowered to Hugr bounded nats, so monomorphization is not needed)
  * From the provided type arguments, the context picks the subset that correspond to these to-be-monomorphized parameters, which is done via the `partially_monomorphize_args` function. We represent these partial monomorphizations using the `PartiallyMonomorphizedArgs` type.
  * A `mono_args: PartiallyMonomorphizedArgs` is a sequence of `Argument | None`, so that `mono_args[i] == None` means that the argument with index `i` is not part of the monomorphization. Otherwise, `mono_args[i]` specifies the monomorphic instantiation for parameter `i`. So a function with 3 parameters where no monomorphization is required at all, will have `mono_args = (None, None, None)`.
  * The worklist and compilation store in the `CompilerContext` are now indexed via a pair of `DefId` and optional `PartiallyMonomorphizedArgs`. This way, the worklist can contain the same definition multiple times, but with different monomorphized args.
  * The same also applies to the `global_funcs` stored in the context. By taking potential mono-arguments into account, we can create multiple global functions corresponding to the same `GlobalConstId`. This is used in `CustomFunctionDef.load_with_args` to support custom functions with params that require monomorphization.

* The context also has two new methods: `type_var_to_hugr` and `const_var_to_hugr`. Those are used when compiling types and type arguments to Hugr. The context stores the current monomorphisation of the function that is being compiled, and when we request to lower a variable to Hugr, the context makes sure that the de Bruijn indices are down-shifted to account for the monomorphized parameters.
* Similarly, `ExprCompiler.visit_GenericParamValue` looks up the value corresponding to the current monomorphization in the context

BREAKING CHANGE: `CheckedFunctionDef` now implements `MonomorphizableDef` instead of `CompileableDef`
BREAKING CHANGE: `CompilerContext.build_compiled_def` now requires an instantiation for the definitions type parameters
BREAKING CHANGE: The `ToHugrContext` protocol now requires two additional methods:  `type_var_to_hugr` and `const_var_to_hugr`
BREAKING CHANGE: `CompilerContext.{compiled, worklist}` and `CompilationEngine.compiled` are now indexed by a tuple of `DefId` and optional `PartiallyMonomorphizedArgs`

### DIFF
--- a/guppylang/compiler/core.py
+++ b/guppylang/compiler/core.py
@@ -1,10 +1,10 @@
 import itertools
 from abc import ABC
 from collections import defaultdict
-from collections.abc import Iterator
+from collections.abc import Iterator, Sequence
 from contextlib import contextmanager
 from dataclasses import dataclass, field
-from typing import Any, cast
+from typing import TYPE_CHECKING, Any, ClassVar, cast
 
 import tket2_exts
 from hugr import Hugr, Node, Wire, ops
@@ -14,17 +14,38 @@ from hugr.build.dfg import DP, DefinitionBuilder, DfBase
 from hugr.hugr.base import OpVarCov
 from hugr.hugr.node_port import ToNode
 from hugr.std import PRELUDE
+from typing_extensions import assert_never
 
 from guppylang.checker.core import FieldAccess, Globals, Place, PlaceId, Variable
-from guppylang.definition.common import CheckedDef, CompilableDef, CompiledDef, DefId
+from guppylang.definition.common import (
+    CheckedDef,
+    CompilableDef,
+    CompiledDef,
+    DefId,
+    MonomorphizableDef,
+)
 from guppylang.definition.ty import TypeDef
 from guppylang.definition.value import CompiledCallableDef
+from guppylang.diagnostic import Error
 from guppylang.engine import ENGINE
-from guppylang.error import InternalGuppyError
+from guppylang.error import GuppyError, InternalGuppyError
+from guppylang.tys.arg import ConstArg, TypeArg
+from guppylang.tys.builtin import nat_type
 from guppylang.tys.common import ToHugrContext
-from guppylang.tys.ty import StructType, Type
+from guppylang.tys.const import BoundConstVar, ConstValue
+from guppylang.tys.param import ConstParam, Parameter, TypeParam
+from guppylang.tys.subst import Inst
+from guppylang.tys.ty import BoundTypeVar, NumericType, StructType, Type
+
+if TYPE_CHECKING:
+    from guppylang.tys.arg import Argument
 
 CompiledLocals = dict[PlaceId, Wire]
+
+#: Partial instantiation of generic type parameters for monomorphization.
+#:
+#: Note that this is required to be a tuple to ensure hashability.
+PartiallyMonomorphizedArgs = tuple["Argument | None", ...]
 
 
 @dataclass(frozen=True)
@@ -43,6 +64,25 @@ class GlobalConstId:
         return f"{self.base_name}.{self.id}"
 
 
+#: Unique identifier for a partially monomorphized definition
+MonoDefId = tuple[DefId, PartiallyMonomorphizedArgs | None]
+
+#: Unique identifier for global Hugr constants and functions with optional monomorphized
+#: type arguments
+MonoGlobalConstId = tuple[GlobalConstId, PartiallyMonomorphizedArgs | None]
+
+
+@dataclass(frozen=True)
+class EntryMonomorphizeError(Error):
+    title: ClassVar[str] = "Invalid entry point"
+    span_label: ClassVar[str] = (
+        "Function `{name}` is not a valid compilation entry point since the value of "
+        "generic paramater `{param}` is not known"
+    )
+    name: str
+    param: Parameter
+
+
 class CompilerContext(ToHugrContext):
     """Compilation context containing all available definitions.
 
@@ -52,10 +92,16 @@ class CompilerContext(ToHugrContext):
     """
 
     module: DefinitionBuilder[ops.Module]
-    compiled: dict[DefId, CompiledDef]
-    worklist: dict[DefId, None]  # use dict over set for deterministic iteration order
+    compiled: dict[MonoDefId, CompiledDef]
 
-    global_funcs: dict[GlobalConstId, hf.Function]
+    # use dict over set for deterministic iteration order
+    worklist: dict[MonoDefId, None]
+
+    global_funcs: dict[MonoGlobalConstId, hf.Function]
+
+    #: Partial instantiation for some of the type parameters of the current function for
+    #: the purpose of monomorphization
+    current_mono_args: PartiallyMonomorphizedArgs | None
 
     checked_globals: Globals
 
@@ -68,39 +114,85 @@ class CompilerContext(ToHugrContext):
         self.compiled = {}
         self.global_funcs = {}
         self.checked_globals = Globals(None)
+        self.current_mono_args = None
 
-    def build_compiled_def(self, def_id: DefId) -> CompiledDef:
+    @contextmanager
+    def set_monomorphized_args(
+        self, mono_args: PartiallyMonomorphizedArgs | None
+    ) -> Iterator[None]:
+        """Context manager to set the partial monomorphization for the function that is
+        compiled currently being compiled.
+        """
+        old = self.current_mono_args
+        self.current_mono_args = mono_args
+        yield
+        self.current_mono_args = old
+
+    def build_compiled_def(self, def_id: DefId, type_args: Inst | None) -> CompiledDef:
         """Returns the compiled definitions corresponding to the given ID.
 
         Might mutate the current Hugr if this definition has never been compiled before.
         """
-        if def_id not in self.compiled:
-            defn = ENGINE.get_checked(def_id)
-            self.compiled[def_id] = self._compile(defn)
-            self.worklist[def_id] = None
-        return self.compiled[def_id]
+        # TODO: The check below is a hack to support nested function definitions. We
+        #  forgot to insert frames for nested functions into the DEF_STORE, which would
+        #  make the call to `ENGINE.get_checked` below fail. For now, let's just short-
+        #  cut if the function doesn't take any generic params (as is the case for all
+        #  nested functions).
+        #  See https://github.com/CQCL/guppylang/issues/1032
+        if (def_id, ()) in self.compiled:
+            return self.compiled[def_id, ()]
 
-    def _compile(self, defn: CheckedDef) -> CompiledDef:
+        defn = ENGINE.get_checked(def_id)
+        mono_args: PartiallyMonomorphizedArgs | None = None
+        if isinstance(defn, MonomorphizableDef):
+            assert type_args is not None
+            mono_args = partially_monomorphize_args(defn.params, type_args, self)
+        if (def_id, mono_args) not in self.compiled:
+            self.compiled[def_id, mono_args] = self._compile(defn, mono_args)
+            self.worklist[def_id, mono_args] = None
+        return self.compiled[def_id, mono_args]
+
+    def _compile(
+        self, defn: CheckedDef, mono_args: PartiallyMonomorphizedArgs | None
+    ) -> CompiledDef:
         if isinstance(defn, CompilableDef):
+            assert mono_args is None
             return defn.compile_outer(self.module, self)
+        if isinstance(defn, MonomorphizableDef):
+            assert mono_args is not None
+            return defn.monomorphize(self.module, mono_args, self)
         return defn
 
     def compile(self, defn: CheckedDef) -> None:
         """Compiles the given definition and all of its dependencies into the current
         Hugr."""
-        if defn.id in self.compiled:
-            return
+        # Check and compile the entry point
+        checked = ENGINE.get_checked(defn.id)
+        mono_args: PartiallyMonomorphizedArgs | None = None
+        if isinstance(checked, MonomorphizableDef):
+            # Make sure that the entry point doesn't require monomorphization
+            for param in checked.params:
+                if requires_monomorphization(param):
+                    err = EntryMonomorphizeError(
+                        checked.defined_at, checked.name, param
+                    )
+                    raise GuppyError(err)
+            mono_args = tuple(None for _ in checked.params)
+        self.compiled[defn.id, mono_args] = self._compile(checked, mono_args)
+        self.worklist[defn.id, mono_args] = None
 
-        self.compiled[defn.id] = self._compile(defn)
-        self.worklist[defn.id] = None
+        # Compile definition bodies
         while self.worklist:
-            next_id = self.worklist.popitem()[0]
-            with track_hugr_side_effects():
-                next_def = self.build_compiled_def(next_id)
+            next_id, mono_args = self.worklist.popitem()[0]
+            next_def = self.compiled[next_id, mono_args]
+            with track_hugr_side_effects(), self.set_monomorphized_args(mono_args):
                 next_def.compile_inner(self)
 
     def get_instance_func(
-        self, ty: Type | TypeDef, name: str
+        self,
+        ty: Type | TypeDef,
+        name: str,
+        type_args: Inst | None,
     ) -> CompiledCallableDef | None:
         from guppylang.engine import ENGINE
 
@@ -108,7 +200,7 @@ class CompilerContext(ToHugrContext):
         if parsed_func is None:
             return None
         checked_func = ENGINE.get_checked(parsed_func.id)
-        compiled_func = self.build_compiled_def(checked_func.id)
+        compiled_func = self.build_compiled_def(checked_func.id, type_args)
         assert isinstance(compiled_func, CompiledCallableDef)
         return compiled_func
 
@@ -116,21 +208,77 @@ class CompilerContext(ToHugrContext):
         self,
         const_id: GlobalConstId,
         func_ty: ht.PolyFuncType,
+        mono_args: PartiallyMonomorphizedArgs | None = None,
     ) -> tuple[hf.Function, bool]:
         """
         Creates a function builder for a global function if it doesn't already exist,
         else returns the existing one.
         """
-        if const_id in self.global_funcs:
-            return self.global_funcs[const_id], True
+        if (const_id, mono_args) in self.global_funcs:
+            return self.global_funcs[const_id, mono_args], True
         func = self.module.define_function(
             name=const_id.name,
             input_types=func_ty.body.input,
             output_types=func_ty.body.output,
             type_params=func_ty.params,
         )
-        self.global_funcs[const_id] = func
+        self.global_funcs[const_id, mono_args] = func
         return func, False
+
+    def type_var_to_hugr(self, var: BoundTypeVar) -> ht.Type:
+        """Compiles a bound Guppy type variable into a Hugr type.
+
+        Takes care of performing partial monomorphization as specified in the current
+        context.
+        """
+        if self.current_mono_args is None:
+            # If we're not inside a monomorphized context, just return the Hugr
+            # variable with the same de Bruijn index
+            return ht.Variable(var.idx, var.hugr_bound)
+
+        match self.current_mono_args[var.idx]:
+            # Either we have a decided to monomorphize the corresponding parameter...
+            case TypeArg(ty=ty):
+                return ty.to_hugr(self)
+            # ... or we're still want to be generic in Hugr
+            case None:
+                # But in that case we'll have to down-shift the de-Bruijn index to
+                # account for earlier params that have been monomorphized away
+                down_shift = sum(1 for arg in self.current_mono_args[: var.idx] if arg)
+                return ht.Variable(var.idx - down_shift, var.hugr_bound)
+            case _:
+                raise InternalGuppyError("Invalid monomorphization")
+
+    def const_var_to_hugr(self, var: BoundConstVar) -> ht.TypeArg:
+        """Compiles a bound Guppy constant variable into a Hugr type argument.
+
+        Takes care of performing partial monomorphization as specified in the current
+        context.
+        """
+        if var.ty != nat_type():
+            raise InternalGuppyError(
+                "Tried to convert non-nat const type argument to Hugr. This should "
+                "have been erased."
+            )
+        param = ht.BoundedNatParam(upper_bound=None)
+
+        if self.current_mono_args is None:
+            # If we're not inside a monomorphized context, just return the Hugr
+            # variable with the same de Bruijn index
+            return ht.VariableArg(var.idx, param)
+
+        match self.current_mono_args[var.idx]:
+            # Either we have a decided to monomorphize the corresponding parameter...
+            case ConstArg(const=ConstValue(value=int(v))):
+                return ht.BoundedNatArg(n=v)
+            # ... or we're still want to be generic in Hugr
+            case None:
+                # But in that case we'll have to down-shift the de-Bruijn index to
+                # account for earlier params that have been monomorphized away
+                down_shift = sum(1 for arg in self.current_mono_args[: var.idx] if arg)
+                return ht.VariableArg(var.idx - down_shift, param)
+            case _:
+                raise InternalGuppyError("Invalid monomorphization")
 
 
 @dataclass
@@ -230,6 +378,50 @@ def return_var(n: int) -> str:
 def is_return_var(x: str) -> bool:
     """Checks whether the given name is a dummy return variable."""
     return x.startswith("%ret")
+
+
+def requires_monomorphization(param: Parameter) -> bool:
+    """Checks if a type parameter must be monomorphized before compiling to Hugr.
+
+    This is required for some Guppy language features that cannot be encoded in Hugr
+    yet. Currently, this only applies to non-nat const parameters.
+    """
+    match param:
+        case TypeParam():
+            return False
+        case ConstParam(ty=ty):
+            return not isinstance(ty, NumericType) or ty.kind != NumericType.Kind.Nat
+        case x:
+            return assert_never(x)
+
+
+def partially_monomorphize_args(
+    params: Sequence[Parameter],
+    args: Inst,
+    ctx: CompilerContext,
+) -> PartiallyMonomorphizedArgs:
+    """
+    Given an instantiation of all type parameters, extracts the ones that will need to
+    be monomorphized.
+
+    Also takes care of normalising bound variables w.r.t. the current monomorphization.
+    """
+    mono_args: list[Argument | None] = []
+    for param, arg in zip(params, args, strict=True):
+        if requires_monomorphization(param):
+            # The constant could still refer to a bound variable, so we need to
+            # instantiate it w.r.t. to the current monomorphization
+            match arg:
+                case ConstArg(const=BoundConstVar(idx=idx)):
+                    assert ctx.current_mono_args is not None
+                    inst = ctx.current_mono_args[idx]
+                    assert inst is not None
+                    mono_args.append(inst)
+                case arg:
+                    mono_args.append(arg)
+        else:
+            mono_args.append(None)
+    return tuple(mono_args)
 
 
 QUANTUM_EXTENSION = tket2_exts.quantum()
--- a/guppylang/compiler/expr_compiler.py
+++ b/guppylang/compiler/expr_compiler.py
@@ -30,10 +30,12 @@ from guppylang.compiler.core import (
 from guppylang.compiler.hugr_extension import PartialOp
 from guppylang.definition.custom import CustomFunctionDef
 from guppylang.definition.value import (
+    CallableDef,
     CallReturnWires,
     CompiledCallableDef,
     CompiledValueDef,
 )
+from guppylang.engine import ENGINE
 from guppylang.error import GuppyError, InternalGuppyError
 from guppylang.nodes import (
     BarrierExpr,
@@ -75,7 +77,7 @@ from guppylang.std._internal.compiler.tket2_bool import (
     not_op,
     read_bool,
 )
-from guppylang.tys.arg import Argument
+from guppylang.tys.arg import Argument, ConstArg
 from guppylang.tys.builtin import (
     bool_type,
     get_element_type,
@@ -83,6 +85,7 @@ from guppylang.tys.builtin import (
     is_bool_type,
     is_frozenarray_type,
 )
+from guppylang.tys.const import ConstValue
 from guppylang.tys.subst import Inst
 from guppylang.tys.ty import (
     BoundTypeVar,
@@ -251,7 +254,15 @@ class ExprCompiler(CompilerBase, AstVisitor[Wire]):
         return self.dfg[node.place]
 
     def visit_GlobalName(self, node: GlobalName) -> Wire:
-        defn = self.ctx.build_compiled_def(node.def_id)
+        defn = ENGINE.get_checked(node.def_id)
+        if isinstance(defn, CallableDef) and defn.ty.parametrized:
+            # TODO: This should be caught during checking
+            err = UnsupportedError(
+                node, "Polymorphic functions as dynamic higher-order values"
+            )
+            raise GuppyError(err)
+
+        defn = self.ctx.build_compiled_def(node.def_id, type_args=[])
         assert isinstance(defn, CompiledValueDef)
         if isinstance(defn, CompiledCallableDef) and defn.ty.parametrized:
             # TODO: This should be caught during checking
@@ -270,8 +281,16 @@ class ExprCompiler(CompilerBase, AstVisitor[Wire]):
                 )
                 usize = self.builder.add_op(load_nat)
                 return self.builder.add_op(convert_ifromusize(), usize)
-            case _:
-                raise NotImplementedError
+            case ty:
+                # Look up monomorphization
+                assert self.ctx.current_mono_args is not None
+                match self.ctx.current_mono_args[node.param.idx]:
+                    case ConstArg(const=ConstValue(value=v)):
+                        val = python_value_to_hugr(v, ty, self.ctx)
+                        assert val is not None
+                        return self.builder.load(val)
+                    case _:
+                        raise InternalGuppyError("Monomorphized const is not a value")
 
     def visit_Name(self, node: ast.Name) -> Wire:
         raise InternalGuppyError("Node should have been removed during type checking.")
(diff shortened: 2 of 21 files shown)


guppylang/compiler/core.py
```diff
@@ -14,18 +14,39 @@
...
+
+if TYPE_CHECKING:
+    from guppylang.tys.arg import Argument
 
 CompiledLocals = dict[PlaceId, Wire]
 
+#: Partial instantiation of generic type parameters for monomorphization.
+#:
+#: Note that this is required to be a tuple to ensure hashability.
+PartiallyMonomorphizedArgs = tuple["Argument | None", ...]
```
acl-cqc, line 62: I'd document here that the Partially still means - as monomorphized as it's ever gonna get. (Maybe there's another name that indicates it's monomorphized "to the hugr level" or something, I'm not sure.)

acl-cqc: Maybe put an example - like you write in the description "a function with 3 parameters where no monomorphization is required at all, will have mono_args = (None, None, None)." - but nontrivial, here too

guppylang/compiler/core.py
```diff
@@ -52,10 +92,16 @@ class CompilerContext(ToHugrContext):
     """
 
     module: DefinitionBuilder[ops.Module]
-    compiled: dict[DefId, CompiledDef]
-    worklist: dict[DefId, None]  # use dict over set for deterministic iteration order
+    compiled: dict[MonoDefId, CompiledDef]
```
acl-cqc, line 123: Might be worth commenting that (IIUC) a DefId will feature in here *either* only once and without PartiallyMonomorphizedArgs, *or* always with PartiallyMonomorphizedArgs

acl-cqc: In some cases the PartiallyMonomorphizedArgs will be `()` (a type which supports only one value), so I'm not sure when this happens vs `None`. (Maybe only for the "nested-functions" hack below?)

guppylang/compiler/core.py
```diff
@@ -68,70 +114,172 @@ def __init__(
...
+
+        match self.current_mono_args[var.idx]:
+            # Either we have a decided to monomorphize the corresponding parameter...
+            case TypeArg(ty=ty):
+                return ty.to_hugr(self)
+            # ... or we're still want to be generic in Hugr
+            case None:
+                # But in that case we'll have to down-shift the de-Bruijn index to
+                # account for earlier params that have been monomorphized away
+                down_shift = sum(1 for arg in self.current_mono_args[: var.idx] if arg)
```
acl-cqc: Does this work? I think inversion rather than subtraction is probably clearer?
```suggestion
                hugr_idx = sum(1 for arg in self.current_mono_args[: var.idx] if arg is None)
                return ht.Variable(hugr_idx, var.hugr_bound)
```

guppylang/compiler/core.py
```diff
@@ -232,6 +380,50 @@ def is_return_var(x: str) -> bool:
...
+    ctx: CompilerContext,
+) -> PartiallyMonomorphizedArgs:
+    """
+    Given an instantiation of all type parameters, extracts the ones that will need to
+    be monomorphized.
+
+    Also takes care of normalising bound variables w.r.t. the current monomorphization.
+    """
+    mono_args: list[Argument | None] = []
+    for param, arg in zip(params, args, strict=True):
```
acl-cqc, line 458: There's a lot going on here and TBH I think given that you've done a pretty good job :). Nonetheless....I am wondering whether the `case ConstArg` can occur if `requires_monomorphization(param)` is false, or (if it can) whether it would work? e.g. flipping the order of tests, this seems significantly simpler if it works:
```
for param, arg in zip(params, args, strict=True):
    match arg:
        case ConstArg(const=BoundConstVar(idx=idx)):
            assert ctx.current_mono_args is not None
            inst = ctx.current_mono_args[idx]
            assert inst is not None # is this safe ???
            mono_args.append(inst)
        case arg:
            mono_args.append(arg if requires_monomorphization(param) else None)
```

guppylang/compiler/core.py
```diff
@@ -232,6 +380,50 @@ def is_return_var(x: str) -> bool:
...
+        if requires_monomorphization(param):
+            # The constant could still refer to a bound variable, so we need to
+            # instantiate it w.r.t. to the current monomorphization
+            match arg:
+                case ConstArg(const=BoundConstVar(idx=idx)):
+                    assert ctx.current_mono_args is not None
+                    inst = ctx.current_mono_args[idx]
```
acl-cqc, line 465: I guess we *don't* need to do this for other kinds of `arg` (e.g. types) because for those `requires_monomorphization` would be false i.e. we are just appending `None` ?

guppylang/compiler/core.py
```diff
@@ -232,6 +380,50 @@ def is_return_var(x: str) -> bool:
...
+                    assert inst is not None
+                    mono_args.append(inst)
+                case arg:
+                    mono_args.append(arg)
+        else:
+            mono_args.append(None)
```
acl-cqc, line 475: I find myself thinking - should we in fact be returning *two* things here:
* the PartiallyMonomorphizedArgs i.e. the ones that guppy is going to monomorphize for us
* the *remaining* values that still get passed as Hugr type-args?

Presumably the call-sites of `partially_monomorphize_args` will reveal all...

guppylang/compiler/expr_compiler.py
```diff
@@ -251,7 +254,15 @@ def visit_PlaceNode(self, node: PlaceNode) -> Wire:
         return self.dfg[node.place]
 
     def visit_GlobalName(self, node: GlobalName) -> Wire:
-        defn = self.ctx.build_compiled_def(node.def_id)
+        defn = ENGINE.get_checked(node.def_id)
+        if isinstance(defn, CallableDef) and defn.ty.parametrized:
+            # TODO: This should be caught during checking
```
acl-cqc, line 266: What's the TODO here? You mean that checking doesn't catch this yet??

guppylang/compiler/expr_compiler.py
```diff
@@ -270,8 +281,16 @@ def visit_GenericParamValue(self, node: GenericParamValue) -> Wire:
                 )
                 usize = self.builder.add_op(load_nat)
                 return self.builder.add_op(convert_ifromusize(), usize)
-            case _:
-                raise NotImplementedError
+            case ty:
+                # Look up monomorphization
```
acl-cqc, line 288: So I think you still need the case above (for NumericType.Kind.Nat) because the current mono-arg for that position could be `None` (and we are compiling a Hugr-polymorphic method). But if you went the whole way and didn't use Hugr polymorphism, we could actually just remove the old case above too and use this?

Might be worth adding a comment why we still need the above.

guppylang/decorator.py
```diff
@@ -380,7 +380,15 @@ def compile_function(
...
             raise TypeError(f"Object is not a Guppy definition: {obj}")
         compiled_module = ENGINE.compile(obj.id)
-        compiled_def = ENGINE.compiled[obj.id]
+
+        # Look up how many generic params the function has so we can create an empty
+        # partial monomorphization to look up in the context
+        checked_def = ENGINE.checked[obj.id]
+        mono_args: PartiallyMonomorphizedArgs | None = None
+        if isinstance(checked_def, MonomorphizableDef):
+            mono_args = tuple(None for _ in checked_def.params)
```
acl-cqc, line 390: (ALAN check use?) What does this mean we are going to compile? A Hugr definition polymorphic over all the args, including those guppy will actually monomorphize instead? Or is the entry in ENGINE some kind of proxy?

guppylang/compiler/core.py
```diff
@@ -232,6 +380,50 @@ def is_return_var(x: str) -> bool:
...
+    """Checks if a type parameter must be monomorphized before compiling to Hugr.
+
+    This is required for some Guppy language features that cannot be encoded in Hugr
+    yet. Currently, this only applies to non-nat const parameters.
+    """
+    match param:
+        case TypeParam():
+            return False
+        case ConstParam(ty=ty):
+            return not isinstance(ty, NumericType) or ty.kind != NumericType.Kind.Nat
```
acl-cqc: Ooof ;). A bit of DeMorgan might help here, or even `if isinstance(ty. NumericType): ... else ...` with some `return ty.kind == NumericType.Kind.Nat` and/or constants

guppylang/definition/common.py
```diff
@@ -155,6 +186,21 @@ def compile_inner(self, ctx: "CompilerContext") -> None:
...
 
 
+@dataclass(frozen=True)
+class MonomorphizedDef(CompiledDef):
+    """Abstract base class for definitions that have been added to a Hugr and have been
+    partially monomorphized.
+
+    Args:
+        id: The unique definition identifier.
+        name: The name of the definition.
```
acl-cqc: i.e. the same name for all monomorphizations of one original defn?

guppylang/definition/common.py
```diff
@@ -155,6 +186,21 @@ def compile_inner(self, ctx: "CompilerContext") -> None:
...
+        defined_at: The AST node where the definition was defined.
```
acl-cqc: the AST node of the original polymorphic definition?

guppylang/definition/custom.py
```diff
@@ -229,21 +235,31 @@ def load_with_args(
...
+                returns = self.compile_call(args, generic_ty_args, func_dfg, ctx, node)
+                func.set_outputs(*returns.regular_returns, *returns.inout_returns)
 
         # Finally, load the function into the local DFG
         mono_ty = self.ty.instantiate(type_args).to_hugr(ctx)
-        hugr_ty_args = [arg.to_hugr(ctx) for arg in type_args]
+        hugr_ty_args = [
+            arg.to_hugr(ctx)
+            for param, arg in zip(self.ty.params, type_args, strict=True)
+            if not requires_monomorphization(param)
```
acl-cqc: Ah-ha, yes, this might be a good thing to build in `partially_monomorphize_args` as you go

guppylang/definition/declaration.py
```diff
@@ -51,6 +66,10 @@ def parse(self, globals: Globals, sources: SourceMap) -> "CheckedFunctionDecl":
         ty = check_signature(func_ast, globals)
         if not has_empty_body(func_ast):
             raise GuppyError(BodyNotEmptyError(func_ast.body[0], self.name))
+        # Make sure we won't need monomorphization to compile this declaration
```
acl-cqc, line 69: What kind of error does this correspond to? If the guppy-user has written a function that will need monomorphization, how does the compiler *not* get here? (Should this be an internal error rather than a user-facing error?)

guppylang/engine.py
```diff
@@ -201,10 +205,7 @@ def compile(self, id: DefId) -> ModulePointer:
         from guppylang.compiler.core import CompilerContext
 
         ctx = CompilerContext(graph)
-        # Iterate over copy of the checked values since in some cases, compiling can
-        # kick of checking calls that would change the size of `self.checked`
-        for defn in list(self.checked.values()):
-            ctx.compile(defn)
+        ctx.compile(self.checked[id])
```
acl-cqc, line 208: I would say that `CompilerContext.compile(....)` is now the main driver (where you go through the worklist)

guppylang/tys/param.py
```diff
@@ -212,8 +216,14 @@ def to_hugr(self, ctx: ToHugrContext) -> ht.TypeParam:
             case NumericType(kind=NumericType.Kind.Nat):
                 return ht.BoundedNatParam(upper_bound=None)
             case _:
-                assert isinstance(self.ty.to_hugr(ctx), ht.ExtType)
-                return ht.StringParam()
+                raise InternalGuppyError(
+                    "Tried to convert non-nat const type parameter to Hugr. This "
+                    "should have been erased."
```
acl-cqc:
```suggestion
                    "should have been removed by monomorphization."
```
This is very much not type erasure ;)

tests/error/poly_errors/polymorphic_decl.py
```diff
@@ -0,0 +1,17 @@
+from typing import Generic
+
+from guppylang.decorator import guppy
+
+F = guppy.const_var("F", "float")
+
+
+@guppy.struct
+class Struct(Generic[F]):
```
acl-cqc, line 9: Argh.....I mean, how is a struct possibly polymorphic over a float!?

I guess we can't rule that out, or at least, it wouldn't be consistent. But it does make for something of a confusing test.

tests/error/poly_errors/polymorphic_entrypoint.py
```diff
@@ -0,0 +1,18 @@
...
+
+
+@guppy.struct
+class Struct(Generic[F]):
+    pass
+
+
+@guppy
+def main(_: Struct[F]) -> float:
+    return F
```
acl-cqc, line 15: The comparison with previous seems a bit odd. Here we replace the `...` with a body, which looks valid to me (?). The error message doesn't seem to relate to that - surely
``Function declaration `main` is not allowed to be generic over `F : float` `` is still true and is still an error?

guppylang/compiler/core.py
```diff
@@ -68,70 +114,172 @@ def __init__(
...
+        if (def_id, ()) in self.compiled:
+            return self.compiled[def_id, ()]
+
+        defn = ENGINE.get_checked(def_id)
+        mono_args: PartiallyMonomorphizedArgs | None = None
+        if isinstance(defn, MonomorphizableDef):
+            assert type_args is not None
+            mono_args = partially_monomorphize_args(defn.params, type_args, self)
+        if (def_id, mono_args) not in self.compiled:
+            self.compiled[def_id, mono_args] = self._compile(defn, mono_args)
```
acl-cqc: I'm a bit suspicious of this `_compile`. A few options
* Include the worklist-pushing in there. I'm not sure if you can include the already-in-`compiled` check too
* rename it to `_compile_outer` - it took a certain amount of digging/waiting-until-end-of-review to see why recursion woould work ;-).
* Inline `_compile` - the two possibilities in there are always separated by an immediately preceding `if`. So we could do:
```
if isinstance(defn, MonomorphizableDef):
   assert type_args is not None
   mono_args = partially_monomorphize_args(defn.params, type_args, self)
   if (def_id, mono_args) not in self.compiled:
      self.compile[def_id, mono_args] = defn.monomorphize(self.module, mono_args, self)
else:
   assert type_args is None # Better than asserting in self._compile: the error originates above here
   mono_args = None # So, no need to initialize above
   if (def_id, args) not in self.compiled:
      self.compiled[def_id, args] = defn.compile_outer(self.module, self) if isinstance(defn, CompilableDef) else defn
self.worklist[def_id, mono_args] = None
return self.compiled[def_id, mono_args]
```

in fact I'd be quite tempted to use a higher-order function...
```
compile_outer: Callable[[], CompiledDef]
if isinstance(defn, MonomorphizableDef):
  assert type_args is not None
  mono_args = partially_monomorphize_args(defn.params, type_args, self)
  compile_outer = lambda: defn.monomorphize(self.module, mono_args, self)
else:
  assert type_args is None
  mono_args = None
  compile_outer = lambda: defn.compile_outer(self.module, mono_args, self) if isinstance(defn, CompilableDef) else defn
if (def_id, mono_args) not in self.compiled:
  self.compiled[def_id, mono_args] = compile_outer()
  self.worklist[def_id, mono_args] = None
return self.compiled[def_id, mono_args]
```
* how about changing `CompilableDef::compile_outer` to take optional type-args; existing impls of that method assert that they are None; and then `MonomorphizableDef` is a subclass of `CompilableDef` (so `monomorphize` *is* just `compile_outer`, and has an assert that args are not None)? I think perhaps I like that the best....

 (And I think I've previously suggested changing CheckableDef to implement a default-return-self `compile_outer` and renaming `CompilableDef` to `DefNeedsCompiling`....)

guppylang/compiler/core.py
```diff
@@ -68,70 +114,172 @@ def __init__(
...
-        if def_id not in self.compiled:
-            defn = ENGINE.get_checked(def_id)
-            self.compiled[def_id] = self._compile(defn)
-            self.worklist[def_id] = None
-        return self.compiled[def_id]
-
-    def _compile(self, defn: CheckedDef) -> CompiledDef:
+        # TODO: The check below is a hack to support nested function definitions. We
+        #  forgot to insert frames for nested functions into the DEF_STORE, which would
+        #  make the call to `ENGINE.get_checked` below fail. For now, let's just short-
```
acl-cqc, line 169: This doesn't seem a bad shortcut to take anyway TBH.... and whilst it's a shame that any shortcut is necessary, this one is mercifully short!

guppylang/compiler/core.py
```diff
@@ -68,70 +114,172 @@ def __init__(
...
+        #  nested functions).
+        #  See https://github.com/CQCL/guppylang/issues/1032
+        if (def_id, ()) in self.compiled:
+            return self.compiled[def_id, ()]
+
+        defn = ENGINE.get_checked(def_id)
+        mono_args: PartiallyMonomorphizedArgs | None = None
+        if isinstance(defn, MonomorphizableDef):
+            assert type_args is not None
+            mono_args = partially_monomorphize_args(defn.params, type_args, self)
```
acl-cqc: Hmmmm, so in `partially_monomorphize_args` I suggested you add an extra return value (the args which were dropped because `requires_monomorphization` was false) - if one had those here, could one assert that they were all none (because this is a generic decl over those args)? I think that would be a good thing to do to define what the contract of `build_compiled_def` is....

guppylang/compiler/core.py
```diff
@@ -68,70 +114,172 @@ def __init__(
...
+        type_args: Inst | None,
     ) -> CompiledCallableDef | None:
         from guppylang.engine import ENGINE
 
         parsed_func = self.checked_globals.get_instance_func(ty, name)
         if parsed_func is None:
             return None
         checked_func = ENGINE.get_checked(parsed_func.id)
-        compiled_func = self.build_compiled_def(checked_func.id)
+        compiled_func = self.build_compiled_def(checked_func.id, type_args)
```
acl-cqc: I am a little nervous here - this means `get_instance_func` is returning a CompiledCallableDef that will still need to be given *some* type args (a subset of the original ones), yet does not tell its caller which ones (we rely on the caller of `get_instance_func` performing the same partial-monomorphization logic to determine which). Noting there are at least two completely-separate `get_instance_func` methods 👎, can we change signature of this to return more info (like, the remaining type-args that still need to be passed)?

guppylang/definition/custom.py
acl-cqc: Or if not, then at least, common up in `CompiledFunctionDef` in a helper method alongside+used-by `load_with_args` and the `compile_call` there

guppylang/compiler/expr_compiler.py
```diff
@@ -642,9 +661,10 @@ def _build_method_call(
         args: list[Wire],
         type_args: Sequence[Argument] | None = None,
     ) -> CallReturnWires:
-        func = self.ctx.get_instance_func(ty, method)
+        type_args = type_args or []
+        func = self.ctx.get_instance_func(ty, method, type_args)
         assert func is not None
-        return func.compile_call(args, type_args or [], self.dfg, self.ctx, node)
+        return func.compile_call(args, type_args, self.dfg, self.ctx, node)
```
acl-cqc: This is the case I'm thinking of - if `self.ctx.get_instance_func` returned the *remaining* args still required and you passed them into `func.compile_call` so it didn't need to filter/etc. itself - does that mess up the other uses?

guppylang/compiler/core.py
```diff
@@ -68,70 +114,172 @@ def __init__(
...
+    def const_var_to_hugr(self, var: BoundConstVar) -> ht.TypeArg:
+        """Compiles a bound Guppy constant variable into a Hugr type argument.
+
+        Takes care of performing partial monomorphization as specified in the current
+        context.
+        """
+        if var.ty != nat_type():
+            raise InternalGuppyError(
+                "Tried to convert non-nat const type argument to Hugr. This should "
+                "have been erased."
```
acl-cqc:
```suggestion
                "have been removed by monomorphization."
```
*not* type erasure (that would mean, runtime behaviour does not depend upon the type!)

guppylang/compiler/core.py
```diff
@@ -68,70 +114,172 @@ def __init__(
...
+
+        match self.current_mono_args[var.idx]:
+            # Either we have a decided to monomorphize the corresponding parameter...
+            case ConstArg(const=ConstValue(value=int(v))):
+                return ht.BoundedNatArg(n=v)
+            # ... or we're still want to be generic in Hugr
+            case None:
+                # But in that case we'll have to down-shift the de-Bruijn index to
+                # account for earlier params that have been monomorphized away
+                down_shift = sum(1 for arg in self.current_mono_args[: var.idx] if arg)
```
acl-cqc: as previous re. subtraction, might be worth commoning up the "Downshifting" logic

acl-cqc (commented):
Thanks @mark-koch, seriously meaty PR here but I hope I have digested it successfully - good work :). A few thoughts in summary:
* Is it really worth using *both* guppy and Hugr monomorphization? I think the complexity of doing such (updating "deBruijn" indices, etc.) is quite high - wouldn't it be much easier just to do everything in guppy? What are the counterarguments?
   * The main one I can see is maybe it reduces change/surprise if/when Hugr is updated such that we don't need monomorphization to be done in guppy after all. Reduces *a bit*, in that with the hybrid approach we would only be switching how non-nat/not-type args are monomorphized, rather than all arguments. But the nat/type cases seem like the easy/predictable ones (right?)....I mean, even if we monomorphize nats in guppy, we don't *need* to replace `loadnat<55>` with `loadconstant(Const(55))`.
* However, I guess I'm willing to accept the idea that partial monomorphization might be the best way for ease of reversion...
* Indexing worklist/compilation store/`global_funcs` by `DefId` and (`Partially``MonomorphizedArgs sounds like the right thing to do either way
* There are some funny things going on with error messages
* I made a couple of suggestions of refactoring away MonomorphizableDef and `_compile`, also a lot of places seem to rely on using the same technique encoded in `partially_monomorphize_args` - is it possible to process the type args once and pass the results from that around, rather than rely on many places all (remembering to) do the same processing? (This is the idea of returning the stripped args as a second result from `partially_monomorphize_args` )
* I'm not clear on empty-tuple PartialMonomorphizedArgs vs None, I think this is probably just me, but would be good to spell out (as an "illustrative example" perhaps or just an "e.g.")

guppylang/compiler/core.py
mark-koch: `()` means it's a kind of definition that can be monomorphised (instance of `MonomorphizableDef`), but it happens to have no generic parameters (e.g. a non-generic function definition). `None` is for definitions that are not instances of `MonomorphizableDef` (e.g. constant definitions).

Does this distinction make sense?

guppylang/compiler/core.py
mark-koch: > I am wondering whether the case ConstArg can occur if requires_monomorphization(param) is false

Yes it can occur: `nat` const generics are not monomorphized, so will have `requires_monomorphization == False` but are still instances of `ConstArg`. So your proposed

```python
assert inst is not None # is this safe ???
```

would not be safe I think.

guppylang/compiler/core.py
mark-koch: Yes, but actually thinking about it, this will change if we implement Guppy type classes via monomorphisation before we have proper Hugr support.

I'll refactor a bit and add an explicit check to make sure we correctly handle this later

guppylang/compiler/expr_compiler.py
mark-koch: Oops this is not a new check, I just moved it and forgot to delete the original.

> You mean that checking doesn't catch this yet??

Yes, ideally there should be no `GuppyError`s thrown during Hugr compilation, but there are still a few places around where this stuck around in for convenience...

guppylang/compiler/expr_compiler.py
mark-koch: Correct! I'll add a comment

tests/error/poly_errors/polymorphic_decl.py
mark-koch: Agreed, the main confusion comes from the fact that we don't support the Python 3.12 style

```python
@guppy.declare
def main[F: float]() -> float: ...
```

syntax yet (coming in #1051). Therefore, for now the only way to introduce a generic parameter into a signature is by actually *using* the parameter in the signature. To do that, we need a type that is generic over a float - hence the struct definition.

guppylang/definition/declaration.py
mark-koch: The point is that we disallow users from calling `guppy.compile` on definitions that would require monomorphization (e.g. a function that uses non-nat const generics).

Also see https://github.com/CQCL/guppylang/pull/1035/files/890d960ccc12322be12cfe9ca3922bc217159f46#r2166139634 for the example

mark-koch: The user would need to give us the concrete generic instantiation they want to compile, but that would make the `guppy.compile` API more difficult...

tests/error/poly_errors/polymorphic_entrypoint.py
mark-koch: This is the same point as https://github.com/CQCL/guppylang/pull/1035#discussion_r2166126805

When compiling `main`, we'd definitely have to monomorphize because we couldn't represernt the `F: float` parameter in Hugr. But when we call `guppy.compile(main)` below, there is no way for us to know what value we should pick for `F`.

I'd be happy for suggestions to make this error message easier to understand

guppylang/tys/param.py
mark-koch: Hmm I'd argue it is: If we have a type

```
F = guppy.const_var("F", "float")

@guppy.struct
class Struct(Generic[F]):
    pass
```

then the corresponding Hugr type is `()` (Unit) - the float const generic paramater has been erased. But maybe I'm misusing terminology here, happy to change it if you think it would be clearer otherwise?

guppylang/compiler/core.py
mark-koch: > (a subset of the original ones)

No, we're currently always passing *all* of the type args to `compile_call` and friends. We never filter type args at the call site, only inside `compile_call`

guppylang/compiler/core.py
mark-koch: >  if one had those here, could one assert that they were all none (because this is a generic decl over those args)?

I don't think so, or maybe I misunderstand what you're saying? Here `defn.params` can contain parameters that requires monomorphization and ones that don't, so it's not guaranteed that we would get all `None`s?

guppylang/definition/custom.py
mark-koch: I think it's tricky to carry this information all the way through. I went with a helper that builds this based on the current `PartiallyMonomorphizedArgs`. Does this seem reasonable?

guppylang/compiler/core.py
mark-koch: > how about changing CompilableDef::compile_outer to take optional type-args [...]

Mhmm I personally prefer the distinction between definitions that involve monomorphization and ones that don't, i.e. that it's impossible to call `monomorphize` without providing `mono_args`, but I don't feel super strongly.

I like your higher-order function suggestion, went with that for now. What do you think?

guppylang/decorator.py
mark-koch: The entry is a (potentially generic) function where none of the generic parameters require monomorphization. Thus, the partial monomorphization is empty - i.e. a tuple of `None`s.

Added a comment to clarify

guppylang/compiler/core.py
```diff
@@ -14,18 +14,52 @@
...
+#:
+#: When compiling polymorphic definitions to Hugr, some of their generic parameters will
+#: need to be monomorphized (e.g. to support language features that cannot be encoded
+#: in Hugr, see `requires_monomorphization` for details). However, other generic
+#: parameters can be faithfully captured in Hugr. Thus, we want to perform a *partial*
+#: monomorphization when compiling to Hugr.
+#:
+#: A `mono_args: PartiallyMonomorphizedArgs` represents such a partial monomorphization
+#: of generic parameters as a sequence of `Argument | None`. Concretely,
+#: `mono_args[i] == None` means that the argument with index `i` is not part of the
```
acl-cqc:
```suggestion
#: `mono_args[i] == None` means that the argument with index `i` is left as a polymorphic Hugr type parameter.
```

guppylang/compiler/core.py
acl-cqc: Maybe even more concrete - say "no monomorphization is required at all" -> "all of kind Type handled as polymorphic in Hugr"

guppylang/compiler/core.py
acl-cqc: Possibly. I wonder whether making non-generic functions *not* be MonomorphizableDef might be better, but I suspect that that would be odd elsewhere in the compiler and it is def. better to be consistent. I am not sure whether it makes sense to use `None` for constant definitions rather than an empty list; presumably there are some checks that when we have a constant the PartiallyMonomorphisedArgs are None already and we could update those to check against `()` rather than `None`...

guppylang/compiler/core.py
```diff
@@ -232,6 +415,79 @@ def is_return_var(x: str) -> bool:
...
+    """
+    assert mono_args is None or len(mono_args) == len(inst)
+    return [
+        arg.to_hugr(ctx)
+        for i, arg in enumerate(inst)
+        if mono_args is None or mono_args[i] is None
+    ]
+
+
+def compile_de_bruijn_idx(idx: int, mono_args: PartiallyMonomorphizedArgs) -> int:
```
acl-cqc: Good function, good comment. But we should probably not use the words "de bruijn" - they are just indices; de Bruijn only applies when there are nested binding sites, whereas in Hugr all type vars in scope are bound at the same site. (Quite a few of these up higher above)

Suggestions
* `compute_partially_monomorphized_index` - too long
* `hugr_var_idx`
* `(compute_)partial_(var_)idx`
* `downshift_index`/idx - not the most obvious, but establishes our own term which if we're consistent about it could work well

guppylang/compiler/expr_compiler.py
```diff
@@ -642,9 +657,10 @@ def _build_method_call(
         args: list[Wire],
         type_args: Sequence[Argument] | None = None,
     ) -> CallReturnWires:
-        func = self.ctx.get_instance_func(ty, method)
+        type_args = type_args or []
```
acl-cqc: IIUC, might be worth commenting "convert to PartiallyMonomorphizedArgs" or similar, i.e. we need to do this because `type_args` and `PartiallyMonomorphizedArgs` use `None` for different purposes...

guppylang/compiler/core.py
acl-cqc: Ok, I'm confused here. I think things would be failing obviously if they weren't right, but perhaps a documentation issue. `build_compiled_def` returns something from `self.compiled[id, mono_args]` where `mono_args` are the partially-monomorphized version of `type_args` that we're passing in here. For example, a  `MonomorphizableDef.monomorphize(...mono_args...)`. So *that* thing still takes the entire set of args, including those that have already been baked into it (from `mono_args`) as well as those that are still needed? And we then apply the same policy to the entire set of args again to compute which of those are still needed as Hugr TypeArgs?

guppylang/compiler/core.py
```diff
@@ -43,6 +77,35 @@ def name(self) -> str:
...
+#: should be `None`.
+#:
+#: Note the following subtlety: Functions are instances of `MonomorphizableDef`, even if
+#: they aren't actually generic! This means that for non-generic function definitions we
+#: will have an empty tuple `()` as `PartiallyMonomorphizedArgs`, but never `None`.
+#: `None` only shows up for kinds of definitions that are never monomorphized (e.g.
+#: definitions of constants).
+MonoDefId = tuple[DefId, PartiallyMonomorphizedArgs | None]
+
+#: Unique identifier for global Hugr constants and functions with optional monomorphized
```
acl-cqc: Hang on. Three lines up we say that definitions of constants are never monomorphized. So this parses as "Unique identifier for (global Hugr constants) and (functions with optional monomorphized type arguments)"? Lets write "and functions, the latter with optional...."

acl-cqc: Does look quite similar to the previous I have to say

guppylang/compiler/core.py
acl-cqc: Is it a possibility for `get_instance_func` and `build_compiled_def` to return, for a function that takes type arguments, not just the MonomorphizedDef (i.e. the guppy-time monomorphization), but a *wrapper* (not a member per se - something transient) around that that inserts only the Hugr type arguments? Then you actually would not need to pass the type arguments to `compile_call` (I mean, this does seem redundant, given you passed them all to get_intsance_func?)....??

guppylang/compiler/expr_compiler.py
acl-cqc: Or maybe you split the args up here, rather than sending `type_args` into two different directions which then separately transform them in two different ways that later have to agree

guppylang/compiler/expr_compiler.py
mark-koch: Good point. Actually, I think it's better to be explicit and just get rid of the `None` case here

guppylang/compiler/core.py
acl-cqc: Nice! :)

guppylang/compiler/core.py
acl-cqc: I think the distinction makes sense. I was wondering whether switching to empty list rather than None for non-generic things (like we just did to `build_compiled_def`) would be better, but TBH I don't see any need to, I think it is fine for this to work differently than the interface to `build_compiled_def` (rather, this follows the interface for `partially_monomorphize_args`).

So feel free to hit "resolve" :)

acl-cqc (approved):
Looks great, thanks @mark-koch. Thought the "pair programming" worked quite well too ;).

Only one remaining suggestion - search for "been erased" - there are about 3 hits - find a different terminology that does not suggest "type erasure", e.g. "monomorphized away" / "removed by monomorphization" / *something* :)

PR #1038, opened 2025-06-17
feat: Allow indexing on tuples

### LINKED ISSUE #711: Allow indexing on tuples
current state:

```
Error: Not subscriptable (at <In [?]>:8:8)
  | 
7 |     a = (1, 2)
8 |     b = a[0]
  |         ^ Expression of type `(int, int)` is not subscriptable
```

tatiana-s:
Closes #711


guppylang/checker/linearity_checker.py
```diff
@@ -426,6 +430,17 @@ def visit_SubscriptAccessAndDrop(self, node: SubscriptAccessAndDrop) -> None:
         self.scope.assign(node.item)
         self.visit(node.getitem_expr)
 
+    def visit_TupleAccessAndDrop(self, node: TupleAccessAndDrop) -> None:
+        # A tuple access on a value that is not a place. This means the value can no
+        # longer be accessed after the element has been projected out. Thus, this is
+        # only legal if there are no remaining linear elements in the tuple
+        self.visit(node.value)
+        for elem_ty in node.tuple_ty.element_types:
+            if not elem_ty.droppable:
```
mark-koch: This will also error if only the projected index is non-droppable. But that should be allowed:

```python
q = (1, 2, qubit())[2]  # Ok!
```

guppylang/compiler/core.py
```diff
@@ -163,19 +170,35 @@ def __getitem__(self, place: Place) -> Wire:
...
+                FieldAccess(place, field, None) for field in place.ty.fields
+            ]
+            child_types = [child.ty.to_hugr() for child in field_children]
+            child_wires = [self[child] for child in field_children]
+            wire = self.builder.add_op(ops.MakeTuple(child_types), *child_wires)[0]
+            for f_child in field_children:
+                if f_child.ty.linear:
+                    self.locals.pop(f_child.id)
+            self.locals[place.id] = wire
+            return wire
```
mark-koch: Is there a way to refactor this to avoid some of the code duplication with the struct case below? Maybe something like

```python
if isinstance(ty, TupleType):
    children = [
        TupleAccess(place, elem, idx, None)
        for idx, elem in enumerate(place.ty.element_types)
    ]
elif isinstance(ty, StructType):
    children = [
        FieldAccess(place, field, None) for field in place.ty.fields
    ]
else:
    raise InternalGuppyError(f"Couldn't obtain a port for `{place}`")
child_types = ...
# ...
```

tests/integration/test_tuples.py
mark-koch, line 1: Could you add some more tests for linearity checking across control-flow and involving borrows similar to the struct tests in `test_inout.py`. E.g.

```python
t = (qubit(), qubit(), 0)
if b:
    h(t[0])
else:
    h(t[1])
discard(t[0])
discard(t[1])
```

```python
ts = array((qubit(), qubit())
h(ts[0][0])
h(ts[0][1])
# cx(ts[0][0], ts[0][1])  # I think this would fail at runtime?
for (q1, q2) in ts:
    discard(q1)
    discard(q2)
```

```python
t = (qubit(), qubit(), 0)
if b:
    q = mem_swap(t[0], qubit())
    discard(q)
else:
    q = mem_swap(t[1], qubit())
    discard(q)
h(t[0])
discard(t[0])
discard(t[1])
```

```python
t = (qubit(), qubit(), 0)
while True:
    q = mem_swap(t[0], qubit())
    discard(q)
# No discard needed because unreachable
```

tatiana-s: The last one only works when passed to the function, but fields have the same issue so https://github.com/CQCL/guppylang/issues/1052

mark-koch (approved):
Thanks, looks great!

PR #1039, opened 2025-06-17
feat: Add guppy version metadata to hugr entrypoint

### LINKED ISSUE #1037: Put guppy version in the metadata of the generated hugr

croyzor:
Resolves #1037

### DIFF
--- a/guppylang/engine.py
+++ b/guppylang/engine.py
@@ -238,7 +238,7 @@ class CompilationEngine:
             for ext in all_used_extensions
         ]
         graph.hugr.module_root.metadata["__generator"] = {
-            "name": "guppy",
+            "name": "guppylang",
             "version": guppylang.__version__,
         }
         return ModulePointer(Package(modules=[graph.hugr], extensions=extensions), 0)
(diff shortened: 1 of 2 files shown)


croyzor:
> Why at the entrypoint rather than the `module_root`? This should be common to the full hugr

I thought it was the same thing 😅 I'll update it

croyzor:
> Why at the entrypoint rather than the `module_root`? This should be common to the full hugr

This is where metadata gets written to [by default](https://cqcl.github.io/hugr/generated/hugr.build.function.Module.html#hugr.build.function.Module.metadata) in the hugr-py api

guppylang/engine.py
```diff
@@ -196,6 +196,7 @@ def compile(self, id: DefId) -> ModulePointer:
         # Prepare Hugr for this module
         graph = hf.Module()
         graph.metadata["name"] = "__main__"
+        graph.metadata["guppy_version"] = guppylang.__version__
```
ss2165: We should use a generic field like 
```suggestion
        graph.metadata["generator"] = f"guppylang@{guppylang.__version__}"
```
so the hugr tooling can report it with error messages etc.

ss2165: might be using some indication that this is a reserved metadata keyword with "__generator" or the like

PR #1045, opened 2025-06-23
feat: Add manual registration of extensions

jake-arkinstall:
At present we have the following in CompilationEngine:

```
        # TODO: Currently the list of extensions is manually managed by the user.
        #  We should compute this dynamically from the imported dependencies instead.
        #
        # The hugr prelude and std_extensions are implicit.
        from guppylang.std._internal.compiler.tket2_exts import TKET2_EXTENSIONS

        extensions = [*TKET2_EXTENSIONS, guppylang.compiler.hugr_extension.EXTENSION]
```

This PR does _not_ achieve this. Instead, it provides an intermediate step that allows extensions to be registered manually.

This can be done by passing an `Extension` to `ENGINE.register_extension()`.

### DIFF
--- a/guppylang/engine.py
+++ b/guppylang/engine.py
@@ -3,6 +3,7 @@ from types import FrameType
 
 import hugr.build.function as hf
 from hugr.package import ModulePointer, Package
+from hugr.ext import Extension
 
 import guppylang
 from guppylang.definition.common import (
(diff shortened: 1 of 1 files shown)


tatiana-s (approved):

guppylang/engine.py
```diff
@@ -107,9 +109,15 @@ def reset(self) -> None:
         """Resets the compilation cache."""
         self.parsed = {}
         self.checked = {}
+        self.compiled = {}
```
jake-arkinstall, line 117: note to reviewers: this is a drive-by change, as the original omission of a `self.compiled` reset appeared accidental.

Please check whether or not this is the case. I don't want to cause a performance regression if resetting the compiled dictionary is avoided for e.g. caching.

PR #1047, opened 2025-06-23
chore: Use tket2_exts ops instead of custom ops

### LINKED ISSUE #1029: Use results op from `tket2_exts` instead hardcoding it as `Custom` ops

tatiana-s:
Closes #1029

### DIFF
--- a/guppylang/compiler/expr_compiler.py
+++ b/guppylang/compiler/expr_compiler.py
@@ -1,5 +1,5 @@
 import ast
-from collections.abc import Iterable, Iterator, Sequence
+from collections.abc import Iterator, Sequence
 from contextlib import AbstractContextManager, ExitStack, contextmanager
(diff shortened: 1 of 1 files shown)


guppylang/compiler/expr_compiler.py
```diff
@@ -554,12 +556,17 @@ def visit_ResultExpr(self, node: ResultExpr) -> Wire:
...
-            extra_args=extra_args,
+
+        sig = ht.FunctionType(
+            input=[hugr_ty],
+            output=[],
         )
+        args = [
+            ht.StringArg(node.tag),
+            *extra_args,
+        ]
```
mark-koch: ultra nit: We could clean up the formatting a bit as a drive-by

```suggestion
        sig = ht.FunctionType(input=[hugr_ty], output=[])
        args = [ht.StringArg(node.tag), *extra_args]
```

PR #1051, opened 2025-06-24
feat: Support Python 3.12 generic syntax

### LINKED ISSUE #823: Support Python 3.12 generic syntax
```python
def foo[T: Copyable](x: T) -> T: ...
```

mark-koch:
Closes #823

### DIFF
--- a/guppylang/tys/parsing.py
+++ b/guppylang/tys/parsing.py
@@ -1,4 +1,5 @@
 import ast
+import sys
 from collections.abc import Sequence
 from types import ModuleType
 
@@ -30,6 +31,7 @@ from guppylang.tys.errors import (
     InvalidTypeArgError,
     InvalidTypeError,
     LinearComptimeError,
+    LinearConstParamError,
     ModuleMemberNotFoundError,
     NonLinearOwnedError,
 )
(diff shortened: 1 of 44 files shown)


tests/error/poly_errors_py312/free_var_in_bound2.err
```diff
@@ -0,0 +1,11 @@
+Error: Free type variable (at $FILE:7:18)
+  | 
+5 | 
+6 | @guppy.struct
+7 | class MyStruct[x: T]:
+  |                   ^ Type variable `T` is unbound
+
+Note: Only struct and function definitions can be generic. Other generic values
+or nested types are not supported.
```
croyzor, line 9: he real error here is that `T` isn't `nat` or a mix of type bounds though, right?

guppylang/tys/parsing.py
```diff
@@ -279,6 +281,58 @@ def parse_function_io_types(
...
+                return TypeParam(
+                    idx, node.name, must_be_copyable=True, must_be_droppable=False
+                )
+            case ast.Name(id="Drop"):
+                return TypeParam(
+                    idx, node.name, must_be_copyable=False, must_be_droppable=True
+                )
+            # Copy and drop is annotated as `T: (Copy, Drop)`
+            # TODO: Should we also allow `T: Copy + Drop`? Mypy would complain about it
+            case ast.Tuple(elts=[ast.Name(id=id1), ast.Name(id=id2)]) if {id1, id2} == {
```
croyzor, line 309: nice

guppylang/tys/parsing.py
```diff
@@ -279,6 +281,58 @@ def parse_function_io_types(
...
+                )
+            # Otherwise, it must be a const parameter
+            case bound:
+                # For now, we don't allow the types of const params to refer to previous
+                # parameters, so we pass an empty dict as the `param_var_mapping`.
+                # TODO: In the future we might want to allow stuff like
+                #   `def foo[T, XS: array[T, 42]]` and so on
+                ty = type_from_ast(bound, globals, {}, allow_free_vars=False)
+                if not ty.copyable or not ty.droppable:
+                    raise GuppyError(LinearConstParamError(bound, ty))
```
croyzor, line 324: I think this check/error just confuses things - for now, at least - I'd get rid of it

croyzor (approved):

tests/error/poly_errors_py312/free_var_in_bound2.err
mark-koch: Hmm I don't want to mention `nat` since that's only a temporary restriction until the monomorphization PR is merged. 

In principle, we should allow every copyable&droppable type here, so this message is coming from trying to parse a type and encountering a type variable that is not declared as a parameter. Not sure if there is an easy way to improve this

Issue #1056, opened 2025-06-25
Add support for measure leaked


narmlu:
Add measure leaked qsystem op.

Tracking issue: https://github.com/quantinuum-dev/hugrverse/issues/127

PR #1057, opened 2025-06-25
feat: add qsystem op for measure leaked

narmlu:
Closes: #1056

### DIFF
--- a/guppylang/std/qsystem/__init__.py
+++ b/guppylang/std/qsystem/__init__.py
@@ -107,3 +107,8 @@ def _rz(q: qubit, angle: float) -> None:
     See `guppylang.std.qsystem.rz` for a public definition that
     accepts angle parameters.
     """
+
+@guppy.hugr_op(quantum_op("LazyMeasureLeaked", ext=QSYSTEM_EXTENSION))
+@no_type_check
+def measure_leaked(q: qubit @ owned) -> int:
+    """Measure the quibit or return 2 if it is leaked."""
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -88,7 +88,7 @@ miette-py = { workspace = true }
 
 # Uncomment these to test the latest dependency version during development
 # hugr = { git = "https://github.com/CQCL/hugr", subdirectory = "hugr-py", rev = "7f233fa" }
-# tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", rev = "652a7d0" }
+tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", branch = "905-measure-leaked" }
 # tket2 = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-py", rev = "fcb2131" }
 
 
--- a/tests/integration/test_qsystem.py
+++ b/tests/integration/test_qsystem.py
@@ -4,6 +4,7 @@ from guppylang.decorator import guppy
 from guppylang.std.angles import angle
 from guppylang.std.builtins import owned, array
 from guppylang.std.qsystem.random import make_discrete_distribution, RNG
+
 from guppylang.std.qsystem.utils import get_current_shot
 from guppylang.std.quantum import qubit, measure_array
 from guppylang.std.qsystem.functional import (
@@ -57,5 +58,16 @@ def test_qsystem_random(validate):  # type: ignore[no-untyped-def]
         rng.discard()
 
         return rint, rfloat, rint_bnd, rint_discrete, rangle, rcangle
+
+    validate(guppy.compile(test))
+
+def test_measure_leaked(validate):  # type: ignore[no-untyped-def]
+    """Compile the measure_leaked operation."""
+
+    @guppy
+    def test(q: qubit @ owned) -> bool:
+        q = qubit()
+        m = measure_leaked(q)
+        return m
 
     validate(guppy.compile(test))
--- a/uv.lock
+++ b/uv.lock
@@ -633,7 +633,7 @@ requires-dist = [
     { name = "networkx", specifier = ">=3.2.1,<4" },
     { name = "pydantic", specifier = ">=2.7.0b1,<3" },
     { name = "pytket", marker = "extra == 'pytket'", specifier = ">=1.34" },
-    { name = "tket2-exts", specifier = "~=0.9.1" },
+    { name = "tket2-exts", git = "https://github.com/CQCL/tket2?subdirectory=tket2-exts&branch=905-measure-leaked" },
     { name = "typing-extensions", specifier = ">=4.9.0,<5" },
 ]
 provides-extras = ["pytket"]
@@ -2648,13 +2648,9 @@ wheels = [
 [[package]]
 name = "tket2-exts"
 version = "0.9.1"
-source = { registry = "https://pypi.org/simple" }
+source = { git = "https://github.com/CQCL/tket2?subdirectory=tket2-exts&branch=905-measure-leaked#1fb17989f380de01b6337999ae3c7d85d7df6f21" }
 dependencies = [
     { name = "hugr" },
-]
-sdist = { url = "https://files.pythonhosted.org/packages/1e/e3/d00691b4d0754bdde56c13822e326fb89fc13ac45cfbb1ae358b24a4f406/tket2_exts-0.9.1.tar.gz", hash = "sha256:31ff8cf833545ac2de466446180581aacd010e298538e49ece0c517b3ad5dfb9", size = 12410 }
-wheels = [
-    { url = "https://files.pythonhosted.org/packages/1a/e0/0c8477db1aab499cef96bee9761c138b9d634f7b4826317c77a5273b475e/tket2_exts-0.9.1-py3-none-any.whl", hash = "sha256:5aea9fa9fbdeded4bc93b7e87f4a9ff56b348ef6c961e11dd9e1163c78880aa6", size = 18282 },
 ]
 
 [[package]]


tests/integration/test_qsystem.py
```diff
@@ -59,3 +60,14 @@ def test() -> tuple[int, float, int, int, angle, angle]:
         return rint, rfloat, rint_bnd, rint_discrete, rangle, rcangle
 
     validate(guppy.compile(test))
+
+def test_measure_leaked(validate):  # type: ignore[no-untyped-def]
+    """Compile the measure_leaked operation."""
+
+    @guppy
+    def test(q: qubit @ owned) -> bool:
```
ss2165, line 71:
```suggestion
    def test(q: qubit @ owned) -> int:
```
?

guppylang/std/qsystem/__init__.py
```diff
@@ -107,3 +107,9 @@ def _rz(q: qubit, angle: float) -> None:
     See `guppylang.std.qsystem.rz` for a public definition that
     accepts angle parameters.
     """
+
+
+@guppy.hugr_op(quantum_op("LazyMeasureLeaked", ext=QSYSTEM_EXTENSION))
+@no_type_check
+def measure_leaked(q: qubit @ owned) -> int:
```
ss2165: Do we want to a more structured type returned that wraps the integer? A struct with `to_measure_result() -> Option[bool]` and `is_leaked() -> bool` methods?

tests/integration/test_qsystem.py
narmlu: Yeah this should be `int`, thanks.

guppylang/std/qsystem/__init__.py
narmlu: Is it possible (like physics-wise) to detect leakage without actually measuring?

If yes, then one could use `is_leaked() -> bool` to check whether or not leakage has occurred and then potentially measure later, keeping the value of the qubit live. If no, then presumably qfree has to be called.

The reason I ask, is because if we wrap the int, then the user has to be aware that calling `is_leaked()` is actually measuring and qfree'ing, right?

ss2165: I meant methods on the object after measurement:

```python
def measure_leaked(q: qubit @ owned) -> MaybeLeaked: ...

class MaybeLeaked:
    _measured: int
    def is_leaked(self) -> bool: ...
    def to_result(self) -> Option[bool]: ...

# ...

ml = measure_leaked(q)
if ml.is_leaked():
   panic("leaked, aborting")
else:
   result("c", ml.to_result())
```

narmlu: Ah, yeah that's nice. Where should `MaybeLeaked` live?

Also, pretty sure the failing tests are due to tket2 (no future integers yet). I'm going to hold off on this PR until that is done.

guppylang/std/qsystem/__init__.py
```diff
@@ -74,6 +76,42 @@ def reset(q: qubit) -> None: ...
...
+    """Measure the quibit or return 2 if it is leaked."""
+
+
+@guppy
+def measure_leaked(q: qubit @ owned) -> "MaybeLeaked":
+    fm = _measure_leaked(q)
+    return MaybeLeaked(fm)
+
+@guppy
+class MaybeLeaked:
```
mark-koch: This should be a struct
```suggestion
@guppy.struct
class MaybeLeaked:
```

guppylang/std/qsystem/__init__.py
```diff
@@ -74,6 +76,42 @@ def reset(q: qubit) -> None: ...
...
+    """A class representing a measurement that may have leaked.
+
+    This is used to represent the result of `measure_leaked`, which can either
+    return a boolean measurement result or indicate that the qubit has leaked.
+    """
+
+    _measurement: Future[int]
+
+    def __init__(self, value: Future[int]):
+        self._measurement = value
```
mark-koch: No need to define `__init__` on guppy structs

```suggestion
```

guppylang/std/qsystem/__init__.py
```diff
@@ -74,6 +76,42 @@ def reset(q: qubit) -> None: ...
...
+
+    def is_leaked(self) -> bool:
```
mark-koch: Methods must be annotated as Guppy and `self` needs a type annotation:

```suggestion
    @guppy
    def is_leaked(self: "MaybeLeaked @ owned") -> bool:
```

guppylang/std/qsystem/__init__.py
```diff
@@ -74,6 +76,42 @@ def reset(q: qubit) -> None: ...
...
+        """Check if the measurement indicates a leak."""
+        return self._measurement.read() == 2
+    
+    def to_result(self) -> Option[bool]:
```
mark-koch: Same here

narmlu:
Everything seems to be working. The `MaybeLeaked` wrapper struct around the `_measurement: Future[int]` value has three funcs:
1. `is_leaked()` calls future's `.copy().read()` (i.e., does ___not___ consume it)
2. `to_result()` calls `.read()` (consumes it)
3. `discard()` to consume the future, which needs to be called if `to_result()` was never called because it was leaked.

I'm not sure that we can omit discard, because I don't think that `is_leaked` should consume `_measurement`. Concrete example is in test_qsystem.

guppylang/std/_internal/compiler/futures.py
narmlu, line 1: These are Marks changes in https://github.com/CQCL/guppylang/pull/1075. I'm happy to rebase to that branch if you want, but I'm inclined to just wait until that's merged.

tests/integration/test_qsystem.py
```diff
@@ -59,3 +62,20 @@ def test() -> tuple[int, float, int, int, angle, angle]:
...
+        ml: MaybeLeaked = measure_leaked(q)
+        if ml.is_leaked():
+                ml.discard()
+                q = qubit()
+                m = measure(q)
+                return m
```
qartik: What's the intention of this code pattern?

guppylang/std/qsystem/__init__.py
```diff
@@ -74,6 +76,52 @@ def reset(q: qubit) -> None: ...
...
 
+@guppy.hugr_op(quantum_op("LazyMeasureLeaked", ext=QSYSTEM_EXTENSION))
+@no_type_check
+def _measure_leaked(q: qubit @ owned) -> Future[int]:
+    """Measure the quibit or return 2 if it is leaked."""
+
+
+@guppy
+@no_type_check
+def measure_leaked(q: qubit @ owned) -> "MaybeLeaked":
```
mark-koch, line 87: Could you add a dosctring here?

guppylang/std/qsystem/__init__.py
```diff
@@ -74,6 +76,52 @@ def reset(q: qubit) -> None: ...
...
+    @guppy
+    @no_type_check
+    def is_leaked(self: "MaybeLeaked") -> bool:
+        """Check if the measurement indicates a leak."""
+        return self._measurement.copy().read() == 2
+
+    @guppy
+    @no_type_check
+    def to_result(self: "MaybeLeaked @ owned") -> Option[bool]:
+        """Get the measurement result if not leaked."""
```
mark-koch: Maybe something like this would be more accurate?
```suggestion
        """Returns the measurement result or `nothing` if leaked."""
```

mark-koch (approved):
Looks good, but we should wait to merge until #1075 is merged

> I'm not sure that we can omit discard, because I don't think that is_leaked should consume _measurement. Concrete example is in test_qsystem.

I think `discard` is nice to have 👍

tests/integration/test_qsystem.py
narmlu: This doesn't represent a "real" pattern. It is intended to make sure that things type check and that the variable `q` was properly consumed when `measure_leaked(q)` was called.

narmlu:
@mark-koch - it looks like #1075 was merged. Anything else we should wait on for this one?

tests/integration/test_qsystem.py
qartik: Might be nice to document that. Or remove it.

narmlu: Removed.

pyproject.toml
```diff
@@ -88,7 +88,7 @@ miette-py = { workspace = true }
 
 # Uncomment these to test the latest dependency version during development
 # hugr = { git = "https://github.com/CQCL/hugr", subdirectory = "hugr-py", tag = "hugr-v0.20.2" }
-# tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", rev = "652a7d0" }
+tket2-exts = { git = "https://github.com/CQCL/tket2", subdirectory = "tket2-exts", rev = "38d1c6f" }
```
mark-koch: Is this rev dependency still needed? If yes we should wait for a tket-exts release before merging

narmlu: Yes, good catch, we can wait until then (corresponding tket2 release PR https://github.com/CQCL/tket2/pull/952)

ss2165: I think this is the one you need: https://github.com/CQCL/tket2/pull/951

narmlu: Ah yes, exts. Got it, thanks.

ss2165: now released

narmlu: Bumped exts to the new release. Will merge when you remove `wait to merge` in case there are any other gotchas :).

ss2165: the only concern is selene support, @jake-arkinstall thoughts on merging?

narmlu: From @jake-arkinstall in [teams](https://teams.microsoft.com/l/message/19:5c7c78e33ea64b928b354d0ad980c4cf@thread.v2/1752277664602?context=%7B%22contextType%22%3A%22chat%22%7D):

> I have no concerns, merge away. It won’t work with Selene and it’ll link error if used, but it won’t break existing usage (if it isn’t used it isn’t emitted). Selene can play catch up next week.

narmlu: @ss2165 - I think this is ready to merge. I'll do so tomorrow morning (Colorado time) unless you tell me otherwise.

ss2165: :shipit:

PR #1063, opened 2025-06-30
feat: set hugr entrypoint to compiled function

ss2165:
use to avoid passing function names to execute_llvm

### DIFF
--- a/execute_llvm/src/lib.rs
+++ b/execute_llvm/src/lib.rs
@@ -66,6 +66,7 @@ fn compile_module<'a>(
 
     let emitter =
         hugr::llvm::emit::EmitHugr::new(ctx, llvm_module, namer.into(), extensions.into());
+    // TODO use hugr.fat_root after hugr 0.20.2
     let hugr_module = hugr.try_fat(hugr.module_root()).unwrap();
     let emitter = emitter
         .emit_module(hugr_module)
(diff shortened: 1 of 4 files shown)


execute_llvm/src/lib.rs
```diff
@@ -86,7 +66,7 @@ fn compile_module<'a>(
 
     let emitter =
         hugr::llvm::emit::EmitHugr::new(ctx, llvm_module, namer.into(), extensions.into());
-    let hugr_module = hugr.fat_root().unwrap();
+    let hugr_module = hugr.try_fat(hugr.module_root()).unwrap();
```
aborgna-q, line 70: I think `fat_root` was already a helper doing this?

tests/integration/conftest.py
```diff
@@ -86,8 +87,17 @@ def f(
             if not fn:
                 pytest.skip("Skipping llvm execution")
 
+            # set the module entrypoint to the one specified by `fn_name`
+            for mod in module.package.modules:
```
aborgna-q: What happens if we try to execute a package with multiple hugrs, where the function only exists in one of them?

It looks like `run_function` panics if given a multi-hugr package, but here it looks like it'd be supported. Should we check and error-out early instead?

aborgna-q (approved):

execute_llvm/src/lib.rs
ss2165: it uses entrypoint rather than module_root

aborgna-q: There's `fat_root` and `fat_entrypoint`, the former uses `module_root` (maybe it could be renamed)

PR #1064, opened 2025-06-30
fix: Fix nested function definitions in Python 3.12

### LINKED ISSUE #1061: Tests fail for Python 3.12
Something must have gone wrong with #1051

mark-koch:
Fixes #1061

The issue was that the `NestedFunctionDef.type_params` field was not being set. Now, we rely on `ast.iter_fields` to copy all fields of the `ast.FunctionDef` into our `NestedFunctionDef` node.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -19,10 +19,14 @@ env:
 
 jobs:
   check:
-    name: Check Python (3.10)
+    name: Check Python (${{ matrix.python-version }})
     runs-on: ubuntu-latest
     env:
-      PYTHON_VERSION: '3.10'
+      PYTHON_VERSION: ${{ matrix.python-version }}
+
+    strategy:
+      matrix:
+        python-version: [ '3.10', '3.12' ]
 
     steps:
       - uses: actions/checkout@v4
--- a/guppylang/cfg/builder.py
+++ b/guppylang/cfg/builder.py
@@ -259,12 +259,7 @@ class CFGBuilder(AstVisitor[BB | None]):
             cfg,
             func_ty,
             docstring=docstring,
-            name=node.name,
-            args=node.args,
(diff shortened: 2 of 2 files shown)


guppylang/cfg/builder.py
```diff
@@ -259,12 +259,7 @@ def visit_FunctionDef(
             cfg,
             func_ty,
             docstring=docstring,
-            name=node.name,
-            args=node.args,
-            body=node.body,
-            decorator_list=node.decorator_list,
-            returns=node.returns,
-            type_comment=node.type_comment,
+            **dict(ast.iter_fields(node)),
```
acl-cqc, line 262: Seems legit to me, `NestedFunctionDef` extends `ast.FunctionDef` so can support all the same fields

.github/workflows/pull-request.yaml
```diff
@@ -19,10 +19,14 @@ env:
...
-    name: Check Python (3.10)
+    name: Check Python (${{ matrix.python-version }})
     runs-on: ubuntu-latest
     env:
-      PYTHON_VERSION: '3.10'
+      PYTHON_VERSION: ${{ matrix.python-version }}
+
+    strategy:
+      matrix:
+        python-version: [ '3.10', '3.12' ]
```
acl-cqc, line 29: IIUC this was previously working on 3.10 and 3.13 but failing on 3.12? So 3.13 clearly changes *something* important in this area - should you include 3.13 also?

acl-cqc (approved):
Simple! :-D

.github/workflows/pull-request.yaml
mark-koch: Python 3.13 is already used for coverage, so all tests are run: https://github.com/CQCL/guppylang/actions/runs/15969900959/job/45038555192?pr=1064

PR #1075, opened 2025-07-02
feat: Add `Future` type

mark-koch:


guppylang/std/futures.py
```diff
@@ -0,0 +1,30 @@
...
+    """A value of type `T` that is computed asynchronously."""
+
+    @guppy.hugr_op(future_op("Read"))
+    @no_type_check
+    def read(self: "Future[T]" @ owned) -> T:
+        """Reads a value from a future, consuming it."""
+
+    @guppy.hugr_op(future_op("Dup"))
+    @no_type_check
+    def copy(self: "Future[T]") -> "Future[T]":
```
croyzor, line 24: This should return two futures!

croyzor (changes requested):

guppylang/std/futures.py
mark-koch: No, it borrows the input!

```python
future2 = future1.copy()
x1 = future1.read()
x2 = future2.read()
```

croyzor: 🤯

croyzor (approved):

Issue #1074, opened 2025-07-02
Optional `emulator` feature wrapping over selene


ss2165:
**Discussed in https://github.com/CQCL/guppylang/discussions/1062**

Key questions to resolve are how much of selene's flexibility is used directly and how much is wrapped. 

Installation:

```shell
pip install guppylang[emulator]
uv add guppylang[emulator]
```

Toy code:

```python
@guppy
def foo() -> None:
    q = qubit()
    state_result("my_state", q)
    result("c", measure(q))

runner: EmulatorInstance = guppy.build_emulator(foo)

# default statevector
result: QysResult = runner.run(n_qubits=1, n_shots=100)

# configure deterministic stabilizer
result: QysResult = runner.run(simulator=Stim(random_seed=34), n_qubits=1, n_shots=100)
print(result.collate_counts())

state_results: EmulatorState = result.extract_states()
state: np.ndarray = result.get_single_state("my_state")
print(state)
```

Tasks:
- [ ] Add selene_sim as optional dependency
- [ ] Wrap build and run methods
- [ ] Replace `execute_llvm`
- [ ] Add state interfaces
- [ ] User validation

CalMacCQ:
Is it a good idea to bundle statevectors and shots into the `QsysResult` object? This reminds me of  `BackendResult`  in pytket.

I guess `QsysResult` always contains at least one shot.

ss2165:
Slightly different scope to `BackendResult`, EmulatorResult (note more generic name) captures the output from every `*result` call in the program, including `state_result`. Does that make more sense?

ss2165:
Thoughts from user feedback session:
- run should take n_qubits since program execution _could_ depend on available qubits.
- `run` method/callable with build caching
- how to do tket2 optimisation? concensus: don't add more build args, have separate chaining for tket2/passes
- guppy.<thing>(foo) or foo.<thing>() ?
- Would be useful to provide runtime inputs to binaries

PR #1081, opened 2025-07-08
feat: add selene via optional feature and use for testing

ss2165:
Replaces validator and executor

Part of #1074 but user facing API will be a follow up

Final commit adds a new test that makes sure constants are loaded by llvm lowering - this cannot be tested in hugr-py using validation only

there's a lot of noise from ruff format in the second commit so might be easier to go one at a time after that one

### DIFF
--- a/tests/integration/test_frozenarray.py
+++ b/tests/integration/test_frozenarray.py
@@ -2,7 +2,7 @@ from guppylang import guppy
 from guppylang.std.builtins import frozenarray, panic, py
 
 
-def test_len(validate):
+def test_len(run_int_fn):
     @guppy
     def foo(xs: frozenarray[int, 42]) -> int:
         return len(xs)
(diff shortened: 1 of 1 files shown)


devenv.nix
mark-koch, line 1: I'm not using devenv so I'll have to trust you on this one 😅

tests/integration/conftest.py
```diff
@@ -73,69 +42,65 @@ class LLVMException(Exception):
...
+            result("_test_output", o)
+
+        @guppy.comptime
+        def flt_entry() -> None:
+            o: float = f(*args)
+            result("_test_output", o)
+
+        entry = flt_entry if is_flt else int_entry
+
+        # em = guppy.build_emulator(entry, n_qubits=1)
```
mark-koch:
```suggestion
```

tests/integration/conftest.py
```diff
@@ -73,69 +42,65 @@ class LLVMException(Exception):
...
+        em = guppy.compile(entry)
+        instance = build(em)
```
mark-koch, line 68: Is this checking that the Hugr is valid?

tests/integration/test_frozenarray.py
```diff
@@ -16,10 +16,10 @@ def main() -> int:
     validate(compiled)
     # TODO: Enable execution test once LLVM lowering is done:
     #  https://github.com/CQCL/hugr/issues/1973
-    #run_int_fn(compiled, 42)
+    # emulate_int_fn(compiled, 42)
```
mark-koch: This is done now, so we could enable the tests. Happy for that to be a different PR though

mark-koch (approved):

tests/integration/conftest.py
ss2165: yes sorry should have said in PR description, the `check_hugr` function is called by the `compile_to_llvm` (which is called by `build`)

tests/integration/test_frozenarray.py
ss2165: done

Issue #1011, opened 2025-06-09
ci: Tests that look at behaviour when tket is not installed are not run in CI


mark-koch:
See #1010

Issue #1010, opened 2025-06-09
Test `comptime_expr_errors.tket_not_installed` fails on main because it still uses the old `GuppyModule`


mark-koch:
This also means this test is not run in CI! We should investigate this

PR #1083, opened 2025-07-10
feat!: update to hugr-py v0.13

ss2165:
Requires a selene release with support for hugr v0.21 to pass tests


tests/error/comptime_expr_errors/tket2_not_installed.err
mark-koch, line 1: This looks broken. Probably also related to #1010, #1011

PR #1091, opened 2025-07-14
feat: Decorator to auto lower to hugr extension

### LINKED ISSUE #797: Auto defining extensions with attached lowerings
The compiler already features the @guppy.custom pattern for marking a function as lowering to an existing hugr extension.

It would be useful to be able to declare types and functions that are mapped to a generated hugr extension. Compilation would produce a HUGR defined over the new extension and an extra module containing the definitions that can be used to link/lower those operations. 

This is a convenient and quick way to define useful compile-time abstractions that can be removed further down the compilation stack.

hsemenenko:
Closes #797

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/lowerable.py
+++ b/guppylang-internals/src/guppylang_internals/definition/lowerable.py
@@ -35,16 +35,16 @@ class RawLowerableFunctionDef(RawCustomFunctionDef):
         call_checker: The custom call checker.
         call_compiler: The custom call compiler.
         higher_order_value: Whether the function may be used as a higher-order value.
-        signature: User-provided signature.
+        signature: Optional User-provided signature.
     """
 
     def parse(self, globals: "Globals", sources: SourceMap) -> "CustomFunctionDef":
-        """Parses and checks the signature of the custom function.
+        """Parses and checks the signature of the lowerable function.
 
         The signature is optional if custom type checking logic is provided by the user.
-        However, note that a signature must be provided by either annotation or as an
-        argument, if we want to use the function as a higher-order value. If a signature
-        is provided as an argument, this will override any annotation.
+        However, a signature *must* be provided to use the function as a higher-order
+        value (either by annotation or as an argument). If a signature is provided as an
+        argument, this will override any annotation.
 
         If no signature is provided, we fill in the dummy signature `() -> ()`. This
         type will never be inspected, since we rely on the provided custom checking
@@ -54,7 +54,7 @@ class RawLowerableFunctionDef(RawCustomFunctionDef):
         from guppylang_internals.definition.function import parse_py_func
 
         func_ast, _ = parse_py_func(self.python_func, sources)
-        sig = self._get_signature(func_ast, globals)
+        sig = self.signature or self._get_signature(func_ast, globals)
         ty = sig or FunctionType([], NoneType())
         return CustomFunctionDef(
             self.id,


guppylang-internals/src/guppylang_internals/definition/lowerable.py
```diff
@@ -0,0 +1,69 @@
...
+        call_compiler: The custom call compiler.
+        higher_order_value: Whether the function may be used as a higher-order value.
+        signature: User-provided signature.
+    """
+
+    def parse(self, globals: "Globals", sources: SourceMap) -> "CustomFunctionDef":
+        """Parses and checks the signature of the custom function.
+
+        The signature is optional if custom type checking logic is provided by the user.
+        However, note that a signature must be provided by either annotation or as an
```
acl-cqc: Whilst I think this is correct as it stands, I wonder if we should reorder for emphasis:
```suggestion
        However, a signature *must* be provided to use the function as a higher-order value (either by annotation or as an argument).
```
Or maybe
"The signature may be omitted if both (a) the function is not used as a higher-order value and (b) custom type-checking logic is provided instead"

Something like that?
(Or, if I've misunderstood....well, then I guess that's the point!! 😆 )

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -121,6 +124,62 @@ def hugr_op(
...
+                lower_funcs=[
+                    he.FixedHugr(
+                        ht.ExtensionSet([ext.name for ext in compiled_defn.extensions]),
+                        module,
+                    )
+                    for module in compiled_defn.modules
+                ],
+            )
+
+            hugr_ext.add_op_def(op_def)
```
acl-cqc: I'm a bit nervous here - if we have multiple uses of the same (decorated) function, do we call `add_op_def` multiple times? Or does this happen only once, when python encounters the `@lowered_op`? If the latter, what happens if I have two different but same-named `@lowered_op`s? (We would want that to be an error, methinks)

tests/integration/test_lower_op.py
```diff
@@ -0,0 +1,30 @@
...
+    @lowerable_op(test_hugr_ext)
+    def entangle(q0: qubit, q1: qubit) -> None:
+        h(q0)
+        cx(q0, q1)
+
+    @guppy
+    def main() -> None:
+        q0 = qubit()
+        q1 = qubit()
+        entangle(q0, q1)
```
acl-cqc, line 23: I reckon it'd be better if you called this a couple of times just to check that that still works...

tests/integration/test_lower_op.py
```diff
@@ -0,0 +1,30 @@
...
+        measure(q0)
+        measure(q1)
+
+    hugr = main.compile()
+
+    hugr.extensions.append(test_hugr_ext)
```
acl-cqc, line 31: Should we check that `test_hugr_ext` has the correct number of "entangle" ops registered, something like that?

tests/integration/test_lower_op.py
```diff
@@ -0,0 +1,30 @@
...
+
+
+def test_auto_hugr_lowering(validate):
+    test_hugr_ext = he.Extension("test_hugr_ext", SemanticVersion(0, 1, 0))
+
+    @lowerable_op(test_hugr_ext)
+    def entangle(q0: qubit, q1: qubit) -> None:
+        h(q0)
+        cx(q0, q1)
+
```
acl-cqc, line 17: So can we check in here that `test_hugr_ext` has the "entangle" op registered, something like that? That should happen now, right?

guppylang-internals/src/guppylang_internals/definition/lowerable.py
```diff
@@ -0,0 +1,69 @@
...
+    with the provided checker and compiler), without inspecting the signature.
+
+    Args:
+        id: The unique definition identifier.
+        name: The name of the definition.
+        defined_at: The AST node where the definition was defined.
+        call_checker: The custom call checker.
+        call_compiler: The custom call compiler.
+        higher_order_value: Whether the function may be used as a higher-order value.
+        signature: User-provided signature.
```
acl-cqc:
```suggestion
        signature: Optional user-provided signature.
```
Perhaps "required if `higher_order_value` is true"

guppylang-internals/src/guppylang_internals/definition/lowerable.py
```diff
@@ -0,0 +1,69 @@
...
+        However, note that a signature must be provided by either annotation or as an
+        argument, if we want to use the function as a higher-order value. If a signature
+        is provided as an argument, this will override any annotation.
+
+        If no signature is provided, we fill in the dummy signature `() -> ()`. This
+        type will never be inspected, since we rely on the provided custom checking
+        code. The only information we need to access is that it's a function type and
+        that there are no unsolved existential vars.
+        """
+        from guppylang_internals.definition.function import parse_py_func
```
acl-cqc, line 54: Hmmmm, ok. So the differences against the `RawCustomFunctionDef.parse` that we're overriding here are:
* we suppress the error if there is a body
* we don't look at `self.signature` before we call `self._get_signature(func_ast, globals)`....is that right?
* we don't assign the local variable `docstring` that AFAICS in the original/overridden method is defined but not used(?!)

There is definitely an opportunity here for both commoning-up, and possibly cleaning up....do you feel like taking it?

tests/integration/test_lower_op.py
acl-cqc: obvs would be even better if we could check the lowering is indeed saved in there

guppylang-internals/src/guppylang_internals/definition/lowerable.py
hsemenenko: Yes, these are the difference between the two methods. As suppressing the error if there is no body is required for the lowering case, it was suggested that these implementations were kept separate. This does come with a small side affect that `self.signature` is not used.

guppylang-internals/src/guppylang_internals/decorator.py
hsemenenko: Good observation. In the case where the function was called multiple times the second usage would override the first. I have moved the op definition outside this function so that it is only called a single time.

guppylang-internals/src/guppylang_internals/definition/lowerable.py
acl-cqc: Yes, agreed that you need suppress the warning, and since you are now using `self.signature`, looks good to me. You could possibly add a private helper that does the `sig = .....; ty = .....; return CustomFunctionDef(self...., self....) etc.` (taking just `self` and `func_ast` as parameters) but fine as it stands.

PR #1100, opened 2025-07-17
fix: Fix hugr conversion and bounds checks on numeric literals

mark-koch:


guppylang/std/_internal/compiler/arithmetic.py
```diff
@@ -12,6 +13,33 @@
 
 INT_T = int_t(NumericType.INT_WIDTH)
 
+
+@dataclass
+class UnsignedIntVal(val.ExtensionValue):
+    """Custom value for an unsigned integer.
+
+    TODO: Upstream this to hugr-py?
```
ss2165: move to comment rather than docstring

ss2165 (commented):
nice! just a comment about a comment

ss2165 (approved):

PR #1105, opened 2025-07-18
feat: Add float parameter inputs to symbolic pytket circuits

### LINKED ISSUE #1076: Handle free float parameters in pytket circuits

tatiana-s:
Closes #1076

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
+++ b/guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
@@ -412,7 +412,7 @@ def _signature_from_circuit(
                 assert isinstance(qubit, GuppyDefinition)
                 qubit_ty = cast(TypeDef, qubit.wrapped).check_instantiate([])
 
-                angle_defn = ENGINE.get_checked(angle.id)
+                angle_defn = ENGINE.get_checked(angle.id)  # type: ignore[attr-defined]
                 assert isinstance(angle_defn, TypeDef)
                 angle_ty = angle_defn.check_instantiate([])
 


guppylang/definition/pytket_circuits.py
```diff
@@ -355,6 +353,11 @@ def _signature_from_circuit(
             try:
                 import tket2  # type: ignore[import-untyped, import-not-found, unused-ignore]  # noqa: F401
 
+                param_inputs = [
+                    FuncInput(float_type(), InputFlags.NoFlags)
```
mark-koch: It might be better to use our `angle` type instead of floats?

tests/integration/test_pytket_circuits.py
```diff
@@ -260,3 +260,30 @@ def test_compile_load(validate):
     pytket_func = guppy.load_pytket("guppy_circ", circ, use_arrays=False)
 
     validate(pytket_func.compile())
+
+
+@pytest.mark.skipif(not tket2_installed, reason="Tket2 is not installed")
+def test_symbolic(validate):
```
mark-koch, line 266: You could turn this into an execution test, now that we have the selene dependency. E.g. an X rotation by pi should flip |0> into |1> deterministicly

guppylang/definition/pytket_circuits.py
```diff
@@ -370,13 +373,13 @@ def _signature_from_circuit(
                         for c_reg in input_circuit.c_registers
                     ]
                     circuit_signature = FunctionType(
-                        inputs,
+                        inputs + param_inputs,
```
mark-koch: Maybe for another PR, but if `use_array == True`, I would expect that the float parameters are also passed as arrays?

guppylang/definition/pytket_circuits.py
```diff
@@ -355,6 +353,11 @@ def _signature_from_circuit(
             try:
                 import tket2  # type: ignore[import-untyped, import-not-found, unused-ignore]  # noqa: F401
 
+                param_inputs = [
+                    FuncInput(float_type(), InputFlags.NoFlags)
+                    for symbol in input_circuit.free_symbols()
```
mark-koch: nit: I got initially confused by the fact that `Circuit.free_symbols` [returns a set](https://docs.quantinuum.com/tket/api-docs/circuit_class.html#pytket.circuit.Circuit.free_symbols) with no guaranteed order. I guess it's fine here since `symbol` is not actually used. Maybe it would be clearer to instead iteratate over `range(len(input_circuit.free_symbols()))`?

mark-koch: As an aside, from a user-perspective the ordering of these parameters it totally opaque, so I expect we'll run into some issues there. Not even sure how the pytket -> hugr converter picks the order?

tests/integration/test_pytket_circuits.py
```diff
@@ -260,3 +260,30 @@ def test_compile_load(validate):
...
+
+    AutoRebase({OpType.CX, OpType.Rz, OpType.H}).apply(circ)
+
+    guppy_circ = guppy.load_pytket("guppy_circ", circ)
+
+    @guppy
+    def foo(reg: array[qubit, 2]) -> array[bool, 2]:
+        alpha = 0.3
+        beta = 1.2
+        return guppy_circ(reg, alpha, beta)
```
mark-koch: E.g. here it's not clear if the first argument is actually alpha or beta

guppylang/definition/pytket_circuits.py
tatiana-s: How do you get an instance of the angle type here? Since `angle` is defined as a struct I have trouble turning it into a `TypeDef` here.

tests/integration/test_pytket_circuits.py
tatiana-s: Good idea - though for some reason this particular example doesn't end up being deterministic for me?

guppylang/definition/pytket_circuits.py
tatiana-s: Very good point - so turns out the order is defined by the order the parameters appear in the circuit. Since this is hard to determine by users, I think the best solution is to impose lexicographic order on the user (same constraint already exists for registers) and then internally assign the wires based on the actual order which can be taken from the hugr metadata added by the converter.

guppylang/definition/pytket_circuits.py
mark-koch: Similar to how we get `qubit_ty` above. You have to import `angle` and then `cast(TypeDef, angle.wrapped).check_instantiate([])`

tests/integration/test_pytket_circuits.py
mark-koch: What is the exact Guppy code you're running? We have a bug somehwere if `rx(pi)` is not deterministic

guppylang/definition/pytket_circuits.py
```diff
@@ -210,9 +210,21 @@ def compile_outer(
                     )
 
                 # Symbolic parameters get passed at the end.
-                param_wires = list(
-                    outer_func.inputs()[len(self.input_circuit.q_registers) :]
-                )
+                # We assume they are given in lexicographic order by the user.
+                # We need to sort them according to the metadata order.
+                # TODO: Allow to be passed as array if `use_arrays=True`.
```
mark-koch: Can you create/link an issue for this?

guppylang/definition/pytket_circuits.py
```diff
@@ -210,9 +210,21 @@ def compile_outer(
...
+                param_wires = []
+                if "TKET1.input_parameters" in hugr_func.metadata:
```
mark-koch: We should probably error out if the circuit has parameters but this metadata is missing?

tests/integration/test_pytket_circuits.py
```diff
@@ -289,20 +290,45 @@ def foo(reg: array[qubit, 2]) -> array[bool, 2]:
...
+    circ.measure_all()
+
+    AutoRebase({OpType.CX, OpType.Rz, OpType.H}).apply(circ)
+
+    flip = guppy.load_pytket("flip", circ, use_arrays=False)
+
+    @guppy
+    def main() -> int:
+        q = qubit()
+        res = int(flip(q, 0))
```
mark-koch: So this test fails if you set `alpha = pi`?

mark-koch: Ahh I guess it's because we're using the coin-flip simulator!

guppylang/definition/pytket_circuits.py
tatiana-s: That doesn't work, fails with `guppylang.error.InternalGuppyError: Tried to instantiate raw struct definition`

tests/integration/test_pytket_circuits.py
tatiana-s: Hm no I am using `Quest` if qubits are involved - I was using `comptime(math.pi)` (because it's currently floats and not angles), does that make a difference?

tatiana-s: It's fine now, I think it didn't work before because of another bug

guppylang/definition/pytket_circuits.py
tatiana-s: Decided to add it in this PR

guppylang/definition/pytket_circuits.py
mark-koch: Oh right, you'd need to get the `CheckedStructDef` from the engine:

```python
angle_defn = ENGINE.get_checked(angle.id)
assert isinstance(angle_defn, TypeDef)
angle_ty = angle_defn.check_instantiate([])
```

PR #1106, opened 2025-07-21
feat: Improve codegen for array unpacking

### LINKED ISSUE #1004: Inefficient array unpacking
```python
a, b, c = array(1, 2, 3)
```

The current unpacking logic for iterable RHS is to collect it into an array. But if it's already an array, this is unnecessary.

mark-koch:
Closes #1004

Guppy supports unpacking of arbitrary iterable types by collecting them into an array:

```python
x, *ys, z = range(10)

# Desugars to
arr = array(i for i in range(10))
x = arr.pop_left()
z = arr.pop_right()
ys = arr
```

However, since arrays are iterable types themselves, we applied the same logic when unpacking arrays:

```python
x, *ys, z = array(1, 2, 3, 4)

# Desugars to
arr = array(i for i in array(1, 2, 3, 4))  # Unnecessary comprehension
...
```

This PR adds a short-cut to avoid the array collection if the RHS is already an array.


tests/error/type_errors/unpack_generic_size2.err
mark-koch, line 1: New test since the old test is now triggered by the array check

croyzor (approved):

PR #1109, opened 2025-07-22
fix: Fix Jupyter notebook diagnostic rendering

### LINKED ISSUE #937: notebook: errors on struct methods can point to wrong method
Example:

```python
from guppylang.std.builtins import owned
@guppy.struct
class Struct:
    q: qubit
    @guppy
    def wrong(self: "Struct" @ owned) -> None:
        pass # self.discard() would work
    @guppy
    def discard(self: "Struct" @ owned) -> None:
        discard(self.q)

    
guppy.compile_module()
```

Throws correct error in wrong place:
```
Error: Drop violation (at <In [?]>:2:14)
  | 
1 |     @guppy
2 |     def discard(self: "Struct" @ owned) -> None:
  |               ^^^^^^^^^^^^^^^^^^^^^^ Field `self.q` with non-droppable type `array[qubit, 2]` is
  |                                      leaked

Help: Make sure that `self.q` is consumed or returned to avoid the leak

```

Problem is not seen in python script. Notebook issue.

### LINKED ISSUE #1053: Custom TB Handler failed, unregistering.....
This is the error message that popped up when I ran the following test program.

```
def test_prog(pauli_lst:list, sign_int:int):

    sq_table=array(array(0, 1, 2, 3), array(1, 0, 3, 2), array(2, 3, 0, 1), array(3, 2, 1, 0))
    sq_sign_table=array(array(0, 0, 0, 0), array(0, 0, 1, 3), array(0, 3, 0, 1), array(0, 1, 3, 0))

    n = guppy.nat_var("n")

    @guppy
    def get_combined_sq(pauli_list: array[int,n], sign_in:int) -> tuple[int, int]:
        pauli=0
        sign_out=0
        pauli=comptime(sq_table)[pauli_list[0],pauli_list[1]]
        sign_out=sign_in+ comptime(sq_sign_table)[pauli_list[0],pauli_list[1]]
        if n>2:
            for ii in range(2,n):
                sign_out=sign_out+comptime(sq_sign_table)[pauli,pauli_list[ii]]
                pauli= comptime(sq_table)[pauli,pauli_list[ii]]

        sign_out=sign_out % 4

        return pauli, sign_out
        

    @guppy
    def main() -> None:
        pauli, sign = get_combined_sq(comptime(pauli_lst),comptime(sign_int))
    
        qs= array(qubit(), qubit() )

        if pauli==1:
            x(qs[0])
        elif pauli==2:
            y(qs[0])
        elif pauli== 3:
            z(qs[0])
        bs = measure_array(qs)
        for b in bs:
            result("c", b)
            
    return main.compile()


prog = test_prog(pauli_lst=[0,1,3,2,1,2],sign_int=2)
runner = build(prog)

shots = QsysResult(runner.run_shots(Stim(random_seed=2), n_qubits=2, n_shots=100))

Counter(get_bitstrings(shots))

```
Image

mark-koch:
Fixes #1053 and fixes #937

I finally found a way to fix our notebook problems in a way that's hopefully more stable than previous hacks. It relies on the internal [`format_code_name`](https://ipython.readthedocs.io/en/9.1.0/api/generated/IPython.core.compilerop.html#IPython.core.compilerop.CachingCompiler.format_code_name) IPython method that allows us to translate dummy file names generated by IPython into their corresponding cells. This allows us to get rid of previous attempts to find definitions by walking the AST of cells.

### DIFF
--- a/guppylang/definition/function.py
+++ b/guppylang/definition/function.py
@@ -30,7 +30,6 @@ from guppylang.definition.common import (
 )
 from guppylang.definition.value import CallableDef, CallReturnWires, CompiledCallableDef
 from guppylang.error import GuppyError
-from guppylang.ipython_inspect import find_ipython_def, is_running_ipython
 from guppylang.nodes import GlobalCall
 from guppylang.span import SourceMap
 from guppylang.tys.subst import Inst, Subst
@@ -236,27 +235,10 @@ def compile_call(
 def parse_py_func(f: PyFunc, sources: SourceMap) -> tuple[ast.FunctionDef, str | None]:
     source_lines, line_offset = inspect.getsourcelines(f)
     source, func_ast, line_offset = parse_source(source_lines, line_offset)
-    # In Jupyter notebooks, we shouldn't use `inspect.getsourcefile(f)` since it would
-    # only give us a dummy temporary file
-    file: str | None
-    if is_running_ipython():
-        file = "<In [?]>"
-        if isinstance(func_ast, ast.FunctionDef):
-            defn = find_ipython_def(func_ast.name)
-            if defn is not None:
-                file = f"<{defn.cell_name}>"
-                sources.add_file(file, defn.cell_source)
-            else:
-                # If we couldn't find the defining cell, just use the source code we
-                # got from inspect. Line numbers will be wrong, but that's the best we
-                # can do.
-                sources.add_file(file, source)
-                line_offset = 0
-    else:
-        file = inspect.getsourcefile(f)
-        if file is None:
-            raise GuppyError(UnknownSourceError(None, f))
-        sources.add_file(file)
+    file = inspect.getsourcefile(f)
+    if file is None:
+        raise GuppyError(UnknownSourceError(None, f))
+    sources.add_file(file)
     annotate_location(func_ast, source, file, line_offset)
     if not isinstance(func_ast, ast.FunctionDef):
         raise GuppyError(ExpectedError(func_ast, "a function definition"))
--- a/guppylang/span.py
+++ b/guppylang/span.py
@@ -7,6 +7,7 @@ from typing import TypeAlias
 
 from guppylang.ast_util import get_file, get_line_offset
 from guppylang.error import InternalGuppyError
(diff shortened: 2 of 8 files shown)


tests/integration/notebooks/comptime.ipynb
```diff
@@ -362,15 +362,15 @@
    "outputs": [
     {
      "ename": "GuppyComptimeError",
-     "evalue": "Value with non-copyable type `qubit` was already used\n\nPrevious use occurred in <In [10]>:3 as an argument to `cx`",
+     "evalue": "Value with non-copyable type `qubit` was already used\n\nPrevious use occurred in <In[10]>:3 as an argument to `cx`",
```
mark-koch, line 365: IPython gives us the cell name `In[...]` without the space

guppylang/definition/function.py
```diff
@@ -236,27 +235,10 @@ def compile_call(
...
-    # In Jupyter notebooks, we shouldn't use `inspect.getsourcefile(f)` since it would
-    # only give us a dummy temporary file
-    file: str | None
-    if is_running_ipython():
-        file = "<In [?]>"
-        if isinstance(func_ast, ast.FunctionDef):
-            defn = find_ipython_def(func_ast.name)
-            if defn is not None:
-                file = f"<{defn.cell_name}>"
-                sources.add_file(file, defn.cell_source)
```
mark-koch, line 248: We now delay the cell lookup to the point where we actually print the diagnostic. Inside the compiler, we'll use the dummy names generated by IPython

guppylang/span.py
```diff
@@ -23,7 +24,8 @@ class Loc:
 
     def __str__(self) -> str:
         """Returns the string representation of this source location."""
-        return f"{self.file}:{self.line}:{self.column}"
+        file = normalize_ipython_dummy_files(self.file)
+        return f"{file}:{self.line}:{self.column}"
```
mark-koch, line 28: This makes sure that the IPython dummy names are not visible to the user and are instead replaced with nice `In[...]` names

qartik (approved):

PR #1118, opened 2025-07-28
chore: ruff fixes

ss2165:
`uv run ruff check` now passes

### DIFF
--- a/miette-py/miette_py.pyi
+++ b/miette-py/miette_py.pyi
@@ -1,4 +1,3 @@
-
 class PyDiagnostic:
     def __init__(
         self,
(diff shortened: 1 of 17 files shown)


uv.lock
```diff
@@ -2455,27 +2455,27 @@ wheels = [
 
 [[package]]
 name = "selene-core"
-version = "0.2.0rc1"
+version = "0.2.0rc2"
```
tatiana-s, line 2458: Assuming this release candidate bump doesn't have any influence on guppy?

ss2165: no its just a uv update

PR #1126, opened 2025-08-04
refactor!: Split up into `guppylang_internals` package

### LINKED ISSUE #1102: Pull compiler in to separate `guppylang-core` package and version independently of language
For end users the most important thing to version is the guppy language, but currently we use semver to indicate changes also to compiler internals. 

This issue proposes separating those internals in to a separate package, leaving a smaller API surface to version with the language in `guppylang`. This elements are:

1. Language itself - i.e. two `guppylang` versions are compatible if guppy source accepted by one is accepted by the other.
2. Standard library.
3. `@guppy` decorator and sub-decorators (e.g. comptime, load_pytket, declare, etc.)
4. Compiler interface: `check, compile, (build? run?)` https://github.com/CQCL/guppylang/issues/1074

This should allow guppylang to reach v1

mark-koch:
Closes #1102

BREAKING CHANGE: All compiler-internal and non-userfacing functionality is moved into a new `guppylang_internals` package


guppylang-internals/pyproject.toml
```diff
@@ -0,0 +1,47 @@
+[project]
+name = "guppylang-internals"
+version = "0.20.0"
```
aborgna-q, line 3:
```suggestion
version = "0.0.0"
```

pyproject.toml
```diff
@@ -34,10 +33,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "typing-extensions>=4.9,<5",
-    "tket-exts ~= 0.10.0",
-    "hugr ~= 0.13.0",
-    "selene-sim>=0.2.0rc4, <0.3",
+    "guppylang-internals",
```
aborgna-q:
```suggestion
    "guppylang-internals ~= 0.0.0",
```

The wheel building workflow may be broken an `push` to `main` until that package gets published since it specifies a non-existing lib.

Also note that the release PRs for `-internals` won't update this dependency automatically. (We may be able to hack something with the `extra-files` option, but I haven't looked into it yet).

guppylang-internals/pyproject.toml
aborgna-q: It may be more intuitive to just start from `0.20`, but we have to update the other occurrences to match.

PR #1127, opened 2025-08-04
feat!: Top level compile + emulate Interface

### LINKED ISSUE #1074: Optional `emulator` feature wrapping over selene

ss2165:
Closes #1074 

See example notebooks for usage

- Moves check/compile interface to `GuppyDefinition` (fear not the large diff, mostly noise from this change)
- Introduces a `foo.emulator(1).run()` interface for running local emulation
- Selene interface is wrapped with minimal starting top level interface. Can be extended in future
- The HUGR `QsysResult` class is sub-classed to allow future de-coupled extensions/changes
- State interface is kept generic over state representation to allow extension to other simulation backends

TODO:

- [ ] Unit tests
- [ ] Docstrings with examples
- [ ] Module level docs

BREAKING CHANGE: `guppy.compile(foo)` and `guppy.check(foo)` replaced with `foo.check()` and `foo.compile()`


guppylang/decorator.py
```diff
@@ -378,43 +376,16 @@ def extern(
...
-    def check(self, obj: Any) -> None:
-        """Compiles a Guppy definition to Hugr."""
-        from guppylang.engine import ENGINE
-
-        if not isinstance(obj, GuppyDefinition):
-            raise TypeError(f"Object is not a Guppy definition: {obj}")
-        return ENGINE.check(obj.id)
-
+    @deprecated(
+        "foo.compile() is deprecated and will be removed in a future version"
```
cqc-alec:
```suggestion
        "foo.compile() is deprecated and will be removed in a future version:"
```

guppylang/defs.py
```diff
@@ -20,10 +27,15 @@
...
     """A general Guppy definition."""
 
-    def compile(self) -> ModulePointer:
-        from guppylang.decorator import guppy
+    def compile(self) -> Package:
+        """Compile a Guppy definition to HUGR."""
+        return ENGINE.compile(self.id).package
+
+    def check(self) -> None:
+        """Type check a Guppy definition."""
```
cqc-alec:
```suggestion
        """Type-check a Guppy definition."""
```

guppylang/defs.py
```diff
@@ -34,10 +46,41 @@ class GuppyFunctionDefinition(GuppyDefinition, Generic[P, Out]):
...
+        compiled_module = ENGINE.compile(self.id)
+
+        # Look up how many generic params the function has so we can create an empty
+        # partial monomorphization to look up in the context
+        checked_def = ENGINE.checked[self.id]
+        mono_args: PartiallyMonomorphizedArgs | None = None
+        if isinstance(checked_def, MonomorphizableDef):
+            mono_args = tuple(None for _ in checked_def.params)
+
+        _ = ENGINE.compiled[self.id, mono_args]
```
cqc-alec: What is the point of this line? Does it do anything?

guppylang/emulator/state.py
```diff
@@ -0,0 +1,109 @@
...
+
+StateVector = npt.NDArray[np.complexfloating]
+
+
+@dataclass(frozen=True)
+class NotSingleStateError(Exception):
+    """Raised when a Selene state is not a single state, i.e. it has unspecified qubits
+    that have been traced out."""
+
+    total_qubits: int
```
cqc-alec, line 30:
```suggestion
    total_n_qubits: int
```
Just a suggestion (also elsewhere) -- looks like Selene uses `total_qubits` though so maybe better to leave for consistency.

tests/error/comptime_expr_errors/python_err.err
```diff
@@ -9,6 +9,8 @@ Traceback printed below:
 
 Traceback (most recent call last):
   File "<string>", line 1, in <module>
+    import sys;exec(eval(sys.stdin.readline()))
```
cqc-alec: This test seems to be failing.

guppylang/emulator/result.py
```diff
@@ -0,0 +1,29 @@
...
+if TYPE_CHECKING:
+    from selene_quest_plugin.state import SeleneQuestState
+
+
+class EmulatorResult(QsysResult):
+    """A result from running an emulator instance."""
+
+    # TODO more docstring
+
+    def partial_states_dict(self) -> list[dict[str, PartialVector]]:
```
cqc-alec: From the name one might expect this to return a `dict`. Maybe `partial_state_dicts` (or something else)?

guppylang/decorator.py
mark-koch:
```suggestion
        "guppy.compile() is deprecated and will be removed in a future version"
```

guppylang/defs.py
mark-koch: If you're not interested in returning a `FuncDefnPointer`, then there is no need to do any of this and the inherited `compile` implementation should be fine

guppylang/emulator/state.py
```diff
@@ -0,0 +1,109 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+from typing import Protocol, TypeVar
+
+import numpy as np
```
mark-koch, line 6: Should add numpy as a dependency if we're importing it here

PR #1131, opened 2025-08-04
docs: add docstrings for emulator module

ss2165:
First pass, should be improved in patch follow ups

### DIFF
--- a/guppylang/emulator/result.py
+++ b/guppylang/emulator/result.py
@@ -26,7 +26,7 @@ class EmulatorResult(QsysResult):
     convention is used, including conversion to a pytket BackendResult.
 
     Under this convention, tags are assumed to be a name of a bit register unless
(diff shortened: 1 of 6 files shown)


guppylang/emulator/result.py
```diff
@@ -12,14 +16,58 @@
...
+
+
+    Collects data from `result("tag", val)` calls in the guppy program. Includes results
+    for all shots.
+
+    Includes conversions to traditional distributions over bitstrings if a tagging
+    convention is used, including conversion to a pytket BackendResult.
+
+    Under this convention, tags are assumed to be a name of a bit register unless
+    theyfit the regex pattern `^([a-z][\w_]*)\[(\d+)\]$` (like `my_Reg[12]`) in which
```
aborgna-q:
```suggestion
    they fit the regex pattern `^([a-z][\w_]*)\[(\d+)\]$` (like `my_Reg[12]`) in which
```

PR #1142, opened 2025-08-05
docs: Fix docs build

mark-koch:
Fixes some problems with docs builds following #1126

### DIFF
--- a/guppylang-internals/src/guppylang_internals/decorator.py
+++ b/guppylang-internals/src/guppylang_internals/decorator.py
@@ -2,7 +2,7 @@ import builtins
 import inspect
 from collections.abc import Callable, Sequence
 from types import FrameType
-from typing import ParamSpec, TypeVar
+from typing import TYPE_CHECKING, ParamSpec, TypeVar
 
 from hugr import ops
 from hugr import tys as ht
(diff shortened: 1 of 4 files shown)


guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -243,4 +244,11 @@ def wasm(f: Callable[P, T]) -> GuppyFunctionDefinition[P, T]:
...
 
-# TODO sphinx hack for custom decorators
+# Override decorators with dummy versions if we're running a sphinx build
+if not TYPE_CHECKING and sphinx_running():
+    custom_function = _dummy_custom_decorator
+    hugr_op = _dummy_custom_decorator
+    extend_type = _dummy_custom_decorator
+    custom_type = _dummy_custom_decorator
+    wasm_module = _dummy_custom_decorator
+    wasm = _dummy_custom_decorator()
```
croyzor, line 254:
```suggestion
    wasm = _dummy_custom_decorator
```

mark-koch: The wasm decorator takes and returns the function directly:

```python
@wasm
def foo(): ...

# Instead of

@wasm()
def foo(): ...
```

Therefore, I put a call here to directly get the inner decorator

croyzor: ah good point!

croyzor (approved):

PR #1150, opened 2025-08-05
ci: Add release checks for `guppylang`

aborgna-q:
Adds a check run only on `release-please`'s PR for `guppylang` that ensures the library we are about to release works correctly with the `guppylang-internals` in `pypi`.
This should catch two failure modes:
- If we forgot to update the `-internals` dependency version in the man lib's pyproject.
- If the `-internals` package has breaking changes that we updated in `guppylang`, but they haven't been released yet.

Based on [tket's release check](https://github.com/CQCL/tket2/blob/main/.github/workflows/release-checks.yml).

### DIFF
--- a/.github/workflows/release-checks.yml
+++ b/.github/workflows/release-checks.yml
@@ -1,0 +1,56 @@
+# A check that ensures that `guppylang` works correctly with the latest release of `guppylang-internals`.
+#
+# If this fails, it is likely that we either didn't update the dependency version in guppylang's pyproject,
+# or the latest `guppylang-internals` package hasn't been published to pypi.
+
+name: 🚚 guppylang release checks
+
+on:
+  pull_request:
+    branches:
+      - release-please--branches--main--components--guppylang
+  workflow_dispatch: {}
+
+env:
+  # Pinned version for the uv package manager
+  UV_VERSION: "0.6.6"
+  # The highest and lowest supported Python versions, used for testing
+  PYTHON_HIGHEST: "3.13"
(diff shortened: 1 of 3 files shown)


.github/workflows/release-checks.yml
```diff
@@ -0,0 +1,58 @@
...
+# or the latest `guppylang-internals` package hasn't been published to pypi.
+
+name: 🚚 guppylang release checks
+
+on:
+  pull_request:
+    branches:
+      #- release-please--branches--main--components--guppylang
+      # TODO: Remove. Temporary trigger for testing in the PR.
+      - '**'
```
ss2165: reminder to remove

ss2165 (approved):

.github/workflows/release-checks.yml
```diff
@@ -0,0 +1,58 @@
...
+#
+# If this fails, it is likely that we either didn't update the dependency version in guppylang's pyproject,
+# or the latest `guppylang-internals` package hasn't been published to pypi.
+
+name: 🚚 guppylang release checks
+
+on:
+  pull_request:
+    branches:
+      #- release-please--branches--main--components--guppylang
```
aborgna-q: Ah, this is filtering the _target_ branch rather than the source. I'll fix it.

aborgna-q:
The check now fails because there's an unpublished change in `-internals` :) (#1149)
https://github.com/CQCL/guppylang/actions/runs/16748369572/job/47412228267?pr=1150

PR #1154, opened 2025-08-05
feat!: Turn type parameters into dependent telescopes

### LINKED ISSUE #1113: Turn generic parameters into type telescopes
https://ncatlab.org/nlab/show/type+telescope

In other words, allow generic parameters to refer back to previous bound variables:

```python
def fun[T: (Copy, Clone), xs: tuple[T, T]](...) -> ...
```

This will also be important for protocols:

```python
def fun[T](ts: Iterable[T]) -> None

# Desugars to
def fun[T, It: Iterable[T]](ts: It) -> None
```

mark-koch:
Closes #1113

BREAKING CHANGE: `checker.core.requires_monomorphization` renamed into `require_monomorphization` and now operating on all parameters simultaneously
BREAKING CHANGE: `tys.subst.BoundVarFinder` removed. Instead, use the new `bound_vars` property on types, arguments, and consts.
BREAKING CHANGE: `tys.parsing.parse_parameter` now requires a `param_var_mapping`.


tests/error/poly_errors_py312/dependent1.err
```diff
@@ -0,0 +1,8 @@
+Error: Type mismatch (at $FILE:11:12)
+   | 
+ 9 | 
+10 | @guppy
+11 | def main(x: MyStruct[bool, 42]) -> None:
+   |             ^^^^^^^^^^^^^^^^^^ Expected argument of type `bool`, got `nat`
```
croyzor, line 6: why is the `arg_locs` change failing us here?

tests/error/poly_errors_py312/dependent3.err
```diff
@@ -0,0 +1,9 @@
+Error: Type mismatch (at $FILE:17:11)
+   | 
+15 | @guppy
+16 | def main() -> Eq[bool, True, False]:
+17 |     return eq_refl()
+   |            ^^^^^^^^^ Expected expression of type `Eq[bool, True, False]`, got
+   |                      `Eq[?T, ?x, ?x]`
```
croyzor, line 7: "expected" and "got" seem the wrong way round

tests/integration/test_poly_const.py
```diff
@@ -351,3 +351,26 @@ def main() -> None:
...
+        foo(True)
+        foo(False)
+
+    compiled = main.compile()
+    validate(compiled)
+
+    # Check we have main, and 3 monomorphisations of foo. The two nat versions should
+    # correspond to the same monomorphisation since the bounded nat args are preserved
+    # in Hugr!
+    assert len(funcs_defs(compiled.modules[0])) == 4
```
croyzor, line 376: nice

tests/error/poly_errors_py312/dependent3.err
croyzor: but hell yeah to the example

tests/integration/test_poly_py312.py
```diff
@@ -112,3 +112,78 @@ def main[T: Drop]() -> None:
...
+
+    @guppy
+    def get[T: (Copy, Drop), x: T](self: "Phantom[T, x]") -> T:
+        return x
+
+
+def test_dependent_struct(validate):
+    @guppy
+    def make_phantom[T: (Copy, Drop)](x: T @ comptime) -> "Phantom[T, x]":
+        return Phantom()
```
croyzor, line 141: 🚀

tests/integration/test_poly_py312.py
```diff
@@ -112,3 +112,78 @@ def main[T: Drop]() -> None:
...
+
+    @guppy
+    def main(x: Phantom[bool, True]) -> float:
+        return 0.0 if x.get() else make_phantom(42).get() + make_phantom(1.5).get()
+
+    validate(main.compile())
```
croyzor: this could be a `run_float_fn` test

tests/integration/test_poly_py312.py
```diff
@@ -112,3 +112,78 @@ def main[T: Drop]() -> None:
...
+
+
+def test_dependent_comptime(validate):
+    T = guppy.type_var("T", copyable=True, droppable=True)
+
+    @guppy
+    def foo(x: T @ comptime, y: "Phantom[T, x]") -> T:
+        return y.get()
```
croyzor, line 159: 🚀

croyzor (approved):
nice!!!!!
Approved in principle, but please look at my comments on error messages

tests/error/poly_errors_py312/dependent3.err
mark-koch: As discussed, this should be fine as is

tests/error/poly_errors_py312/dependent1.err
mark-koch: Because I didn't propagate the `arg_loc` change through the entire compiler - would have been out of scope for this PR

In this case here, the call to `check_all_args` comes from `ParsedStructDef.check_instantiate` which doesn't the spans of the type argumentys :/

PR #1157, opened 2025-08-05
feat: Allow custom start and step in `range`

### LINKED ISSUE #791: Stdlib: More range functions

mark-koch:
Closes #791

This is doable now thanks to our function overloading. Furthermore, we can get rid of the custom checker and replace it with comptime arguments 🎉

### DIFF
--- a/tests/integration/test_range.py
+++ b/tests/integration/test_range.py
@@ -30,7 +30,7 @@ def test_range(run_int_fn):
         return sum(x + 100 for x in r)
 
     run_int_fn(stop, args=[5], expected=expected(builtins.range(5)))
-    run_int_fn(stop, args=[-3], expected=expected(builtins.range(-1)))
+    run_int_fn(stop, args=[-3], expected=expected(builtins.range(-3)))
     run_int_fn(start, args=[2, 7], expected=expected(builtins.range(2, 7)))
     run_int_fn(start, args=[-2, 5], expected=expected(builtins.range(-2, 5)))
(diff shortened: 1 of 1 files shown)


tests/integration/test_range.py
```diff
@@ -1,35 +1,40 @@
...
 
-    run_int_fn(main, expected=510)
-
-    run_int_fn(negative, expected=0)
+    def expected(r) -> int:
+        return sum(x + 100 for x in r)
 
-    run_int_fn(non_static, expected=510)
+    run_int_fn(stop, args=[5], expected=expected(builtins.range(5)))
+    run_int_fn(stop, args=[-3], expected=expected(builtins.range(-1)))
```
qartik: This one is confusing as the args differ on LHS vs RHS. Best to add a comment that the output is 0 with any negative argument.

mark-koch: Oops yes it be the same on the left and right side 👍

PR #1164, opened 2025-08-06
feat: Make collections iterable and discardable

### LINKED ISSUE #1088: Make collections discardable

### LINKED ISSUE #1089: Make collections interable

mark-koch:
Closes #1088 and closes #1089.

As a drive-by, I also deleted a duplicate implementation of `PriorityQueue` in `stack.py`. To make this non-braking, I replaced with a reexport, but we should remove this in the future!

### DIFF
--- a/guppylang/std/collections/priority_queue.py
+++ b/guppylang/std/collections/priority_queue.py
@@ -41,6 +41,25 @@ class PriorityQueue(Generic[T, MAX_SIZE]):  # type: ignore[misc]
 
     @guppy
     @no_type_check
+    def __iter__(
+        self: PriorityQueue[T, MAX_SIZE] @ owned,
+    ) -> PriorityQueue[T, MAX_SIZE]:
+        """Returns an interator over the queued elements in priority order."""
+        return self
+
+    @guppy
+    @no_type_check
+    def __next__(
+        self: PriorityQueue[T, MAX_SIZE] @ owned,
+    ) -> Option[tuple[tuple[int, T], PriorityQueue[T, MAX_SIZE]]]:
+        if len(self) == 0:
+            self.discard_empty()
+            return nothing()
+        prio, val, new_queue = self.pop()
+        return some(((prio, val), new_queue))
+
+    @guppy
+    @no_type_check
     def push(
         self: PriorityQueue[T, MAX_SIZE] @ owned,
         value: T @ owned,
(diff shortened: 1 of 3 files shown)


guppylang/std/collections/priority_queue.py
```diff
@@ -133,6 +152,19 @@ def peek(
...
+    def discard_empty(self: PriorityQueue[T, MAX_SIZE] @ owned) -> None:
+        """Discards a priority queue of potentially non-droppable elements assuming that
+        the queue is empty.
+
+        Panics if the queue is not empty.
+        """
+        if self.size > 0:
+            panic("PriorityQueue.discard_empty: priority queue is not empty")
+        for elem in self.buf:
+            elem.unwrap_nothing()
```
acl-cqc: Is it appropriate to just panic here? I mean, `unwrap_nothing` might panic (if it were called), or not...

tests/integration/std/test_collections.py
```diff
@@ -19,6 +19,23 @@ def main() -> int:
             x, stack = stack.pop()
             s += x * i
             i += 1
+        stack.discard_empty()
+        return s
+
+    run_int_fn(main, sum((i + 1) * x for i, x in enumerate(reversed(list(range(10))))))
```
acl-cqc: wouldn't mind that `# multiplier * value for ordered values in priority queue` comment here (and below) too

guppylang/std/collections/priority_queue.py
```diff
@@ -39,6 +39,25 @@ def __len__(self: PriorityQueue[T, MAX_SIZE]) -> int:
         """Returns the number of elements currently stored in the priority queue."""
         return self.size
 
+    @guppy
+    @no_type_check
+    def __iter__(
+        self: PriorityQueue[T, MAX_SIZE] @ owned,
+    ) -> PriorityQueue[T, MAX_SIZE]:
+        """Returns an interator over the queued elements in priority order."""
```
acl-cqc:
```suggestion
        """Returns an iterator over the queued elements in priority order."""
```

acl-cqc: Actually, should really say that these are the elements *paired with their priorities* or something like that

guppylang/std/collections/priority_queue.py
```diff
@@ -39,6 +39,25 @@ def __len__(self: PriorityQueue[T, MAX_SIZE]) -> int:
...
+        return self
+
+    @guppy
+    @no_type_check
+    def __next__(
+        self: PriorityQueue[T, MAX_SIZE] @ owned,
+    ) -> Option[tuple[tuple[int, T], PriorityQueue[T, MAX_SIZE]]]:
+        if len(self) == 0:
+            self.discard_empty()
+            return nothing()
```
acl-cqc: we can't `return None` here? :(

acl-cqc: Or even just `return self.discard_empty()`....you might not like that for style (fair!)

tests/integration/std/test_collections.py
acl-cqc: well, "values in stack" I guess

guppylang/std/collections/priority_queue.py
mark-koch: No, all these `unwrap_nothing`s *will* be called. We have to actually check all remaining slots in the buffer to make sure there are no linear elements hiding there

guppylang/std/collections/priority_queue.py
mark-koch: `nothing()` is not the same as `None` unfortunately. `None` is the unit type whereas `nothing()` is a constructor for `Option`.

We don't have enums/union types in guppy yet, so we can't use `T | None` :(

PR #1165, opened 2025-08-07
feat: Add unsafe array take and put operations

mark-koch:
This should remove the need for people to use option-arrays.

Release-As: 0.21.8

### DIFF
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -102,9 +102,40 @@ class array(builtins.list[_T], Generic[_T, _n]):
         ```
         """
 
+    @custom_function(ArrayGetitemCompiler())
+    def take(self: array[L, n], idx: int) -> L:
+        """Takes an element out of the array.
+
+        While regular indexing into an array only allows borrowing of elements, `take`
+        actually *extracts* the element and transfers ownership to the caller. This
+        makes this operation inherently unsafe: elements may no longer be accessed after
+        they are taken out. Attempting to do so will result in a runtime panic.
+
+        The complementary `array.put` method may be used to return an element back into
+        the array to make it accessible again.
+
+        Panics if the provided index is negative or out of bounds, or if the element has
+        already been taken out.
+
+        Also see `array.try_take` for a version of this function that does not panic if
+        the element has already been taken out.
+
+        # Example
+
+        ```
+        qs = array(qubit() for _ in range(10))
+        h(qs[3])
+        q = qs.take(3)
+        measure(q)  # We're allowed to deallocate since we own `q`
+        # h(qs[3])  # Would panic since qubit 3 has been taken out
+        qs.put(qubit(), 3) # Put a fresh qubit back into the array
+        h(qs[3])
+        ```
+        """
+
     @guppy
     @no_type_check
-    def take(self: array[L, n], idx: int) -> Option[L]:
+    def try_take(self: array[L, n], idx: int) -> Option[L]:
         """Tries to take an element out of the array.
 
         While regular indexing into an array only allows borrowing of elements, `take`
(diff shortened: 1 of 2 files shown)


guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -410,6 +410,9 @@ def compile(self, args: list[Wire]) -> list[Wire]:
 class ArraySetitemCompiler(ArrayCompiler):
     """Compiler for the `array.__setitem__` function."""
 
+    def __init__(self, elem_first: bool = False):
```
ss2165, line 339: please document the kwarg

ss2165 (commented):
I like

`take_unsafe` and `return_unsafe`

guppylang/src/guppylang/std/either.py
```diff
@@ -16,7 +16,7 @@
 from guppylang_internals.tys.param import TypeParam
 
 from guppylang import guppy
-from guppylang.std.builtins import owned
+from guppylang.std.lang import owned
```
mark-koch, line 19: This is needed to avoid cyclic imports (since `builtins` also imports `array`)

tests/integration/test_array.py
```diff
@@ -644,3 +645,42 @@ def test2() -> None:
...
+
+def test_take_put(validate):
+    @guppy
+    def foo(q: qubit, x: None) -> None:
+        pass
+
+    @guppy
+    def main() -> int:
+        qs = array(qubit() for _ in range(10))
+        result("a", qs.is_borrowed(3))  # False
```
ss2165: suggest using more meaningful result tags to make the outcome easier to interpret

guppylang/src/guppylang/std/array.py
```diff
@@ -80,6 +82,109 @@ def __iter__(self: array[L, n] @ owned) -> SizedIter[ArrayIter[L, n], n]:
...
+
+        # Example
+
+        ```
+        qs = array(qubit() for _ in range(10))
+        h(qs[3])
+        result("a", qs.is_borrowed(3))  # False
+        q = qs.take(3).unwrap()
+        result("a", qs.is_borrowed(3))  # True
+        qs.return_unsafe(qubit(), 3).unwrap()
```
ss2165:
```suggestion
        qs.put(qubit(), 3).unwrap()
```

guppylang/src/guppylang/std/array.py
```diff
@@ -80,6 +82,109 @@ def __iter__(self: array[L, n] @ owned) -> SizedIter[ArrayIter[L, n], n]:
...
+        """
+
+    @guppy
+    @no_type_check
+    def take(self: array[L, n], idx: int) -> Option[L]:
+        """Tries to take an element out of the array.
+
+        While regular indexing into an array only allows borrowing of elements, `take`
+        actually *extracts* the element and transfers ownership to the caller. This
+        makes this operation inherently unsafe: If the array elements are non-copyable,
```
ss2165:
```suggestion
        makes this operation inherently unsafe: if the array elements are non-copyable,
```

guppylang/src/guppylang/std/array.py
```diff
@@ -80,6 +82,109 @@ def __iter__(self: array[L, n] @ owned) -> SizedIter[ArrayIter[L, n], n]:
...
+    def put_unchecked(self: array[L, n], elem: L @ owned, idx: int) -> None:
+        """Panicking version of `put`.
+
+        Equivalent to
+
+        ```
+        xs.put(idx, elem).unwrap()
+        ````
+        """
+
```
ss2165: not sure we should add these to the API, is the reasoning that it saves all the checking code?

guppylang/src/guppylang/std/array.py
```diff
@@ -80,6 +82,109 @@ def __iter__(self: array[L, n] @ owned) -> SizedIter[ArrayIter[L, n], n]:
     def copy(self: array[T, n]) -> array[T, n]:
         """Copy an array instance. Will only work if T is a copyable type."""
 
+    @custom_function(ArrayIsBorrowedCompiler())
+    def is_borrowed(self: array[L, n], idx: int) -> bool:
+        """Checks if a non-copyable element has been taken out of the array.
```
ss2165: what is the behaviour if the element is copyable? Add a test for this?

mark-koch: Oh the behaviour is actually the same for copyable and non-copyable since we use the same lowering. I'll fix the docstrings

guppylang/src/guppylang/std/array.py
mark-koch: Yes, it avoids doing the check twice. Also, those are only a single op which might be nice?

On the other hand, we could also make them private for now if we're not sure and decide later. Don't feel strongly about this

PR #1166, opened 2025-08-07
feat!: Use `borrow_array` instead of `value_array` for array lowering

### LINKED ISSUE #1072: Use `borrow_array` instead of `value_array` extension for array lowering

tatiana-s:
Closes #1072

BREAKING CHANGE: (guppy-internals) Arrays are now lowered to `borrow_array`s instead of `value_array`s so elements do no longer need to be wrapped in options during lowering.

~~Requires a tket2 release with https://github.com/CQCL/tket2/pull/975 for the execution tests to pass~~ Requires a hugr-llvm release for the lowering + hugr with copy discard handler + tket2 with adjustments + hugr-qis compiler with adjustments

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
+++ b/guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
@@ -210,7 +210,8 @@ class ParsedPytketDef(CallableDef, CompilableDef):
                 ]
 
                 # Symbolic parameters (if present) get passed after qubits and bools.
-                has_params = len(self.input_circuit.free_symbols()) != 0
+                num_params = len(self.input_circuit.free_symbols())
+                has_params = num_params != 0
                 if has_params and "TKET1.input_parameters" not in hugr_func.metadata:
                     raise InternalGuppyError(
                         "Parameter metadata is missing from pytket circuit HUGR"
@@ -223,8 +224,7 @@ class ParsedPytketDef(CallableDef, CompilableDef):
                     if self.use_arrays:
                         unpack_result = outer_func.add_op(
                             array_unpack(
-                                ht.Tuple(float_type().to_hugr(ctx)),
-                                q_reg.size,
+                                ht.Tuple(float_type().to_hugr(ctx)), num_params
                             ),
                             lex_params[0],
                         )
--- a/tests/integration/test_pytket_circuits.py
+++ b/tests/integration/test_pytket_circuits.py
@@ -298,7 +298,7 @@ def test_symbolic_array(validate):
     a = Symbol("alpha")
     b = Symbol("beta")
 
-    circ = Circuit(2)
+    circ = Circuit(3)
     circ.YYPhase(b, 0, 1)
     circ.Rx(a, 0)
     circ.measure_all()
@@ -308,7 +308,7 @@ def test_symbolic_array(validate):
     guppy_circ = guppy.load_pytket("guppy_circ", circ)
 
     @guppy
-    def foo(reg: array[qubit, 2]) -> array[bool, 2]:
+    def foo(reg: array[qubit, 3]) -> array[bool, 3]:
         params = array(angle(0.3), angle(1.2))
         return guppy_circ(reg, params)
 


guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -252,156 +279,54 @@ def compile(self, args: list[Wire]) -> list[Wire]:
             return [self.build_classical_array(args)]
 
 
-ARRAY_GETITEM_CLASSICAL: Final[GlobalConstId] = GlobalConstId.fresh(
```
tatiana-s, line 255: This is the version where we stop inlining, I have a version where we keep the function calls stashed away locally, if we want to merge this before improving the lowering in during execution I can put it back here and make a separate branch with this version

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -63,13 +63,16 @@
...
+    convert_itousize,
 )
 from guppylang_internals.std._internal.compiler.array import (
+    array_clone,
     array_convert_from_std_array,
     array_convert_to_std_array,
     array_map,
     array_new,
-    array_repeat,
+    array_new_all_borrowed,
```
acl-cqc: So all of these are borrow-array...not very obvious from the name...

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -556,23 +559,30 @@ def visit_ResultExpr(self, node: ResultExpr) -> Wire:
             op_name = f"result_array_{base_name}"
             size_arg = node.array_len.to_arg().to_hugr(self.ctx)
             extra_args = [size_arg, *extra_args]
-            # Remove the option wrapping in the array
-            unwrap = array_unwrap_elem(self.ctx)
-            unwrap = self.builder.load_function(
-                unwrap,
-                instantiation=ht.FunctionType([ht.Option(base_ty)], [base_ty]),
-                type_args=[ht.TypeTypeArg(base_ty)],
+            # As `borrow_array`s used by Guppy are linear, we need to clone it (knowing
```
acl-cqc, line 562:
```suggestion
            # As `borrow_array`s are linear, we need to clone it (the elements
            # must be copyable as they are output as results) to pass one copy
            # to the result operation and return the other
```
what are the chances of defining *return* as a pass-through operation, i.e. such that it returns the array? (much like `result` in Rust would only borrow the array not consume it)?

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -945,31 +900,21 @@ def apply_array_op_with_conversions(
     output array.
 
     Transformations:
-    1. Unwraps / wraps elements in options.
-    3. (Optional) Converts from / to opaque bool to / from Hugr bool.
+    1. (Optional) Converts from / to opaque bool to / from Hugr bool.
```
acl-cqc, line 903: on next line: `value array` -> `borrow array`

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -489,4 +491,34 @@ class CopyInoutCompiler(CustomInoutCallCompiler):
     """Call compiler for functions that are noops but only want to borrow arguments."""
 
     def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
+        assert len(self.ty.input) == 1
```
acl-cqc, line 494: Do I understand correctly that `args` must therefore have only one element too? (Seems you are passing `*args` into `clone`...)

In which case, I'd pass `Wire` not `list[Wire]` to `_handle_affine_type`; I quite like this shorthand that throws if there are the wrong number of elements:
```
(inp_ty,) = self.ty.input[0]
(arg,) = args
```

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -489,4 +491,34 @@ class CopyInoutCompiler(CustomInoutCallCompiler):
...
+            return CallReturnWires(
+                regular_returns=[copies[0]], inout_returns=[copies[1]]
+            )
         return CallReturnWires(regular_returns=args, inout_returns=args)
+
+    # Affine types in Guppy backed by a linear Hugr type need to be copied explicitly.
+    def _handle_affine_type(self, ty: ht.Type, args: list[Wire]) -> list[Wire]:
+        match ty:
+            case ht.ExtType(type_def=type_def, args=type_args):
+                # TODO: Handle affine extension types more generally.
```
acl-cqc: Is it that this linear BorrowArray type is the first case that's actually shown up, then?

I think I'd put this TODO comment with the function-level one.

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -489,4 +491,34 @@ class CopyInoutCompiler(CustomInoutCallCompiler):
     """Call compiler for functions that are noops but only want to borrow arguments."""
```
acl-cqc: I'd say this was for a function that borrows its argument and copies it, rather than being a noop

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -489,4 +491,34 @@ class CopyInoutCompiler(CustomInoutCallCompiler):
...
+                regular_returns=[copies[0]], inout_returns=[copies[1]]
+            )
         return CallReturnWires(regular_returns=args, inout_returns=args)
+
+    # Affine types in Guppy backed by a linear Hugr type need to be copied explicitly.
+    def _handle_affine_type(self, ty: ht.Type, args: list[Wire]) -> list[Wire]:
+        match ty:
+            case ht.ExtType(type_def=type_def, args=type_args):
+                # TODO: Handle affine extension types more generally.
+                if qualified_name(type_def) == qualified_name(
```
acl-cqc, line 510: If this `if` fails, the `_handle_affine_type` will just return `[]` - is that right??
I think you probably want:
```
match ty:
   case ht.ExtType(type_def=type_def, args=type_args) if qualified_name(type_def) == qualified_name(BORROW_ARRAY_EXTENSION.get_type("borrow_array"):
      assert len(type_args) == 2
      .....
  case _:
      raise InternalGuppyError(....)
# no fallthrough (unreachable)
````

File: guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py [hunk 9]
```diff
@@ -50,10 +42,10 @@ def _instantiate_array_op(
 def array_type(elem_ty: ht.Type, length: ht.TypeArg) -> ht.ExtType:
     """Returns the hugr type of a fixed length array.
 
-    This is the copyable `value_array` type used by Guppy.
+    This is the `borrow_array` type used by Guppy.
```
acl-cqc:
```suggestion
    This is the linear `borrow_array` type used by Guppy.
```

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -163,8 +155,8 @@ def array_repeat(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
 
 
 def array_convert_to_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
```
acl-cqc: how about calling these `array_to_std_array` and `std_array_to_array` rather than `convert to/from`? would that make the difference easier to spot?

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -163,8 +155,8 @@ def array_repeat(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
...
-    """Returns an array operation to convert the `value_array` type used by Guppy into
-    the regular linear `array` in Hugr.
+    """Returns an array operation to convert the `borrow_array` type used by Guppy into
+    the regular `array` in Hugr.
```
acl-cqc: both array types are in Hugr! IIUC borrow-array is defined in Hugr not in the guppy extension, which may be a mistake (! - much easier to reuse code where it is though). So I think you have to say `the Hugr borrow_array type used by Guppy into a standard array` maybe with `where all elements are always present`

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -163,8 +155,8 @@ def array_repeat(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
...
     """
     return EXTENSION.get_op("to_array").instantiate(
```
acl-cqc, line 161: Would it make sense to pass the arguments to the op (and the builder) in here, so we can check the caller provides the correct number(/type)?

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -176,7 +168,7 @@ def array_convert_to_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtO
 
 def array_convert_from_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
     """Returns an array operation to convert the `array` type used by Hugr into the
-    `value_array` type used by Guppy.
+    `borrow_array` type used by Guppy.
```
acl-cqc, line 171: as above, Hugr doesn't really use `array` any more than `borrow_array` (does it? sorry if I'm mistaken)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -186,6 +178,41 @@ def array_convert_from_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.Ex
...
+def array_return(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
+    """Returns an array `return` operation."""
+    arr_ty = array_type(elem_ty, length)
+    return _instantiate_array_op(
+        "return", elem_ty, length, [arr_ty, ht.USize(), elem_ty], [arr_ty]
+    )
+
+
+def array_discard_all_borrowed(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
+    """Returns an array `borrow` operation."""
```
acl-cqc: Nope

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -186,6 +178,41 @@ def array_convert_from_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.Ex
...
+def array_borrow(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
+    """Returns an array `borrow` operation."""
+    arr_ty = array_type(elem_ty, length)
+    return _instantiate_array_op(
+        "borrow", elem_ty, length, [arr_ty, ht.USize()], [elem_ty, arr_ty]
+    )
+
+
+def array_return(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
+    """Returns an array `return` operation."""
```
acl-cqc, line 190:
```suggestion
    """An array `return` operation."""
```

acl-cqc: There are about three calls, so passing the arguments here might help (even if they are just names of type `Wire`: array_return(idx=, array=, elem=`)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -186,6 +178,41 @@ def array_convert_from_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.Ex
...
+
+
+def array_discard_all_borrowed(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
+    """Returns an array `borrow` operation."""
+    arr_ty = array_type(elem_ty, length)
+    return _instantiate_array_op("discard_all_borrowed", elem_ty, length, [arr_ty], [])
+
+
+def array_new_all_borrowed(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
+    """Returns an array `borrow` operation."""
```
acl-cqc: nor here

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -186,6 +178,41 @@ def array_convert_from_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.Ex
...
+    arr_ty = array_type(elem_ty, length)
+    return _instantiate_array_op("new_all_borrowed", elem_ty, length, [], [arr_ty])
+
+
+def array_clone(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
+    """Returns an array `clone` operation."""
```
acl-cqc: Might be good to say that this clones an array none of whose elements are borrowed

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
acl-cqc: What do we want to do now we have BorrowArray codegen'd in hugr-llvm? Inline??

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -252,156 +274,54 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
         )
-        type_args = [ht.TypeTypeArg(self.elem_ty), self.length]
-        func_call = self.builder.call(
-            func.parent_node,
+
+    def _build_linear_getitem(self, array: Wire, idx: Wire) -> CallReturnWires:
+        """Constructs `array.__getitem__` for linear arrays."""
+        idx = self.builder.add_op(convert_itousize(), idx)
+        elem, arr = self.builder.add_op(
+            array_borrow(self.elem_ty, self.length),
```
acl-cqc: I'm confused....

so we're compiling guppy-source `x = array[5]` to `x = array.borrow(5)` ?? (I thought that was ruled out in source code except for: whereas `h(array[5])` (a borrow) is turned into `_temp = array.borrow(5)` then call `h` then `array.return(5,x)`...

In particular IIUC if we have `cx(qs[0][1], qs[2][3])` then we have to return `qs[0]` before borrowing `qs[2]` (both before and after calling cx), which then looks a bit different (and even less like what one would expect if the above is "assignment"). I guess I'm feeling this shouldn't be `ArrayGetItemCompiler` or `ArraySetItemCompiler` - those should be for copyable elements only, and we should have some separate thing for compiling borrow/return.

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -410,160 +330,60 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
-            ],
-            output=[array_type(ht.Option(self.elem_ty), self.length)],
-        )
-        type_args = [ht.TypeTypeArg(self.elem_ty), self.length]
-        func_call = self.builder.call(
-            func.parent_node,
+        """Constructs `array.__setitem__` for linear arrays."""
+        idx = self.builder.add_op(convert_itousize(), idx)
+        arr = self.builder.add_op(
+            array_return(self.elem_ty, self.length),
```
acl-cqc: yeah, as above, this feels a bit scary

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -410,160 +330,60 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
-            if not already_exists:
-                self._build_classical_setitem(func)
-        return self._build_call_setitem(func=func, array=array, idx=idx, elem=elem)
+            return self._build_classical_setitem(array, idx, elem)
 
     def compile(self, args: list[Wire]) -> list[Wire]:
         raise InternalGuppyError("Call compile_with_inouts instead")
 
 
 class ArrayIterAsertAllUsedCompiler(ArrayCompiler):
```
acl-cqc: Rename to....`ArrayDiscardAllUsedCompiler` ?? (Not sure whether the `Iter` is still relevant...maybe remove that)

acl-cqc: and it is all-used not all-borrowed, right?

guppylang-internals/src/guppylang_internals/tys/param.py
```diff
@@ -83,8 +83,11 @@ def instantiate_bounds(self, inst: "PartialInst") -> Self:
 class TypeParam(ParameterBase):
     """A parameter of kind type. Used to define generic functions and types."""
 
+    # TODO: This should be refactored as this is hard to reason about. The optional
+    # field is used as a hack to deal with lowering affine types to linear HUGR types.
     must_be_copyable: bool
     must_be_droppable: bool
+    is_affine: bool | None = None
```
acl-cqc, line 90: can this be a property?
```
@property
def is_affine(self) -> bool:
   return self.must_be_droppable and not self.must_be_copyable
```

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
acl-cqc: It's reasonable to name these after the features they are implementing rather than the implementation they are backing onto, but I wonder about renaming to `...._internal.compiler.barray`, say, at least. (If we did name everything `barray` then you could drop the `std_` on the conversion methods, tho....)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
acl-cqc: Ok, now I see this is used for `_array_unsafe_getitem`, so perhaps it is not exposed to the user? In which case I am less unhappy but would still rather see it renamed/split. (The classical version isn't particularly unsafe, or is it?)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -252,156 +274,54 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
-        elem_opt_ty = ht.Option(elem_ty)
-        none = func.add_op(ops.Tag(0, elem_opt_ty))
-        idx = func.add_op(convert_itousize(), func.inputs()[1])
         # As copyable elements can be used multiple times, we need to swap the element
-        # back after initially swapping it out for `None` to get the value.
-        initial_result = func.add_op(
-            array_set(elem_opt_ty, length),
-            func.inputs()[0],
+        # back after initially borrowing it to get the value.
+        # (`array_get` cannot be used to to current bool lowering limitations)
```
acl-cqc:
```suggestion
        # (`array_get` cannot be used due to current bool lowering limitations)
```

but...can't it? BorrowArray::get definitely exists for copyable elements. Presumably it panics if the element is borrowed?

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
tatiana-s: We inline - sorry the comment above is wrong it should have said "stop using function calls" or "go back to inlining"

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
tatiana-s: It would require the signature of the result which is a breaking change in the extension which was a subject of much discussion and though it was implemented it was then reverted again so I don't think we are doing that for now

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
tatiana-s: hm I think given the default arrays in guppy are borrow array it is fine to call them that (at was the same case with value array, and it made it easier to change the backing than if it was all named specifically after value array) but I agree it may be not completely clear where we draw the boundary

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
tatiana-s: It does (for things like results for example - hugr extensions that internally use arrays to represent data just use standard arrays) - but I agree that this phrasing isn't quite right

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
tatiana-s: I think it makes sense to rename the ops that only borrow arrays have to start with `barray` (I did also consider unsafe?) so I've done that but not sure that is the most ideal solution either

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
tatiana-s: Maybe this could be a separate array refactor

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
tatiana-s: You wouldn't be able to do `x = array[5]` unless the elements were copyable so yes it is not exposed to the user. This is equivalent to this code putting a `None` after taking the item previously, and these custom compilers are only used by the compiler itself which should always ensure they are used correctly.  I do see the concern but not sure what alternative naming would be better?

guppylang-internals/src/guppylang_internals/tys/param.py
tatiana-s: `is_affine` is not equal to `must_be_droppable and not self.must_be_copyable`

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
tatiana-s: When copyable bools are lowered to non-copyable option(future(bool), bool)s this apparently causes issues, I can't remember the details

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
acl-cqc: Ah OK. Yes inlining is now good....I think close the comment unless I've misunderstood.

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
acl-cqc: ah right yes I see. That makes sense. I realize I don't have a good picture as to how borrow/return works for classical values (I mean, does it make sense to "borrow" the value over a function call? Aren't classical values always `@Owned` in which case you never really borrow them?), including the case where you have a function polymorphic over T (allowing linear T) which is then used with T==some_classical_type....

If array_get exists, works, and is appropriate here for everything except bools, the best thing to do (make an issue maybe) might be to change the ReplaceBools pass to lower array_get of bool, to borrow+copy+return, for bool only. That would leave the code here cleaner and better for non-bool cases (or if not lazifying)

guppylang-internals/src/guppylang_internals/tys/param.py
acl-cqc: Ok, so `not self.must_be_copyable and self.must_be_droppable`, is the only thing I can see being assigned, [here](https://github.com/CQCL/guppylang/blob/60665baf97d1511d978c1417fbef624e6a3540b5/guppylang-internals/src/guppylang_internals/definition/parameter.py#L48).

There are plenty of other places that use the default of `None` I think? Which seems a little scary; we are relying on implicit None->bool conversion??

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -162,9 +154,9 @@ def array_repeat(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
     )
 
 
-def array_convert_to_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
-    """Returns an array operation to convert the `value_array` type used by Guppy into
-    the regular linear `array` in Hugr.
+def array_to_std_array(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
+    """Returns an array operation to convert the `borrow_array` type used by Guppy into
+    the a standard `array`.
```
acl-cqc: "the a" isn't right, but otherwise this sounds good now :)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
acl-cqc: I guess what I'm wondering is whether we can split this class into two, one that only does classical arrays (and might then be called ArrayGetItemCompiler - maybe eventually it can use `array::get` if we fix the bool-lowering to swap in a borrow+return) and one that is called, y'know, ArrayBorrowCompiler, and either only does linear, or is only used for linear. Then the *frontend* chooses which to use, because clearly, the frontend knows the difference anyway (in order to generate get before + return afer for linear, but only get for classical)....?

acl-cqc: That said, this is also used for compiling `def __getitem__` of `class array`, so I'm not sure I understand everything that's going on here. Is `class array` only used for classical elements??? @mark-koch

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
acl-cqc: Ok so following https://github.com/CQCL/tket2/pull/1181 how about the following:
```
class ArrayBorrowCompiler(ArrayCompiler):
   def _build(self, elem_ty: Type, array: Wire, idx: Wire) -> CallReturnWires:
      # ignore elem_ty
      ...just a borrow (plus index conversion)...

   def compile_inouts(...):
      ...extract array, size, etc...
      return self._build(elem_ty, array, idx)

@guppy.custom_func(ArrayBorrowCompiler)
def _barray_unsafe_borrow(...) # used in ArrayIter, in place of _array_unsafe_getitem, no other uses

class ArrayGetitemCompiler(ArrayBorrowCompiler):
   def _build_classical_getitem(self, array, idx) -> CallReturnWires:
      ...emit array.get (could be borrow + return without tket#1181)

   @override
   def _build(self, elem_ty: Type, array: Wire, idx: Wire) -> CallReturnWires:
     if elem_ty.copyable():
       return self._build_classical_getitem(...)
     else:
       return super().build(elem_ty, array, idx
```

There are other ways to structure the methods - you could separate out `build_borrow` that doesn't take `elem_ty` so the default `build` just delegates to `build_borrow` without the type, and the overriding `build` then doesn't `super`. Or you could override `call_inouts` with slight duplication. Or many other variations, all fine by me.

acl-cqc: In fact it seems that the ArrayIter only ever gets output for the linear version (copyable never happens) - because we output ArrayIter as polymorphic *in Hugr*, so in the guppy compiler, the type is always a type variable (with bound Linear).

So you could even make ArrayBorrowCompiler test whether the element type is copyable, and switch between two methods; the classical method, in ArrayBorrowCompiler. then panics (GuppyInternalError or something), and you override just that in ArrayGetitemCompiler.

acl-cqc: I'm gonna try and make another PR doing this now!

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
acl-cqc: See #1303.

guppylang-internals/src/guppylang_internals/tys/param.py
acl-cqc: Ok so I went to `class TypeParam` in  `guppylang-internals/src/guppylang_internals/tys/param.py`....
1. Tried adding this:
```
   def __post_init__(self) -> None:
      assert False
```
to verify that this did, indeed, break tests (even collection fails) - i.e., `__post_init__` is called immediately after constructing each TypeParam.
2. changed to:
```
    def __post_init__(self) -> None:
        affine = False if self.is_affine is None else self.is_affine
        assert affine == ((not self.must_be_copyable) and self.must_be_droppable)
```
Note the bracketting required; this passes many tests but breaks about ten, for example this call:
https://github.com/CQCL/guppylang/blob/6990d850170e6901f60ef1d1e718c99349105b56/guppylang-internals/src/guppylang_internals/tys/parsing.py#L320-L323

acl-cqc: A further change in 3635907edfcb64009f102c06ab7c5de2fa430e5e demonstrates that in 8 of the 10 tests that were broken by the assert above, `is_affine` was never read. There are only two cases where `is_affine` is actually read, and is not equal to `(not self.must_be_copyable and self.must_be_droppable)`: these are `test_drop_bound` and `test_reference_inside`, both in `test_poly_py312.py`

acl-cqc: In these tests, it looks like a type variable bound by `T: Drop` is currently output in Hugr as Copyable. Changing `is_affine` to `true` in these cases, means the type variable is output as Linear, whereupon validation fails with an UnconnectedPort

acl-cqc: Looking specifically at `test_drop_bound`
* Using code in this PR as it stands, we have `is_affine==False, must_be_copyable==False, must_be_droppable==True`. The generated Hugr contains a polymorphic `main` function `forall T: Copyable. T -> ()` that contains no Drops, so is valid.

* If I change `is_affine` to `not self.must_be_copyable and self.must_be_droppable`, the generated Hugr contains a polymorphic `main` function: `forall T: Linear. T -> ()` that still contains no Drops, so is invalid.

acl-cqc: However, if we set `is_affine` to `False` in *all* cases, then while those tests pass, others fail with invalid Hugrs where we instantiate polymorphic-over-Copyable functions with "Linear" arguments like BorrowArray...

acl-cqc: https://github.com/CQCL/guppylang/issues/1306

PR #1169, opened 2025-08-07
fix!: Fix globals vs locals scoping behaviour to match Python

### LINKED ISSUE #827: Allow shadowing of extern variables
For example (see test in `test_extern.py`):
```
guppy.extern("x", ty="int", module=module)

@guppy(module)
def main(b: bool) -> int:
    if b:
       x = 4
    return x
```

This currently crashes (`KeyError: Variable.Id(name='x')` in the linearity checker during `[scope[x] for x in live_before[bb]]`), but it would be nice if this worked.

### LINKED ISSUE #1168: Shadowing of generic parameters is broken
```python
@guppy
def main[n: nat](b: bool) -> nat:
    if b:
        n = nat(42)
    return n
```

Fails with

```
guppylang_internals.error.InternalGuppyError: Cannot find definition `n`
```

This is very similar to the scoping problems we had with globals before. See #772

mark-koch:
Fixes #1168.
Fixes #827.

In the past we had some aspirations to allow conditional shadowing of globals with locals:

```python
x = 1

def foo(b: bool) -> int:
    if b:
        x = 2
    return x
```

Since implementation proved difficult, we decided to produce an error until this is implemented (#815). However, we now have the same problems with generic parameters (#1168). Furthermore, allowing this would also go against Python's behaviour: `foo(False)` above throws an `UnboundLocalError`!

This PR updates our scoping logic to resemble Python more closely and prevent the case above. This also allows us to get rid of the hacky solution in `check_rows_match`.

BREAKING CHANGE: `check_rows_match` no longer takes `globals`
BREAKING CHANGE: Deleted `GlobalShadowError` and `BranchTypeError.GlobalHint`

### DIFF
--- a/tests/error/errors_on_usage/shadow_global4.err
+++ b/tests/error/errors_on_usage/shadow_global4.err
@@ -1,0 +1,11 @@
+Error: Variable not defined (at $FILE:13:15)
+   | 
+11 | 
+12 |     def inner() -> float:
+13 |         return x
+   |                ^ `x` might be undefined ...
+   | 
+ 9 |     if b:
+   |        - ... if this expression is `False`
+
+Guppy compilation failed due to 1 previous error
--- a/tests/error/errors_on_usage/shadow_global4.py
+++ b/tests/error/errors_on_usage/shadow_global4.py
@@ -1,0 +1,16 @@
+from guppylang import guppy
+from hugr.std.float import FloatVal
+
+x = guppy.constant("x", "float", FloatVal(4.2))
+
+
+@guppy
+def main(b: bool) -> None:
+    if b:
+        x = 1.0
+
+    def inner() -> float:
+        return x
+
+
+main.compile()


tests/error/errors_on_usage/shadow_global3.py
```diff
@@ -0,0 +1,23 @@
...
+
+@guppy
+def bar() -> None:
+    pass
+
+
+@guppy
+def main(b: bool) -> None:
+    if b:
+        bar = foo
```
acl-cqc, line 17: Do we have a case where the LHS (but not RHS) `bar` is renamed to say `baz` and the program then works? (I mean, it seems kinda obvious that it should, but....)

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
```diff
@@ -245,23 +229,26 @@ def check_bb(
...
+            # Check that the variables requested by the successor are defined. If `x` is
+            # a local variable, then we must be able to find it in the context.
+            # Following Python, locals are exactly those variables that are defined
+            # somewhere in the function body.
+            if x in cfg.assigned_somewhere:
+                if x not in ctx.locals:
+                    # If the variable is defined on *some* paths, we can give a more
+                    # informative error message
+                    if x in cfg.maybe_ass_before[use_bb]:
+                        err = VarMaybeNotDefinedError(use_bb.vars.used[x], x)
```
acl-cqc: Suggest moving this `err` outside of the nearest `if x in cfg.maybe_ass_before`...

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
```diff
@@ -245,23 +229,26 @@ def check_bb(
...
+                        if bad_branch := diagnose_maybe_undefined(use_bb, x, cfg):
+                            branch_expr, truth_value = bad_branch
+                            note = VarMaybeNotDefinedError.BadBranch(
+                                branch_expr, x, truth_value
+                            )
+                            err.add_sub_diagnostic(note)
+                        raise GuppyError(err)
```
acl-cqc: ....then you can skip this `raise`....

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
```diff
@@ -245,23 +229,26 @@ def check_bb(
...
+                    raise GuppyError(VarNotDefinedError(use_bb.vars.used[x], x))
```
acl-cqc: ....and just `raise GuppyError(err)` here (so it's the same `err` for all `if x not in ctx.locals` but maybe with a sub-diagnostic `if x in cfg.maybe_ass_before`)

guppylang-internals/src/guppylang_internals/cfg/cfg.py
```diff
@@ -101,6 +105,10 @@ def analyze(
         inout_vars: list[str],
     ) -> dict[BB, VariableStats[str]]:
         stats = {bb: bb.compute_variable_stats() for bb in self.bbs}
+        # Locals are variables that are assigned somewhere inside the function
```
acl-cqc, line 108: Or somewhere in a containing function (`maybe_ass_before`) - this handles the case where the use of an assigned-only-on-some-paths variable is inside a nested function (after the merge of paths)...

Which is a very corner case, kudos for spotting it 👍 . Is that one tested? 😉

guppylang-internals/src/guppylang_internals/cfg/cfg.py
```diff
@@ -101,6 +105,10 @@ def analyze(
...
+        self.assigned_somewhere = def_ass_before.union(
+            maybe_ass_before, (x for bb in self.bbs for x in stats[bb].assigned)
```
acl-cqc, line 110: ✅

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
mark-koch: The errors are not the same! `VarMaybeNotDefinedError` vs `VarNotDefinedError`

But maybe it's clearer to assign `err` in the else branch

PR #1171, opened 2025-08-11
feat: Add `Result` type

### LINKED ISSUE #1170: Add `Result` type

mark-koch:
Closes #1170


guppylang/std/err.py
ss2165, line 1: why err rather than `result`

ss2165: Answered my own question - confusion with `result` function

guppylang/std/err.py
```diff
@@ -0,0 +1,70 @@
...
+from guppylang.std.builtins import owned
+from guppylang.std.either import Either
+
+T = guppy.type_var("T", copyable=False, droppable=False)
+E = guppy.type_var("E", copyable=False, droppable=False)
+
+_params = [TypeParam(0, "T", False, False), TypeParam(1, "E", False, False)]
+
+
+@custom_type(either_to_hugr, params=_params)
```
ss2165: why custom rather than just wrap an either value? Smaller codegen?

ss2165: this is very neat though

ss2165 (approved):
Happy with this so approving but a few questions for clarity

guppylang/std/err.py
mark-koch: Yes, this might even be a reason to pick a different name than `Result`, but nothing good came to mind

guppylang/std/err.py
mark-koch: > Smaller codegen?

Yes, the produced Hugr will be nicer imo. Also less code since we can reuse the `Either` compilers one-to-one

guppylang/std/err.py
ss2165: Yes, worth thinking about. I couldn't find a good example that doesn't just use `Result`. 

Maybe add it to the top level and struct docstrings to clarify this is nothing to do with result outputs?

CalMacCQ: Minor point but sphinx autosummary doesn't tend to like classes which only differ by captialisation. It generates a bunch of files for the API docs and in a case sensitive file system there can be issues.

That said, can probably hack around it.

PR #1178, opened 2025-08-11
ci: Check for breaking changes

### LINKED ISSUE #1175: Check for breaking changes in CI
This seems like it should do the job: https://mkdocstrings.github.io/griffe/guide/users/checking/

mark-koch:
Closes #1175

See  #1180 for a test PR that triggers the check


.github/workflows/semver-checks.yml
```diff
@@ -0,0 +1,139 @@
+name: Python semver-checks
```
aborgna-q: Important change to match `rs-semver-checks`

```suggestion
name: Python semver-checks 🐍
```

.github/workflows/semver-checks.yml
```diff
@@ -0,0 +1,139 @@
...
+        shell: bash
+        run: |
+          # Don't fail the workflow when griffe returns a non-zero exit code.
+          set +e
+
+          uv run griffe check guppylang --against main --verbose 2> diagnostic.txt
+          guppylang_breaking=$?
+
+          uv run griffe check guppylang_internals --search guppylang-internals/src --against main --verbose 2>> diagnostic.txt
+          guppylang_internals_breaking=$?
```
aborgna-q: If making this into a composite action, we'll need to specify an array of packages as input and loop on them here.

Github doesn't allow array inputs, but this workaround looks fine
https://stackoverflow.com/a/75420778

ss2165:
can merge now?

PR #1181, opened 2025-08-12
chore: Move `guppylang` package into its own directory

### LINKED ISSUE #1179: chore: Move `guppylang` to its own dir
Having one of the libraries at the repo's root is messy, and causes problems with `release-please`'s component detection for merged commits.

mark-koch:
Closes #1179


guppylang/README.md
aborgna-q, line 1: This is the file that gets rendered on the pypi landing page. We should make an issue to improve it a bit.

aborgna-q: The root README looks like a better base than what's being used here.

guppylang-internals/README.md
ss2165, line 1: Isn't this the wrong one? (internals)

.github/workflows/python-wheels.yml
```diff
@@ -34,11 +34,11 @@ jobs:
       matrix:
         target:
           - {
-            dir: ".",
+            dir: guppylang,
             component: guppylang,
             package: guppylang,
             lib: guppylang,
-            dist: dist
```
ss2165, line 41: this can now be removed, was only added because both packages weren't in subdirectories

guppylang-internals/README.md
mark-koch: Ooops 😅

ss2165 (approved):

PR #1187, opened 2025-08-13
docs: expand Guppylang changelog

CalMacCQ:
Updating this changelog with hand written elements so that there is a single source of truth for the langauge changelog. We can then use this markdown page in the docs website.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,16 +19,42 @@ As of August 2025, The Guppy language is undergoing rapid development and is cur
 
 ### Features
 
-* Add float parameter inputs to symbolic pytket circuits ([#1105](https://github.com/CQCL/guppylang/issues/1105)) ([34c546c](https://github.com/CQCL/guppylang/commit/34c546c3b5787beb839687fdbf4db8bc94f36c4a)), closes [#1076](https://github.com/CQCL/guppylang/issues/1076)
-* Allow custom start and step in `range` ([#1157](https://github.com/CQCL/guppylang/issues/1157)) ([a1b9333](https://github.com/CQCL/guppylang/commit/a1b9333712c74270d5efaaa72f83d6b09047c068))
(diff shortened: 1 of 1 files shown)


CHANGELOG.md
```diff
@@ -1,20 +1,60 @@
+---
+file_format: mystnb
+kernelspec:
+  name: python3
+---
+
```
aborgna-q, line 6: Note that this file is auto-generated by `release-please`. Your modifications to older releases won't be overwritten, but I'm not sure about the header.
(Can't find anything on the docs/source code so we'll have to merge and see).

CHANGELOG.md
```diff
@@ -1,20 +1,60 @@
...
 # Changelog
 
+The Guppy language is distributed via the [guppylang](https://pypi.org/project/guppylang/) pypi package.
+
+This changelog documents user-facing changes to the Guppy language excluding changes to the compiler internals. The Guppy compiler is versioned with the [guppylang-internals](https://pypi.org/project/guppylang-internals/) package.
```
aborgna-q, line 11: Changes get added here automatically on release, one per each merged commit affecting the package and a _Breaking changes_ section at the start if applicable.

I thought we wanted to have a higher level hand written changelog with user-friendly explanations of the main changes?
Are we meant to write that on the release PRs in the future?
(and fight with `release-please` as it overwrites the PR changes each time it's rerun)

CHANGELOG.md
CalMacCQ: The header can easily be removed I think. Its not very important for this page. We can just use standard Python markdown cells rather than the mystnb stuff.

CHANGELOG.md
CalMacCQ: Worth a chat with @ss2165  I think. We were keen to have a single source of truth for the changelog but maybe it doesn't play so nicely with automated releases.

aborgna-q (approved):

PR #1192, opened 2025-08-14
feat: Infer type of `self` arguments

### LINKED ISSUE #1059: Allow omission of `self` type annotations

mark-koch:
Closes #1059

I recommend reviewing commit by commit. Most of them are refactors, the main logic is added in 9f3f155

BREAKING CHANGE: `TypeDef`s now require a `params` field
BREAKING CHANGE: `guppylang_internals.ty.parsing.parse_function_io_types` replaced with `parse_function_arg_annotation` and `check_function_arg`


guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -176,7 +226,9 @@ def check_nested_func_def(
     return with_loc(func_def, checked_def)
 
 
-def check_signature(func_def: ast.FunctionDef, globals: Globals) -> FunctionType:
+def check_signature(
+    func_def: ast.FunctionDef, globals: Globals, def_id: DefId | None = None
+) -> FunctionType:
```
croyzor, line 242: Can we have the arguments explained in the docstring?

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -211,22 +263,30 @@ def check_signature(func_def: ast.FunctionDef, globals: Globals) -> FunctionType
...
     input_names = []
-    for inp in func_def.args.args:
-        ty_ast = inp.annotation
-        if ty_ast is None:
-            raise GuppyError(MissingArgAnnotationError(inp))
-        input_nodes.append(ty_ast)
+    for i, inp in enumerate(func_def.args.args):
+        # Special handling for `self` arguments. Note that `__new__` is excluded here
+        # since it's not a method so doesn't take `self`.
+        if self_defn and i == 0 and func_def.name != "__new__":
```
croyzor, line 294: Are we not allowed to define other class methods that don't take self? Shouldn't we try to parse self and and report either error/valid-but-not-self/success?

croyzor (approved):

guppylang-internals/src/guppylang_internals/checker/func_checker.py
mark-koch: No, currently `__new__` is the only supported static method and hardcoded :/ Would be good to allow user-defined static methods in the futuree though

> Shouldn't we try to parse self and and report either error/valid-but-not-self/success?

Not sure I fully get what you mean here?

croyzor: Ah okay, I was just trying to say it's easy to remove this restriction for when we do support static methods

PR #1194, opened 2025-08-14
test: move over integration tests from guppy-integration

ss2165:


tests/integration/test_emulator.py
```diff
@@ -87,3 +106,161 @@ def main() -> None:
...
+    return program.emulator(n_qubits).with_shots(n_shots).with_seed(seed).run()
+
+
+def test_zeros():
+    N = 2
+
+    @guppy
+    def main() -> array[qubit, 2]:
+        q = array(qubit(), qubit())
+        for i in range(comptime(N)):
```
acl-cqc, line 128: Do you prefer this to `for qubit in q: result("c", project_z(q))` (or rename `q` to `qs`)? It seems less natural to me, but you may be meaning to test this specifically

acl-cqc: also, if you remove this use of `N`....then perhaps go further....I mean, if the program were actually parametric in `N` this would be great, but it's not really...so you might consider removing `N`?

tests/integration/test_emulator.py
```diff
@@ -87,3 +106,161 @@ def main() -> None:
...
+        # bell measurement
+        cx(aux, tgt)
+        h(aux)
+
+        result("aux", measure(aux))
+        discard(tgt)
+
+    res = _build_run(main, n_qubits=2, n_shots=1, seed=12).results[0].entries
+    # deterministic - should always be 0
+    # with buggy crz implementation which assumes angle is equivalent modulo 2pi,
```
acl-cqc, line 166: I'm not sure this comment is really helpful here, I mean the test is such that the bug doesn't matter here right

tests/integration/test_emulator.py
```diff
@@ -11,6 +12,24 @@
...
+    maybe_qubit,
+    project_z,
+    z,
+    s,
+    cx,
+    crz,
+    discard,
+)
+
+# from selene_sim import build
```
acl-cqc: looks to me you can drop this comment now?

tests/integration/test_emulator.py
```diff
@@ -11,6 +12,24 @@
 from selene_sim.event_hooks import NoEventHook
 
 
+import pytest
+from guppylang.std.angles import pi
+from guppylang.std.builtins import array, comptime, exit, panic
+from guppylang.std.qsystem.utils import get_current_shot
+from guppylang.std.quantum import (
```
acl-cqc: nit: good to combine this with the other `from guppylang.std.quantum import` above (and same for `...builtins`)

tests/integration/test_emulator.py
```diff
@@ -87,3 +106,161 @@ def main() -> None:
...
+        reset(q1)
+        qfree(q1)
+        b2 = project_z(q0)
+        result("c0", b1)
+        result("c1", b2)
+        result("c2", measure(q0))
+
+    res = _build_run(main, n_qubits=2, n_shots=1, seed=12).results[0].entries
+    # deterministic - should always be 0
+    # with buggy crz implementation which assumes angle is equivalent modulo 2pi,
```
acl-cqc: ok, so here the bug applies to c0 but not c1/c2

tests/integration/test_emulator.py
```diff
@@ -87,3 +106,161 @@ def main() -> None:
...
+        z(tgt)
+
+        # bell measurement
+        cx(aux, tgt)
+        h(aux)
+
+        result("aux", measure(aux))
+        discard(tgt)
+
+    res = _build_run(main, n_qubits=2, n_shots=1, seed=12).results[0].entries
```
acl-cqc: does the seed affect this test? Doesn't look like it should, so drop it?

tests/integration/test_emulator.py
```diff
@@ -87,3 +106,161 @@ def main() -> None:
...
+    @guppy
+    def main() -> None:
+        i = get_current_shot()
+        if i % 2 == 0:
+            exit("even!", 0)
+        result("c", i)
+
+    res = _build_run(main, n_qubits=2, n_shots=N)
+    assert res == EmulatorResult(
+        [[("c", i)] if i % 2 != 0 else [("exit: even!", 0)] for i in range(10)]
```
acl-cqc:
```suggestion
        [[("c", i)] if i % 2 != 0 else [("exit: even!", 0)] for i in range(N)]
```

tests/integration/test_emulator.py
acl-cqc: Of course, it would be good to have a test where seed did matter, but I guess that's for another PR...

tests/integration/test_emulator.py
ss2165: `for qubit in q` takes ownership so wouldn't work, non-owning iteration is a big guppy
TODO. The range is a workaround since indexing is borrowing.

PR #1200, opened 2025-08-15
feat: export PartialState from emulator top level

ss2165:


guppylang/src/guppylang/emulator/__init__.py
```diff
@@ -9,13 +9,14 @@
 from .builder import EmulatorBuilder
 from .instance import EmulatorInstance
 from .result import EmulatorResult
-from .state import PartialVector, StateVector, TracedState
+from .state import PartialState, PartialVector, StateVector, TracedState
```
CalMacCQ, line 12: Should `NotSingleStateError` be part of `__all__` as well?

CalMacCQ (approved):

guppylang/src/guppylang/emulator/__init__.py
ss2165: don't think so

CalMacCQ:
Guess I'm not on the Guppy github team  o.0

PR #1208, opened 2025-08-18
docs: use results sequence protocol for simplicity

ss2165:


uv.lock
mark-koch, line 1: Why the big lockfile diff?

ss2165: I suspect uv version, every `upload_time` has become `upload_time`

PR #1215, opened 2025-08-18
chore!: Delete unused old iterator AST nodes

mark-koch:
These are not used and should have been deleted in #833

BREAKING CHANGE: Deleted `guppylang_internals.nodes.{IterHasNext, IterEnd}`


acl-cqc (approved):
Obvs the python typechecker would complain if we were referencing IterEnd / IterHasNext anywhere, right...

PR #1227, opened 2025-08-21
docs: use root README for pypi page

CalMacCQ:
I think the README in the repository root with the logo, links etc is a better thing to put on the pypi page than the one inside the `guppylang` subdirectory.

Relevant comment from @aborgna-q 
https://github.com/CQCL/guppylang/pull/1181#discussion_r2269374023

Relative link should work I think? 
https://packaging.python.org/en/latest/specifications/pyproject-toml/#readme

### DIFF
--- a/guppylang/pyproject.toml
+++ b/guppylang/pyproject.toml
@@ -4,7 +4,7 @@ version = "0.21.3"
 requires-python = ">=3.10,<4"
 description = "Pythonic quantum-classical programming language"
 license = { file = "LICENCE" }
-readme = "README.md"
+readme = "../README.md"
 authors = [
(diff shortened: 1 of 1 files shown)


CalMacCQ:
> The repository README is a landing page for the whole repo, and normally a bit more oriented towards people wanting to get into the internals of the library.
> 
> Could we instead copy the interesting parts into the package README?

I see your point. However I think all of the stuff on the root README concerns guppylang rather than guppy-internals. IMO all of the content which is in the root README is the kind of thing that makes sense to include in a pypi page. 

I think its important to have a README in the root. Copying over the interesting stuff seems like just having two identical READMEs in different places.

PR #1228, opened 2025-08-22
test: Add pytket->selene execution test

aborgna-q:
Adds a couple tests:
- Running the emulator on a program with `qsystem` primitives, and complex angle expressions.
- Running the emulator on a program translated from `pytket`, with the same qsystem primitives and expressions.

The second case fails with `tket 0.12.2` :)

~Kept as draft until we update tket with the fix.~

### DIFF
--- a/tests/integration/test_pytket_circuits.py
+++ b/tests/integration/test_pytket_circuits.py
@@ -408,20 +408,21 @@ def test_qsystem_exec():
     circ.H(1)
 
     # Full rotation, just an identity
-    # ZZMax() ∘ ZZPhase(3𝜋/2) = ZZPhase(2𝜋) = I
+    # ZZMax() ∘ ZZPhase(-7/2) = ZZPhase(-4) = I
     circ.ZZMax(qubit0=1, qubit1=0)
-    circ.ZZPhase(angle=sympify("(pi/2) * 3"), qubit0=0, qubit1=1)
+    circ.ZZPhase(angle=sympify("-(7/2)"), qubit0=0, qubit1=1)
     # Another id operation
-    # PhasedX(2𝜋, -𝜋/3) = I
+    # PhasedX(2, -1/3) = I
     circ.PhasedX(
-        angle0=sympify("(3 * pi/2) / 3 - (-3 * pi/2)"),
-        angle1=sympify("-(pi/3)"),
+        angle0=sympify("(3/2) / 3 - (-3 * 1/2)"),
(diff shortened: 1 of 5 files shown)


tests/integration/test_pytket_circuits.py
```diff
@@ -394,3 +396,49 @@ def foo(q1: qubit, q2: qubit) -> None:
...
+
+
+def _build_run(
+    program: GuppyFunctionDefinition,
+    n_qubits: int = 4,
+    n_shots: int = 1,
+    seed: int | None = None,
+) -> EmulatorResult:
+    return program.emulator(n_qubits).with_shots(n_shots).with_seed(seed).run()
+
```
ss2165: this is a leftover  from some legacy integration tests, inline

tests/integration/test_pytket_circuits.py
```diff
@@ -394,3 +396,49 @@ def foo(q1: qubit, q2: qubit) -> None:
...
+@pytest.mark.skipif(not tket_installed, reason="Tket is not installed")
+def test_qsystem_exec():
+    from pytket import Circuit
+    from sympy import sympify
+
+    circ = Circuit(2)
+    circ.H(0)
+    circ.H(1)
+    circ.ZZMax(qubit0=1, qubit1=0)
+    circ.ZZPhase(angle=sympify("(pi/2) * 3"), qubit0=0, qubit1=1)
```
ss2165: some comments about what the circuit is doing would be good (why it is identity)

tests/integration/test_pytket_circuits.py
```diff
@@ -394,3 +396,48 @@ def foo(q1: qubit, q2: qubit) -> None:
...
+    def main() -> None:
+        a, b = qubit(), qubit()
+
+        guppy_circ(a, b)
+
+        result("a", measure(a))
+        result("b", measure(b))
+
+    # deterministic - should always be 0
+    res = main.emulator(n_qubits=2).with_shots(3).with_seed(42).run()
```
ss2165: if its deterministic could leave the seed unspecified?

aborgna-q: It's a bit annoying for reproducibility if the test eventually fails, but it's true that removing the seed covers more cases.

aborgna-q: Case in point >.<
https://github.com/CQCL/guppylang/actions/runs/17269645068/job/49010582469#step:10:254

PR #1232, opened 2025-08-26
fix: Remove module from big array benchmark

### LINKED ISSUE #1132: `big_array` benchmark is broken
It still uses explicit modules

tatiana-s:
Closes #1132


ferbetanzo (approved):
LGTM. Thx for adding the executable test case :)

tests/benchmarks/test_big_array.py
```diff
@@ -1,281 +1,281 @@
...
+    benchmark.extra_info["bytes"] = len(b)
+
+
+def test_big_array_executable(benchmark) -> None:
+    def big_array_executable():
+        return main.emulator(n_q+n_a*n_q)
 
-    _json_str: str = benchmark(big_array_json, big_array_compile())
-    benchmark.extra_info["bytes"] = len(_json_str.encode("utf-8"))
+    benchmark(big_array_executable)
```
ferbetanzo, line 281: If the intention is to benchmark the emulation with the default Selene-quest plugin, then maybe we're missing a `.run()` here? Otherwise, we would just be benchmarking the creation of the `EmulatorBuilder`, I think.

croyzor (approved):

PR #1237, opened 2025-08-28
fix(emulator): update selene compiler to pull in cz fix

ss2165:


uv.lock
```diff
@@ -1,5 +1,5 @@
 version = 1
-revision = 3
+revision = 2
```
aborgna-q: Pls `uv self update`

PR #1240, opened 2025-08-29
fix: comptime should act as identity in python context 

### LINKED ISSUE #1239: comptime/py does not act as identity in python context
Because it calls `tuple` on inputs

ss2165:
Closes #1239

drive-by: Ruff fixes

### DIFF
--- a/tests/integration/tracing/test_basic.py
+++ b/tests/integration/tracing/test_basic.py
@@ -89,3 +89,17 @@ def test_inner_scope(validate):
     foo, bar = make(42)
     validate(foo.compile())
     validate(bar.compile())
+
+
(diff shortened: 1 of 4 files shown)


tests/integration/tracing/test_basic.py
```diff
@@ -89,3 +89,17 @@ def bar(xs: array[int, comptime(n)]) -> None:
...
+
+    def make(n: int):
+        @guppy.comptime
+        def foo() -> int:
+            return comptime(n)
+
+        return foo
+
+    foo = make(42)
+    validate(foo.compile())
```
croyzor: Can we check here that 42 is returned with selene?

PR #1241, opened 2025-09-01
chore: set up continuous benchmarking with bencher

### LINKED ISSUE #795: Run benchmarks on CI and export to bencher
Currently benchmarks are disabled on CI because they are slow to run with other tests, we should enable them as a non-required check and export the results.

ferbetanzo:
Closes #795.

Includes the following changes:
- New `bench_upload` just command that uses bencher for uploading latency and custom metrics.
- New `bencher.py` that creates a file in Bencher Metric Format (BMF) with the custom metrics (hugr_bytes and hugr_nodes) extracted from the `pytest-benchmark` report.
- New workflows for continuous benchmarking of base (main) on push, and PRs against base.
- Set up `bencher_cli` in devenv (went with installation via `curl` instead of `cargo install` as the latter was slow, but there might be a better way of installing the tool).
- Add `check` and `executable` latency benchmark tests in `test_ctrl_flow.py`.


justfile
```diff
@@ -63,3 +63,24 @@ bench *PYTEST_FLAGS:
...
     uv run pytest --benchmark-only --benchmark-storage={{path}} --benchmark-save={{name}}
+
+# Run benchmarks and upload the results using the bencher `python_pytest` adapter
+# (https://bencher.dev/docs/explanation/adapters/#-python-pytest).
+# Also, create a bencher BNF with custom metrics for the bytes and nodes of the
+# generated hugr, and upload it using the bencher `json` adapter.
+# Note: Needs the BENCHER_API_TOKEN env variable
+NOW := `date +%s%n | tr -d '\n'`
+BENCHER_PROJECT := "guppylang-benchmarks"
+bench_upload *BENCHER_FLAGS:
```
aborgna-q: Move the comment to just before the recipe, so it gets shown when running `just` / `just help`.

justfile
```diff
@@ -63,3 +63,24 @@ bench *PYTEST_FLAGS:
...
+    uv run pytest --benchmark-only --benchmark-json="{{NOW}}-pytest-benchmark.json"
```
aborgna-q, line 73: Add a check to see if `bencher` is installed first? So we can point towards the installation instructions
https://bencher.dev/docs/tutorial/quick-start/

We should also list that requirement in DEVELOPMENT.md

.github/workflows/benchmarks_base.yml
aborgna-q, line 1: How stable are wall-clock benchmarks in CI?

One of the main advantages of using codspeed (apart from the UI/UX and github integration) was that it automatically [sets up single-shot benchmarking](https://codspeed.io/docs/benchmarks/python) in CI, so it gives really stable results.
What that is missing though is non-time benchmark metrics.

I don't know how hard it is to setup valgrind-based tests in python otherwise.

We could perhaps do a mix of timing tests in codspeed and other metrics in bencher?

justfile
ferbetanzo: I'm doing that in `devenv.nix > enterShell` as I think all required tools are installed/set up there.

.github/workflows/benchmarks_base.yml
ferbetanzo: I've just done some manual checks to see how the benchmarks would behave, specifically when checking PRs in the CI. It seems if we use the default average (`mean`) when comparing the latency there are some outliers in the `test_import_guppy` benchmark that make the comparison fail (for example it sometimes takes seconds to import guppy-related modules). But if you use the `median` metric, those outliers do not impact the final number and then it's ok. The other benchmarks do not seem to have that much variation between runs, in terms of latency.

The other custom metrics hugr_bytes and hugr_nodes should, in principle, have no variation between runs, so the threshold there is set somewhat tight and we would get a failure anytime the numbers change.

We could indeed use codspeed for latency, and bencher for the hugr-related metrics, but I'm afraid that would make running the benchmarks and comparing them locally a bit more convoluted.

justfile
ferbetanzo: I've added a section on how to run the benchmarks to the DEVELOPMENT.MD too.

ferbetanzo:
> This PR contains breaking changes to the public Python API. Please deprecate the old API instead (if possible), or mark the PR with a `!` to indicate a breaking change.
> 
> Breaking changes summary
> ```
> 
> ```

@aborgna-q would you know why it's identifying this as a breaking change?

aborgna-q:
> RuntimeError: Could not create git worktree: fatal: invalid reference: 2f0648ecb676ee441cfc0199001e9f671e8c7975

Weird. It's failing to fetch a (valid) commit from github??

aborgna-q (approved):

PR #1245, opened 2025-09-03
feat: Preserve previous results upon error

jake-arkinstall:
Main changes:
- Catch exceptions that occur on the inner iterator of selene's run_shots
- Raise a new EmulatorError that captures existing results
- Replace panic tests with tests that verify this behaviour upon guppy-encoded errors and internal errors

Controversial:
- Mocking previously wanted the mock object to be passed directly to EmulatorResult, but this is no longer viable as we need to iterate over it.
  - The mock object cannot (I believe?) be iterated over.
  - I don't know the best way to manage that so I've broken the 4th wall, which I'm not proud of.
  - A TODO comment is left to note this.
  - Suggestions welcome.

Drive-by:
- Add a progress bar when chained with `.with_progress_bar()`

### DIFF
--- a/guppylang/src/guppylang/emulator/instance.py
+++ b/guppylang/src/guppylang/emulator/instance.py
@@ -256,11 +256,7 @@ class EmulatorInstance:
     ) -> Iterator[Iterator[TaggedResult]]:
         """Iterate over the shots in the result stream, optionally displaying a progress
         bar."""
-        if hasattr(result_stream, "assert_called"):
-            # catch mocks
-            # TODO: this is breaking the 4th wall, and I don't like it.
-            return iter([])
-        elif self._options._display_progress_bar:
+        if self._options._display_progress_bar:
             return cast(
                 "Iterator[Iterator[TaggedResult]]",
                 tqdm(result_stream, total=self.shots, desc="Emulating shots"),
--- a/tests/emulator/test_instance.py
+++ b/tests/emulator/test_instance.py
@@ -11,6 +11,7 @@ from selene_sim.backends.bundled_simulators import Coinflip
 from selene_soft_rz_runtime_plugin import SoftRZRuntimePlugin
 
 from guppylang.emulator.instance import EmulatorInstance
+from guppylang.emulator.result import EmulatorResult
 
 
 def test_emulator_instance_init():
@@ -296,7 +297,7 @@ def test_emulator_instance_run(mock_iterate):
     mock_result_stream = Mock()
     mock_selene_instance.run_shots.return_value = mock_result_stream
 
-    instance.run()
+    assert instance.run() == EmulatorResult()
 
     # Check that EmulatorResult was created with the result stream
     mock_iterate.assert_called_once_with(mock_result_stream)


guppylang/src/guppylang/emulator/instance.py
```diff
@@ -206,9 +216,22 @@ def run(self) -> EmulatorResult:
...
+                for tag, value in shot:
+                    shot_results.append(tag, value)
+            except Exception as e:  # noqa: BLE001
+                # In this case, casting a wide net on exceptions is
+                # suitable.
+                raise EmulatorError(
+                    completed_shots=EmulatorResult(all_results),
+                    failing_shot=shot_results,
+                    underlying_exception=e,
+                ) from None
```
ss2165, line 232: why not `from e`? too much noise?

guppylang/src/guppylang/emulator/exceptions.py
```diff
@@ -0,0 +1,20 @@
...
+
+from .result import EmulatorResult
+
+
+class EmulatorError(Exception):
+    completed_shots: EmulatorResult
+    failing_shot: QsysShot
+    underlying_exception: Exception | None
+
+    def __init__(
```
ss2165, line 11: suggestion: a `failed_shot_index` property that tells you where you stopped

guppylang/src/guppylang/emulator/instance.py
ss2165: Answer: underlying error is stored inside re-raise, nicer message

guppylang/src/guppylang/emulator/result.py
```diff
@@ -64,9 +63,7 @@ class EmulatorResult(QsysResult):
 
     # Re-define parent methods for documentation purposes
 
-    def __init__(
-        self, results: Iterable[QsysShot | Iterable[TaggedResult]] | None = None
-    ):
+    def __init__(self, results: list[QsysShot] | None = None):
```
ss2165, line 66: why change this annotation? it matches the parent type

guppylang/src/guppylang/emulator/instance.py
```diff
@@ -227,3 +250,20 @@ def _run_instance(self) -> Iterator[Iterator[TaggedResult]]:
...
             n_processes=self.n_processes,
         )
+
+    def _iterate_shots(
+        self, result_stream: Iterator[Iterator[TaggedResult]]
+    ) -> Iterator[Iterator[TaggedResult]]:
+        """Iterate over the shots in the result stream, optionally displaying a progress
+        bar."""
+        if hasattr(result_stream, "assert_called"):
+            # catch mocks
```
ss2165: you can tell the mock object what to return when __iter__ is called

https://stackoverflow.com/a/66065390

ss2165 (changes requested):
Requesting changes for the mock 4th wall break

guppylang/src/guppylang/emulator/result.py
jake-arkinstall: The generic iterable interface was required for iterating through the emulator's generator, but it is now always just passed in directly as a concrete list. It seemed better to use the more concrete type, but I can revert

PR #1247, opened 2025-09-04
fix: Update adaptive-qpe.ipynb to add pip install guppylang at the start of the notebook

nelslindahlx:
fix: needed to add pip install guppylang at the start of the notebook


ss2165:
We prefer to not affect user's environments when running notebooks

PR #1258, opened 2025-09-11
fix!: Fix `nat` constructor in comptime functions

### LINKED ISSUE #1250: `@guppy.comptime` error when defining `nat` variable
```python
n_var = 10

@guppy.comptime
def main() -> None:
    nat(n_var)
    
main.compile()
```

Error:
```
Cell In[11], line 7
      3 @guppy.comptime
      4 def main() -> None:
      5     nat(n_var)
----> 7 main.compile()

Cell In[11], line 5
      3 @guppy.comptime
      4 def main() -> None:
----> 5      nat(n_var)

GuppyComptimeError: Unsupported Python expression: Expression of type `<class 'type'>` is not supported
```

mark-koch:
Fixes #1250

The problem was that the imported `nat` is a class declared using `@extend_type`. This decorator didn't wrap the result into a `GuppyDefinition`, so was not equipped with comptime behaviour.

I changed the `@extend_type` decorator to now return `GuppyObject`s by default. Note that we still need an escape hatch to return the unwrapped class to support the `array` type whose comptime behaviour is customised to act like lists. This can be achieved by setting the new `return_class=True` flag in the decotator.

Note that this change is not breaking for `guppylang` since the deprecated `@guppy.extend_type` decorator still behaves as before.

drive-by: Added a `run_nat_function` pytest fixture to test nat functions.

BREAKING CHANGE: `guppylang_internals.decorator.extend_type` now returns a `GuppyDefinition` by default. To get the previous behaviour of returning the annotated class unchanged, pass `return_class=True`.

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -3,7 +3,8 @@ from hugr.package import Package, PackagePointer
 
 from pathlib import Path
 import pytest
-from typing import Any, Literal, assert_never
+from typing import Any, Literal
+from typing_extensions import assert_never
 
 from selene_hugr_qis_compiler import check_hugr
 


croyzor:
> Note that this change is not breaking since the deprecated @guppy.extend_type decorator still behaves as before

But before it returned a python class and now it returns a `GuppyDefinition` by default?

mark-koch:
> But before it returned a python class and now it returns a GuppyDefinition by default?

Yeah, this would be breaking for `guppylang_internals` but not for `guppylang` (since the `extend_type` decorator is in `guppylang_internals` and `guppy.extend_type` always sets `return_class=True`)

Unfortunately, release-please can't make that distinction if the PR modifies files in both packages 🫠

PR #1267, opened 2025-09-19
feat: Bump selene-hugr-qis-compiler version

croyzor:
Bump selene testing dep to a version that uses the same tket version as guppy, so doesn't have a wasm extension version mismatch


ss2165 (approved):

qartik (approved):

croyzor:
Bencher CI is failing because of 1c06d94d755c1ec889cf9ecdcc35a804c9763cc8. Something went wrong with adding LLVM to the environment, which caused the benchmarking to fail, hence there's no existing result for the current HEAD of `main`, leading to a 404 error when trying to build a comparison for this PR.

Merging this - reverting the changes that required an llvm install - should bring us back to the status quo

PR #1269, opened 2025-09-22
fix: update logo display to work on pypi page.

CalMacCQ:
in #1227 I made an update to use the root README for the guppylang package description. After the 0.21.4 release I see this has worked and the text is updated.

As you can see, the fish logo isn't showing properly. 

Screenshot 2025-09-22 at 11 42 13

According to https://stackoverflow.com/questions/41983209/how-do-i-add-images-to-a-pypi-readme-that-works-on-github the fix is to link to the raw github user content rather than just the path to the directory. Hopefully this works.


croyzor (approved):
:shipit:

PR #1271, opened 2025-09-22
feat: `compile_entrypoint` and `compile_function` methods

### LINKED ISSUE #1162: `GuppyFunctionDefinition.compile` should optionally check for arguments to entrypoint

ss2165:
Closes #1162

Introduces `GuppyFunctionDef` methods `compile_function` which behaves exactly like `compile` did before, and `compile_entrypoint` which checks the entrypoint function has no arguments. `compile` now is an alias for `compile_entrypoint`

This technically breaks existing behaviour even if API change is non-breaking. If `compile` was used to produce HUGR rather than executables, they would now need to call `compile_function` instead.
This is not deemed common enough to introduce a breaking version, but should still be made clear in the CHANGELOG.

Tests are updated to call the new function.
In tests of correct error paths, usually no change is required since this error is triggered after all other compilation is complete.

### DIFF
--- a/guppylang/src/guppylang/defs.py
+++ b/guppylang/src/guppylang/defs.py
@@ -4,12 +4,12 @@ These are the objects returned by the `@guppy` decorator. They should not be con
 with the compiler-internal definition objects in the `definitions` module.
 """
 
+import ast
 from collections.abc import Sequence
 from dataclasses import dataclass
 from typing import Any, ClassVar, Generic, ParamSpec, TypeVar, cast
 
 import guppylang_internals
-from guppylang_internals.definition.function import CheckedFunctionDef
 from guppylang_internals.definition.value import CompiledCallableDef
 from guppylang_internals.diagnostic import Error
 from guppylang_internals.engine import ENGINE, CoreMetadataKeys
@@ -45,7 +45,7 @@ def _update_generator_metadata(hugr: Hugr[Any]) -> None:
 class EntrypointArgsError(Error):
     title: ClassVar[str] = "Entrypoint function has arguments"
     span_label: ClassVar[str] = (
-        "Entrypoint function must have no input parameters, found {input_names}."
+        "Entrypoint function must have no input parameters, found ({input_names})."
     )
     args: Sequence[str]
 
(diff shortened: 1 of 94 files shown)


guppylang/src/guppylang/defs.py
```diff
@@ -35,11 +41,23 @@ def _update_generator_metadata(hugr: Hugr[Any]) -> None:
     }
 
 
+@dataclass(frozen=True)
+class EntrypointArgsError(Error):
+    title: ClassVar[str] = "Entrypoint function has arguments"
+    span_label: ClassVar[str] = "Entrypoint function " "has arguments: {input_names}."
```
qartik:
```suggestion
    span_label: ClassVar[str] = "Entrypoint function has arguments: {input_names}."
```
any particular reason why this string is broken into two?

ss2165: copy paste artefact - fixed

guppylang/src/guppylang/defs.py
```diff
@@ -78,11 +96,49 @@ def emulator(
...
+        pack = super().compile(entrypoint=entrypoint)
+        # entrypoint cannot be polymorphic
+        monomorphized_id = (self.id, ())
+        compiled_def = ENGINE.compiled.get(monomorphized_id)
+        if (
+            entrypoint
+            and isinstance(compiled_def, CompiledCallableDef)
+            and len(compiled_def.ty.inputs) > 0
+        ):
+            # Check if the entrypoint being has arguments
```
mark-koch:
```suggestion
            # Check if the entrypoint has arguments
```

guppylang/src/guppylang/defs.py
```diff
@@ -78,11 +96,49 @@ def emulator(
...
+            checked = cast(CheckedFunctionDef, ENGINE.checked[self.id])
```
mark-koch: This may could also be another kind of callable definition (e.g. comptime function, pytket, etc). Also, why not extract the span straight from `compiled_def`?

guppylang/src/guppylang/defs.py
```diff
@@ -35,11 +41,23 @@ def _update_generator_metadata(hugr: Hugr[Any]) -> None:
     }
 
 
+@dataclass(frozen=True)
+class EntrypointArgsError(Error):
+    title: ClassVar[str] = "Entrypoint function has arguments"
+    span_label: ClassVar[str] = "Entrypoint function has arguments: {input_names}."
```
mark-koch: Maybe something like "Entrypoint function is not allowed to take arguments" would be clearer?

Also a hint telling the user about `.compile(entrypoint=False)` could be good?

qartik: alternatively, "Entrypoint function must have no input parameters (found ...)"

guppylang/src/guppylang/defs.py
```diff
@@ -83,6 +103,52 @@ def emulator(
...
+
+        Returns:
+            Package: The compiled package object.
+        Raises:
+            GuppyError: If the entrypoint has arguments.
+        """
+
+        return self.compile_entrypoint()
+
+    def compile_entrypoint(self) -> Package:
```
mark-koch, line 129: Missing docstring

PR #1274, opened 2025-09-23
chore: fix exporting integration tests

ferbetanzo:
There's an issue with the exporting of hugrs generated while running integration tests. We just need to change the `validate` fixture to have `scope="function"`  (the default one), otherwise the hugr files are overwritten as the `request.node_name` is always None.


acl-cqc (approved):
Good spot Fernando, thanks!

Issue #1130, opened 2025-08-04
Snapshot error when parallelizing tests in Python >=3.13


ss2165:
[`python_err.py`](https://github.com/CQCL/guppylang/blob/main/tests/error/comptime_expr_errors/python_err.py) fails with

```
E       AssertionError: value does not match the expected value in snapshot tests/error/comptime_expr_errors/python_err.err
E         (run pytest with --snapshot-update to update snapshots)
E       assert 'Error: Pytho...vious error\n' == 'Error: Pytho...vious error\n'
E         
E         Skipping 291 identical leading characters in diff, use -v to show
E            <module>
E         +     import sys;exec(eval(sys.stdin.readline()))
E         +     ^^^^^
E           ZeroDivisionError: division by zero
E           
E           Guppy compilation failed due to 1 previous error
```

when running `just test` (uv run pytest -n auto -x)

Appears to be a problem with parallel testing

mark-koch:
Looks to be Python 3.13 specific: https://github.com/pytest-dev/pytest/blob/f3da9187d0b7c2b0a1131ba4002cc55fda34bd5c/testing/code/test_excinfo.py#L379-L381

nicolaassolini-qntm:
Some inconsistency related to this issue:

- On local machines:`uv run pytest -k python_err` fails while `just test` not
- While `just test` succeed local, `python_err.py` fails on CI

PR #1275, opened 2025-09-24
Protocol definition parsing

tatiana-s:
Closes [#1209](https://github.com/CQCL/guppylang/issues/1209) (to be merged into `protocol-main` as the clean protocols work in progress branch)

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/protocol.py
+++ b/guppylang-internals/src/guppylang_internals/definition/protocol.py
@@ -1,0 +1,149 @@
+import ast
+import sys
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, ClassVar
+
+from guppylang_internals.ast_util import AstNode, has_empty_body
+from guppylang_internals.checker.core import Globals
+from guppylang_internals.checker.errors.generic import (
+    UnexpectedError,
+    UnsupportedError,
+)
+from guppylang_internals.definition.common import (
+    CompiledDef,
+    DefId,
+    Definition,
+    ParsableDef,
+)
+from guppylang_internals.definition.struct import (
+    NonGuppyMethodError,
+    RedundantParamsError,
+    params_from_ast,
+    parse_py_class,
+    try_parse_generic_base,
+)
+from guppylang_internals.diagnostic import Help
+from guppylang_internals.engine import DEF_STORE
+from guppylang_internals.error import GuppyError
+from guppylang_internals.span import SourceMap, Span, to_span
+from guppylang_internals.tys.arg import Argument
+from guppylang_internals.tys.param import Parameter, check_all_args
+from guppylang_internals.tys.protocol import ProtocolInst
+
+if TYPE_CHECKING:
+    from guppylang_internals.diagnostic import Error
+
+if sys.version_info >= (3, 12):
+    from guppylang_internals.tys.parsing import parse_parameter
+
+
+@dataclass(frozen=True)
+class ProtocolDef(Definition):
+    """Abstract base class for protocol definitions."""
+
+    description: str = field(default="protocol", init=False)
+
+
+@dataclass(frozen=True)
+class DeclarationHint(Help):
+    message: ClassVar[str] = (
+        "Add a `@guppy.declare` annotation and leave the body empty to turn this into "
+        "a declaration"
+    )
+
+
+@dataclass(frozen=True)
+class RawProtocolDef(ProtocolDef, ParsableDef):
+    """A raw protocol definition that has not been parsed yet."""
+
+    python_class: type
+
+    def parse(self, globals: Globals, sources: SourceMap) -> "CheckedProtocolDef":
+        """Parses the raw class object into an AST and checks that it is well-formed."""
+        # Mostly copied from RawStructDef.parse, but only allowing function declarations
+        # in the body and allowing `Protocol` as a base class.
+        frame = DEF_STORE.frames[self.id]
+        cls_def = parse_py_class(self.python_class, frame, sources)
+        if cls_def.keywords:
+            raise GuppyError(UnexpectedError(cls_def.keywords[0], "keyword"))
+
+        # Look for generic parameters from Python 3.12 style syntax.
+        params = []
+        params_span: Span | None = None
+        if sys.version_info >= (3, 12):
+            if cls_def.type_params:
+                first, last = cls_def.type_params[0], cls_def.type_params[-1]
+                params_span = Span(to_span(first).start, to_span(last).end)
+                params = [
+                    parse_parameter(node, idx, globals)
+                    for idx, node in enumerate(cls_def.type_params)
+                ]
+
+        match cls_def.bases:
+            case []:
+                pass
+            # We allow `Generic[...]` to specify  parameters with the legacy syntax.
+            case [base] if elems := try_parse_generic_base(base):
+                # Complain if we already have Python 3.12 generic params
+                if params_span is not None:
+                    err: Error = RedundantParamsError(base, self.name)
+                    err.add_sub_diagnostic(RedundantParamsError.PrevSpec(params_span))
+                    raise GuppyError(err)
+                params = params_from_ast(elems, globals)
+            # Specifying `Protocol` is redundant but we allow it optionally.
+            case [base] if isinstance(base, ast.Name) and base.id == "Protocol":
+                pass
+            case bases:
+                err = UnsupportedError(bases[0], "Protocol inheritance", singular=True)
+                raise GuppyError(err)
+
+        func_defs: dict[str, DefId] = {}
+        for i, node in enumerate(cls_def.body):
+            match i, node:
+                # Docstrings are fine if they occur at the start.
+                case 0, ast.Expr(value=ast.Constant(value=v)) if isinstance(v, str):
+                    pass
+                # Ensure that all function definitions are Guppy declarations.
+                case _, ast.FunctionDef(name=name) as node:
+                    from guppylang.defs import GuppyDefinition
+
+                    v = getattr(self.python_class, name)
+                    if not isinstance(v, GuppyDefinition):
+                        raise GuppyError(NonGuppyMethodError(node, self.name, name))
+                    if not has_empty_body(node):
+                        err = UnexpectedError(
+                            node, "function body", unexpected_in="protocol definition"
+                        )
+                        err.add_sub_diagnostic(DeclarationHint(None))
+                        raise GuppyError(err)
+                    # Store the definition ID and then get the type through the engine
+                    # during protocol checking, avoiding re-parsing the signature here.
+                    func_defs[name] = v.wrapped.id
+                # Fields are not allowed in protocols.
+                case _, ast.AnnAssign(target=ast.Name(_)) as node:
+                    err = UnsupportedError(
+                        node, "fields", unsupported_in="a protocol definition"
+                    )
+                    raise GuppyError(err)
+                case _, node:
+                    err = UnexpectedError(
+                        node, "statement", unexpected_in="protocol definition"
+                    )
+                    raise GuppyError(err)
+
+        return CheckedProtocolDef(self.id, self.name, cls_def, params, func_defs)
+
+
+@dataclass(frozen=True)
+class CheckedProtocolDef(ProtocolDef, CompiledDef):
+    defined_at: ast.ClassDef
+    params: Sequence[Parameter]
+    members: Mapping[str, DefId]
+
+    def check_instantiate(
+        self, args: Sequence[Argument], loc: AstNode | None = None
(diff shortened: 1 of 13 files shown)


guppylang-internals/src/guppylang_internals/definition/protocol.py
```diff
@@ -0,0 +1,149 @@
...
+            # We allow `Generic[...]` to specify  parameters with the legacy syntax.
+            case [base] if elems := try_parse_generic_base(base):
+                # Complain if we already have Python 3.12 generic params
+                if params_span is not None:
+                    err: Error = RedundantParamsError(base, self.name)
+                    err.add_sub_diagnostic(RedundantParamsError.PrevSpec(params_span))
+                    raise GuppyError(err)
+                params = params_from_ast(elems, globals)
+            # Specifying `Protocol` is redundant but we allow it optionally.
+            case [base] if isinstance(base, ast.Name) and base.id == "Protocol":
```
mark-koch, line 104: Python also allows specifying legacy generic parameters via a `Protocol[T]` base: https://typing.python.org/en/latest/spec/protocol.html#generic-protocols

Do we also want this?

guppylang-internals/src/guppylang_internals/definition/protocol.py
```diff
@@ -0,0 +1,149 @@
...
+    python_class: type
+
+    def parse(self, globals: Globals, sources: SourceMap) -> "CheckedProtocolDef":
+        """Parses the raw class object into an AST and checks that it is well-formed."""
+        # Mostly copied from RawStructDef.parse, but only allowing function declarations
+        # in the body and allowing `Protocol` as a base class.
+        frame = DEF_STORE.frames[self.id]
+        cls_def = parse_py_class(self.python_class, frame, sources)
+        if cls_def.keywords:
+            raise GuppyError(UnexpectedError(cls_def.keywords[0], "keyword"))
```
mark-koch, line 72: Could you add a test that triggers this error?

mark-koch: Same for the other errors in this file without coverage

guppylang-internals/src/guppylang_internals/definition/protocol.py
```diff
@@ -0,0 +1,149 @@
...
+                # Docstrings are fine if they occur at the start.
+                case 0, ast.Expr(value=ast.Constant(value=v)) if isinstance(v, str):
+                    pass
+                # Ensure that all function definitions are Guppy declarations.
+                case _, ast.FunctionDef(name=name) as node:
+                    from guppylang.defs import GuppyDefinition
+
+                    v = getattr(self.python_class, name)
+                    if not isinstance(v, GuppyDefinition):
+                        raise GuppyError(NonGuppyMethodError(node, self.name, name))
```
mark-koch: Do we actually want to require `@guppy.declare` annotations on protocol methods? We could also leave them unannotated and parse them here?

guppylang/src/guppylang/decorator.py
```diff
@@ -157,6 +158,34 @@ def add_fields(self: "MyStruct") -> int:
...
+        frame = get_calling_frame()
+        DEF_STORE.register_def(defn, frame)
+        for val in cls.__dict__.values():
+            if isinstance(val, GuppyDefinition):
+                DEF_STORE.register_impl(defn.id, val.wrapped.name, val.id)
+        # Prior to Python 3.13, the `__firstlineno__` attribute on classes is not set.
+        # However, we need this information to precisely look up the source for the
+        # class later. If it's not there, we can set it from the calling frame:
+        if not hasattr(cls, "__firstlineno__"):
+            cls.__firstlineno__ = frame.f_lineno  # type: ignore[attr-defined]
```
mark-koch: This is obscure enough that it might be worth putting into a helper function now that we have this twice

tests/error/protocol_errors/field.err
```diff
@@ -0,0 +1,8 @@
+Error: Unsupported (at $FILE:5:4)
+  | 
+3 | @guppy.protocol
+4 | class MyProto:
+5 |     id: int
+  |     ^^^^^^^ fields are not supported in a protocol definition
```
mark-koch: Nit: should be capitalised to be consistent with other errors

tests/error/struct_errors/protocol_base.err
```diff
@@ -0,0 +1,10 @@
+Error: Unsupported (at $FILE:6:14)
+  | 
+4 | 
+5 | @guppy.struct
+6 | class MyProto(Protocol):
+  |               ^^^^^^^^ Protocol base is not supported
+
+Help: Add a `@guppy.protocol` annotation to turn this struct into a protocol
```
mark-koch, line 8: Nice!

guppylang-internals/src/guppylang_internals/definition/protocol.py
tatiana-s: If we just parse them here the issue is that you can't register them with the compilation engine at this point - would it be fine to just store function types in the protocol independent from anything else? I am not sure if that will cause issues with parameters

mark-koch: > would it be fine to just store function types in the protocol independent from anything else?

Mhmm it depends how we want to go about checking/compiling calls to protocol methods. My gut feeling is that storing the signatures in the protocol definition should be fine, but maybe good to discuss this in more detail?

One further subtlety is that it used to be the case that type definitions needed to be parsed and checked before functions (in order to resolve their signature). But with the compilation engine I'm not sure if that still applies since things are parsed/checked on demand rather than all in one go.

mark-koch: Either way, I think function declarations are maybe the wrong fit - either we store them in the signature or we need a new kind of `ProtocolMethodDef` function definition?

tests/error/comptime_expr_errors/python_err.err
```diff
@@ -9,6 +9,8 @@ Traceback printed below:
 
 Traceback (most recent call last):
   File "<string>", line 1, in <module>
+    import sys;exec(eval(sys.stdin.readline()))
+    ^^^^^
```
mark-koch:
```suggestion
```

This is a bug when running pytest in parallel, see #1130

tests/error/protocol_errors/function_body.err
```diff
@@ -0,0 +1,13 @@
...
+  | 
+5 | class MyProto:
+6 |     @guppy.declare
+7 |     def foo(self: "MyProto", x: int) -> str:
+  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+8 |         return "abcdef"
+  | ^^^^^^^^^^^^^^^^^^^^^^^ Unexpected function body in protocol definition
+
+Help: Add a `@guppy.declare` annotation and leave the body empty to turn this
+into a declaration
```
mark-koch: This help is a bit confusing since it's already annotated as `@guppy.declare`? I guess also depends how we proceed with the declare discussion

guppylang-internals/src/guppylang_internals/definition/protocol.py
```diff
@@ -113,7 +115,16 @@ def parse(self, globals: Globals, sources: SourceMap) -> "ParsedProtocolDef":
                     pass
                 # Parse the function definitions into types.
                 case _, ast.FunctionDef(name=name) as node:
+                    from guppylang.defs import GuppyDefinition
+
                     py_func = getattr(self.python_class, name, None)
+
+                    if isinstance(py_func, GuppyDefinition):
+                        err = UnexpectedError(
+                            node, "Guppy function", unexpected_in="protocol definition"
```
mark-koch: nit: You could use `py_func.wrapped.description` here instead of hard coding `"function"`

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -518,9 +519,10 @@ def __call__(self, *args: Any) -> Any:
     def __getitem__(self, item: Any) -> Any:
         # If this is a type definition, then `__getitem__` might be called when
         # specifying generic arguments
-        if isinstance(self.wrapped, TypeDef):
+        if isinstance(self.wrapped, TypeDef | ProtocolDef):
             # It doesn't really matter what we return here since we don't support types
             # as comptime values yet, so just give back the definition
+            # TODO: How do protocols interact with comptime?
```
mark-koch, line 525: Good question, can you create an issue and add it to the protocol tracking issue?

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -200,11 +207,37 @@ def _arg_from_instantiated_defn(
...
+    """Parses a protocol definition with type args into an argument."""
+    proto_args = [arg_from_ast(arg_node, ctx) for arg_node in arg_nodes]
+    if not proto_args:
+        proto_args = [param.to_bound() for param in proto_defn.params]
+    inst = proto_defn.check_instantiate(proto_args, node)
+    param = TypeParam(
+        len(ctx.param_var_mapping.keys()),
+        proto_defn.name,
+        must_be_copyable=False,
+        must_be_droppable=False,
```
mark-koch, line 232: I think this is the correct assumption to make for now, but in the future we probably want a way to specify that all protocol implementors must be copyable/droppable (maybe by making the protocol definition inherit from `Copy`/`Drop`?)

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -330,13 +363,42 @@ def parse_parameter(
...
+                        bound.slice.elts
+                        if isinstance(bound.slice, ast.Tuple)
+                        else [bound.slice]
+                    )
+                    proto_args = [arg_from_ast(arg_node, ctx) for arg_node in arg_nodes]
+                else:
+                    proto_defn = _try_parse_defn(bound, ctx.globals)
+                if isinstance(proto_defn, ParsedProtocolDef):
+                    if proto_args == []:
+                        proto_args = [param.to_bound() for param in proto_defn.params]
```
mark-koch: Don't we need to add these new params to the `param_var_mapping`?

Also, we probably need to specify a manual de Bruijn index, i.e. `param.to_bound(idx=len(ctx.param_var_mapping))`?

tests/error/comptime_expr_errors/python_err.err
```diff
@@ -9,8 +9,6 @@ Traceback printed below:
 
 Traceback (most recent call last):
   File "<string>", line 1, in <module>
-    import sys;exec(eval(sys.stdin.readline()))
-    ^^^^^
```
mark-koch:
```suggestion
```

tests/integration/test_protocol_py312.py
```diff
@@ -3,20 +3,72 @@
...
+
+
+def test_def_parameterised():
     @guppy.protocol
     class MyProto[T]:
-        def foo(self: "MyProto", x: T) -> str: ...
+        def foo(self: "MyProto", x: T) -> T: ...
+
+        # TODO: Implement Self support for protocols.
+        # def bar(self: Self) -> "MyProto": ...
```
mark-koch, line 24: Can you add an issue for this?

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -200,11 +207,37 @@ def _arg_from_instantiated_defn(
...
+    node: AstNode,
+    ctx: TypeParsingCtx,
+) -> Argument:
+    """Parses a protocol definition with type args into an argument."""
+    proto_args = [arg_from_ast(arg_node, ctx) for arg_node in arg_nodes]
+    if not proto_args:
+        proto_args = [param.to_bound() for param in proto_defn.params]
+    inst = proto_defn.check_instantiate(proto_args, node)
+    param = TypeParam(
+        len(ctx.param_var_mapping.keys()),
```
mark-koch:
```suggestion
        len(ctx.param_var_mapping),
```

guppylang-internals/src/guppylang_internals/tys/parsing.py
mark-koch: So if we have a generic `Proto[T]`, this means we're allowed to just write `Proto` and the `T` is added implicitly?

Not sure if this is a good thing to do (e.g. Rust doesn't allow this). Thoughts? @croyzor

Either way, if we create new params here, they will also need to be added to the `param_var_mapping`

mark-koch: Also, `param.to_bound()` will most likely give you the wrong de Bruijn index. I think it should be `param.to_bound(idx=len(param_var_mapping))`?

croyzor: I prefer to be more explicit, require writing `Proto[T]` :+1:

PR #1276, opened 2025-09-25
refactor: Move `get_instance_func` from `Globals` to `CompilationEngine`

tatiana-s:
Doesn't actually require information only available in globals and this allows you to use it without having a globals object (needed in protocol checker implementation) but not entirely sure about this


maximilianruesch:
Superseeded by #1628

PR #1277, opened 2025-09-25
chore: Add `guppylang-internals` to `just fix` and `just format` commands

tatiana-s:


ss2165 (commented):
I think we can actually just remove both packages - that is the check we run on CI after all

qartik:
> I think we can actually just remove both packages from the commands - that is the check we run on CI after all

It's probably better to encourage developers to do local checks before hitting the CI. So, I'd suggest keeping them.

PR #1281, opened 2025-09-30
feat: modifiers in CFG and its type checker (experimental)

hkengo-qtnm:
This is the first PR for implementation of modifiers (control/dagger/power).
The modifiers are implemented with python's context-manager syntax as 
```python
with control(c), dagger:
    s(t)
```
whose semantics is controlled-Sdg gate applied to (c,t).

## Contents
- AST-to-CFG lowering
- Type checking for modifiers
- 18 error tests (wish snapshots) under `tests/error/modifier_errors`, which might be a good starting point to review.
- Slightly unrelated bug fix [tests/error/util.py](https://github.com/CQCL/guppylang/compare/feat/modifier-type_check?expand=1#diff-2df045859ffae9afb594e5c0e044d96500c047b9ace5d398b711f49d9baa951a)
  + This fixes the behaviour of error tests of when no errors are found.
  + If this should be in a different commit, please exclude 35711c89ca2b4521e1b5ebff1217a0e573614e96

## Not included in this PR
- Compiler (CFG-to-Hugr lowering) for modifiers
- Unitarity checker (type checker for functions that can be called in modifier contexts) and annotations for it


guppylang-internals/src/guppylang_internals/ast_util.py
```diff
@@ -106,6 +106,12 @@ def return_nodes_in_ast(node: Any) -> list[ast.Return]:
     return cast(list[ast.Return], found)
 
 
+def loop_in_ast(node: Any) -> list[ast.For | ast.While]:
+    """Returns all `For` and `While` nodes occurring in an AST."""
+    found = find_nodes(lambda n: isinstance(n, ast.For | ast.While), node)
```
mark-koch: You probably don't want to look into nested functions?

```suggestion
    found = find_nodes(lambda n: isinstance(n, ast.For | ast.While), node, {ast.FunctionDef})
```

guppylang-internals/src/guppylang_internals/ast_util.py
```diff
@@ -117,6 +123,17 @@ def breaks_in_loop(node: Any) -> list[ast.Break]:
     return cast(list[ast.Break], found)
 
 
+def loop_controls_in_loop(node: Any) -> list[ast.Break | ast.Continue]:
+    """Returns all `Break` nodes occurring in a loop.
```
mark-koch:
```suggestion
    """Returns all `Break` and `Continue` nodes occurring in a loop.
```

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -422,3 +423,119 @@ def __init__(
...
+
+    def __init__(self, node: ast.Call, iter: ast.expr) -> None:
+        super().__init__(**node.__dict__)
+        self.iter = iter
+
+
+ModifierKind = Dagger | Control | Power
+
+
+class Modifier(ast.With):
```
mark-koch: In terms of naming, it might be clearer to use `Modifier = Dagger | Control | Power` (since dagger, control & power *are* modifiers) and call the `with` statement something else. Maybe `ModifiedSection`, `ModifiedBlock`?

What do you think?

guppylang-internals/src/guppylang_internals/cfg/bb.py
```diff
@@ -219,3 +223,27 @@ def visit_NestedFunctionDef(self, node: NestedFunctionDef) -> None:
...
+            for x, using_bb in live[node.cfg.entry_bb].items()
+            if x not in assigned_before_in_bb
+        }
+
+    def visit_Control(self, node: Control) -> None:
+        for item in node.ctrl:
+            self.visit(item)
+    
+    def visit_Power(self, node: Power) -> None:
+        self.visit(node.iter)
```
mark-koch: You don't need these visitors if they are just recursing into the structure. The `NodeVisitor` will be able to handle those automatically

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -422,3 +423,119 @@ def __init__(
...
+
+    def __init__(self, node: ast.Call, ctrl: list[ast.expr]) -> None:
+        super().__init__(**node.__dict__)
+        self.ctrl = ctrl
+        self.qubit_num = None
+
+
+class Power(ast.Call):
+    """The power modifier"""
+    iter: ast.expr
```
mark-koch: Could you add a `_fields` field similar to other nodes in this file? Then visitors will be able to traverse them automatically:

```suggestion
class Power(ast.Call):
    """The power modifier"""
    iter: ast.expr

    _fields = ("iter",)
```

Some for the other nodes

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -422,3 +423,119 @@ def __init__(
...
         self.ty = ty
         self.captured = captured
+
+
+class Dagger(ast.expr):
+    """The dagger modifier"""
+    def __init__(self, node: ast.expr) -> None:
+        super().__init__(**node.__dict__)
+        
+class Control(ast.Call):
```
mark-koch, line 435: Have you considered inheriting from `ast.expr` instead of `ast.Call` here? Are you still looking up the `Call` fields later on?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -265,6 +280,85 @@ def visit_FunctionDef(
...
+
+        e = node.context_expr
+        match e:
+            case ast.Name(id="dagger"):
+                return Dagger(e)
+            case ast.Call(func=ast.Name(id="dagger")):
+                if len(e.args) != 0:
+                    span = Span(to_span(e.args[0]).start, to_span(e.args[-1]).end)
+                    raise GuppyError(
+                        WrongNumberOfArgsError(span, 0, len(e.args)))
```
mark-koch: Could you add a test for this case?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -265,6 +280,85 @@ def visit_FunctionDef(
...
+                return Dagger(e)
+            case ast.Call(func=ast.Name(id="control")):
+                for arg in e.args:
+                    arg, bb = ExprBuilder.build(arg, self.cfg, bb)
```
mark-koch: Careful: We're not allowed to throw the `bb` returned by `ExprBuilder.build` away since that's the bb to continue building in. Probably best to also return `BB` from `_visit_withitem`

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -265,6 +280,85 @@ def visit_FunctionDef(
...
+            case ast.Call(func=ast.Name(id="power")):
+                if len(e.args) == 0:
+                    span = Span(to_span(e.func).end, to_span(e).end)
+                    raise GuppyError(
+                        WrongNumberOfArgsError(span, 1, len(e.args)))
+                elif len(e.args) != 1:
+                    span = Span(to_span(e.args[1]).start, to_span(e.args[-1]).end)
+                    raise GuppyError(
+                        WrongNumberOfArgsError(span, 1, len(e.args)))
+                return Power(e, e.args[0])
```
mark-koch: `e.args[0]` also needs to be visited by `ExprBuilder`

mark-koch: Maybe it would be better to do that in `visit_With`? That would also handle the bb return problem above. Something like

```python
def visit_With(self, node: ast.With, bb: BB, jumps: Jumps) -> BB | None:
    # ....
    for i in range(len(node.items)):
        node.items[i], bb = ExprBuilder.build(node.items[i], self.cfg, bb)
        match self._handle_with_item(node.items[i]):
           # ...
    # ...
    return bb
```

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
```diff
@@ -43,3 +43,43 @@ class ExpectedError(Error):
...
+
+@dataclass(frozen=True)
+class AssignUnderDagger(Error):
+    title: ClassVar[str] = "Assignment in dagger"
+    span_label: ClassVar[str] = (
+        "Assignment found in a dagger context"
+    )
+    @dataclass(frozen=True)
+    class Modifier(Note):
+        span_label: ClassVar[str] = "dagger modifier is used here"
```
mark-koch: Nice! I like the hint pointing to the modifier 👍

Have you thought about adding a similar hint to `ReturnUnderModifierError` and `LoopCtrlUnderModifierError`?

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
```diff
@@ -43,3 +43,43 @@ class ExpectedError(Error):
...
+
+
+# TODO: It might be better to merge this error with AssignUnderDagger
```
mark-koch: Would slightly prefer merging them but I don't feel strongly

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
```diff
@@ -0,0 +1,110 @@
...
+        modifier.power,
+        **dict(ast.iter_fields(modifier)),
+    )
+    return with_loc(modifier, checked_modifier)
+
+
+def _set_inout_if_linear(
+    var: Variable
+) -> Variable:
+    """Set the `inout` flag if the variable is linear."""
```
mark-koch: Based on the implementation:

```suggestion
def _set_inout_if_affine(
    var: Variable
) -> Variable:
    """Set the `inout` flag if the variable is affine."""
```

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -398,6 +403,45 @@ def visit_NestedFunctionDef(self, node: NestedFunctionDef) -> ast.stmt:
         self.ctx.locals[func_def.name] = Variable(func_def.name, func_def.ty, func_def)
         return func_def
 
+    def visit_Modifier(self, node: Modifier) -> ast.stmt:
+        from guppylang_internals.checker.modifier_checker import check_modifier
```
mark-koch: Can't this be a global import?

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -398,6 +403,45 @@ def visit_NestedFunctionDef(self, node: NestedFunctionDef) -> ast.stmt:
...
+        for control in modifier.control:
+            ctrl = control.ctrl
+            ctrl[0], ty = self._synth_expr(ctrl[0])
+
+            if is_array_type(ty):
+                if len(ctrl) > 2:
+                    assert isinstance(control, ast.Call)
+                    span = Span(to_span(control.func).end, to_span(control).end)
+                    raise GuppyError(
+                        WrongNumberOfArgsError(span, 1, len(control.args)))
```
mark-koch: Could you add a test for this case?

guppylang-internals/src/guppylang_internals/cfg/builder.py
mark-koch: Also, you need to write the resulting `arg` back into `e.args`

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
mark-koch: Also, in the dagger and power case, you perform this check in the CFG builder. I'd prefer if they were all in a consistent place (slightly prefer them here over the builder)

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -398,6 +403,45 @@ def visit_NestedFunctionDef(self, node: NestedFunctionDef) -> ast.stmt:
...
+        if not self.bb:
+            raise InternalGuppyError("BB required to check with block!")
+
+        # check the mody of the modifier.
+        modifier = check_modifier(node, self.bb, self.ctx)
+
+        # check the arguments of the control and power.
+        for control in modifier.control:
+            ctrl = control.ctrl
+            ctrl[0], ty = self._synth_expr(ctrl[0])
```
mark-koch, line 420: You need to check that `len(ctrl) > 0` before indexing

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
```diff
@@ -0,0 +1,110 @@
...
+    # The other checks are done in unitary checking.
+    # e.g. call to non-unitary function in a unitary modifier.
+
+    # Construct inputs for checking the body CFG
+    inputs = [v for v, _ in captured.values()]
+    inputs = linear_front_others_back(inputs)
+    def_id = DefId.fresh()
+    globals = ctx.globals
+
+    # TODO: Ad hoc name for the new function
```
mark-koch, line 70: Can you create an issue so we don't forget about this todo?

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
```diff
@@ -0,0 +1,117 @@
...
+                    AssignUnderDagger.Modifier(modifier.span_ctxt_manager())
+                )
+                raise GuppyError(assign_err)
+
+    # The other checks are done in unitary checking.
+    # e.g. call to non-unitary function in a unitary modifier.
+
+    # Construct inputs for checking the body CFG
+    inputs = [v for v, _ in captured.values()]
+    inputs = linear_front_others_back(inputs)
```
mark-koch: Why do these need to be sorted?

guppylang-internals/src/guppylang_internals/experimental.py
```diff
@@ -90,3 +90,8 @@ def check_lists_enabled(loc: AstNode | None = None) -> None:
 def check_capturing_closures_enabled(loc: AstNode | None = None) -> None:
     if not EXPERIMENTAL_FEATURES_ENABLED:
         raise GuppyError(UnsupportedError(loc, "Capturing closures"))
+
+
+def check_modifiers_enabled(loc: AstNode | None = None) -> None:
+    if not EXPERIMENTAL_FEATURES_ENABLED:
+        raise GuppyError(ExperimentalFeatureError(loc, "Modifiers"))
```
mark-koch, line 97: Could you add a test for this under `experimental_errors`?

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -422,3 +423,122 @@ def __init__(
...
+        self.cfg = cfg
+        self.ty = ty
+        self.captured = captured
+        self.dagger = dagger
+        self.control = control
+        self.power = power
+
+    def __str__(self) -> str:
+        # generate a function name from the def_id
+        return f"__WithBlock__({self.def_id})"
```
mark-koch, line 539: Looks like you're not actually using this name?

guppylang-internals/src/guppylang_internals/tys/qubit.py
```diff
@@ -0,0 +1,61 @@
...
+        return False
+
+    @visit.register
+    def _visit_TypeArg(self, arg: TypeArg) -> bool:
+        arg.ty.visit(self)
+        return True
+
+
+def contain_qubit_ty(ty: Type) -> bool:
+    """Checks if the given type contains the qubit type."""
```
mark-koch: This also seems unused?

tests/error/modifier_errors/captured_var_inout_own.err
```diff
@@ -0,0 +1,13 @@
...
+   | 
+12 |     a = qubit()
+13 |     with dagger:
+14 |         discard(a)
+   |                 ^ Function `discard` wants to take ownership of this argument,
+   |                   but `__modified__()` doesn't own `a`
+   | 
+12 |     a = qubit()
+   |     - Argument `a` is only borrowed. Consider taking ownership:
+   |       `a: qubit @owned`
```
mark-koch, line 11: This is quite misleading, but not trivial to fix. Can you create an issue to improve this error in the future?

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -621,6 +622,74 @@ def _check_comprehension(
...
+
+        # reassign controls
+        for ctrl in node.control:
+            for arg in ctrl.ctrl:
+                match arg:
+                    case PlaceNode(place=place):
+                        self._reassign_single_inout_arg(place, place.defined_at or arg)
+                    case arg:
+                        # This is not supposed to happen
+                        raise InternalGuppyError("Cannot reassign non-place control")
```
mark-koch: Consider a bit less verbose:

```suggestion
                assert isinstance(arg, PlaceNoce)  # Checked above
                self._reassign_single_inout_arg(arg.place, arg.place.defined_at or arg)
```

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -621,6 +622,74 @@ def _check_comprehension(
...
+        for power in node.power:
+            if isinstance(power.iter, PlaceNode):
+                self.visit_PlaceNode(
+                    power.iter, use_kind=UseKind.CONSUME, is_call_arg=None
+                )
+            else:
+                self.visit(power.iter)
+
+        # check captured variables
+        for var, use in node.captured.values():
```
mark-koch, line 661: It's not enough to check the captured variables, you'll need to visit the body in a new nested scope. Otherwise, you wouldn't catch problems like

```python
q = qubit()
with control(q):
    xs = array(1, 2, 3)
    foo(xs, xs)  # This should error because local variable xs is borrowed twice
```

guppylang-internals/src/guppylang_internals/tys/qubit.py
```diff
@@ -0,0 +1,61 @@
+import functools
+from typing import Any, cast
+
+from guppylang_internals.definition.ty import TypeDef
+from guppylang_internals.tys.arg import TypeArg
+from guppylang_internals.tys.common import Visitor
+from guppylang_internals.tys.ty import OpaqueType, Type
+
+
+def qubit_ty() -> Type:
```
mark-koch, line 9: Consider using https://docs.python.org/3/library/functools.html#functools.cache to memoise this function

tests/error/modifier_errors/power_arg_type.py
```diff
@@ -0,0 +1,11 @@
+from guppylang.decorator import guppy
+
+
+# TODO: Currently, `with power(0.2):` passes typecheck as it only checks numeric type.
```
mark-koch: Can you turn this into an issue?

tests/error/modifier_errors/return_found.err
```diff
@@ -0,0 +1,8 @@
+Error: Unexpected return (at $FILE:7:8)
+  | 
+5 | def test() -> None:
+6 |     with dagger:
+7 |         return
+  |         ^^^^^^ Return statement found in a with block
```
mark-koch: nit: Can you change the message so "with" is in backticks `

tests/error/modifier_errors/unexpected_ctxt_maneger.err
```diff
@@ -0,0 +1,8 @@
+Error: Unexpected context manager (at $FILE:6:9)
+  | 
+4 | @guppy
+5 | def test() -> None:
+6 |     with True:
+  |          ^^^^ Unexpected context manager in a context manager
```
mark-koch: This reads a bit akwardly. Maybe something like "Invalid modifier, expected one of `dagger`, `control(...)`, or `power(...)`" would be clearer?

guppylang-internals/src/guppylang_internals/nodes.py
hkengo-qtnm: That makes a lot of sense. Let me do that.

guppylang-internals/src/guppylang_internals/nodes.py
hkengo-qtnm: `ast.expr` is enough as it only tracks the span of the original python code later on. Is it better to inherit `ast.expr`?

guppylang-internals/src/guppylang_internals/cfg/builder.py
hkengo-qtnm: I don't think we can pass an expression like `control(c)` to `ExprBuilder` since the function `control` is undefined (and results in UnknownVariable). Maybe it's possible if we add something like
```python
@contextmanager
@custom_function
@no_type_check
def control(c: array[qubit, n]) ...
```
to the standard libraries, but this would require a strict rule to prevent `control` from being used anywhere except in `with_item`.

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
hkengo-qtnm: I have thought, but just haven't done it yet.

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
hkengo-qtnm: That is true. I'll fix it.

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
hkengo-qtnm: This caused a circular import. I don't know why, but it must be the same reason as why `check_nested_func_def` is imported locally.

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
hkengo-qtnm: I added a test and actually found a bug!

I said true, but probably it's not that straight forward. Here are the gaps between Control and Power/Dagger:
- The check for control can only work during type-checking since it's overloaded (takes 1 array or more than 1 qubits)
- Argument number checking for `Power` should be done in CFG builder if we want to have this definition:
  ```python
  class Power(ast.Call):
    iter: ast.expr
  ```
  otherwise, we need to keep list of arguments as `iter: list[ast.expr]`, although it is an obvious error if the length is not 1.

I agree this implementation is not nice, but just postponing the check for Power/Dagger still does not sound like the best way for me. What about defining all of these argument-number-checker somewhere in one place (in a new file or `modifier_checker.py`) and calling them in a different stage of compilation?

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
hkengo-qtnm: This is due to the definition of modifier extension ops in Hugr. They take two `TypeArgs` each of them referring to in-out arguments and input-only arguments.

guppylang-internals/src/guppylang_internals/nodes.py
mark-koch: If you inherit from `ast.Call`, you still have the `func` and `args` fields around. If you no longer need those, it's imo better to not inherit them - then you can also get rid of the constuctor

guppylang-internals/src/guppylang_internals/cfg/builder.py
mark-koch: `ExprBuilder` shouldn't care about unbound variables, they will be preserved as `ast.Name` nodes and not inspected?

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
mark-koch: Ah yes makes sense 👍

guppylang-internals/src/guppylang_internals/nodes.py
hkengo-qtnm: Not yet in this PR, but it'll be used while generating Hugr. This will be the name of the function generated from a modified block. Should I separate this in the coming compiler PR?

guppylang-internals/src/guppylang_internals/tys/qubit.py
hkengo-qtnm: This will be used in the unitarity checking, when I add UnitaryFlag in future PR.

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
hkengo-qtnm: The `check_cfg` for the `ModifierBlock` is called here, so this example fails with some correct error messages. I'll add some tests that shows type/linearity checks are working correctly.
https://github.com/CQCL/guppylang/blob/6c741356d01835b04fed4c9a315e35cbbeb4d82a/guppylang-internals/src/guppylang_internals/checker/modifier_checker.py#L76

tests/error/modifier_errors/power_arg_type.py
hkengo-qtnm: Let me just fix it.

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
hkengo-qtnm: I did it!

hkengo-qtnm:
Still working on. I'll fix the rest tomorrow.

guppylang-internals/src/guppylang_internals/nodes.py
hkengo-qtnm: Oh, I found that I am actually using `ast.Call` here (this `assert` is not needed though)
https://github.com/CQCL/guppylang/blob/f35c53d7369addc7e8dad09815737f2e57869de8/guppylang-internals/src/guppylang_internals/checker/stmt_checker.py#L424-L425

guppylang-internals/src/guppylang_internals/cfg/builder.py
hkengo-qtnm: Oh, yes, you are right. It worked with this definition:
```python
def _handle_withitem(self, node: ast.withitem) -> Modifier:
```

tests/error/modifier_errors/power_arg_type.py
hkengo-qtnm: Fixed!
https://github.com/CQCL/guppylang/blob/547ce515cb86ba8a9dce7451f879cf0f0c25310c/tests/error/modifier_errors/power_arg_type.err#L1-L8

tests/error/modifier_errors/ctrl_no_arg.py
```diff
@@ -0,0 +1,12 @@
+from guppylang.decorator import guppy
+
+
+# TODO: `control` is overloaded, so the number of arguments need not be fixed to 1.
+# However, the error message says "expected 1 arguments, got 0" when no argument is given.
```
mark-koch: I think this is fine 👍

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -421,14 +424,19 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> ast.stmt:
...
                 if not is_qubit_ty(element_ty):
-                    raise GuppyTypeError(
-                        WithArgTypeMismatchError(ctrl[0], ty, "type `array[qubit, _]`")
+                    dummy_array_ty = array_type_def.check_instantiate(
+                        [
+                            TypeArg(qubit_ty()),
+                            ConstParam(
+                                1, "n", NumericType(NumericType.Kind.Nat)
+                            ).to_existential()[0],
+                        ]
```
mark-koch: You can use the `guppylang_internals.tys.builtin.array_type` helper function to construct an array type

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -438,9 +446,9 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> ast.stmt:
 
         for power in node.power:
             power.iter, ty = self._synth_expr(power.iter)
-            if not isinstance(ty, NumericType):
+            if not isinstance(ty, NumericType) or not ty.kind == NumericType.Kind.Nat:
                 raise GuppyTypeError(
-                    WithArgTypeMismatchError(power.iter, ty, "numeric type")
+                    TypeMismatchError(power.iter, NumericType(NumericType.Kind.Nat), ty)
                 )
             power.iter, subst = self._check_expr(power.iter, ty)
```
mark-koch: If you want to check exactly for `nat` then `_check_expr` is enough:

```suggestion
            power.iter, subst = self._check_expr(power.iter, NumericType(NumericType.Kind.Nat))
```

guppylang-internals/src/guppylang_internals/nodes.py
mark-koch: In general it's better to separate, to avoid merging code without tests. But I've already looked at it now, so I'm fine with leaving it here if it's too annoying to move

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
mark-koch: Oops yes I missed that, sorry!

guppylang-internals/src/guppylang_internals/nodes.py
mark-koch: Ok, I don't mind either way. Just go for what you think is better 👍

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
mark-koch: Mhmm ok I see - in that case I'll leave it to you. Putting the checks into `modifier_checker.py` seems reasonable, or just leaving as is seems also fine 👍

mark-koch (commented):
Thanks, looks great! Just a few tiny suggestions and then I'm happy to merge 👍

guppylang-internals/src/guppylang_internals/tys/qubit.py
hkengo-qtnm: I will separate this since this function need to be tested. (__str__ might be ok since it's trivial, but this one is not)

PR #1282, opened 2025-10-02
feat: add custom emulator build arguments

jake-arkinstall:
This adds `builder.with_build_arg(name: str, value: Any)`, which allows for build customisation.

# Background
Selene builds are managed by the selene_core BuildPlanner. This is a graph comprising:
- nodes: artifact kinds (e.g. "a .hugr file", "a .bc file with external calls to __foo()", "a string containing llvm")
- edges: steps that each map from one kind to another, e.g. compilation steps, along with associated `cost`.

It is configured with sensible defaults, but one can add kinds and steps to the default build planner with custom extensions.

Upon providing an input to build(), the artifact _kind_ is determined, and djikstra provides the walk of least total cost from the input kind to a selene executable.
 
Costs and steps can be customised in the definition of steps. If custom steps are added into the build planner, a common pattern is to enable them (by reducing their `cost`) upon presence of custom build arguments.

# Concrete Example

The build steps undertaken during a walk of least cost can be inspected by looking at the 'steps' entry in `${build_dir}/selene.yaml`, which for a vanilla run looks like:
```yaml
steps:
- class: HUGRPackageToHUGREnvelopeBytesStep
  module: selene_core.build_utils.builtins.hugr
- class: SeleneCompileHUGRToLLVMBitcodeStringStep
  module: selene_helios_qis_plugin.build
- class: LLVMBitcodeStringToLLVMBitcodeFileStep
  module: selene_core.build_utils.builtins.helios
- class: HeliosLLVMBitcodeFileToHeliosObjectFileStep
  moydule: selene_core.build_utils.builtins.helios
- class: HeliosObjectFileToSeleneObjectFileStep_Linux
  module: selene_core.build_utils.builtins.helios
- class: SeleneObjectToSeleneExecutable
  module: selene_core.build_utils.builtins.selene
 ```

A HUGR Envelope Bytes [can be compiled to .bc or .ll](https://github.com/CQCL/selene/blob/main/selene-ext/interfaces/helios_qis/python/selene_helios_qis_plugin/build.py) based on a 'build_method' parameter. Passing `build_method='via-llvm-ir'` to selene_sim.build will thus provide human-readable LLVM IR in the build artifacts, which is great for inspection.

By doing so, you get these steps instead:
```yaml
steps:
- class: HUGRPackageToHUGREnvelopeBytesStep
  module: selene_core.build_utils.builtins.hugr
- class: SeleneCompileHUGRToLLVMIRStringStep
  module: selene_helios_qis_plugin.build
- class: LLVMIRStringToLLVMIRFileStep
  module: selene_core.build_utils.builtins.helios
- class: HeliosLLVMIRFileToHeliosObjectFileStep
  module: selene_core.build_utils.builtins.helios
- class: HeliosObjectFileToSeleneObjectFileStep_Linux
  module: selene_core.build_utils.builtins.helios
- class: SeleneObjectToSeleneExecutable
  module: selene_core.build_utils.builtins.selene
  ```

# Adding this capability to the guppylang emulator interface

This PR adds the ability to pass custom arguments to selene_sim.build function via the builder method.

Usage looks like:
```python
builder = EmulatorBuilder() \
            .with_build_dir("./prefer_text") \
            .with_build_arg("build_method", "via-llvm-ir")

print(main.emulator(1, builder).run())
```

In `./prefer_text/artifacts` LLVM IR rather than bitcode will be present. This demonstrates a customisation point - boring in itself, but the possibilities with such customisation are quite exciting. One can pass custom args to compilation (e.g. optimisation level, verbosity levels, or choose a different compiler entirely), steer builds to using other machine targets, etc.

### DIFF
--- a/guppylang/src/guppylang/emulator/builder.py
+++ b/guppylang/src/guppylang/emulator/builder.py
@@ -81,14 +81,16 @@ class EmulatorBuilder:
         to selene_sim.build as kwargs.
 
         For example:
-        ```
+
+        .. code-block:: python
+
             .with_build_arg("build_method", "via-llvm_ir")
-        ```
 
         is equivalent to
-        ```
+
+        .. code-block:: python
+
             selene_sim.build(..., build_method="via-llvm_ir")
-        ```
 
         which saves LLVM IR into the build directory rather than
         saving bitcode.


guppylang/src/guppylang/emulator/builder.py
```diff
@@ -69,6 +75,29 @@ def with_verbose(self, value: bool) -> Self:
         """Set whether to print verbose output during the build process."""
         return replace(self, _verbose=value)
 
+    def with_build_arg(self, key: str, value: Any) -> Self:
+        """Selene builds may support additional customisable arguments,
+        e.g. to override the choice of compilation route. This is passed
+        to selene_sim.build as kwargs.
+
+        For example:
+        ```
```
CalMacCQ: In the sphinx docs
```
.. code-block:: python

    .with_build_arg("build_method", "via-llvm_ir")
```

Will probably render better

jake-arkinstall: I'm not familiar with that format so, just to check, is this correct?

```python
    def with_build_arg(self, key: str, value: Any) -> Self:
        """Selene builds may support additional customisable arguments,
        e.g. to override the choice of compilation route. This is passed
        to selene_sim.build as kwargs.

        For example:
        .. code-block:: python
            .with_build_arg("build_method", "via-llvm_ir")

        is equivalent to
        .. code-block:: python
            selene_sim.build(..., build_method="via-llvm_ir")

        which saves LLVM IR into the build directory rather than
        saving bitcode.
        """
        if self._custom_args is None:
            return replace(self, _custom_args={key: value})
        else:
            return replace(self, _custom_args=self._custom_args | {key: value})

```

CalMacCQ: I checked quickly and sphinx seems most happy with some extra spaces in there

```python
def with_build_arg(self, key: str, value: Any) -> Self:
        """Selene builds may support additional customisable arguments,
        e.g. to override the choice of compilation route. This is passed
        to selene_sim.build as kwargs.

        For example:

        .. code-block:: python

            .with_build_arg("build_method", "via-llvm_ir")

        is equivalent to

        .. code-block:: python

            selene_sim.build(..., build_method="via-llvm_ir")

        which saves LLVM IR into the build directory rather than
        saving bitcode.
        """
        if self._custom_args is None:
            return replace(self, _custom_args={key: value})
        else:
            return replace(self, _custom_args=self._custom_args | {key: value})

```

PR #1284, opened 2025-10-02
feat: Add selene plugin to run *.gpy.py files

### LINKED ISSUE #1283: Add a selene plugin allowing selene to consume `*.gpy.py` files
Selene has a build planning feature allowing it to run programs of various forms. For example it can consume serialised hugrs or compiled bitcode.

I propose we add a plugin in this repo allowing selene to consume python files with the extension `.gpy.py`.

Selene should execute this file via `subprocess.run(sys.executable, [file])` and interpret it's output on stdout as a serialised hugr package.

This will support users in separating their guppy programs from the rest of their experiments. Users are writing complicated experiments and embedding guppy programs in them. It is difficult to isolate a single guppy program from these more complicated programs for the purposes of debugging that specific program.

doug-q:
Closes #1283

### DIFF
--- a/devenv.lock
+++ b/devenv.lock
@@ -3,10 +3,10 @@
     "devenv": {
       "locked": {
         "dir": "src/modules",
-        "lastModified": 1756398161,
+        "lastModified": 1759406738,
         "owner": "cachix",
         "repo": "devenv",
-        "rev": "b1283b164f915ac4724c0f6d22ca1f35f0a95c03",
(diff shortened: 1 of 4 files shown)


ss2165:
Thoughts:
1. Agree that avoiding the user having to include the sys.stdout line would be nice
2. The natural interface to this feels like a cli to me:
```sh
gpy emulate prog.gpy.py --n-shots=10 --res-file=results.json
gpy compile prog.gpy.py | hugr validate
```

PR #1286, opened 2025-10-03
refactor!: Store function argument names in `FuncInput`

mark-koch:
It's a lot nicer to have the name of function arguments in the same object as the type an flags instead of the separate list we have at the moment - not adding the name was more of an oversight when we introduced `FuncInput` 😅

BREAKING CHANGE: `FunctionType` constructor no longer accepts the `input_names` argument. Instead, input names should be provided as an optional argument to `FuncInput`

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/ty.py
+++ b/guppylang-internals/src/guppylang_internals/tys/ty.py
@@ -392,7 +392,7 @@ class FuncInput:
     flags: InputFlags
 
     #: Name of this input, or `None` if it is an unnamed argument (e.g. inside a
-    #: higher-order `Callable` type)
+    #: `Callable`). We use `compare=False` because names are not visible to the caller.
     name: str | None = field(default=None, compare=False)
 
 


guppylang-internals/src/guppylang_internals/tys/printing.py
```diff
@@ -165,10 +165,10 @@ def _wrap(s: str, inside_row: bool) -> str:
 
 def signature_to_str(name: str, sig: FunctionType) -> str:
     """Displays a function signature in Python syntax including the function name."""
-    assert sig.input_names is not None
```
croyzor, line 168: this line shouldn't have to change

acl-cqc: because `input_names` is a property of type `Sequence[str] | None` that returns `None` if any elements are unnamed...yes, true, but not necessarily that obvious . See comment on `def input_names` but maybe go with Craig ;)

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -449,6 +450,16 @@ def bound_vars(self) -> set[BoundVar]:
             return set()
         return super().bound_vars
 
+    @cached_property
+    def input_names(self) -> Sequence[str] | None:
+        """Names of all inputs or `None` if there are unnamed inputs."""
```
acl-cqc, line 468: Is it worth adding a `def __post_init__` to check that either all have names or none have names? I'm not sure I see any cases where only some might be named....perhaps positional-only arguments??

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -389,6 +389,10 @@ class FuncInput:
     ty: "Type"
     flags: InputFlags
 
+    #: Name of this input, or `None` if it is an unnamed argument (e.g. inside a
+    #: higher-order `Callable` type)
+    name: str | None = field(default=None, compare=False)
```
acl-cqc, line 396: Why `compare=False`? So function types are equal if one has names and the other does not? Not strictly correct - the former can be called with kwargs syntax whereas the latter does not - I'm not sure how much of that stays in guppy but if we do support that, it *might* be worth going to a more-full-blown FunctionType "generalizes" check allowing assignment of named to nameless but not the other way around?

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -194,10 +194,14 @@ def check_nested_func_def(
...
-        for i, (x, inp) in enumerate(
-            zip(func_ty.input_names, func_ty.inputs, strict=True)
+        Variable(
+            cast(str, inp.name),
+            inp.ty,
+            func_def.args.args[i],
+            inp.flags,
+            is_func_input=True,
         )
+        for i, inp in enumerate(func_ty.inputs)
```
acl-cqc:
```suggestion
        for arg, inp in zip(func_def.args.args, func_ty.inputs)
```
possibly with `strict=True`

guppylang-internals/src/guppylang_internals/tys/ty.py
mark-koch: Guppy currently doesn't support kwargs, so function names are irrelevant when it comes to calling. The same is also true when e.g. trying to unify to function types - we just ignore the names

acl-cqc: Ah! Very good. In that case quite right. Maybe worth a `# not exposed to caller` here then

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -391,6 +391,10 @@ class FuncInput:
     ty: "Type"
     flags: InputFlags
 
+    #: Name of this input, or `None` if it is an unnamed argument (e.g. inside a
```
acl-cqc, line 394:
```suggestion
    #: Name of this input, or `None` if it is an unnamed argument (e.g. inside a `Callable`).
    #: We use compare=False because names are not visible to the caller
```

PR #1287, opened 2025-10-06
feat: compiler for modifiers

hkengo-qtnm:
This PR implements the Guppy compiler for modifiers.

## Major changes
- Compiler implemented in [modifier_compiler.py](guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py)
- Imported modifier extension ops [tket_exts.py](guppylang-internals/src/guppylang_internals/std/_internal/compiler/tket_exts.py)
- Some validation tests [test_modifier.py](tests/integration/test_modifier.py)

## Issues
- As the python syntax, current syntax of modifiers are not valid (dagger(...)/control(...)/power(...) are undefined) and need to be fixed. (See [here](https://github.com/CQCL/guppylang/compare/main...feat/modifier-compiler#diff-3c552a04d91da2c418c6760e6571a84119343ebe3d6cdd0d9547b65c181a24c6R7-R11))
- Haven't tested, but I think type variables in modified block are not supported at this point. (See [here](https://github.com/CQCL/guppylang/compare/main...feat/modifier-compiler#diff-e33dd79c941364a4444ffbfde4354c998ea0e0fe362f39a312a5e24a29f526f2R44-R46))

## Future implementation
- UnitaryFlags, the annotations for functions that can be called in modified block.
- Integrated test including emulation.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
@@ -1,24 +1,19 @@
 """Hugr generation for modifiers."""
 
-from hugr import Node, Wire, ops
+from hugr import Wire, ops
 from hugr import tys as ht
 
 from guppylang_internals.ast_util import get_type
 from guppylang_internals.checker.modifier_checker import non_copyable_front_others_back
 from guppylang_internals.compiler.cfg_compiler import compile_cfg
 from guppylang_internals.compiler.core import CompilerContext, DFContainer
-from guppylang_internals.compiler.expr_compiler import (
-    ExprCompiler,
-    array_unwrap_elem,
-    array_wrap_elem,
-)
+from guppylang_internals.compiler.expr_compiler import ExprCompiler
 from guppylang_internals.nodes import CheckedModifiedBlock, PlaceNode
 from guppylang_internals.std._internal.compiler.array import (
-    array_convert_from_std_array,
-    array_convert_to_std_array,
-    array_map,
     array_new,
+    array_to_std_array,
     standard_array_type,
+    std_array_to_array,
     unpack_array,
 )
 from guppylang_internals.std._internal.compiler.tket_exts import MODIFIER_EXTENSION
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
```diff
@@ -0,0 +1,205 @@
...
+    POWER_OP_NAME = "PowerModifier"
+
+    dagger_op_def = MODIFIER_EXTENSION.get_op(DAGGER_OP_NAME)
+    control_op_def = MODIFIER_EXTENSION.get_op(CONTROL_OP_NAME)
+    power_op_def = MODIFIER_EXTENSION.get_op(POWER_OP_NAME)
+
+    body_ty = modified_block.ty
+    # TODO: Shouldn't this be `to_hugr_poly` since it can contain
+    # a variable with a generic type?
+    hugr_ty = body_ty.to_hugr(ctx)
```
mark-koch, line 41: Yes, this will most likely break, but afaik nested functions that reference type variables from the parent function also aren't handled correctly.

I guess the best solution for now would be to detect this case and emit an error telling the user that this isn't supported yet.

To properly handle this, we'll need to turn the modifier into a `generic` function and then instantiate it with the variables from the parent function...

tests/integration/test_modifier.py
```diff
@@ -0,0 +1,135 @@
...
+from guppylang.std.quantum import qubit
+from guppylang.std.num import nat
+from guppylang.std.builtins import owned
+from guppylang.std.array import array
+
+# Dummy variables to suppress Undefined name
+# TODO: `ruff` fails when without these, which need to be fixed
+dagger = object()
+control = object()
+power = object()
```
mark-koch, line 11: I think it would be good to add these to `guppylang.std.lang` and then reexport in `guppylang.std.builtins`.

But I'm also happy if you prefer to do this in a separate PR

hkengo-qtnm: I forgot to mention, but I made an issue for this (https://github.com/CQCL/guppylang/issues/1290)
I would rather think it should be left as a future feature to be handled in a better manner.

guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
hkengo-qtnm: Is it okay to merge this PR without such a proper error handling?

mark-koch: Yes, I think it's fine. We just have to wait for the tket release

PR #1292, opened 2025-10-09
feat!: Unitarity annotations for functions

hkengo-qtnm:
This PR introduces optional `unitary, dagger, control, power` keyword arguments to Guppy decorators that can be used to annotate a function that tells which modifiers are applicable to it.

Note that this PR is only breaking for `guppylang_internals`, not `guppylang` itself.

BREAKING CHANGE: `modifier_checker.check_modified_block_signature` now requires the `ModifiedBlock` as first argument

## Major chage
- Defined `guppylang_internals.tys.ty.UnitaryFlags`
- Implemented [unitary_checker.py](https://github.com/CQCL/guppylang/compare/feat/modifier-compiler...feat/modifier-unitarity_checker#diff-556fa94f8eaa5854d7859b46069010e23cbb4f1b17c4fd82f48fe7e49dd5cb94) which mainly checks that a function called in a modified context satisfies the unitarity constraint.
- `guppylang_internals.tys.ty.FunctionType` now contains UnitaryFlags
- Added flags to quantum gates defined in `std.quantum`
- Metadata of flags to the generated Hugr function.

## FIXME
- I think UnitaryFlags are not propagated to child CFGs properly (see [here](https://github.com/CQCL/guppylang/blob/26e00e2a33dabe70aacc6348111b4b18f5de3fe2/guppylang-internals/src/guppylang_internals/cfg/builder.py#L311-L316)).
  - This is not affecting type checking (see [this test](https://github.com/CQCL/guppylang/blob/26e00e2a33dabe70aacc6348111b4b18f5de3fe2/tests/error/modifier_errors/flags_nested.py)), but can cause an issue later in tket.

### DIFF
--- a/tests/error/modifier_errors/captured_var_inout_own.py
+++ b/tests/error/modifier_errors/captured_var_inout_own.py
@@ -1,5 +1,5 @@
 from guppylang.decorator import guppy
-from guppylang.std.quantum import qubit, owned, UnitaryFlags
+from guppylang.std.quantum import qubit, owned
 
 
 @guppy.declare(dagger=True)
--- a/tests/error/modifier_errors/captured_var_inout_reassign.py
+++ b/tests/error/modifier_errors/captured_var_inout_reassign.py
@@ -1,5 +1,5 @@
 from guppylang.decorator import guppy
-from guppylang.std.quantum import qubit, UnitaryFlags
+from guppylang.std.quantum import qubit
 
 
 @guppy.declare(dagger=True)
--- a/tests/error/modifier_errors/ctrl_arg_copy.py
+++ b/tests/error/modifier_errors/ctrl_arg_copy.py
@@ -1,5 +1,5 @@
 from guppylang.decorator import guppy
-from guppylang.std.quantum import qubit, owned, UnitaryFlags
+from guppylang.std.quantum import qubit, owned
 
 
 @guppy.declare
--- a/tests/error/modifier_errors/flag_call.py
+++ b/tests/error/modifier_errors/flag_call.py
@@ -1,5 +1,5 @@
 from guppylang.decorator import guppy
-from guppylang.std.quantum import qubit, UnitaryFlags
+from guppylang.std.quantum import qubit
 
 
 @guppy.declare
--- a/tests/error/modifier_errors/flag_dagger_assign.err
+++ b/tests/error/modifier_errors/flag_dagger_assign.err
@@ -1,8 +1,8 @@
-Error: Invalid expression in dagger (at $FILE:7:4)
+Error: Invalid expression in dagger (at $FILE:6:4)
   | 
-5 | @guppy(dagger=True)
-6 | def test() -> None:
-7 |     x = 3
+4 | @guppy(dagger=True)
+5 | def test() -> None:
+6 |     x = 3
   |     ^^^^^ Assignment found in a dagger context
 
 Guppy compilation failed due to 1 previous error
--- a/tests/error/modifier_errors/flag_dagger_assign.py
+++ b/tests/error/modifier_errors/flag_dagger_assign.py
@@ -1,5 +1,4 @@
 from guppylang.decorator import guppy
-from guppylang.std.quantum import UnitaryFlags
 
 
 @guppy(dagger=True)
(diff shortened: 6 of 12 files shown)


guppylang-internals/src/guppylang_internals/checker/unitary_checker.py
```diff
@@ -0,0 +1,138 @@
...
+    def visit_ResultExpr(self, node: ResultExpr) -> None:
+        # Result is always allowed
+        pass
+
+    def visit_StateResultExpr(self, node: StateResultExpr) -> None:
+        # StateResult is always allowed
+        pass
+
+    def _check_assign(self, node: Any) -> None:
+        assert isinstance(node, ast.Assign | ast.AnnAssign | ast.AugAssign)
```
mark-koch:
```suggestion
    def _check_assign(self, node: ast.Assign | ast.AnnAssign | ast.AugAssign) -> None:
```

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -134,6 +137,7 @@ def parse(self, globals: "Globals", sources: SourceMap) -> "CustomFunctionDef":
             raise GuppyError(BodyNotEmptyError(func_ast.body[0], self.name))
         sig = self.signature or self._get_signature(func_ast, globals)
         ty = sig or FunctionType([], NoneType())
+        object.__setattr__(ty, "unitary_flags", self.unitary_flags)
```
mark-koch: We shouldn't use this

guppylang/src/guppylang/decorator.py
```diff
@@ -168,7 +170,7 @@ def type_var(
         .. code-block:: python
             from guppylang import guppy
 
-            T = guppy.type_var("T")
+            T = guppy.ty
```
mark-koch: Bad change

guppylang/src/guppylang/decorator.py
```diff
@@ -447,6 +449,47 @@ def foo(default_reg: array[qubit, 2],
...
+    ) -> Callable[[GuppyFunctionDefinition[P, T]], GuppyFunctionDefinition[P, T]]:
+        """Wrap a Guppy function with specific unitarity annotations.
+
+        .. code-block:: python
+
+            from guppylang import guppy
+            from guppylang.std.quantum import qubit, h, UnitaryFlags
+
+            @guppy.with_unitary_flags(UnitaryFlags.Unitary)
+            @guppy
```
mark-koch: I think passing the flags to `@guppy` directly would be better. Also, imo we shouldn't expose the internal `UnitaryFlags` type here. I'd prefer individual `unitary=True`, or more fine grained `controllable=True` args

tests/error/modifier_errors/captured_var_inout_own.err
mark-koch, line 1: Why are the line numbers changing??

tests/integration/test_modifier.py
```diff
@@ -101,6 +101,7 @@ def bar(q: qubit) -> None:
 def test_free_linear_variable_in_modifier(validate):
     T = guppy.type_var("T", copyable=False, droppable=False)
 
+    @guppy.with_unitary_flags(UnitaryFlags.Control)
     @guppy.declare
```
mark-koch: Do we want to allow unitary modifiers on function declarations?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -300,6 +308,14 @@ def visit_With(self, node: ast.With, bb: BB, jumps: Jumps) -> BB | None:
             modifier = self._handle_withitem(item)
             new_node.push_modifier(modifier)
 
+        # FIXME: Currently, the unitary flags is not set correctly if there are nested
+        # `with` blocks. This is because the outer block's unitary flags are not
+        # propagated to the outer block. The following line should calculate the sum
+        # of the unitary flags of the outer block and modifiers applied in this
+        # `with` block.
+        unitary_flags = new_node.flags()
+        object.__setattr__(cfg, "unitary_flags", unitary_flags)
```
mark-koch: We shouldn't use `__setattr__`

PR #1303, opened 2025-10-16
refactor: Split ArrayBorrowCompiler off from ArrayGetitemCompiler

acl-cqc:
The goal here is to (a) make it clearer what's going on, (b) give a bit more protection, via asserts, against changes elsewhere in the compiler accidentally generating bad code.

ArrayGetitemCompiler compiles classical and linear variants separately; and has two uses: for `fn __getitem__` of `class array` and for `fn _array_unsafe_getitem` (used only in `ArrayIter`).

First thought was that these can now be separate: ArrayIter ends in `discard_all_borrowed`, so just a borrow would suffice even in the classical case. In fact it turns out that the ArrayIter code only ever uses the linear case, because it's output as polymorphic in Hugr; so outputting `get` for the classical case would be just as efficient - it never happens. But that would be quite misleading, I argue this refactor makes it clearer what's going on.

Passes mypy. Not done any execution/integration tests yet...

### DIFF
--- a/guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
+++ b/guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
@@ -276,7 +276,9 @@ class NewArrayCompiler(ArrayCompiler):
 
 
 class ArrayGetitemCompiler(ArrayCompiler):
-    """Compiler for the `array.__getitem__` function."""
+    """Compiler for the `array.__getitem__` function, used for both classical and
+    linear arrays e.g. when pulling out arguments to pass to a function.
+    (In the linear, non-owned, case they will be replaced afterwards.)"""
 
     def _build_classical_getitem(self, array: Wire, idx: Wire) -> CallReturnWires:
         """Constructs `__getitem__` for classical arrays."""
@@ -293,8 +295,8 @@ class ArrayGetitemCompiler(ArrayCompiler):
             inout_returns=[arr],
         )
 
-    def _build_linear_getitem(self, array: Wire, idx: Wire) -> CallReturnWires:
-        """Constructs `array.__getitem__` for linear arrays."""
+    def _build_borrow(self, array: Wire, idx: Wire) -> CallReturnWires:
+        """Constructs `borrow_array.borrow` for linear-typed arrays."""
         idx = self.builder.add_op(convert_itousize(), idx)
         arr, elem = self.builder.add_op(
             barray_borrow(self.elem_ty, self.length),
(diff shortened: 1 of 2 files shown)


guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -275,8 +275,49 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
+        # HOWEVER note that even if we monomorphized in guppy and needed to handle
+        # copyable ArrayIter, we could still use just a borrow (not the more complex
+        # _build_classical_getitem), as ArrayIter ends with a discard-all-borrowed.
+        return self._build_borrow(array, idx)
+
+    def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
+        [array, idx] = args
+        [elem_ty_arg, _] = self.type_args
+        assert isinstance(elem_ty_arg, TypeArg)
+        assert elem_ty_arg.ty.copyable == (
```
acl-cqc: Probably overkill but verifies we've got the same branching behaviour as in #1166

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -275,8 +275,49 @@ def compile(self, args: list[Wire]) -> list[Wire]:
...
+        raise InternalGuppyError("Call compile_with_inouts instead")
+
+
+class ArrayGetitemCompiler(ArrayBorrowCompiler):
+    """Compiler for the `array.__getitem__` function, used for both classical and
+    linear arrays e.g. when pulling out arguments to pass to a function.
+    (In the linear, non-owned, case they will be replaced afterwards.)"""
 
     def _build_classical_getitem(self, array: Wire, idx: Wire) -> CallReturnWires:
         """Constructs `__getitem__` for classical arrays."""
```
acl-cqc, line 284: This could become (borrow_)array::get following https://github.com/CQCL/tket2/pull/1181 but looks like that's non-conflicting/orthogonal (yay!)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -276,7 +276,9 @@ def compile(self, args: list[Wire]) -> list[Wire]:
 
 
 class ArrayGetitemCompiler(ArrayCompiler):
-    """Compiler for the `array.__getitem__` function."""
+    """Compiler for the `array.__getitem__` function, used for both classical and
+    linear arrays e.g. when pulling out arguments to pass to a function.
+    (In the linear, non-owned, case they will be replaced afterwards.)"""
```
acl-cqc, line 281: I have just been looking at how this happens in the frontend - `check_place_assignable` filling in the `getitem_call` field of a SubscriptAccess - it is nontrivial, otherwise I would love to have separated out the two different kinds of GetItem (linear with setitem to follow, vs classical without) first

PR #1308, opened 2025-10-21
fix: compilation of affine-bounded type variables

### LINKED ISSUE #1306: Functions polymorphic over `[T:Drop]` compile wrongly
Specifically, they compile to polymorphic-in-Hugr functions polymorphic over `Copyable` and without any `Drop` operations.

This bug is exposed by #1166, which puts in some hacky workarounds to preserve compilation/execution on existing test cases. However, the following (an extension of the same-named test in `test_poly_py312.py`) fails:
```
def test_drop_bound(validate):
    @guppy
    def helper[T: Drop](s: T @ owned) -> None:
        pass
    @guppy
    def main(s: array[int, 5] @ owned) -> None:
        helper(s)

```
because the generated Hugr tries to instantiate `helper` (polymorphic over `Copyable`) with `value_array` (linear).
 
Prior to #1166, this example worked, because the `value_array` class is Copyable; later value_array would be lowered to array (linear) with linearization applied (inserting drops even where they were not present in the Hugr).

acl-cqc:
fixes #1306

The first part of the fix is to make affine-bounded type parameters (declared `[T : Drop]`) compile to Hugr TypeParams bounded by linear rather than Copyable; this fixes the error in the issue (invalid instantiation of Copyable-bounded function with type-arg `borrow_array<....>`), but raises a different problem: guppy does not insert the necessary `drop` operations, thus leading to `UnconnectedPort` validation errors (as the ports are now linear).

This was because the `requires_drop` function used when inserting drops (#1090) inspects the `hugr_bound` property of type variables which also needs updating that affine-in-guppy things should become linear-in-Hugr. In this area, update some docs (RIP equatable), refactor using `super()`, and standardize use of `ht` for `hugr.tys`.

Note that (contrary to python-semver-checks) this is not breaking: the APIs removed (optional extra parameter to TypeParam, and settable TypeParam::is_affine field) were only added in #1166, which has not yet been released.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/param.py
+++ b/guppylang-internals/src/guppylang_internals/tys/param.py
@@ -87,11 +87,6 @@ class TypeParam(ParameterBase):
     must_be_droppable: bool
 
     @property
-    def is_affine(self) -> bool:
-        """True if the type may not be copied, but can definitely be dropped."""
-        return not self.must_be_copyable and self.must_be_droppable
-
-    @property
     def can_be_linear(self) -> bool:
         """Whether this type should be treated linearly."""
         return not self.must_be_copyable and not self.must_be_droppable
@@ -159,9 +154,9 @@ class TypeParam(ParameterBase):
     def to_hugr(self, ctx: ToHugrContext) -> ht.TypeParam:
         """Computes the Hugr representation of the parameter."""
         return ht.TypeTypeParam(
-            bound=ht.TypeBound.Linear
-            if self.can_be_linear or self.is_affine
-            else ht.TypeBound.Copyable
+            bound=ht.TypeBound.Copyable
+            if self.must_be_copyable
+            else ht.TypeBound.Linear
         )
 
     def __str__(self) -> str:


guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -57,14 +57,11 @@ def affine(self) -> bool:
...
-        """The Hugr bound of this type, i.e. `Any`, `Copyable`, or `Equatable`.
-
-        This needs to be specified explicitly, since opaque nonlinear types in a Hugr
-        extension could be either declared as copyable or equatable. If we don't get the
-        bound exactly right during serialisation, the Hugr validator will complain.
-        """
+        """The Hugr bound of this type, i.e. `Any` or `Copyable`."""
+        if self.linear or self.affine:
+            return ht.TypeBound.Linear
+        return ht.TypeBound.Copyable
```
mark-koch, line 64: What about

```suggestion
    return ht.Typebound.Copyable if self.copyable else ht.TypeBound.Linear
```

guppylang-internals/src/guppylang_internals/tys/param.py
```diff
@@ -83,11 +83,13 @@ def instantiate_bounds(self, inst: "PartialInst") -> Self:
...
-    # TODO: `is_affine` here is not right, but suffices to hide
-    # https://github.com/CQCL/guppylang/issues/1306 in the current testsuite.
     must_be_copyable: bool
     must_be_droppable: bool
-    is_affine: bool | None = None
+
+    @property
+    def is_affine(self) -> bool:
+        """True if the type may not be copied, but can definitely be dropped."""
+        return not self.must_be_copyable and self.must_be_droppable
```
mark-koch: The naming here is a bit inconsistent with `can_be_linear` below. As far as I can see, this is only used in the `to_hugr` method below? Would prefer to just replace the

```python
if self.can_be_linear or self.is_affine
```

in `to_hugr` with

```python
if not self.must_be_copyable
```

acl-cqc: Yes, so compared to before #1166 we've switch from checking `can_be_linear` to `not self.must_be_copyable`, which indeed has the desired effect (same for copyable/pure-linear things, but flipped for affine).

TBH I would argue against having a method `can_be_linear`. TMHO The relevant questions are: `can_be_copied` (true if copyable, false for affine, false for linear) and `can_be_dropped` (true if copyable, true if affine, false for linear). (Alternatively, rephrase `can_be_dropped` as `must_be_used` and invert.)

Better still *might* be to introduce an enum like hugr typebound but with 3 variants, but maybe never mind that ;)

acl-cqc: Oh, of course "can I copy it" is `must_be_copyable` and "can I drop it" is `must_be_droppable`. So maybe I am arguing for removal of now-unused `can_be_linear`. I think that would actually be a breaking change vs previous release, whereas this isn't, though.

PR #1309, opened 2025-10-22
fix: multiline loop arguments 

### LINKED ISSUE #1301: Multiline for loop header causes assertion failure
Attempting to compile the following guppy example encounters an assertion failure at line 32 of `guppylang_internals/span.py`. As far as I can tell this is valid guppy which should be accepted by the compiler.

```
from guppylang import guppy
from guppylang.std.builtins import array, result
from guppylang.std.quantum import qubit, x, h, measure_array

@guppy
def main() -> None:
    q = array( qubit() for _ in range(4) )

    for i in range(4): # This form is fine
        h(q[i])

    for i in range(
        4
    ): # Assertion failure here
        x(q[i])

    result("q", measure_array(q))

if __name__ == "__main__":
    hugr = main.compile()

```
This is easy to work around, so not an urgent issue from my perspective.

Versions:
`guppylang` : v0.21.5
`guppylang-internals`: v0.24.0

hsemenenko:
Fixes #1301

### DIFF
--- a/tests/integration/test_for.py
+++ b/tests/integration/test_for.py
@@ -134,7 +134,7 @@ def test_loop_multiline_args(validate):
     def foo() -> None:
         for _ in range(
             1
-        ):  # Really really really long comment to force linter to split line
+        ):  # fmt: skip
             pass
 
     validate(foo)


tests/integration/test_for.py
```diff
@@ -127,3 +127,12 @@ def foo(xs: list[int], ys: list[int]) -> int:
         return p
 
     validate(foo)
+
+
+def test_loop_multiline_args(validate):
+    @compile_guppy
+    def foo() -> None:
+        for _ in range(1):
```
ss2165: is this supposed to be multi-line? Formatter?

hsemenenko: Yes, fixed.

ss2165: any way to stop auto-formatters chewing it up in future?

hsemenenko: Added a long comment to force the linter to break the line. This what caused the bug initially for me.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -465,8 +465,13 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
         # A `value.attr` attribute access. Unfortunately, the `attr` is just a string,
         # not an AST node, so we have to compute its span by hand. This is fine since
         # linebreaks are not allowed in the identifier following the `.`
+        # Attributes for loops are not confined to a single line. If the node span is
+        # not confined to a single line, we return the entire attribute span.
```
mark-koch: Suggest to add a bit more detail here:

```suggestion
        # The only exception are attributes accesses that are generated during
        # desugaring (for example for iterators in `for` loops). Since those just
        # inherit the span of the sugared code, we could have line breaks there.
        # See https://github.com/CQCL/guppylang/issues/1301
```

tests/error/misc_errors/loop_attribute_error.py
```diff
@@ -0,0 +1,11 @@
+from guppylang import guppy
+
+@guppy
+def foo() -> None:
+    for i in range(
+        a
+    ): # Really really really really long comment to cause linter to split line
```
mark-koch: Does this work?

```suggestion
    ): # fmt: skip
```

PR #1311, opened 2025-10-22
fix: Improve track_hugr_side_effects, adding Order edges from/to Input/Output

### LINKED ISSUE #1310: Result ops are not reachable via Order edges
`tket.result` ops are not reachable by any kind of edge, so DCE has no reason to preserve them.

Clearly they cannot be reachable by value edges (until #1266, which is the best but longest-term, solution), so we need to connect them to Output nodes (and their containers, if themselves dataflow nodes, to the Output nodes of their containers, and so on).

Example:
```
    @guppy
    def main() -> None:
        q = qubit()
        q = f(q)
        b = measure(q)

        result("b", b)

    @guppy
    def f(q: qubit @ owned) -> qubit:
        return q


    program = main.compile()
    with open("measure_result.hgr", "wb") as f:
        f.write(program.to_bytes())
    validate(program)
```
produces this Hugr:

Image
Note we do have edges from `QAlloc` to `Call(f)` to `MeasureFree` to `result_bool`, but no further.

acl-cqc:
Fixes #1310

`track_hugr_side_effects` already linked stateful nodes in each DataflowBlock, however
* When the last of these was a node with no outputs (e.g. `tket.result`), the entire chain could still be unreachable from Output. So, add an Order edge to the Output.
* Also add an Order edge to the *first* side-effecting node from Input; some interpretations of Hugr mean that without this, we could execute such nodes before e.g. DFG containers "start".
* Recursion (marking parents as side-effectful) stopped at CFGs, meaning an entire CFG could be unreachable (e.g. if it had no outputs), or otherwise be unconnected to other stateful ops in an ancestor. Remove this assumption by recursing through CFGs/Conditionals (up to FuncDefns only) - skipping, rather than stopping, at non-dataflow nodes.

I've not tested the last, it is a bit of a corner case.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/compiler/core.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/core.py
@@ -582,6 +582,9 @@ def may_have_side_effect(op: ops.Op) -> bool:
             # precise answer
             return True
         case _:
+            # There is no need to handle TailLoop (in case of non-termination) since
+            # TailLoops are only generated for array comprehensions which must have
+            # statically-guaranteed (finite) size. TODO revisit this for lists.
             return False
 
 


guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -596,9 +595,7 @@ def track_hugr_side_effects() -> Iterator[None]:
     # Remember original `Hugr.add_node` method that is monkey-patched below.
     hugr_add_node = Hugr.add_node
     # Last node with potential side effects for each dataflow parent
-    prev_node_with_side_effect: defaultdict[Node, Node | None] = defaultdict(
-        lambda: None
-    )
+    prev_node_with_side_effect: dict[Node, tuple[Node, Hugr[Any]]] = {}
```
acl-cqc, line 601: Storing the Hugr in here is annoying, but I don't see another way (?)

guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -619,22 +616,41 @@ def handle_side_effect(node: Node, hugr: Hugr[OpVarCov]) -> None:
...
     # Monkey-patch the `add_node` method
     Hugr.add_node = hugr_add_node_with_order  # type: ignore[method-assign]
     try:
         yield
+        for parent, (last, hugr) in prev_node_with_side_effect.items():
+            # Connect the last side-effecting node to Output
+            outp = hugr.children(parent)[1]
+            assert isinstance(hugr[outp].op, ops.Output)
+            assert last != outp
+            hugr.add_order_link(last, outp)
```
mark-koch, line 655: Should this be in `finally`?

tests/integration/test_side_effect_ordering.py
```diff
@@ -53,6 +53,28 @@ def check_order(hugr: Hugr, nodes: list[Node]) -> None:
     assert len(nodes) == 0
 
 
+def test_input_result_output(validate):
```
mark-koch, line 56: Could you also update the `test_nested` case to check that all inner outputs are connected?

guppylang-internals/src/guppylang_internals/compiler/core.py
mark-koch: Mhmm yeah I also don't see a better way...

guppylang-internals/src/guppylang_internals/compiler/core.py
acl-cqc: I went through a few opinions there. Kinda felt that if building the Hugr falls over with an exception, the less messing around we do afterwards the easier it will be to figure out why the exception happened....but OTOH if anyone's gonna recover from that exception and still use the Hugr then it'd be better to have finished the job. What do you think is most likely?!

mark-koch: I don't have a strong opinion, happy to leave it as is. Just wondering if you had considered it

tests/integration/test_side_effect_ordering.py
acl-cqc: Good call, done!

PR #1314, opened 2025-10-23
chore: benchmark PR against last main metrics

ferbetanzo:
Using `--start-point-max-versions 1` so PR benchmarks are only compared against the last main metrics.

Also, I noticed that when exporting hugrs from the integration tests some files were being overwritten because they have the same function/test case name. I've added the module name to the file name of the exported files to prevent this.


ss2165 (approved):
thanks

PR #1318, opened 2025-10-23
docs: Update bibentries to use arXiv

qartik:
- https://arxiv.org/pdf/2510.12582
- https://arxiv.org/pdf/2510.13082


aborgna-q (approved):
Should we do a cff file instead? The .bib gets auto-generated from it
https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/about-citation-files#about-citation-files

PR #1327, opened 2025-11-03
feat!: Allow dynamic tag and signal in `panic`/`exit`

### LINKED ISSUE #1058: Use `MakeError` op when panicking instead of requiring constant

tatiana-s:
Closes #1058

BREAKING CHANGE: `guppylang_internals.std._internal.compiler.prelude.build_error` renamed to `build_static_error`
BREAKING CHANGE: `signal` and `msg` fields in `PanicExpr` node are now expressions


guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
```diff
@@ -90,7 +98,7 @@ def build_panic(
     return builder.add_op(op, err, *args)
 
 
-def build_error(builder: DfBase[P], signal: int, msg: str) -> Wire:
+def build_static_error(builder: DfBase[P], signal: int, msg: str) -> Wire:
```
mark-koch, line 101: I think this rename technically makes this a breaking change for `guppy_internals`. Not sure why the semver check didn't catch it?

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -275,8 +275,8 @@ class PanicExpr(ast.expr):
     """A `panic(msg, *args)` or `exit(msg, *args)` expression ."""
 
     kind: ExitKind
-    signal: int
-    msg: str
+    signal: ast.expr
+    msg: ast.expr
```
mark-koch, line 279: I guess this is also breaking?

PR #1331, opened 2025-11-04
docs: Add example notebook on QAOA

CalMacCQ:
EDIT: As of 2nd March I've finished off the explanatory text and acknowledged some of the performance limitations. 

--------------------

Implementation of QAOA in Guppy. The code was based on an pytket example notebook implementing QAOA which I ported over to Guppy.

The implementation works and gives the correct answer for the maxcut instnace for a seven vertex graph. However there are some outstanding issues.

1. The example takes a long time to run (over 1 minute)
```
Guppy performance (with Selene)
CPU times: user 31.8 s, sys: 3.21 s, total: 35 s
Wall time: 1min 10s
```

```
pytket with AerBackend 
(Note that compilation essentially plays no role here, not exactly apples to apples)
CPU times: user 3.35 s, sys: 1.35 s, total: 4.69 s
Wall time: 511 ms
```
I expect the overhead is in the repeated compilation of Guppy functions on every QAOA iteration (100 iterations in this case, 2000 shots, 7 qubits).

This was running afoul of the 120s timeout imposed by C.I. (The runners are slower than my local machine) This in turn means that I've had to skip the notebook execution test here which is unfortunate. See the changes I made to the `test_notebooks.py` file.

2. The classical optimisation strategy is very naive... I'm using the most basic possible strategy here. I'm just sampling parameter vales from a uniform distribution see `solve_maxcut_instance`. Note that this is not the cause of the slow runtime as the same optimisation strategy is applied in the pytket version.

 I'd like to use a scipy optimizer or something but that would require a fair bit of refactoring. I only get a float back from `scipy.optimize.minimize` so I need to refactor to account for the fact that I'm not getting a `QsysResult` back in the loop.

### DIFF
--- a/tests/integration/test_notebooks.py
+++ b/tests/integration/test_notebooks.py
@@ -6,6 +6,10 @@ from pathlib import Path
 example_notebooks = list(
     (Path(__file__).parent.parent.parent / "examples").glob("*.ipynb")
 )
+
+example_notebooks.remove(
+    Path(__file__).parent.parent.parent / "examples" / "qaoa_maxcut_example.ipynb"
+)
 # Turn paths into strings, otherwise pytest doesn't display the names
 example_notebooks = [str(f) for f in example_notebooks]
 
@@ -21,10 +25,6 @@ def test_example_notebooks(nb_regression, notebook: Path):
 
 integration_notebooks = list((Path(__file__).parent / "notebooks").glob("*.ipynb"))
 
-
-integration_notebooks.remove(
-    Path(__file__).parent.parent.parent / "examples" / "qaoa_maxcut_example.ipynb"
-)
 
 # Turn paths into strings, otherwise pytest doesn't display the names
 integration_notebooks = [str(f) for f in integration_notebooks]


examples/qaoa_maxcut_example.ipynb
qartik, line 1:
```
❯ typos
error: `demonstarting` should be `demonstrating`
   ╭▸ ./examples/qaoa_maxcut_example.ipynb:13:43
   │
13 │     "In this notebook tutorial we will be demonstarting how to solve the maxcut problem using the quantum approximate optimisation algorithm (QAOA). This is a variational algorithm where we run a par…
   ╰╴                                          ━━━━━━━━━━━━━
error: `paramaterised` should be `parameterised`
   ╭▸ ./examples/qaoa_maxcut_example.ipynb:95:62
   │
95 │     "QAOA is a variational algorithm, which is to say that a paramaterised state is prepared, with the parameters varied to improve the solution. We will have $2p$ parameters where $p$ is our number …
   ╰╴
```


Similarly, "optimsing" on line 13.

CalMacCQ: Thanks, fixed.

qartik (approved):

PR #1339, opened 2025-11-11
feat!: Validate signatures against wasm file

### LINKED ISSUE #1093: Validation for WASM signatures
Follow up to #755.

It will be a nicer user experience if WASM signatures from the supplied wasm file/module were validated at guppy compile time to match against the user-specified signatures in a `@guppy.wasm_module`.

croyzor:
Resolves #1093 

BREAKING CHANGE:
* The `@wasm_module` decorator now raises a compile time error unless it points to a matching `.wasm` file in the users filesystem
* `@wasm` functions raise errors unless they match the corresponding functions in the `.wasm` file
* Redo the `FirstArgNotModule` error class, which is used for WASM errors

### DIFF
--- a/guppylang-internals/src/guppylang_internals/decorator.py
+++ b/guppylang-internals/src/guppylang_internals/decorator.py
@@ -291,7 +291,11 @@ def ext_module_decorator(
                                 WasmFunctionNotInFile(
                                     wasm_def.defined_at,
                                     wasm_def.name,
-                                    pathlib.Path(wasm_sigs.filename).name,
+                                ).add_sub_diagnostic(
+                                    WasmFunctionNotInFile.WasmFileNote(
+                                        None,
+                                        wasm_sigs.filename,
+                                    )
                                 )
                             )
                     if isinstance(wasm_sig_or_err, FunctionType):
--- a/guppylang-internals/src/guppylang_internals/wasm_util.py
+++ b/guppylang-internals/src/guppylang_internals/wasm_util.py
@@ -47,10 +47,14 @@ class WasmFileNotFound(Error):
 @dataclass(frozen=True)
 class WasmFunctionNotInFile(Error):
     title: ClassVar[str] = (
-        "Declared wasm function `{function}` isn't exported by wasm file `{file}`"
+        "Declared wasm function `{function}` isn't exported by wasm file"
     )
     function: str
-    file: str
+
+    @dataclass(frozen=True)
+    class WasmFileNote(Note):
+        message: ClassVar[str] = "Wasm file: {filename}"
+        filename: str
 
 
 @dataclass(frozen=True)
@@ -115,7 +119,9 @@ def decode_wasm_functions(filename: str) -> ConcreteWasmModule:
                     case [output]:
                         data = decode_sig(fun_ty.params, output)
                     case _multi:
-                        data = f"Multiple output types in function `{fn.name}`"
+                        data = (
+                            f"Multiple output types unsupported in function `{fn.name}`"
+                        )
             case _:
                 data = f"`{fn.name}` is not exported as a function"
         function_sigs[fn.name] = data
--- a/tests/error/wasm_errors/wasm_multi_output.err
+++ b/tests/error/wasm_errors/wasm_multi_output.err
@@ -1,6 +1,6 @@
 Error: Invalid signature for @wasm function `multiple_outputs`
 in wasm file: `$WASM`
 
-Note: Multiple output types in function `multiple_outputs`
+Note: Multiple output types unsupported in function `multiple_outputs`
 
 Guppy compilation failed due to 1 previous error
--- a/tests/error/wasm_errors/wasm_no_func.err
+++ b/tests/error/wasm_errors/wasm_no_func.err
@@ -1,3 +1,5 @@
-Error: Declared wasm function `foo` isn't exported by wasm file `test.wasm`
+Error: Declared wasm function `foo` isn't exported by wasm file
+
+Note: Wasm file: $WASM
 
 Guppy compilation failed due to 1 previous error


tests/resources/arith.wasm
croyzor, line 1: This is the same test file as used in eldarion, I'm only using the methods `two: int` and `add: int, int -> int`

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -211,20 +227,29 @@ def type_def_wrapper(
...
+
+    return inner_fun
 
 
 def ext_module_decorator(
     type_def: Callable[[DefId, str, ast.AST | None, str, str | None], OpaqueTypeDef],
     init_compiler: CustomInoutCallCompiler,
     discard_compiler: CustomInoutCallCompiler,
     init_arg: bool,  # Whether the init function should take a nat argument
+    wasm_sigs: ConcreteWasmModule
```
croyzor, line 249: This is a change to the `guppylang-internals` api that's used by `gpuppy`, but should be fine because we give a default value for the new argument

tests/integration/test_wasm.py
```diff
@@ -1,99 +1,80 @@
...
-        x = mod.foo()
-        y = mod.bar(x)
-        mod.discard()
-        return x
-
-    mod = main.compile_function()
-    validate(mod)
-
-
-def test_wasm_types(validate):
```
croyzor, line 52: This test doesn't work anymore since we have to provide real `.wasm` files. Also the types we're allowed to use in wasm are more restrictive than in this test

guppylang-internals/src/guppylang_internals/decorator.py
qartik: To be clear, gpuppy is now guppy-gpu: https://github.com/CQCL/guppy-gpu We should consider merging it upstream as part of guppylang at some point.

tests/integration/test_wasm.py
```diff
@@ -1,99 +1,80 @@
...
     class MyWasm:
-        @wasm
+        @wasm(1)
         def foo(self: "MyWasm") -> int: ...
 
-        @guppy
-        def bar(self: "MyWasm", x: int) -> int:
-            return x + 1
+        @wasm(0)
+        def bar(self: "MyWasm", x: int, y: int) -> int: ...
```
qartik, line 48: How do these non-wasm methods work in the test? Don't they fail validation?

tests/resources/arith.wasm
qartik: We may want to add tests for all supported WASM types: i64 (64-bit integer) and double (64-bit floating-point value)

guppylang-internals/pyproject.toml
```diff
@@ -36,6 +36,7 @@ dependencies = [
     "typing-extensions >=4.9.0,<5",
     "tket-exts ~= 0.12.0",
     "hugr ~= 0.14.1",
+    "wasm-tob>=1.0.1",
```
qartik: This package seems unmaintained, could you use an alternative like wasmtime that's officially supported by Bytecode Alliance? https://pypi.org/project/wasmtime/

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -201,6 +209,14 @@ def dec(c: type[T]) -> type[T]:
 def wasm_module(
     filename: str,
 ) -> Callable[[builtins.type[T]], GuppyDefinition]:
+    wasm_file = pathlib.Path(filename)
+    if wasm_file.is_file():
+        with pathlib.Path.open(wasm_file, "rb") as f:
+            wasm_bytes = f.read()
+            wasm_sigs = decode_wasm_functions(filename, wasm_bytes)
+    else:
+        raise GuppyError(WasmFileNotFound(None, filename))
```
mark-koch, line 217: I think this might be better as a regular Python exception since it's invoked when the decorator is called, not when the program is compiled?

guppylang-internals/src/guppylang_internals/wasm_util.py
```diff
@@ -0,0 +1,143 @@
...
+
+    from wasm_tob.decode import ModuleFragment
+
+
+# The thing to be imported elsewhere
+@dataclass
+class ConcreteWasmModule:
+    filename: str
+    # Function names in order for looking up by index
+    functions: dict[int, str]
```
mark-koch: Why is this a dict instead of a list? Are the indices not contiguous?

guppylang-internals/src/guppylang_internals/wasm_util.py
```diff
@@ -0,0 +1,143 @@
...
+    function_sigs: dict[str, FunctionType | str]
```
mark-koch, line 23: Is there a reason you want to put a string here instead of just `FunctionType | None`. Are there situations where this error string is shown to the user?

guppylang-internals/src/guppylang_internals/wasm_util.py
```diff
@@ -0,0 +1,143 @@
...
+    funcs: dict[str, FunctionType | str] = {}
+    fun_tys: list[FunctionType | str] = []
+    fun_ptrs: list[int] = []
+    ordered_func_names: dict[int, str] = {}
+    for _, frag_data in wasm_module:
+        # If we find a type section, add each type to our ordered list of types
+        # in `fun_tys`
+        if frag_data.id == SEC_TYPE:
+            for ty in frag_data.payload.entries:
+                # TODO: Do we need to worry about the form?
```
mark-koch: Sorry, what does this mean? 😅

pyproject.toml
```diff
@@ -45,6 +45,13 @@ allow_redefinition = true
 [[tool.mypy.overrides]]
 module = ["miette_py"]
 
+[[tool.mypy.overrides]]
+module = ["wasm_tob"]
+ignore_missing_imports = true
```
mark-koch: Why is this needed?

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -502,7 +504,28 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+    def setup(self, wasm_sigs: ConcreteWasmModule) -> None:
+        self.wasm_sigs = wasm_sigs
+
     def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
+        type_in_wasm: FunctionType = DEF_STORE.wasm_functions[self.func.id]
+
+        # Drop the first arg because it should be "self"
+        expected_type = FunctionType(self.func.ty.inputs[1:], self.func.ty.output)
+
+        if expected_type != type_in_wasm:
```
mark-koch: Maybe this check would make more sense in `RawWasmFunctionDef.parse` instead of checking every time we call the function?

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -502,7 +504,28 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+            raise GuppyTypeError(
+                WasmSigMismatchError(self.node)
```
mark-koch: At least the error location should point to the declaration instead of the call I think. Then, maybe the ` WasmSigMismatchError.Declaration` hint wouldn't be necessary since the span is already on the signature?

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -244,6 +269,36 @@ def dec(cls: builtins.type[T]) -> GuppyDefinition:
...
+                                WasmFunctionNotInFile(
+                                    wasm_def.defined_at,
+                                    wasm_def.name,
+                                    pathlib.Path(wasm_sigs.filename).name,
+                                )
+                            )
+                    if isinstance(wasm_sig_or_err, FunctionType):
+                        DEF_STORE.register_wasm_function(wasm_def.id, wasm_sig_or_err)
+                    elif isinstance(wasm_sig_or_err, str):
+                        raise InternalGuppyError(wasm_sig_or_err)
```
mark-koch: Why is this an internal error? Couldn't users trigger this by providing a bad wasm file?

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -244,6 +269,36 @@ def dec(cls: builtins.type[T]) -> GuppyDefinition:
             for val in cls.__dict__.values():
                 if isinstance(val, GuppyDefinition):
                     DEF_STORE.register_impl(ext_module.id, val.wrapped.name, val.id)
+                    wasm_def: RawWasmFunctionDef
+                    if isinstance(val, RawWasmFunctionDef):
+                        wasm_def = val
```
mark-koch: Do we actually need this case? Afaik everything is always wrapped into a `GuppyFunctionDefinition`?

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -244,6 +269,36 @@ def dec(cls: builtins.type[T]) -> GuppyDefinition:
...
+                    elif isinstance(val, GuppyFunctionDefinition) and isinstance(
+                        val.wrapped, RawWasmFunctionDef
+                    ):
+                        wasm_def = val.wrapped
+                    else:
+                        continue
+                    assert wasm_sigs is not None
```
mark-koch, line 281: Should this be a user error instead of an assert?

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -244,6 +269,36 @@ def dec(cls: builtins.type[T]) -> GuppyDefinition:
...
+                        else:
+                            raise GuppyError(
+                                WasmFunctionNotInFile(
+                                    wasm_def.defined_at,
+                                    wasm_def.name,
+                                    pathlib.Path(wasm_sigs.filename).name,
+                                )
+                            )
+                    if isinstance(wasm_sig_or_err, FunctionType):
+                        DEF_STORE.register_wasm_function(wasm_def.id, wasm_sig_or_err)
```
mark-koch, line 302: Have you considered doing the signature checking at the same time as the parsing of the wasm file? Then you wouldn't need to hold on to the signatures in the `DEF_STORE` and the checking logic would be more self contained?

guppylang-internals/src/guppylang_internals/wasm_util.py
```diff
@@ -0,0 +1,143 @@
...
+@dataclass(frozen=True)
+class WasmFileNotFound(Error):
+    title: ClassVar[str] = "Wasm file `{file}` not found"
+    file: str
+
+
+@dataclass(frozen=True)
+class WasmFunctionNotInFile(Error):
+    title: ClassVar[str] = (
+        "Declared wasm function `{function}` isn't exported by wasm file {file}"
```
mark-koch:
```suggestion
        "Declared wasm function `{function}` isn't exported by wasm file `{file}`"
```

guppylang-internals/src/guppylang_internals/decorator.py
croyzor: Yes -- the downside is that there's no locaation info for the error messages, that's why i went this route

tests/resources/arith.wasm
croyzor: Have made a bespoke `test.wasm` file which deals in floats as well, and included the `.wat` source in the resources directory

tests/integration/test_wasm.py
croyzor: These are the same wasm methods, they're just being looked up by their index in the file (`1`, `0`) rather than by their name

guppylang-internals/pyproject.toml
croyzor: Good point! At first glance I had thought that `wasmtime-py` didn't have the functionality we needed, but happily I was wrong!

guppylang-internals/src/guppylang_internals/decorator.py
croyzor: I have tried it that way in a commit here, but the resulting error is nastier so I reverted it

guppylang-internals/src/guppylang_internals/wasm_util.py
croyzor: Now it's a `list[str]`. (It used to be a `dict` because the entries weren't added in order but I suppose I could've initialised it to the right size...)

guppylang-internals/src/guppylang_internals/decorator.py
croyzor: If it's a user error, it will get caught by another mechanism. Have added a comment

guppylang-internals/src/guppylang_internals/decorator.py
croyzor: Good point!

guppylang-internals/src/guppylang_internals/wasm_util.py
croyzor: Yes the main purpose of the string is to bubble up to the user to say the wasm file functions don't match the format we expect, allowing non-compliant functions in the file as long as they're not exposed via guppy.
lemme add some tests for this...

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
croyzor: Yes, now refactored and much nicer! I thought this wouldn't work because of functions getting parsed before their containing class, but I now realise it does 😁

tests/error/wasm_errors/wasm_type_mismatch.err
```diff
@@ -0,0 +1,13 @@
+Error: Wasm signature mismatch (at $FILE:10:4)
+   | 
+ 8 | class Foo:
+ 9 |     @wasm
+10 |     def add(self: "Foo", x: int, y: int, z: int) -> int: ...
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Signature of wasm function didn't match that in provided
+   |                                                              file
```
qartik:
```suggestion
   |                                                              wasm file
```
I think this is less ambiguous.

tests/error/wasm_errors/wasm_type_mismatch.err
```diff
@@ -0,0 +1,13 @@
...
+
+Note: Declared signature: (int, int, int) -> int
+
+Note: Signature in wasm:  (int, int) -> int
+
```
qartik, line 12: Do we need this many blank lines around `Note`s? I'd prefer it to have no new line between the two `Note`s.

tests/error/wasm_errors/wasm_non_function.err
```diff
@@ -0,0 +1,6 @@
+Error: Invalid signature for @wasm function `non_fn`
+in wasm file: `$WASM`
```
qartik, line 2: Is `$WASM` supposed to be the name of the WASM file here?

tests/error/wasm_errors/wasm_no_func.err
```diff
@@ -0,0 +1,3 @@
+Error: Declared wasm function `foo` isn't exported by wasm file test.wasm
```
qartik:
```suggestion
Error: Declared wasm function `foo` isn't exported by wasm file `test.wasm`
```
For consisetency in errors.

tests/error/wasm_errors/wasm_multi_output.err
```diff
@@ -0,0 +1,6 @@
+Error: Invalid signature for @wasm function `multiple_outputs`
+in wasm file: `$WASM`
```
qartik, line 2: Same as the other comment, `$WASM` should be expanded to the filename.

tests/error/wasm_errors/wasm_bad_output.err
```diff
@@ -0,0 +1,6 @@
+Error: Invalid signature for @wasm function `bad_output_type`
+in wasm file: `$WASM`
+
+Note: Invalid output: f32
```
qartik:
```suggestion
Note: Unsupported type: f32
```
or "Unsupported output type"

tests/error/wasm_errors/wasm_bad_output.err
```diff
@@ -0,0 +1,6 @@
+Error: Invalid signature for @wasm function `bad_output_type`
+in wasm file: `$WASM`
```
qartik, line 2: I am still confused, is `$WASM` referring to the guppy file or wasm file?

tests/error/wasm_errors/wasm_bad_input.err
```diff
@@ -0,0 +1,6 @@
+Error: Invalid signature for @wasm function `bad_input_type`
+in wasm file: `$WASM`
+
+Note: Invalid param: i32
```
qartik:
```suggestion
Note: Unsupported type: i32
```
or "Unsupported param type"

qartik (changes requested):
Looking good, left some feedback for improvements to error messages.

tests/error/wasm_errors/wasm_type_mismatch.err
croyzor: I don't have a strong opinion 🤷

tests/error/wasm_errors/wasm_non_function.err
croyzor: Yes, `$WASM` is an abbreviation of the full path of the wasm file, so that the golden file looks the same regardless of filesystem structure. This is implemented in `tests/error/util.py`

tests/error/wasm_errors/wasm_bad_output.err
croyzor: `$WASM` is the wasm file, for users it would be displayed as e.g. `/home/username/guppy/tests/resources/test.wasm`

tests/error/wasm_errors/wasm_multi_output.err
```diff
@@ -0,0 +1,6 @@
...
+
+Note: Multiple output types in function `multiple_outputs`
```
qartik:
```suggestion
Note: Multiple output types unsupported in function `multiple_outputs`
```

tests/error/wasm_errors/wasm_non_function.err
qartik: I see, why is it that in some of the errors, the name of the wasm file is reported while in some others it is reported as `$WASM`? See https://github.com/CQCL/guppylang/pull/1339/files#diff-f8b76a0b2eb3a18229b3ead461754622a3c1950d739f36e1685f729b229f3628 (tests/error/wasm_errors/wasm_no_func.err)

croyzor: i actually have no idea... I assume the name isn't getting abstracted to `$WASM` because its not printing the full path. It not printing the full path means that we don't _need_ to abstract it anyway, but I'm not sure why there's the discrepancy between printing the full path and just printing the filename

croyzor: Okay, I worked it out and have updated the code so that the full path to the wasm file is printed in both cases (hence, abstracted to `$WASM`)

tests/error/wasm_errors/wasm_multi_output.err
croyzor: Updated

tests/error/wasm_errors/wasm_no_func.err
```diff
@@ -0,0 +1,5 @@
+Error: Declared wasm function `foo` isn't exported by wasm file
+
+Note: Wasm file: $WASM
```
qartik, line 3:
```suggestion
Error: Declared wasm function `foo` isn't exported by wasm file: $WASM
```

Since you have access to the name of the wasm file in the error context, I think it makes sense to avoid a trivial `Note`.

croyzor: I did this for a sneaky reason: the renaming to `$WASM` happens _after_ the error message is rendered.
A line break may be inserted depending on the full path of the test file, which can lead to differences in golden files between my machine and CI, which I wanted to avoid!

qartik: I believe such testing considerations should not dictate the quality of error messages. If it's too hard to avoid this hack, then I care less.

croyzor: Sure, I just consider the two to be of the same quality - the name of the wasm file is not a crucial part of the error message (while we only allow one module anyway), but an extra piece of information that will help in some cases, that's why it made sense to me for it to be a note

PR #1358, opened 2025-11-26
fix: Handle subscript borrows involving index coercions

### LINKED ISSUE #1356: Borrow error when indexing with nat
```
@guppy.struct
class Iceberg(Generic[n]):
    top: qubit
    bottom: qubit
    block: array[qubit, n]

    @guppy
    def cx(self, q: nat) -> None:
        cx(self.bottom, self.block[q])
```

gives

```
Error: Not able to borrow subscripted elements (at /home/doug/code/iceberg-repro/repro.gpy.py:21:24)
   | 
19 |     @guppy
20 |     def cx(self, q: nat) -> None:
21 |         cx(self.bottom, self.block[q])
   |                         ^^^^^^^^^^^^^ Expression of type `array[qubit, n]` is not able to borrow
   |                                       subscripted elements

Help: Fix signature of method `array[qubit, n].__setitem__`:  Expected
`(array[qubit, n], nat, qubit @owned) -> None`, got `forall L, n: nat. (array[L,
n], int, L @owned) -> None`
```

changing `q: nat` to `q: int` fixes the problem.

mark-koch:
Fixes #1356

### DIFF
--- a/tests/error/inout_errors/subscript_not_setable.err
+++ b/tests/error/inout_errors/subscript_not_setable.err
@@ -6,7 +6,7 @@ Error: Not able to borrow subscripted elements (at $FILE:19:8)
    |         ^^^^ Expression of type `MyImmutableContainer` is not able to
    |              borrow subscripted elements
 
-Help: Implement missing method: `__setitem__: (MyImmutableContainer, int, qubit
+Help: Implement missing method: `__setitem__: (MyImmutableContainer, ?T, qubit
 @owned) -> None`
 
 Guppy compilation failed due to 1 previous error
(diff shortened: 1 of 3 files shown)


tests/error/inout_errors/subscript_not_setable.err
```diff
@@ -6,7 +6,7 @@ Error: Not able to borrow subscripted elements (at $FILE:19:8)
    |         ^^^^ Expression of type `MyImmutableContainer` is not able to
    |              borrow subscripted elements
 
-Help: Implement missing method: `__setitem__: (MyImmutableContainer, int, qubit
+Help: Implement missing method: `__setitem__: (MyImmutableContainer, ?T, qubit
```
mark-koch, line 9: This help message is getting slightly worse, but imo acceptable to fix the bug

Note that our handling of subscripts will probably change anyways once we go ahead with #1260

Issue #1352, opened 2025-11-18
More dynamic result labels


croyzor:
This is a little bit blocked down the stack because the "result" op takes the label argument as a type arg instead of a string wire. But we should do it anyway!

PR #1361, opened 2025-11-26
refactor!: Implement `result` using overloads instead of a custom node

### LINKED ISSUE #1357: Implement `result` using guppy source
`result`, though it looks like a standard guppy function call, is actually implemented as an expression with custom type checking and compilation.

Initially guppy source was not capable of capturing it's signature: 
- static string tag
- set of possible value types (numbers and arrays of numbers)

With two recent guppy features this should now be expressible as source, saving compiler code and making the function more consistent and flexible. These are:
- `@comptime` function parameters, for the string tag
- function overloading to express the set of value types

Each overload would call some implementation function which compiles to the respective hugr up (using @hugr_op ideally).

mark-koch:
Closes #1357

Depends on #1360

Note that this PR is only breaking for `guppylang-internals`

BREAKING CHANGE: Removed `guppylang_internals.nodes.ResultExpr`
BREAKING CHANGE: Moved `guppylang_internals.std._internal.checker.{TAG_MAX_LEN, TooLongError}` to `guppylang_internals.std._internal.compiler.platform`


tests/error/result_errors/result_tag_not_static.err
```diff
@@ -1,8 +1,19 @@
-Error: Expected a string literal (at $FILE:7:11)
+Error: Invalid call of overloaded function (at $FILE:7:11)
   | 
 5 | @compile_guppy
 6 | def foo(y: bool, x: int) -> None:
 7 |     result("foo" if y else "bar", x)
-  |            ^^^^^^^^^^^^^^^^^^^^^ Expected a string literal
+  |            ^^^^^^^^^^^^^^^^^^^^^^^^ No variant of overloaded function `result` takes arguments
+  |                                     `str`, `int`
```
mark-koch, line 7: This error is arguable worse

On the other hand this can be dropped anyways of we go with #1352

tests/error/result_errors/result_tag_too_big.err
```diff
@@ -1,9 +1,9 @@
-Error: Tag too long (at $FILE:7:11)
+Error: Tag too long (at $FILE:7:4)
   | 
 5 | def foo(y: bool) -> None:
 6 |     # each tick or cross is 3 bytes. The cross sends the tag over the limit.
 7 |     result("✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅✅❌", y)
-  |            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Result tag is too long
+  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Result tag is too long
```
mark-koch, line 6: The error now spans the entire call instead of just the tag argument, but I don't see a way to improve this without a bigger refactor of custom compiler :(

tests/error/result_errors/result_array_not_numeric.err
```diff
@@ -1,11 +1,19 @@
...
 5 | @compile_guppy
 6 | def foo(x: array[tuple[int, bool], 42]) -> None:
 7 |     result("foo", x)
-  |                   ^ Expression of type `array[(int, bool), 42]` is not a valid
-  |                     result.
+  |            ^^^^^^^^ No variant of overloaded function `result` takes arguments
+  |                     `str`, `array[(int, bool), 42]`
 
-Note: Only numeric values or arrays thereof are allowed as results
+Note: Available overloads are:
```
ss2165, line 9: this is a nice error outcome

tests/error/result_errors/result_tag_too_big.err
ss2165: doesn't seem like a big issue

PR #1378, opened 2025-12-11
feat!: Add qubit hints on Guppy functions, allowing elision when building emulators

### LINKED ISSUE #1297: User hint for maximum number of qubits allocated
## Suggested syntax

```python
from guppylang.hints import max_qubits

N_QB = 4
@max_qubits(N_QB)
@guppy
def foo() -> None:
    a = array(qubit() for _ in range(py(N_QB))
    discard_array(a)
```

## Compilation
Attach as standard metadata key to entrypoint op. Key should be specified in tket2.

## Usage

n_qubits parameter to emulator can now be optional:

```python
foo.emulator().run() # will raise error if `foo` is not annotated
```

maximilianruesch:
Added a keyword argument to `@guppy` decorator for annotation of maximum qubits required for the function, which allows the emulator to be built without specifying a number of qubits (if given on the entrypoint function):
```python
@guppy(max_qubits=1)
def main() -> None:
    result("c", measure(qubit()))

main.emulator().run()
```
The feature (still) complains if you:
- pass no qubit count anywhere
- pass a qubit count to the emulator that is insufficient to cover the qubit count hinted on the entrypoint

The emulation will still crash if you do sufficient hinting, but the allocated number of qubits is too low for the number of required qubits.

Additionally (or perhaps more importantly for future developments), this PR adds a notion of metadata to the Guppy compiler which gets passed onto the HUGR. This metadata is currently only attached to function definitions, but could be extended to live on all definitions (and thus be able to be attached to all HUGR nodes that are created from Guppy). For a full list of TODOs, see below.

BREAKING CHANGE:
- The first argument to `add_unitarity_metadata` is now named `node` instead of `func`, since its type was raised to allow for more HUGR nodes to be fed. Migration is trivial.

> [!NOTE]
> This PR is only "breaking" `guppylang-internals`, and not `guppylang`

Before this PR is merged (or relatively soon thereafter), the following things must/should/will happen:
- HUGR adds type-safe metadata access and refactors its core keys: https://github.com/Quantinuum/hugr/pull/2755
- HUGR exports type-safe access and core keys to Python bindings: https://github.com/Quantinuum/hugr/issues/2757
- Guppy uses the new core keys internally
- TKET uses the new pattern to define a unitary key (e.g.`tket.modify.unitarity`, replacing the [string literal](https://github.com/Quantinuum/tket2/blob/2918a67c6c5ce688ccd184192284355205047091/tket/src/modifier.rs#L66)) and defines a new `max_qubits` key (e.g. `tket.hint.max_qubits`), exporting both to Python bindings
  - The rust side: https://github.com/Quantinuum/tket2/pull/1328 
- Guppy adds TKET as a direct dependency
- Guppy uses the new keys internally to replace `"unitary"` as well as the existing string literal for MaxQubits
- Guppy adds the ability to propagate metadata to any HUGR node, attaching it to any def (?)

Closes #1297

Release-As: 0.21.8

### DIFF
--- a/tests/integration/test_emulator.py
+++ b/tests/integration/test_emulator.py
@@ -17,6 +17,7 @@ from guppylang.std.quantum import (
     h,
     x,
     t,
+    measure_array,
 )
 from guppylang.std.angles import angle, pi
 from guppylang.std.qsystem import zz_max, zz_phase, phased_x, rz as qsystem_rz
@@ -108,7 +109,7 @@ def test_no_given_qubits() -> None:
             r"`@guppy\(max_qubits=...\)`."
         ),
     ):
-        main.emulator().run()
+        main.emulator().coinflip_sim().with_seed(0).with_shots(1).run()
 
 
 def test_hinted_qubits() -> None:
@@ -116,15 +117,18 @@ def test_hinted_qubits() -> None:
     def main() -> None:
         result("c", measure(qubit()))
 
-    main.emulator().run()
+    shots = main.emulator().coinflip_sim().with_seed(0).with_shots(1).run()
+    assert shots[0].as_dict()["c"] == 1
 
 
 def test_hinted_qubits_with_given_qubits() -> None:
     @guppy(max_qubits=1)
     def main() -> None:
-        result("c", measure(qubit()))
+        qubits = array(qubit() for _ in range(4))
+        result("c", measure_array(qubits))
 
-    main.emulator(n_qubits=4).run()
+    shots = main.emulator(n_qubits=4).coinflip_sim().with_seed(0).with_shots(1).run()
+    assert shots[0].as_dict()["c"] == [1, 0, 1, 0]
 
 
 def test_hinted_qubits_with_insufficient_given_qubits() -> None:
@@ -139,7 +143,7 @@ def test_hinted_qubits_with_insufficient_given_qubits() -> None:
             r"number of qubits hinted on the entrypoint \(3\)."
         ),
     ):
-        main.emulator(n_qubits=1).run()
+        main.emulator(n_qubits=1).coinflip_sim().with_seed(0).with_shots(1).run()
 
 
 def test_statevector() -> None:


maximilianruesch (commented):
Note to self: This needs a changelog entry

guppylang-internals/src/guppylang_internals/definition/metadata.py
```diff
@@ -0,0 +1,84 @@
+from abc import ABC
```
ss2165, line 4: nit: add module docstring

guppylang-internals/src/guppylang_internals/definition/metadata.py
```diff
@@ -0,0 +1,84 @@
...
+    title: ClassVar[str] = "Metadata key already set"
+    message: ClassVar[str] = "Received two values for the metadata key `{key}`"
+    key: str
+
+
+@dataclass(frozen=True)
+class ReservedMetadataKeysError(Fatal):
+    title: ClassVar[str] = "Metadata key is reserved"
+    message: ClassVar[str] = (
+        "The following metadata keys are reserved by Guppy but provided in additional "
```
ss2165:
```suggestion
        "The following metadata keys are reserved by Guppy but also provided in additional "
```

guppylang/src/guppylang/decorator.py
```diff
@@ -639,8 +646,10 @@ def _with_optional_kwargs(
 
 
 @hide_trace
-def _parse_kwargs(kwargs: GuppyKwargs) -> UnitaryFlags:
-    """Parses the kwargs dict specified in the `@guppy` decorator."""
+def _parse_kwargs(kwargs: GuppyKwargs) -> tuple[UnitaryFlags, GuppyMetadata]:
```
ss2165, line 649: not very satisfying to return a pair here, should "UnitaryFlags" be conceptually expanded to "DecoratorHints" or something similar?

guppylang/src/guppylang/defs.py
```diff
@@ -99,7 +101,8 @@ def emulator(
 
 
         Args:
-            n_qubits: The number of qubits to allocate for the function.
+            n_qubits: The number of qubits to allocate for the function. If it is not
+            provided, the function has to declare is maximum number of qubits.
```
ss2165:
```suggestion
            provided, the function has to declare the maximum number of qubits it needs in the decorator, e.g. `@guppy(max_qubits=10)`.
```

tests/integration/test_emulator.py
```diff
@@ -94,6 +95,53 @@ def main() -> None:
...
+    ):
+        main.emulator().run()
+
+
+def test_hinted_qubits() -> None:
+    @guppy(max_qubits=1)
+    def main() -> None:
+        result("c", measure(qubit()))
+
+    main.emulator().run()
```
ss2165:
```suggestion
    main.emulator().coinflip_sim().with_seed(1234).run()
```
and similar for the others - default is costly statevector sim

ss2165: + ideally assert something about the outcome

tests/integration/test_emulator.py
```diff
@@ -94,6 +95,53 @@ def main() -> None:
...
+
+
+def test_hinted_qubits_with_given_qubits() -> None:
+    @guppy(max_qubits=1)
+    def main() -> None:
+        result("c", measure(qubit()))
```
ss2165: you could actually allocate 4 here to show the hint can be overriden

guppylang/src/guppylang/decorator.py
maximilianruesch: UnitaryFlags is used as an annotation to the function _type_, whereas GuppyMetadata lives on the definition. One could introduce a wrapper object that collects all decorator hints, and has accessor methods for retrieving UnitaryFlags / Metadata (for the function definition) when needed. Completely replacing UnitaryFlags with this new object is not something I would do. What are your thoughts on this? We might need to duplicate metadata keys into this new object when we want to receive them from the decorator, although this could also be a good thing...

ss2165: happy to leave it for now - something to revisit when we want to add other kinds of "hints"

ss2165 (approved):
🚀

PR #1382, opened 2025-12-15
fix!: improve diagnostics rendering

nicolaassolini-qntm:
improvement to the error message rendering to answer to [issue 608](https://github.com/Quantinuum/guppylang/issues/608)

BREAKING CHANGE: 

- `guppylang-internals/src/guppylang_internals/diagnostic` `DiagnosticsRenderer.PREFIX_CONTEXT_LINES`: Public object removed
- changed in `guppylang-internals/src/guppylang_internals/diagnostic` various error message
- (guppylang-internals): update a class var on the error

### DIFF
--- a/tests/error/nested_errors/linear_capture.err
+++ b/tests/error/nested_errors/linear_capture.err
@@ -10,6 +10,6 @@ Note:
   | 
 6 | @guppy
 7 | def foo(q: qubit @owned) -> qubit:
-  |         --------------- `q` already defined here
+  |         --------------- `q` defined here
 
 Guppy compilation failed due to 1 previous error


guppylang-internals/src/guppylang_internals/diagnostic.py
```diff
@@ -243,31 +244,85 @@ def render_diagnostic(self, diag: Diagnostic) -> None:
         else:
             span = to_span(diag.span)
             level = self.level_str(diag.level)
+
+            children_with_span = [child for child in diag.children if child.span]
             all_spans = [span] + [
-                to_span(child.span) for child in diag.children if child.span
+                to_span(child.span) for child in children_with_span if child.span
```
croyzor: The `if child.span` is now redundant

guppylang-internals/src/guppylang_internals/diagnostic.py
```diff
@@ -243,31 +244,85 @@ def render_diagnostic(self, diag: Diagnostic) -> None:
...
-            for sub_diag in diag.children:
-                if sub_diag.span:
-                    self.render_snippet(
-                        to_span(sub_diag.span),
-                        sub_diag.rendered_span_label,
-                        max_lineno,
-                        is_primary=False,
-                    )
+
+            if len(children_with_span) > 0:
```
croyzor: At most only the 2 lines below need to be in this if block. It's fine to have the `for` loop such that when the list is empty we do nothing

guppylang-internals/src/guppylang_internals/diagnostic.py
```diff
@@ -243,31 +244,85 @@ def render_diagnostic(self, diag: Diagnostic) -> None:
...
+                is_first = True
+                for sub_diag in children_with_span:
+                    if sub_diag.span is None:
+                        continue
+                    span = to_span(sub_diag.span)
+                    span_start_lineno = span.start.line
+                    span_end_lineno = span.end.line
+
+                    # Is it is the first note, render the header
+                    if is_first:
```
croyzor: I'd prefer that this wasn't part of the loop. You could separate out the first element, like
```python
x, *xs = children_with_span
# Print first diagnostic
...
# Print the rest
for diagnostic in xs:
    ...

File: guppylang-internals/src/guppylang_internals/diagnostic.py [hunk 4]
```diff
@@ -243,31 +244,85 @@ def render_diagnostic(self, diag: Diagnostic) -> None:
...
-                        sub_diag.rendered_span_label,
-                        max_lineno,
-                        is_primary=False,
-                    )
+
+            if len(children_with_span) > 0:
+                prev_span_end_lineno = 0
+                is_first = True
+                for sub_diag in children_with_span:
+                    if sub_diag.span is None:
```
croyzor: This check is redundant

guppylang-internals/src/guppylang_internals/diagnostic.py
croyzor: The calls to `render_snippet` can be commoned up as well. They're mostly the same, but with different prefixes and different arguments to `prefix_lines` which could be assigned in the `if ... else ...` block

guppylang-internals/src/guppylang_internals/diagnostic.py
```diff
@@ -315,17 +372,22 @@ def render_snippet(
...
+            if line_number is None or not print_line_number:
+                ll = ""
+            else:
+                ll = str(line_number)
             self.buffer.append(" " * (ll_length - len(ll)) + ll + " | " + line)
 
-        # One line of padding
-        render_line("")
+        # One line of padding (only with first note or primary span)
+        if is_primary or is_first:
```
croyzor: I think this could be moved into the caller

guppylang-internals/src/guppylang_internals/diagnostic.py
nicolaassolini-qntm: I added to ensure  that mypy does not complain

guppylang-internals/src/guppylang_internals/diagnostic.py
nicolaassolini-qntm: I did like that in the first place, I refactored in this way to ensure that mypy was fine with it (there was complain about the `child.span` type

croyzor: If you've already checked some property you can convince `mypy` using asserts, like
```
children_with_span = [child for child in children if children.span]
...
for child in children_with_span:
    assert child.span
    ...
```

guppylang-internals/src/guppylang_internals/diagnostic.py
croyzor: Nah

tests/diagnostics/snapshots/test_three_labels_formatted.txt
```diff
@@ -2,9 +2,10 @@ Error: Can't compare apples with oranges (at <unknown>:1:6)
...
 1 | apple == orange
   |       ^^ Comparison attempted here
+
+Notes:
   | 
 1 | apple == orange
   | ----- This is an apple
-  | 
-1 | apple == orange
+  | apple == orange
```
croyzor: I would say that the old error message is better here - can we retain it?

tests/error/errors_on_usage/else_expr_type_change.err
```diff
@@ -4,11 +4,13 @@ Error: Different types (at $FILE:8:11)
...
   |            ^ Variable `y` may refer to different types
+
+Notes:
   | 
+6 |     y = 3
 7 |     (y := y + 1) if x else (y := True)
   |      - This is of type `int`
-  | 
-7 |     (y := y + 1) if x else (y := True)
+  |     (y := y + 1) if x else (y := True)
```
croyzor: Same comment here about the old error looking better. I guess what I want is:
If there's two notes for the same line, put a divider between them and print the line number

tests/error/errors_on_usage/if_type_change.err
```diff
@@ -4,10 +4,13 @@ Error: Different types (at $FILE:9:8)
...
 9 |     z = y
   |         ^ Variable `y` may refer to different types
+
+Notes:
   | 
+5 | def foo(x: bool) -> int:
 6 |     y = 3
   |     - This is of type `int`
-  | 
+7 |     if x:
```
croyzor, line 13: Nice change!

tests/error/result_errors/state_result_duplicates.err
```diff
@@ -5,8 +5,11 @@ Error: Copy violation (at $FILE:8:28)
 8 |     state_result("tag", q1, q1)
   |                             ^^ Variable `q1` with non-copyable type `qubit` cannot be
   |                                borrowed ...
+
+Note:
   | 
+7 |     q1 = qubit()
 8 |     state_result("tag", q1, q1)
-  |                         -- since it was already borrowed here
+  |                         -- Variable `q1` borrowed here
```
croyzor: Could we have this say "_already_ borrowed"?

tests/diagnostics/snapshots/test_three_labels_formatted.txt
nicolaassolini-qntm: if you are asking to have a line between 9 and 10, yes we can

tests/error/errors_on_usage/else_expr_type_change.err
nicolaassolini-qntm: I'm agree about the empty line, I'm not sure about printing more times the line number, but can be easily done

tests/error/result_errors/state_result_duplicates.err
nicolaassolini-qntm: yes we can!

nicolaassolini-qntm:
@croyzor implementing your request, I noted that when you have error on the same lime (IMO) is better to print the line number twice without empty line in the midle, like: 
```
Notes:
  | 
1 | apple == orange
  | ----- This is an apple
1 | apple == orange
  |          ------ This is an orange
```
if you prefer 
```
Notes:
  | 
1 | apple == orange
  | ----- This is an apple
  |
1 | apple == orange
  |          ------ This is an orange
```
let me know, one line change

nicolaassolini-qntm:
> I'd prefer with the extra
> 
> > @croyzor implementing your request, I noted that when you have error on the same lime (IMO) is better to print the line number twice without empty line in the midle, like:
> > ```
> > Notes:
> >   | 
> > 1 | apple == orange
> >   | ----- This is an apple
> > 1 | apple == orange
> >   |          ------ This is an orange
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > if you prefer
> > ```
> > Notes:
> >   | 
> > 1 | apple == orange
> >   | ----- This is an apple
> >   |
> > 1 | apple == orange
> >   |          ------ This is an orange
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > let me know, one line change
> 
> I'd prefer the second option with the extra line in the middle 🙂

Ok! I procee now! (I also forgot this https://github.com/Quantinuum/guppylang/pull/1382#discussion_r2631450861 doing that too)

guppylang-internals/src/guppylang_internals/checker/errors/linearity.py
```diff
@@ -270,7 +274,7 @@ class NonCopyableCaptureError(Error):
 
     @dataclass(frozen=True)
     class DefinedHere(Note):
-        span_label: ClassVar[str] = "`{var}` defined here"
+        span_label: ClassVar[str] = "`{var}` already defined here"
```
croyzor:
```suggestion
        span_label: ClassVar[str] = "`{var}` defined here"
```

Issue #1383, opened 2025-12-15
Add some integration tests for optimization on Guppy programs


CalMacCQ:
As of `tket==0.12.13` we now have the ability to run pytket optimization passes on Hugr graphs using the `PytketHugrPass` interface. 

Given that the Guppy compiler sometimes outputs things like order edges and MakeTuple nodes, we should have some tests of optimizations that include generation of Hugrs from Guppy source.

PR #1384, opened 2025-12-15
test: start adding integration tests for optimization on Guppy generated hugr programs

CalMacCQ:
I've started adding some basic tests that validate that `tket` optimizations transform guppy programs as expected. This is part of https://github.com/Quantinuum/guppylang/issues/1383

Note that some key use cases are not supported yet including

1. Some sequences of parameterized gates are not simplified as expected. Hopefully once https://github.com/Quantinuum/hugr/pull/2739 is released and used in `tket` this will be fixed.
2. Programs with arrays don't always simplify as expected (see https://github.com/Quantinuum/tket2/issues/1307)
3. Guppy functions which call other functions will only be optimized at the top level currently.

Once these limitations are removed I think it would be good to add test cases for the above as well.


tests/integration/tket_opt/test_opt.py
```diff
@@ -0,0 +1,66 @@
...
+        cx(q0, q1)
+        t(q1)
+        cx(q0, q1)
+
+        normalized_hugr: Hugr = normalize(
+            pauli_zz_rotation.compile_function().modules[0]
+        )
+
+        # Test that the dataflow block is inlined by NormalizeGuppy
+        assert _count_ops(normalized_hugr, "DataflowBlock") == 0
```
CalMacCQ, line 41: Maybe for explicitness I should assert that `_count_ops(hugr, "DataflowBlock") == 1` prior to the normalization pass so its clear to the reader that the test does something.

Likewise with MakeTuple nodes. Prior to normalization there will be 3 in total. One for each quantum gate.

CalMacCQ:
> Hmm these tests all pass the precommit check locally but there seems to be an issue with importing `NormalizeGuppy` in the C.I. run. Will take a look.

fixed, I needed to ensure that a recent enough version of the `tket` package was used in testing. See [3d9e03d](https://github.com/Quantinuum/guppylang/pull/1384/commits/3d9e03d8682e7ff659c4bdbf1b900dc3347dfd86)

tests/integration/tket_opt/test_opt.py
CalMacCQ: Done, see [7cd5753](https://github.com/Quantinuum/guppylang/pull/1384/commits/7cd575398575f5f7d802e50705e24bf3e66393c1)

tests/integration/tket_opt/test_opt.py
qartik, line 1: can the path here be smaller? instead of `tests/integration/tket_optimization/test_optimization.py` , perhaps `tests/integration/tket_opt/test_opt.py`?

CalMacCQ: thats a fair point. I can make it shorter.

PR #1385, opened 2025-12-15
fix(checker): handle imported ParamDef with aliases in expr_checker

kkoci:
### Problem

When a `ParamDef` like `nat_var` was imported with an alias (e.g., `from module import nat_var as n`), the `_check_global` method in the expression checker didn't know how to handle it, causing an error even though the parameter was valid.

### Solution

Extended `_check_global` to recognize and properly handle `ParamDef` instances:

**File:** `guppylang-internals/src/guppylang_internals/checker/expr_checker.py`

1.  **Line 88** - Added import:
    
    ```python
    from guppylang_internals.definition.parameter import ParamDef
    
    ```
    
2.  **Lines 458-480** - Added new case to handle `ParamDef`:
    
    ```python
    case ParamDef():
        # Check if this parameter is in our generic_params (e.g., used in type signature)
        if name in self.ctx.generic_params:
            param = self.ctx.generic_params[name]
            match param:
                case ConstParam() as param:
                    ast_node = with_loc(
                        node, GenericParamValue(id=name, param=param)
                    )
                    return ast_node, param.ty
                case TypeParam() as param:
                    raise GuppyError(
                        ExpectedError(
                            node, "a value", got=f"type `{param.name}`"
                        )
                    )
                case _:
                    return assert_never(param)
        # If not in generic_params, it's being used outside its scope
        raise GuppyError(
            ExpectedError(node, "a value", got=f"{defn.description} `{name}`")
        )
    
    ```
    

### How It Works

-   When a `ParamDef` is encountered, the checker now verifies if it's present in the function's `generic_params` (populated during type parsing)
-   If found, it's handled identically to locally-defined parameters
-   If not found, it raises an appropriate error for out-of-scope usage

### Testing

Run the existing test suite:

```bash
pytest tests/integration/test_imports.py

```

Consider adding a specific test case for `nat_var` aliases to ensure this behavior is preserved in future releases.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -408,9 +408,7 @@ class ExprSynthesizer(AstVisitor[tuple[ast.expr, Type]]):
             raise GuppyError(IllegalConstant(node, type(node.value)))
         return node, ty
 
-    def _check_generic_param(
-        self, name: str, node: ast.expr
-    ) -> tuple[ast.expr, Type]:
+    def _check_generic_param(self, name: str, node: ast.expr) -> tuple[ast.expr, Type]:
         """Helper method to check a generic parameter (ConstParam or TypeParam)."""
         param = self.ctx.generic_params[name]
         match param:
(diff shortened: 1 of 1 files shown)


kkoci:
> Hi, thanks for your contribution! This looks like the right fix to me 👍
> 
> Could you add a test (maybe to `tests/integration/test_imports.py`) so we avoid regression in the future?
> 
> Also, the code you added to `_check_global` looks quite similar to what we already have in `visit_Name`, so it would be nice to refactor this into a function to avoid duplication

Thanks for your review! Yes, absolutely, give me sometime today and I will update my PR :)

.gitignore
```diff
@@ -156,3 +156,4 @@ devenv.local.nix
 /.envrc
 
 **/.benchmarks/
+
```
mark-koch: Also, could you revert this change as it's unrelated to the PR?

kkoci: Done @mark-koch

kkoci:
> Thanks! Happy to merge once CI is happy. Should just be a matter of running `ruff format` and `ruff check`

Applied ruff linting, CI/CD should be okay with it now, hopefully :) @mark-koch

PR #1399, opened 2025-12-17
fix: Fix another wasm diagnostics rendering issue

mark-koch:
Missed one previously in #1398 

As a side note, I think it would be better if those errors were only triggered when we call `check` instead of when the decorator is called, but that's a separate issue


mark-koch:
Those don't raise `GuppyError`s

PR #1403, opened 2025-12-18
Refactoring of #1382

croyzor:
Probably best looked at commit by commit

### DIFF
--- a/guppylang-internals/src/guppylang_internals/diagnostic.py
+++ b/guppylang-internals/src/guppylang_internals/diagnostic.py
@@ -248,7 +248,8 @@ class DiagnosticsRenderer:
             children_with_span = [
                 (child, to_span(child.span)) for child in diag.children if child.span
             ]
-            max_lineno = span.end.line
+            all_spans = [span] + [span for _, span in children_with_span]
+            max_lineno = max(s.end.line for s in all_spans)
 
             self.buffer.append(f"{level}: {diag.rendered_title} (at {span.start})")
 
@@ -265,7 +266,6 @@ class DiagnosticsRenderer:
                 case []:
                     pass
                 case [(only_child, span)]:
-                    max_lineno = max(span.end.line, max_lineno)
                     self.buffer.append("\nNote:")
                     self.render_snippet(
                         span,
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/diagnostic.py
```diff
@@ -245,10 +245,10 @@ def render_diagnostic(self, diag: Diagnostic) -> None:
...
             level = self.level_str(diag.level)
 
-            children_with_span = [child for child in diag.children if child.span]
-            all_spans = [span] + [
-                to_span(child.span) for child in children_with_span if child.span
+            children_with_span = [
+                (child, to_span(child.span)) for child in diag.children if child.span
             ]
+            all_spans = [span] + [span for _, span in children_with_span]
             max_lineno = max(s.end.line for s in all_spans)
```
nicolaassolini-qntm: we can avoid a loop

```suggestion
            max_lineno = span.end.line
            children_with_span = []

            for child in diag.children:
                if child.span:
                    span = to_span(child.span)
                    children_with_span.append(child, span)
                    max_lineno = max(span.end.line, max_lineno)
```

guppylang-internals/src/guppylang_internals/diagnostic.py
```diff
@@ -262,60 +262,59 @@ def render_diagnostic(self, diag: Diagnostic) -> None:
...
+                        is_first=True,
+                    )
+
+                    prev_span_end_lineno = first_span.end.line
+
+                    for sub_diag, span in children_with_span:
+                        span_start_lineno = span.start.line
+                        span_end_lineno = span.end.line
+
+                        prefix_lines = 0
```
nicolaassolini-qntm: if we move  `prefix_lines` inside the `if` the code is easier to read

guppylang-internals/src/guppylang_internals/diagnostic.py
croyzor: True, I just prefer separating out the first subdiagnostic the `match`. We shouldn't have so many subdiagnostics that we need to worry about extra loops vs readability.

Incorporating the `max_lineno` calculation into the loop and getting rid of `all_spans` seems like a good change to me though

PR #1410, opened 2025-12-23
fix: Stop parsing entrypoints twice

mark-koch:
When calling `check()` on a definition, we accidentally invoked the parsing logic twice

### DIFF
--- a/guppylang-internals/src/guppylang_internals/engine.py
+++ b/guppylang-internals/src/guppylang_internals/engine.py
@@ -220,21 +220,12 @@ class CompilationEngine:
 
         This is the main driver behind `guppy.check()`.
         """
-        from guppylang_internals.checker.core import Globals
-
         # Clear previous compilation cache.
         # TODO: In order to maintain results from the previous `check` call we would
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -220,21 +220,12 @@ def check(self, id: DefId) -> None:
...
-
         # Clear previous compilation cache.
         # TODO: In order to maintain results from the previous `check` call we would
         #  need to store and check if any dependencies have changed.
         self.reset()
 
-        defn = DEF_STORE.raw_defs[id]
-        self.to_check_worklist = {
-            defn.id: (
-                defn.parse(Globals(DEF_STORE.frames[defn.id]), DEF_STORE.sources)
```
mark-koch, line 233: The problem was that we called `parse` here but didn't save the result in `self.parsed`. Using `self.get_parsed` is more concise and correctly handles the caching

nicolaassolini-qntm (approved):

PR #1411, opened 2025-12-24
test: Add test that made hugr-0.25.0 fail

acl-cqc:
This odd `guppy.comptime` + `result` combo seems to invoke the BorrowArray *copying* handler in a way that I haven't managed to otherwise.

The two calls with different-length arrays caused linking problems following https:://github.com/Quantinuum/hugr/pull/2749, fixed in https:://github.com/Quantinuum/hugr/pull/2779.

Note, all testing was based on guppylang 0.21.6, I realize there has been a 0.21.7 since but the testing was quite a faff so I will not repeat it unless you think there is reason guppy compilation may have changed....

### DIFF
--- a/tests/integration/test_constants.py
+++ b/tests/integration/test_constants.py
@@ -1,19 +1,7 @@
 """Constants are only loaded by execution compilers (not validation), test that."""
 
 from guppylang.decorator import guppy
-from guppylang.std.builtins import result, comptime, array
-
-
-# This caused problems with hugr (see https://github.com/Quantinuum/hugr/pull/2779),
-# as it makes the linearizer copy an array (indeed, two arrays of different lengths).
-def test_array_consts(run_int_fn):
-    @guppy
-    def main() -> int:
-        result("x", array(True, False))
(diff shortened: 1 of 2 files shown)


tests/integration/test_constants.py
```diff
@@ -1,7 +1,19 @@
 """Constants are only loaded by execution compilers (not validation), test that."""
 
 from guppylang.decorator import guppy
-from guppylang.std.builtins import result, comptime
+from guppylang.std.builtins import result, comptime, array
+
+
+# This caused problems with hugr (see https://github.com/Quantinuum/hugr/pull/2779),
+# as it makes the linearizer copy an array (indeed, two arrays of different lengths).
+def test_array_consts(run_int_fn):
```
mark-koch: Could you move this to `test_result.py`? The issue is the `result`, not the constant array. The bug should be occurring for all bool array results with differing lengths

See https://github.com/Quantinuum/guppylang/pull/1411#issuecomment-3695925714

mark-koch (approved):

PR #1417, opened 2026-01-05
test: add `tket~=0.12.14` tests for 1q squashing and programs with arrays

CalMacCQ:
Adds some new integration tests after the fixes made in `tket` v0.12.14. Bumped version to `~=0.12.15`.

### DIFF
--- a/tests/integration/tket_opt/test_opt.py
+++ b/tests/integration/tket_opt/test_opt.py
@@ -1,6 +1,7 @@
 from guppylang import guppy
-from guppylang.std.angles import pi
+from guppylang.std.angles import pi, angle
 from guppylang.std.quantum import qubit, cx, h, s, t, rz
+from guppylang.std.qsystem import phased_x
 from guppylang.std.builtins import array
 
 from pytket.passes import RemoveRedundancies, CliffordSimp, SquashRzPhasedX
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -16,7 +16,7 @@ test = [
     "pytest-xdist>=3.8.0",
     "ipykernel >=6.29.5,<7",
     "miette-py",
-    "tket >= 0.12.13",
+    "tket",
```
aborgna-q:
```suggestion
    "tket ~= 0.12.14",
```

You'll also need to update the hugr dep in `guppy-internals` to `0.15.0`

CalMacCQ: Yeah just noticed that at the same time :)

ss2165: > You'll also need to update the hugr dep in guppy-internals to 0.15.0

this is already the case on main

CalMacCQ: Yeah I picked up this change by merging in main to this branch.

tests/integration/tket_opt/test_opt.py
```diff
@@ -70,3 +85,16 @@ def simple_clifford(q0: qubit, q1: qubit) -> None:
         opt_hugr = cliff_pass(my_hugr_graph)
         # test that we can cancel a CX gate by using an implicit swap
         assert _count_ops(opt_hugr, "CX") == 1
+
+
+def test_1q_rz_squashing() -> None:
+    @guppy
+    def redundant_1q_gates(q0: qubit) -> None:
```
CalMacCQ, line 93: Maybe I should add more test cases for squashing. What do you think?

CalMacCQ: added another which includes PhasedX [acc293e](https://github.com/Quantinuum/guppylang/pull/1417/commits/acc293e6256b39a1ac88bcc4094d1c44a1a6f323)

PR #1426, opened 2026-01-08
fix: added deepcopy in `OverloadedFunctionDef.{check_call,synthesize_call}`

### LINKED ISSUE #1405: Regression in generator args to `result`
Seems to be due to [this commit](https://github.com/Quantinuum/guppylang/commit/1da2c5dbb82bf6da35949b505a69f4e2f51acd3b#diff-8d460bf94b2b277a415b377d25ad7bb884ce60477b9d47dd7460bd7320072c93).

Minimal example that fails:
```python
from guppylang import guppy

@guppy
def main() -> None:
    result("fs", array(i * 0.2 for i in range(1)))

main.compile()

```
gives
```
guppylang_internals.error.InternalGuppyError: Invalid iterator assign statement
```

Interestingly, the result is different if the generator element type is `int` instead of `float`

nicolaassolini-qntm:
This PR closes #1405

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/overloaded.py
+++ b/guppylang-internals/src/guppylang_internals/definition/overloaded.py
@@ -89,7 +89,8 @@ class OverloadedFunctionDef(CompiledCallableDef, CallableDef):
             with suppress(GuppyError):
                 # check_call may modify args and node,
                 # thus we deepcopy them before passing in the function
-                node_copy, args_copy = safe_deepcopy(node, args)
+                node_copy = copy.deepcopy(node)
+                args_copy = copy.deepcopy(args)
                 return defn.check_call(args_copy, ty, node_copy, ctx)
         return self._call_error(args, node, ctx, available_sigs, ty)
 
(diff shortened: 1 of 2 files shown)


guppylang-internals/src/guppylang_internals/definition/overloaded.py
```diff
@@ -148,3 +155,25 @@ def load_with_args(
         raise InternalGuppyError(
             "OverloadedFunctionDef.load_with_args shouldn't be invoked"
         )
+
+
+def safe_deepcopy(
+    node: AstNode, args: list[ast.expr]
+) -> tuple[AstNode, list[ast.expr]]:
+    """Attempts to deepcopy an AST node and a list of AST expressions.
```
nicolaassolini-qntm: I'm not 100% happy with this implementation, but seems to work

mark-koch: Mhmm I don't think this is acceptable, what errors were you running into with the regular deepcopy?

nicolaassolini-qntm: when running some tests, like `test_capture` for instance. It involves this function call: [/tests/integration/conftest.py#L102](https://github.com/Quantinuum/guppylang/blob/4724d9039d8dffae8fd939f62ae80ec307d8918a/tests/integration/conftest.py#L102)
and the emulator I guess. (Someone try to duplicate a Guppyobject)

mark-koch: Ah yes, `PlaceNode` can contain a `GuppyObject` if `place` is a `ComptimeVariable`.

Can you try a manual `__deepcopy__` implementation for `GuppyObject`?

Issue #1213, opened 2025-08-18
Add API docs for std modules


CalMacCQ:
Top level `guppylang.std` docs (see `futures`, `collections`, `option`, `qsystem`) 

<img width="774" height="724" alt="Image" src="https://github.com/user-attachments/assets/71a1184b-a363-420f-9156-796170cf9d93" />

`guppylang.std.qsystem` docs

<img width="621" height="164" alt="Image" src="https://github.com/user-attachments/assets/f6d79d05-828f-45d4-969c-964d17869b2a" />

rmfisher26:
Hi @CalMacCQ , I would like to be assigned this issue to improve this documentation.

CalMacCQ:
> Hi [@CalMacCQ](https://github.com/CalMacCQ) , I would like to be assigned this issue to improve this documentation.

You're more than welcome to take this :) thank you @rmfisher26

PR #1428, opened 2026-01-08
docs: Added documentation for Issue #1213 Add API docs for std modules

rmfisher26:
Regarding Issue  #1213 , I have added basic documentation  to the top level guppylang.std docs (see futures, collections, option, qsystem)

If wording should be modified for clearer undertanding, please let me know!

### DIFF
--- a/guppylang/src/guppylang/std/qsystem/utils.py
+++ b/guppylang/src/guppylang/std/qsystem/utils.py
@@ -1,4 +1,5 @@
-"""Guppy standard library for utilities available on Quantinuum Systems."""
+"""Guppy standard library for utilities available on Quantinuum
+Systems."""
 
 from typing import no_type_check
 
--- a/guppylang/src/guppylang/std/qsystem/random.py
+++ b/guppylang/src/guppylang/std/qsystem/random.py
@@ -1,5 +1,5 @@
 """Guppy standard library for random number generation and
- discrete probability distributions."""
+discrete probability distributions."""
 
 # mypy: disable-error-code="no-any-return"
 from typing import Generic, no_type_check


guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -1,3 +1,5 @@
+"""Guppy standard module for quantum system operations."""
```
CalMacCQ: I'd suggest something like "Quantinuum systems device operations" rather than "quantum system operations" here. The qsystem module contains operations which are somehow native to the Quantinuum systems devices.

CalMacCQ (changes requested):
Thanks for the contribution :) Just one suggestion

guppylang/src/guppylang/std/qsystem/__init__.py
rmfisher26: Thanks @CalMacCQ  for your input! I have updated the description based on your suggestion, so it is ready for review once more.

guppylang/src/guppylang/std/qsystem/utils.py
```diff
@@ -1,3 +1,5 @@
+"""Guppy standard library for quantum shot utilities."""
```
ss2165:
```suggestion
"""Guppy standard library for utilities available on Quantinuum Systems."""
```

rmfisher26: @ss2165, Thank you for your feedback. I think the file should now be updated to reflect your text suggestion.  If something still doesn't look right, let me know.

Issue #1363, opened 2025-12-03
Comptime expressions fail in subscript assignment


mark-koch:
```python
n = 5

@guppy
def main(xs: array[int, 10]) -> None:
    xs[comptime(n)] = 0

main.check()
```

fails with

```
Error: Variable not defined (at file.py:204:11)
    | 
202 |     @guppy
203 |     def main(xs: array[int, 10]) -> None:
204 |         xs[comptime(n)] = 0
    |            ^^^^^^^^ `comptime` is not defined

Guppy compilation failed due to 1 previous error
```

Looks like the desugarer during CFG construction is not looking into subscripts in assignment position?

PR #1436, opened 2026-01-13
fix: Support more cases of comptime subscripts

### LINKED ISSUE #1435: Comptime expressions still fail in augmenting subscript assignments
The fix in #1433 only updated the `visit_Assign` case, but not the `visit_AugAssign` one. Thus,

```python
n = 5

@guppy
def main(xs: array[int, 10]) -> None:
    xs[comptime(n)] += 0

main.check()
```

still fails with

```
Error: Variable not defined (at test.py:159:7)
    | 
157 | @guppy
158 | def main(xs: array[int, 10]) -> None:
159 |     xs[comptime(n)] += 0
    |        ^^^^^^^^ `comptime` is not defined

```

Same for annotated assignments:

```python
n = 5

@guppy
def main(xs: array[int, 10]) -> None:
    xs[comptime(n)]: int = 0

main.check()
```

maximilianruesch:
The fix of #1433 (for #1363) was insufficient in so far that it only covered the case of a regular assignment. This PR extends the solution to other types of assignments (those that augment values, and those that are annotated (both at the same time are not allowed in Python, so both cases individually are sufficient). The code has been reworked to more snuggly fit expectations from function names and increase reusability.

~Also added a test to cover multi-target assignments.~ Oop, didnt push that commit.

Closes #1435

### DIFF
--- a/guppylang-internals/src/guppylang_internals/cfg/builder.py
+++ b/guppylang-internals/src/guppylang_internals/cfg/builder.py
@@ -154,31 +154,52 @@ class CFGBuilder(AstVisitor[BB | None]):
         return bb_opt
 
     def _build_node_value(self, node: BBStatement, bb: BB) -> BB:
-        """Utility method for building a node containing a `value` expression.
+        """Utility method for building a nodes `value` expression, if available.
 
         Builds the expression and mutates `node.value` to point to the built expression.
-        Returns the BB in which the expression is available and adds the node to it.
+        Returns the BB in which the expression is available.
         """
         if (
             not isinstance(node, NestedFunctionDef | ModifiedBlock)
             and node.value is not None
         ):
             node.value, bb = ExprBuilder.build(node.value, self.cfg, bb)
-        bb.statements.append(node)
+        return bb
+
+    def _build_node_targets(self, node: BBStatement, bb: BB) -> BB:
+        """Utility method for building a nodes `target` or `targets` expressions,
+        depending on the node type.
+
+        Builds the expressions and mutates the elements of `node.targets` to point to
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -154,31 +154,52 @@ def visit_stmts(self, nodes: list[ast.stmt], bb: BB, jumps: Jumps) -> BB | None:
...
+        Builds the expressions and mutates the elements of `node.targets` to point to
+        the built expressions. Returns the BB in which the expressions are available.
+        """
+        if isinstance(node, ast.Assign):
+            for i, target in enumerate(node.targets):
+                node.targets[i], bb = ExprBuilder.build(target, self.cfg, bb)
+        elif isinstance(node, ast.AugAssign | ast.AnnAssign):
+            new_target, bb = ExprBuilder.build(node.target, self.cfg, bb)
+            if not isinstance(new_target, ast.Name | ast.Attribute | ast.Subscript):
+                raise InternalGuppyError("Unexpected type for built expression.")
```
croyzor, line 182: I don't think this is an _internal_ error, the user could cause it by writing an unsupported AST

maximilianruesch: The type hint for `node.target` reports these exact types, and I expect the same to result from the ExpressionBuilder. In what case could a user write an unsupported AST to make the builder not return one of those types?

croyzor: ah fair enough, you're checking after running `build` :+1:

PR #1439, opened 2026-01-13
fix: Raise type inference error when not all function argument types could be resolved.

### LINKED ISSUE #1437: Unsolved generic type for empty arrays
```python
@guppy
def foo[T, n: nat](xs: array[T, n] @ owned) -> array[T, n]:
    return xs

@guppy
def main() -> None:
    xs: array[float, 0] = foo(array())

main.check()
```

Fails with an assertion error

maximilianruesch:
Changing this from an assertion to a type inference error causes return type information to be incorporated on the next try of type inference for the function in question, which might provide the missing type hints to parameters in case the arguments cannot supply them.

Closes #1437

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -1008,8 +1008,8 @@ def type_check_args(
         new_args.append(a)
     assert next(comptime_args, None) is None
 
-    # If the argument check succeeded, this means that we must have found instantiations
-    # for all unification variables occurring in the input types
+    # Check whether we have found instantiations for all unification variables occurring
+    # in the input types
     for inp in func_ty.inputs:
         if not set.issubset(inp.ty.unsolved_vars, subst.keys()):
             raise GuppyTypeInferenceError(


tests/integration/test_poly.py
```diff
@@ -255,6 +255,20 @@ def main() -> None:
...
+def test_type_apply_empty_array(validate):
+    T = guppy.type_var("T")
+    n = guppy.nat_var("n")
+
+    @guppy.declare
+    def foo(xs: array[T, n]) -> array[T, n]: ...
+
+    @guppy
+    def main() -> None:
+        xs: array[float, 0] = foo(array())
```
mark-koch, line 267: Could you also add a corresponding error test where the annotation is missing?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1010,9 +1010,11 @@ def type_check_args(
 
     # If the argument check succeeded, this means that we must have found instantiations
     # for all unification variables occurring in the input types
```
mark-koch: Could you also update this comment (i.e. we need to check that this is the case rather than asserting that it is)

PR #1440, opened 2026-01-14
feat!: Make tket a non-optional dependency

### LINKED ISSUE #1011: ci: Tests that look at behaviour when tket is not installed are not run in CI

maximilianruesch:
In preparation for the planned upcoming `tket` dependency introduction, to resolve errors / misconfigurations on the CI around testing with / without `tket`, and to simplify the code around `tket` optionality, `tket` has been made a non-optional dependency of `guppylang-internals` (and thus transitively of `guppylang`).

> [!IMPORTANT]
>  The `pytket` extras for `guppylang` and `guppylang-internals` have been cleared, but remain for backwards compatibility / easier transition. The command to include the extras can still be executed, but has no effect.

As a side effect, this trivially resolves recorded issues around tests not running without tket.

Closes #1011

BREAKING CHANGE: Tket is now a non-optional dependency. Most systems already have this installed, rerunning the package manager should fix any issues.

BEGIN_COMMIT_OVERRIDE
feat: Make tket a non-optional dependency (#1440)
END_COMMIT_OVERRIDE

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
+++ b/guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
@@ -3,12 +3,12 @@ from dataclasses import dataclass, field
 from typing import Any, cast
 
 import hugr.build.function as hf
-import pytket
 from hugr import Node, Wire, envelope, ops, val
 from hugr import tys as ht
 from hugr.build.dfg import DefinitionBuilder, OpVar
 from hugr.envelope import EnvelopeConfig
 from hugr.std.float import FLOAT_T
+from pytket.circuit import Circuit
 
 from guppylang.defs import GuppyDefinition
 from guppylang_internals.ast_util import AstNode, has_empty_body, with_loc
@@ -368,15 +368,12 @@ class CompiledPytketDef(ParsedPytketDef, CompiledCallableDef, CompiledHugrNodeDe
 
 
 def _signature_from_circuit(
-    input_circuit: Any,
+    input_circuit: Circuit,
     defined_at: ToSpan | None,
     use_arrays: bool = False,
 ) -> FunctionType:
     """Helper function for inferring a function signature from a pytket circuit."""
     # May want to set proper unitary flags in the future.
-    if not isinstance(input_circuit, pytket.circuit.Circuit):
-        pass
-
     from guppylang.std.angles import angle  # Avoid circular imports
     from guppylang.std.quantum import qubit
 


ss2165:
make the pr title "feat", "chore" won't show up on changelog

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
```diff
@@ -375,63 +374,46 @@ def _signature_from_circuit(
...
-                err.add_sub_diagnostic(TketNotInstalled.InstallInstruction(None))
-                raise GuppyError(err) from None
-        else:
-            pass
-    except ImportError:
-        raise InternalGuppyError(
-            "Pytket error should have been caught earlier"
-        ) from None
+    if not isinstance(input_circuit, pytket.circuit.Circuit):
+        pass
```
ss2165: what is this block for, it is not doing anything?

guppylang-internals/pyproject.toml
```diff
@@ -37,10 +34,12 @@ dependencies = [
     "tket-exts ~= 0.12.0",
     "hugr ~= 0.15.0",
     "wasmtime ~= v38.0.0",
+    "pytket>=1.34",
```
ss2165, line 37: `tket` has a pytket dependency, we should rely on the transitive rather than constrain here
```suggestion
```

justfile
```diff
@@ -2,26 +2,22 @@ set shell := ["bash", "-uc"]
 
 # List the available commands
 help:
-    @just --list --justfile {{justfile()}}
+    @just --list --justfile {{ justfile() }}
```
ss2165: nit: try to avoid including drive-by linting fixes in unrelated PRs. They are better in standalone PRs to preserve useful git blame.

In this file it is hard to find relevant changes.

guppylang-internals/pyproject.toml
maximilianruesch: Since `pytket` is explicitly used, I would argue that it should also be listed explicitly in the dependencies. The `pytket` extra previously listed it with this particular constraint.

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
maximilianruesch: A small oversight. I noticed I can now simply annotate the `Circuit` to the parameter, and this check is not needed anymore.

PR #1444, opened 2026-01-15
fix: upgrade selene-sim version to fix state ordering bug

ss2165:
Pulls in selene-sim 0.2.7, containing a bug fix for qubit ordering in state results


tests/integration/test_emulator.py
```diff
@@ -385,3 +387,38 @@ def main() -> None:
 
     res = _build_run(main, n_qubits=1).results[0].entries
     assert res == [("a", True), ("b", False)]
+
+
+def test_state_qubit_order():
+    def get_statevector(main: GuppyFunctionDefinition[[], None], n_qubits: int):
```
maximilianruesch: Does this get a return type hint? Do we try to have those in tests

maximilianruesch: A bit confusing that the passed guppy function receives `[[], None]`, even though it takes an argument. This is probably a separate thing of the compiler though, similar to how it annotates this "fallback" type on overloaded functions.

tests/integration/test_emulator.py
```diff
@@ -385,3 +387,38 @@ def main() -> None:
...
+    @guppy
+    def swap(state_qreg: array[qubit, 2]) -> None:
+        x(state_qreg[1])
+        mem_swap(state_qreg[0], state_qreg[1])
+
+    state_noswap = get_statevector(noswap, 2)
+    state_swap = get_statevector(swap, 2)
+
+    assert pytest.approx(state_noswap[1]) == 1 + 0j
+    assert pytest.approx(state_swap[2]) == 1 + 0j
```
maximilianruesch: Can you make assertions about the whole array, not just about specific elements?

tests/integration/test_emulator.py
ss2165: got rid of the generic parameters, not necessary

ss2165: mypy is not run on tests at the moment but we try to have them anyway

PR #1445, opened 2026-01-15
test: Add a regex to remove the xdist-execnet-python3.13 bootstrap combo

### LINKED ISSUE #1130: Snapshot error when parallelizing tests in Python >=3.13

maximilianruesch:
Adds a regex to the test runner that strips the bootstrap string from the string checked against the golden file, and refactors the extraction code slightly to make it more readable.

See https://github.com/Quantinuum/guppylang/issues/1130#issuecomment-3755257371 for further explanation on why this change is necessary. Short story:
> Boom

Additionally, reenables parallelism on the coverage CI which was disabled because it was running on Python 3.14. This takes the CI from ~3m 29s to ~2m 53s (see [old](https://github.com/Quantinuum/guppylang/actions/runs/21037212016/job/60488399495?pr=1445) vs [new](https://github.com/Quantinuum/guppylang/actions/runs/21037907363/job/60491000251?pr=1445)). Half a minute less to pay for.

Closes #1130

### DIFF
--- a/tests/error/util.py
+++ b/tests/error/util.py
@@ -4,14 +4,11 @@ import pathlib
 import re
 import sys
 
-from guppylang_internals.decorator import custom_type
 import pytest
-import sys
 from hugr import tys
 from hugr.tys import TypeBound
 
-import guppylang.decorator as decorator
-
+from guppylang_internals.decorator import custom_type
 from tests.util import get_wasm_file
 
 # Regular expression to match the `~~~~~^^^~~~` highlights that are printed in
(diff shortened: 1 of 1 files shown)


ss2165 (approved):
nice find!

PR #1447, opened 2026-01-16
fix: Remove leading tab characters from diagnostics during rendering

### LINKED ISSUE #1101: Caret offset in compile error message
The following code produces an error that has the carets offset by some spaces to the left:

```py
from guppylang.std.quantum import *

@guppy
def main() -> None:
	qreg = qubit()
	measure(qreg)
	discard(qreg)
	return

main.compile()
```

gives error:

```bash
Error: Copy violation (at <In[7]>:9:9)
  | 
7 | 	qreg = qubit()
8 | 	measure(qreg)
9 | 	discard(qreg)
  |          ^^^^ Variable `qreg` with non-copyable type `qubit` cannot be
  |               consumed ...
Note:
  | 
7 | 	qreg = qubit()
8 | 	measure(qreg)
  |          ---- Variable `qreg` already consumed here
```

where the caret under qreg0 is offset

maximilianruesch:
Removes leading tab characters from diagnostics by converting them to 4 tab characters as recommended by PEP8 (stating that `Spaces are the preferred indentation method` and in particular `Python disallows mixing tabs and spaces for indentation`).

This results in actually rendered diagnostics line up with their highlightings again since the columns are explicit with spaces, rather than implicit with tabs.

Closes #1101

### DIFF
--- a/tests/diagnostics/snapshots/test_leading_tab_characters_multi_line_span.txt
+++ b/tests/diagnostics/snapshots/test_leading_tab_characters_multi_line_span.txt
@@ -1,0 +1,7 @@
+Error: Can't compare apples with oranges (at <unknown>:1:1)
+  | 
+1 |     super_apple := apple ** 2
+  |     ^^^^^^^^^^^^^^^^^^^^^^^^^
+  | ...
+3 |     apple == orange
+  | ^^^^^^^^^^^^^^^^^^^ Comparison attempted here
--- a/tests/diagnostics/snapshots/test_leading_tab_characters_multi_line_span_varying_tabs.txt
+++ b/tests/diagnostics/snapshots/test_leading_tab_characters_multi_line_span_varying_tabs.txt
@@ -1,0 +1,7 @@
+Error: Can't compare apples with oranges (at <unknown>:1:2)
+  | 
+1 |         super_apple := apple ** 2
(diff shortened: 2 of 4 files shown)


guppylang-internals/src/guppylang_internals/diagnostic.py
```diff
@@ -384,9 +384,25 @@ def render_line(line: str, line_number: int | None = None) -> None:
...
+        # Convert leading tab characters into four whitespaces each (see PEP 8)
+        for i, line in enumerate(all_lines):
+            line_no_tabs = line.lstrip("\t")
+            num_tabs = len(line) - len(line_no_tabs)
+            all_lines[i] = " " * (num_tabs * 4) + line_no_tabs
+            # Shift span locations, accounting for incorporated \t
+            new_start = span.start
+            new_end = span.end
+            if i == prefix_lines:  # Line is the first line in the span
+                new_start = span.start.shift_right(num_tabs * 3)
```
ss2165, line 400: don't think I follow why this is `* 3` rather than `* 4`

maximilianruesch: The `\t` is already / still incorporated in the span column indices, even though it was removed from the line. Thus, for each tab character, the column is already shifted 1/4 right, so we just need to shift it the remaining 3/4.

Issue #1438, opened 2026-01-13
Compile time array bounds check for literal indices


ss2165:
In general static bounds checking of arrays is not possible since indices can be runtime variables.

However in many common cases indices are literals, or can be cheaply traced back to a statically known value. In this case we should aim to throw an error. For instance

```python

qs = array(qubit() for _ in range(3))
h(qs[3])

```

should throw an error like "index 3 is invalid for array of size 3"

mark-koch:
A good place to implement this would be either:

* In the `ArrayGetitemCompiler` and `ArraySetitemCompiler` in `guppylang_internals.std._internal.compiler.array`. However, this means the error is only caught when lowering to Hugr (i.e. not when calling `check`) which is not ideal

* Alternatively, one could add a new `CustomCallChecker` that handles the indexing check. This would be more effort, but cleaner/more reusable

ss2165:
Prefer option 2

rmfisher26:
@ss2165 @mark-koch , please refer to Pull Request  #1448 regarding the implementation and testing results.

PR #1448, opened 2026-01-16
feat: Compile time array bounds check for literal indices

rmfisher26:
@mark-koch , @ss2165 

Regarding issue #1438 , the proposed pull request implements the ArrayIndexChecker(CustomCallChecker) to determine if a index is out of bounds. 

This contains a new set of tests specifically for checking the index in various positive and negative ways. If an error is caught and IndexOutOfBoundsError exception is thrown with more details.  

I have also run the full pytest tests was tested with 2 tests failing with this exception: 

Here is the full error message I go while running pytest:

```
guppylang-internals/src/guppylang_internals/std/_internal/checker.py:283: GuppyError
====================================== short test summary info ======================================
FAILED tests/integration/test_quantum.py::test_barrier_array - guppylang_internals.error.GuppyError: ArrayIndexChecker.IndexOutOfBoundsError(span=PlaceNode(pla...
FAILED tests/integration/test_state_result.py::test_array_access - guppylang_internals.error.GuppyError: ArrayIndexChecker.IndexOutOfBoundsError(span=PlaceNode(pla...
================= 2 failed, 1202 passed, 23 skipped, 4 xfailed in 80.99s (0:01:20) ==================
```

**1. tests/integration/test_quantum.py::test_barrier_array**

```
def test_barrier_array(validate):
    """Barrier on array/struct access."""

    @guppy
    @no_type_check
    def test() -> None:
        qs = array(qubit() for _ in range(4))
        q.h(qs[1])
        q.h(qs[2])
        barrier(qs[1], qs[2], qs[3])
        barrier(qs[1])
        q.h(qs[3])

        q.cx(qs[1], qs[2])
        barrier(qs[2], qs[3])
        q.cx(qs[3], qs[4]) #IndexOutOfBoundsError here...
        barrier(qs)
        discard_array(qs)

    validate(test.compile_function())
```
Error:
`guppylang_internals.error.GuppyError: ArrayIndexChecker.IndexOutOfBoundsError(span=PlaceNode(place=Variable(name='qs', ty=OpaqueType(args=[TypeArg(ty=OpaqueType(args=[], defn=OpaqueTypeDef(id=DefId(id=263), name='qubit', defined_at=None, description='type', params=[], never_copyable=True, never_droppable=True, to_hugr=<function custom_type.<locals>.<lambda> at 0x705621339d20>, bound=None))), ConstArg(const=ConstValue(ty=NumericType(kind=<Kind.Nat: 1>), value=4))], defn=OpaqueTypeDef(id=DefId(id=10), name='array', defined_at=None, description='type', params=[TypeParam(idx=0, name='T', must_be_copyable=False, must_be_droppable=False), ConstParam(idx=1, name='n', ty=NumericType(kind=<Kind.Nat: 1>), from_comptime_arg=False)], never_copyable=True, never_droppable=False, to_hugr=<function _array_to_hugr at 0x705660db24b0>, bound=None)), defined_at=Name(id='qs', ctx=Store()), flags=<InputFlags.NoFlags: 0>, is_func_input=False)), _parent=None, children=[], index=4, size=4)`

**2. tests/integration/test_state_result.py::test_array_access**

```
def test_array_access(validate):
    @compile_guppy
    def main() -> None:
        qs = array(qubit() for _ in range(4))
        h(qs[1])
        h(qs[2])
        state_result("a", qs[1], qs[2], qs[3])
        state_result("b", qs[1])
        h(qs[3])

        cx(qs[1], qs[2])
        state_result("c", qs[2], qs[3])
        cx(qs[3], qs[4]) #IndexOutOfBoundsError here...
        discard_array(qs)

    validate(main)
```
Error
`guppylang_internals.error.GuppyError: ArrayIndexChecker.IndexOutOfBoundsError(span=PlaceNode(place=Variable(name='qs', ty=OpaqueType(args=[TypeArg(ty=OpaqueType(args=[], defn=OpaqueTypeDef(id=DefId(id=263), name='qubit', defined_at=None, description='type', params=[], never_copyable=True, never_droppable=True, to_hugr=<function custom_type.<locals>.<lambda> at 0x7dd26ba35d20>, bound=None))), ConstArg(const=ConstValue(ty=NumericType(kind=<Kind.Nat: 1>), value=4))], defn=OpaqueTypeDef(id=DefId(id=10), name='array', defined_at=None, description='type', params=[TypeParam(idx=0, name='T', must_be_copyable=False, must_be_droppable=False), ConstParam(idx=1, name='n', ty=NumericType(kind=<Kind.Nat: 1>), from_comptime_arg=False)], never_copyable=True, never_droppable=False, to_hugr=<function _array_to_hugr at 0x7dd2ab4b24b0>, bound=None)), defined_at=Name(id='qs', ctx=Store()), flags=<InputFlags.NoFlags: 0>, is_func_input=False)), _parent=None, children=[], index=4, size=4)`

I have modified those 2 unit tests so that they do not fail, but if they should be modified in some other way, please let me know. 

BEGIN_COMMIT_OVERRIDE
feat: Compile time array bounds check for literal indices (https://github.com/Quantinuum/guppylang/issues/1438)
END_COMMIT_OVERRIDE

### DIFF
--- a/tests/error/array_errors/array_index_equal_size.err
+++ b/tests/error/array_errors/array_index_equal_size.err
@@ -1,6 +1,6 @@
 Error: Index out of bounds (at $FILE:7:14)
   | 
-5 | def main() -> None:
+5 | def main() -> int:
 6 |    arr = array(i for i in range(5))
 7 |    return arr[5]
   |               ^ Array index 5 is out of bounds for array of size 5.
--- a/tests/error/array_errors/array_index_negative_oob.err
+++ b/tests/error/array_errors/array_index_negative_oob.err
@@ -1,6 +1,6 @@
 Error: Index out of bounds (at $FILE:7:14)
   | 
-5 | def main() -> None:
+5 | def main() -> int:
 6 |    arr = array(i for i in range(5))
 7 |    return arr[-5]
   |               ^^ Array index -5 is out of bounds for array of size 5.
--- a/tests/error/array_errors/array_index_nested_array_inner.err
+++ b/tests/error/array_errors/array_index_nested_array_inner.err
@@ -1,6 +1,6 @@
 Error: Index out of bounds (at $FILE:7:23)
   | 
-5 | def main() -> None:
+5 | def main() -> int:
 6 |       matrix = array(array(i + j for i in range(10)) for j in range(5))
 7 |       return matrix[0][20]
   |                        ^^ Array index 20 is out of bounds for array of size 10.
--- a/tests/error/array_errors/array_index_nested_array_outer.err
+++ b/tests/error/array_errors/array_index_nested_array_outer.err
@@ -1,6 +1,6 @@
 Error: Index out of bounds (at $FILE:7:20)
   | 
-5 | def main() -> None:
+5 | def main() -> int:
 6 |       matrix = array(array(i + j for i in range(10)) for j in range(5))
 7 |       return matrix[5][0]
   |                     ^ Array index 5 is out of bounds for array of size 5.
--- a/tests/error/array_errors/array_index_positive_oob.err
+++ b/tests/error/array_errors/array_index_positive_oob.err
@@ -1,6 +1,6 @@
 Error: Index out of bounds (at $FILE:7:14)
   | 
-5 | def main() -> None:
+5 | def main() -> int:
 6 |    arr = array(i for i in range(5))
 7 |    return arr[10]
   |               ^^ Array index 10 is out of bounds for array of size 5.


tests/integration/test_quantum.py
```diff
@@ -198,7 +198,7 @@ def test() -> None:
 
         q.cx(qs[1], qs[2])
         barrier(qs[2], qs[3])
-        q.cx(qs[3], qs[4])
+        # q.cx(qs[3], qs[4]) #IndexOutOfBoundsError here...
```
maximilianruesch: Instead of disabling this line, please adjust the indices of the array accesses instead. Qubit 0 is currently entirely unused, so we can make use of it.

Also for the second occurrence in the other file.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,127 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
         return with_loc(self.node, node), get_type(array_arg)
 
 
+class ArrayIndexChecker(CustomCallChecker):
+    """Custom call checker that performs compile-time bounds
+      checking for array indexing.
```
maximilianruesch:
```suggestion
    """Performs compile-time bounds checking for array indexing.
```
nit: The supertype is visible one line above, we do not need to repeat ourselves here.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,127 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+    When the array size is statically known and the index is a literal constant,
+    this checker validates that the index is within bounds and raises an error
+    at compile time if it's not.
+    """
+
+    @dataclass(frozen=True)
+    class IndexOutOfBoundsError(Error):
+        title: ClassVar[str] = "Index out of bounds"
+        span_label: ClassVar[str] = (
+            "Array index {index} is out of bounds for array of size {size}"
```
maximilianruesch:
```suggestion
            "Array index {index} is out of bounds for array of size {size}."
```
nit

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,127 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+    def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
+        """Check-mode: verify arguments against
+        expected type and perform bounds check."""
+
+        # Run regular type checking for the arguments
+        args, subs, type_args = check_call(self.func.ty, args, ty, self.node, self.ctx)
+
+        # Extract the array length from type arguments
+        # Type args are: [element_type, length]
+        if len(type_args) >= 2:
```
maximilianruesch: Did you encounter any cases where this was not always true? A Guppy array has two type parameters, so I would expect there to be two arguments at any point.

Same goes for the `len(args)` check below: When this checker is called, we should already be in a scenario where two arguments are supplied. We could thus simplify the control flow here.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,127 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+                if len(args) >= 2:
+                    index_expr = args[1]
+
+                    # Extract constant index value
+                    index_value = self._extract_constant_index(index_expr)
+
+                    # Perform bounds check
+                    if index_value is not None and (
+                        index_value < 0 or index_value >= array_length
+                    ):
```
maximilianruesch:
```suggestion
                    index_value = self._extract_constant_index(args[1])

                    if index_value is not None and (
                        index_value < 0 or index_value >= array_length
                    ):
```
Can simplify without losing clarity

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,127 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        return with_loc(self.node, node), subs
+
+    def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
+        """Synthesize-mode: infer return type and perform bounds check."""
+        # Run regular type synthesis for the arguments
+        args, subs, type_args = synthesize_call(self.func.ty, args, self.node, self.ctx)
+
+        # Extract the array length from type arguments
+        # Type args are: [element_type, length]
+        if len(type_args) >= 2:
```
maximilianruesch: This looks really similar to the code above. It can be extracted into a function and thus deduplicated (after the simplifications above). You could consider passing `type_args[1]` and `args[1]` directly if the indices are always valid, to spare the new function know all other (type) arguments.

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+from guppylang_internals.error import GuppyError
+from guppylang.std.quantum import qubit, measure, h, discard_array
+
+# ============================================================================
+# POSITIVE OUT-OF-BOUNDS TESTS (index >= array_length)
+# ============================================================================
+
+
+def test_index_equals_array_size():
+    """Test index equal to array size (first invalid index)."""
```
maximilianruesch:
```suggestion
```
nit: Most of these doc strings are directly inferable from the test name, so they can be omitted.

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+
+import pytest
+from guppylang import guppy
+from guppylang.std.builtins import array
+from guppylang_internals.error import GuppyError
+from guppylang.std.quantum import qubit, measure, h, discard_array
+
+# ============================================================================
+# POSITIVE OUT-OF-BOUNDS TESTS (index >= array_length)
+# ============================================================================
```
maximilianruesch: Suggestion: Instead of adding these, you could simply prefix the test case names with `_oob` or `_negative` or `_valid` and so on.

tests/integration/test_array_index.py
maximilianruesch, line 1: We usually test error cases with `.err` snapshot files, as they allow us to capture the exact diagnostics and assert their style (i.e. what the user will actually see) as well. Especially `tests/error/array_error/*` might be relevant for this.

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+        y = arr[5]
+        z = arr[9]
+        return x + y + z
+
+    main.compile()  # Should succeed
+
+
+# ============================================================================
+# SINGLE ELEMENT ARRAY TESTS
+# ============================================================================
```
maximilianruesch: I do believe these test cases do not add much to the test base, they can be removed. The cases should be covered by more general in-bounds and out-of-bounds tests (which they are).

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+
+    @guppy
+    def main() -> int:
+        arr = array(i for i in range(10))
+        x = 5
+        # Note: depending on optimization, x might be treated as constant
+        # This test validates that if it's not optimized away, it doesn't error
+        return arr[0]  # Use literal to ensure it compiles
+
+    main.compile()  # Should succeed
```
maximilianruesch: I do not understand what this test is supposed to do. `x` does not interact with `arr` in any way here.

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+
+
+# ============================================================================
+# DIFFERENT ARRAY TYPES
+# ============================================================================
```
maximilianruesch: These tests are not adding much (in the same vein as the previous comment). One could consider keeping the nested array check and checking bounds checks for the inner array though.

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+    main.compile()  # Should succeed
+
+
+# ============================================================================
+# MULTIPLE ERRORS IN SAME FUNCTION
+# ============================================================================
+
+
+def test_first_error_caught():
+    """Test that the first out-of-bounds access is caught."""
```
maximilianruesch: Does it really matter that it is the first one, as long as all are caught successively as they are resolved by the user?

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+# ============================================================================
+
+
+def test_method_with_valid_index():
+    """Test array methods with valid constant indices."""
+
+    @guppy
+    def main() -> None:
+        qs = array(qubit() for _ in range(10))
+        result = qs.is_borrowed(3)  # Valid index
```
maximilianruesch, line 114: You did not implement the custom call checker on `is_borrowed`, so this can never fail.

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+        main.compile()
+
+    err = excinfo.value.error
+    assert err.title == "Index out of bounds"
+    assert err.index == 5
+    assert err.size == 5
+
+
+# ============================================================================
+# METHOD CALL CONTEXT (like is_borrowed, take, put)
```
maximilianruesch: `put` is missing here

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,550 @@
...
+
+    err = excinfo.value.error
+    assert err.title == "Index out of bounds"
+    assert err.index == 10
+    assert err.size == 10
+
+
+# ============================================================================
+# BOUNDARY VALUE ANALYSIS
+# ============================================================================
```
maximilianruesch: What does boundary mean here, and what do these cases add to the test base?

tests/integration/test_quantum.py
rmfisher26: I have updated the indices in both of the failed tests so that index 4 is no longer be used.

tests/integration/test_array_index.py
rmfisher26: Will do.  I will apply this suggestion and also remove the doc strings for other tests that follow that pattern.

tests/integration/test_array_index.py
rmfisher26: Thanks, I meant to have the "`x`" variable in the return `arr[x]`, instead so I will be adjusting this test.

tests/integration/test_array_index.py
rmfisher26: Perhaps not. The intent of this test was to show that catches the 1st compilation error and doesn't catch all possible errors.

tests/integration/test_array_index.py
rmfisher26: The purpose of these tests is to show another way to test the lower and upper boundaries in a small array. So testing that index 0 and 1 are valid, but 2 is outside of the boundary.  But, I do understand your point that is may not be adding much value given the other tests. So if you would to have it removed, I can do that.

tests/integration/test_array_index.py
rmfisher26: Understood. I will keep the nested array check with some other variations since that may be more beneficial.

tests/integration/test_array_index.py
rmfisher26: Thanks! I will add that the custom call checker to the `is_borrowed`

tests/integration/test_array_index.py
rmfisher26: I have added a test case for `put `as well now which can be reviewed.

tests/integration/test_array_index.py
rmfisher26: I have removed the tests under this category.

tests/integration/test_array_index.py
rmfisher26: Works for me. I have applied updates to use more prefixes like you suggested and removed those category titles

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: I did not encounter any cases where this wasn't true but wanted to ensure just in case. However, I did not realize that the Guppy array should only have 2 type parameters so I agree that it makes sense to simplify this. 

I have removed the 2 if conditions:  `if len(type_args) >= 2:` and `if len(args) >= 2:` and just assigned the values. I re-ran my same test suite, and did not observe any issues

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: Thanks for recommendation, and I have added a new `_check_constant_index_bounds ` to deduplicate the calls in `check `and `synthesize`  to try to simplify this further.

tests/integration/test_array_index.py
rmfisher26: Sure, I'm unfamiliar with how this is done, but will try to look in other examples in that location.

rmfisher26: @maximilianruesch, I have added 6 examples of array_index tests in `tests/error/array_error/*`, so let me know if this what you are looking for. If you need me to add more `.err` files to be more thorough, I can do that too.

Also, should I remove any "assert" tests such as ones like this from `tests/integration/test_array_index.py`?

```
def test_positive_index_equals_array_size():
    @guppy
    def main() -> int:
        arr = array(i for i in range(5))
        return arr[5]  # Index 5 is invalid for size 5 (valid: 0-4)

    with pytest.raises(GuppyError) as excinfo:
        main.compile()

    err = excinfo.value.error
    assert err.title == "Index out of bounds"
    assert err.index == 5
    assert err.size == 5
```

maximilianruesch: Yes. This is what I was looking for. The idea is to migrate all tests that contain error cases (which can be faithfully reproduced using the snapshots) to the snapshots.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,101 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        """Extract a constant integer value from an index expression if possible.
+
+        Handles both AST constants and PlaceNode structures.
+        """
+        # Case 1: Simple AST constant (e.g., arr[8] - direct Constant node)
+        if isinstance(index_expr, ast.Constant) and isinstance(index_expr.value, int):
+            return index_expr.value
+
+        # Case 2: PlaceNode with constant defined_at (e.g., after type checking)
+        if hasattr(index_expr, "place"):
```
mark-koch: It looks like you're trying to do constant propagation here? I.e. identifying

```
i = 42
xs[i]
```

Unfortunately, looking at the `defined_at` field won't achieve that. I think for this PR it's ok to handle case 1 above and remove this

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,101 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        return None
+
+    def _check_constant_index_bounds(
+        self, index_expr: ast.expr, length_arg: TypeArg | ConstArg
+    ) -> None:
+        """Perform compile-time bounds checking if size and index are constant."""
+
+        # Check if array size is statically known
+        if not (
+            hasattr(length_arg, "const") and isinstance(length_arg.const, ConstValue)
```
mark-koch: Please don't use `hasattr`, you can check `isinstance(length_arg, ConstArg)`

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,101 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        # Extract constant index if possible
+        index_value = self._extract_constant_index(index_expr)
+        if index_value is None:
+            return
+
+        # Perform bounds check
+        if index_value < 0 or index_value >= array_length:
+            raise GuppyError(
+                ArrayIndexChecker.IndexOutOfBoundsError(
+                    self.node,
```
mark-koch: I think it would be better to use `index_expr` as the error span

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,101 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        self._check_constant_index_bounds(args[1], type_args[1])
+
+        # Return the synthesized node and type
+        node = GlobalCall(def_id=self.func.id, args=args, type_args=type_args)
+        return with_loc(self.node, node), subs
+
+    def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
+        """Synthesize-mode: infer return type and perform bounds check."""
+        # Run regular type synthesis for the arguments
+        args, subs, type_args = synthesize_call(self.func.ty, args, self.node, self.ctx)
```
mark-koch:
```suggestion
        args, ty, type_args = synthesize_call(self.func.ty, args, self.node, self.ctx)
```

`synthesize_call` returns a type instead of a substitution

tests/error/array_errors/array_index_set_item_by_variable.err
```diff
@@ -0,0 +1,8 @@
+Error: Type mismatch (at $FILE:8:13)
+  | 
+6 |       arr = array(i for i in range(10))
+7 |       x = 50
+8 |       return arr[x]
+  |              ^^^^^^ Expected return value of type `None`, got `int`
```
mark-koch: This looks like an unrelated error

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,532 @@
...
+"""
+
+import pytest
+from guppylang import guppy
+from guppylang.std.builtins import array
+from guppylang_internals.error import GuppyError
+from guppylang.std.quantum import qubit, measure, h, discard_array
+
+
+def test_positive_index_equals_array_size():
```
mark-koch: Now that you added the snapshot tests, could you remove all the error tests in this file?

tests/error/array_errors/array_index_equal_size.py
```diff
@@ -0,0 +1,9 @@
+from guppylang.decorator import guppy
+from guppylang.std.builtins import array
+
+@guppy
+def main() -> None:
+   arr = array(i for i in range(5))
+   return arr[5]
```
mark-koch, line 7: Could you add a similar test for `array.is_borrowed`, `array.get` and `array.put`?

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: Thanks. This has been updated to also return `'ty`" in the return statement

```
        """Synthesize-mode: infer return type and perform bounds check."""
        # Run regular type synthesis for the arguments
        args, ty, type_args = synthesize_call(self.func.ty, args, self.node, self.ctx)

        # Check the index bounds (first:index expression, second: length_arg)
        self._check_constant_index_bounds(args[1], type_args[1])

        # Return the synthesized node and type
        node = GlobalCall(def_id=self.func.id, args=args, type_args=type_args)
        return with_loc(self.node, node), ty
```

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: Sounds good. This has been applied so now the error span looks like this in 

`tests/error/array_errors/array_index_equal_size.err
`
```
Error: Index out of bounds (at $FILE:7:14)
  | 
5 | def main() -> None:
6 |    arr = array(i for i in range(5))
7 |    return arr[5]
  |               ^ Array index 5 is out of bounds for array of size 5.

```

or like this

`tests/error/array_errors/array_index_nested_array_inner.err`

```
Error: Index out of bounds (at $FILE:7:23)
  | 
5 | def main() -> None:
6 |       matrix = array(array(i + j for i in range(10)) for j in range(5))
7 |       return matrix[0][20]
  |                        ^^ Array index 20 is out of bounds for array of size 10.

```

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: Understood, this has been corrected/committed so that it looks like this: 

```
  # Check if array size is statically known
        if not (
            isinstance(length_arg, ConstArg)
            and isinstance(length_arg.const, ConstValue)
        ):
            return
```

tests/error/array_errors/array_index_set_item_by_variable.err
rmfisher26: This test for `array_index_set_item_by_variable.py` and `array_index_set_item_by_variable.err` has been removed.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: Hi @mark-koch, if I remove the Case 2 block so that I'm left with:

```
   def _extract_constant_index(self, index_expr: ast.expr) -> int | None:
        """Extract a constant integer value from an index expression if possible.

        """
        # Case 1: Simple AST constant (e.g., arr[8] - direct Constant node)
        if isinstance(index_expr, ast.Constant) and isinstance(index_expr.value, int):
            return index_expr.value
```

In my error snapshot tests, such as `tests/error/array_errors/array_index_equal_size.err`

```
Error: Type mismatch (at $FILE:7:10)
  | 
5 | def main() -> None:
6 |    arr = array(i for i in range(5))
7 |    return arr[5]
  |           ^^^^^^ Expected return value of type `None`, got `int`

```


And with current logic (keeping Case 2)

```
def _extract_constant_index(self, index_expr: ast.expr) -> int | None:
        """Extract a constant integer value from an index expression if possible.

        Handles both AST constants and PlaceNode structures.
        """
        # Case 1: Simple AST constant (e.g., arr[8] - direct Constant node)
        if isinstance(index_expr, ast.Constant) and isinstance(index_expr.value, int):
            return index_expr.value

        # Case 2: PlaceNode with constant defined_at (e.g., after type checking)
        if hasattr(index_expr, "place"):
            place = index_expr.place
            if hasattr(place, "defined_at"):
                defined_at = place.defined_at
                # Check if defined_at is a Constant node with a value
                if isinstance(defined_at, ast.Constant) and isinstance(
                    defined_at.value, int
                ):
                    return defined_at.value

        return None
```

```
Error: Index out of bounds (at $FILE:7:14)
  | 
5 | def main() -> None:
6 |    arr = array(i for i in range(5))
7 |    return arr[5]
  |               ^ Array index 5 is out of bounds for array of size 5.

```
So right now, it would seem to me that having Case 2 does help with getting the error message that I expect. Is there a different approach that you had in mind for `_extract_constant_index`?

tests/error/array_errors/array_index_equal_size.py
rmfisher26: @mark-koch, I have added `array.is_borrowed`, `array.get`, `array.take`, and `array.put`.

However this is a failure with
```

=========================== short test summary info ============================
FAILED tests/error/test_array_errors.py::test_array_errors[/home/runner/work/guppylang/guppylang/tests/error/array_errors/array_index_put.py] - Failed: DID NOT RAISE <class 'Exception'>
======= 1 failed, 1168 passed, 23 skipped, 4 xfailed in 89.27s (0:01:29) =======

```
This `tests/error/array_errors/array_index_put.py` looks like this:

```
from guppylang.std.quantum import qubit, measure, discard_array

@guppy
def main() -> None:
   qs = array(qubit() for _ in range(5))
   q = qs.take(0)
   measure(q)
   qs.put(qubit(), 10)  # Out of bounds
   discard_array(qs)

main.compile()
```

Is there something I'm missing in my test for this. I was expecting the out of bounds exception with `qs.put(qubit(), 10) `, but that did not happen.

tests/integration/test_array_index.py
rmfisher26: Yes, of course. I have removed the error tests in` tests/integration/test_array_index.py` now.

tests/integration/test_array_index.py
rmfisher26: I have added more snapshot `.err` files and removed the error tests from `tests/integration/test_array_index.py`

tests/error/array_errors/array_index_equal_size.py
rmfisher26: The error was that the checker did not look in second argument for the index, such as in  `qs.put(qubit(), 10)` so it was  returning `None`. So changing code to try to examine `args[1]` and `args[2]` helped for a Constant index. There may be a better way to loop through the `args `here, but this helped to produce the `tests/error/array_errors/array_index_put.err`

To do something like this:
```
        """Check-mode: verify arguments against
        expected type and perform bounds check."""

        # Run regular type checking for the arguments
        args, subs, type_args = check_call(self.func.ty, args, ty, self.node, self.ctx)

        # Check the index bounds (first:index expression, second: length_arg)
        # Check args[1] and args[2] both against type_args[1]
        for index_expr in args[1:3]:
            self._check_constant_index_bounds(index_expr, type_args[1])

        # Return the synthesized node and type
        node = GlobalCall(def_id=self.func.id, args=args, type_args=type_args)
        return with_loc(self.node, node), subs


```

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,106 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        """Perform compile-time bounds checking if size and index are constant."""
+
+        # Check if array size is statically known
+        if not (
+            isinstance(length_arg, ConstArg)
+            and isinstance(length_arg.const, ConstValue)
+        ):
+            return
+
+        # get the array length
```
maximilianruesch: The code in this function is sufficiently self documenting, so all comments starting here and below (lines 226, 229 and 234) can be removed.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -172,6 +172,106 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+        """Check-mode: verify arguments against
+        expected type and perform bounds check."""
+
+        # Run regular type checking for the arguments
+        args, subs, type_args = check_call(self.func.ty, args, ty, self.node, self.ctx)
+
+        # Check the index bounds (first:index expression, second: length_arg)
+        # Check args[1] and args[2] both against type_args[1]
+        for index_expr in args[1:3]:
+            self._check_constant_index_bounds(index_expr, type_args[1])
```
maximilianruesch: Using a for-loop is incredible cognitive overhead for what you are trying to achieve. I suggest just calling the function twice (once with args[1] and once with args[2], which directly makes the comment above that line superfluous).
Same for the synthesize function.

--- 

That being said, the index of the index_expr should be an argument to the checker anyway (see comment on Marks review).

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,169 @@
+"""
+Comprehensive unit tests for ArrayIndexChecker class.
+Tests cover all edge cases, both valid and invalid scenarios.
```
maximilianruesch:
```suggestion
```
This is no longer true for this file.

tests/error/array_errors/array_index_equal_size.py
maximilianruesch: Instead of specifying to loop through both arguments, it would be useful for the checker to know which argument to extract. This means adding a keyword-only argument (like `expr_index=...`) to the initialisation of the checker class that specifies (for that particular application of the checker) which argument contains the index expression.

tests/error/array_errors/array_index_get.py
```diff
@@ -0,0 +1,14 @@
...
+@guppy
+def main() -> None:
+   qs = array(qubit() for _ in range(5))
+   qs[10]
+   discard_array(qs)
+
+main.compile()
+
+
+
```
maximilianruesch: nit: Lots of trailing whitespace at the bottom of these files. I cannot really force you to edit these though, as we should instead have some lint for these.

tests/integration/test_array_index.py
```diff
@@ -0,0 +1,169 @@
...
+    main.compile()  # Should succeed
+
+
+def test_boundary_size_2_index_1():
+    @guppy
+    def main() -> int:
+        arr = array(i for i in range(2))
+        return arr[1]
+
+    main.compile()  # Should succeed
```
maximilianruesch: Please remove these. You already have "boundary" tests at the top of the file.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
maximilianruesch: If this case still persists, please do some more debugging as to why it fails and elaborate. Especially given the simplicity of the test case, one should not expect the need to handle a wildy different separate check.

As a useful tool for debugging, you might want to temporarily remove the `return` statement from `guppylang_internals/src/guppylang_internals/error.py$pretty_errors$hook` (such as [this one](https://github.com/Quantinuum/guppylang/blob/e6e113cd60efad05baa4abf00fd6bc5504a8ca0a/guppylang-internals/src/guppylang_internals/error.py#L97)). This helps with providing a traceback for the error you are seeing.

maximilianruesch (dismissed):
Some comments left (also two in the previous review from Mark). It is already looking much better! Thanks for the ongoing effort.

tests/integration/test_array_index.py
rmfisher26: Of course! I have removed these lines now.

tests/error/array_errors/array_index_get.py
rmfisher26: Sure, I did just remove the whitespace in the tests/error/array_errors/array_index_get.py so hopefully that looks cleaner now.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: Sounds good. I have removed the comments on those specific lines as requested.

tests/error/array_errors/array_index_equal_size.py
rmfisher26: Thank you for the feedback! I will continue my debugging and development to improve this further.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: Understood.  I'll continue my development to improve this implementation.

tests/error/array_errors/array_index_equal_size.py
rmfisher26: @maximilianruesch, I have added an __init__ section using the expr_index=... in the `checker.py`. The default is position 1 and the allows for something like this:
```
    @custom_function(
        ArraySetitemCompiler(elem_first=True), checker=ArrayIndexChecker(expr_index=2)
    )
    def put(self: array[L, n], elem: L @ owned, idx: int) -> None:
```

This also simplified the check and synthesize to use this'

` self._check_constant_index_bounds(args[self.expr_index], type_args[1])`

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: @maximilianruesch,  similar to above comment, I have added an __init__ section using the expr_index=... in the `checker.py`. The default is position 1 and the allows for something like this:
```
    @custom_function(
        ArraySetitemCompiler(elem_first=True), checker=ArrayIndexChecker(expr_index=2)
    )
    def put(self: array[L, n], elem: L @ owned, idx: int) -> None:
```

This also simplified the check and synthesize to use this'

` self._check_constant_index_bounds(args[self.expr_index], type_args[1])`

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
rmfisher26: The main reason is that the index_expr is not always `ast.Constant`, sometimes it is a `PlaceNode`

When I remove the Case 2 block from that,  I get None returned from     `self._extract_constant_index(index_expr)`. this is skips the next section where it can raise the IndexOutOfBoundsError

So I get this instead 

```
Error: Type mismatch (at $FILE:7:10)
  | 
5 | def main() -> None:
6 |    arr = array(i for i in range(5))
7 |    return arr[5]
  |           ^^^^^^ Expected return value of type `None`, got `int`

```

While trying to prove that Case 1 is not always true, I added a print to show in logs.

```
    def _extract_constant_index(self, index_expr: ast.expr) -> int | None:
        """Extract a constant integer value from an index expression if possible.

        Handles both AST constants and PlaceNode structures.
        """
        # Case 1: Simple AST constant (e.g., arr[8] - direct Constant node)
        if isinstance(index_expr, ast.Constant) and isinstance(index_expr.value, int):
            print("\nCase 1:Constant", index_expr)
            return index_expr.value

        # Case 2: PlaceNode with constant defined_at (e.g., after type checking)
        if hasattr(index_expr, "place"):
            print("\nCase 2:PlaceNode", index_expr)
            place = index_expr.place
            if hasattr(place, "defined_at"):
                defined_at = place.defined_at
                # Check if defined_at is a Constant node with a value
                if isinstance(defined_at, ast.Constant) and isinstance(
                    defined_at.value, int
                ):
                    return defined_at.value

        return None

```

And then running a test:  `test_method_take_put_valid_index`


```
def test_method_take_put_valid_index():
    @guppy
    def main() -> None:
        measure(qs.take(3))          #Case 1:Constant Constant(value=3, kind=None)
        qs.put(qubit(), 3)           #Case 1:Constant Constant(value=3, kind=None)
        h(qs[3])                     #Case 2:PlaceNode PlaceNode(place=Variable(name='%tmp1', ty=NumericType(kind=<Kind.Int: 2>), defined_at=Constant(value=3, kind=None), flags=<InputFlags.NoFlags: 0>, is_func_input=False))
        discard_array(qs)

main.compile()  


```

or another test example

```
def test_valid_index_zero():
    @guppy
    def main() -> int:
        arr = array(42 for _ in range(5))
        return arr[0]  # Case 2:PlaceNode PlaceNode(place=Variable(name='%tmp1', ty=NumericType(kind=<Kind.Int: 2>), defined_at=Constant(value=0, kind=None), flags=<InputFlags.NoFlags: 0>, is_func_input=False))

    main.compile()  
```

maximilianruesch: I investigated a bit, and there are two causes for the problems:

1. Subscript accesses are by [default augmented with temporary variables](https://github.com/Quantinuum/guppylang/blob/4120d590917245258d69ceb07c763c7633c7c5f8/guppylang-internals/src/guppylang_internals/checker/expr_checker.py#L696-L700), even if the access is an integer constant. This is a bit more involved to fix, since compiled subscript accesses are not prepared to take anything other than a (temporary) variable. @mark-koch this probably needs fixing in its own right, and might not need to be done in this PR.

2. You are seeing the other error since (in the tested function), `main` has an annotated return type of `None` even though you are returning an expression. You should check and correct this across all test files.

I will see if we can come up with a better workaround for the current case. In any case, we should avoid using `hasattr` and instead use `isinstance`, since this provides more IDE support for seeing usage of classes. The comments also need updating to explain this (and now the example in case 1 is also not that correct anymore it seems).

rmfisher26: @maximilianruesch, I did modify the code to use `isinstance` only so that it looks like this (hopefully that's an improvement at least)

From this commit: 
https://github.com/Quantinuum/guppylang/pull/1448/commits/99ec009ca524fa3abd8caf02f6285434937e725e

```
def _extract_constant_index(self, index_expr: ast.expr) -> int | None:
        """Extract a constant integer value from an index expression if possible.

        Handles both AST constants and PlaceNode structures.
        """
        # Case 1: Simple AST constant (e.g., arr.take(0))
        if isinstance(index_expr, ast.Constant) and isinstance(index_expr.value, int):
            return index_expr.value

        # Case 2: Subscript accesses (e.g., arr[0])
        if isinstance(index_expr, PlaceNode):
            place = index_expr.place
            if isinstance(place, Variable):
                defined_at = place.defined_at
                if isinstance(defined_at, ast.Constant) and isinstance(
                    defined_at.value, int
                ):
                    return defined_at.value

        return None
```

Regarding your second comment, in the test files I get an error if don't have `None` in the return type for `main` so just leaving that now since that is consistent with other tests I've seen in the `array_errors ` folder.

tests/error/array_errors/array_index_equal_size.py
```diff
@@ -0,0 +1,9 @@
+from guppylang.decorator import guppy
+from guppylang.std.builtins import array
+
+@guppy
+def main() -> None:
```
nicolaassolini-qntm:
```suggestion
def main() -> int:
```

nicolaassolini-qntm: Not important for the test, just to have the syntax coherent

tests/error/array_errors/array_index_negative_oob.py
```diff
@@ -0,0 +1,9 @@
+from guppylang.decorator import guppy
+from guppylang.std.builtins import array
+
+@guppy
+def main() -> None:
```
nicolaassolini-qntm:
```suggestion
def main() -> int:
```

tests/error/array_errors/array_index_nested_array_inner.py
```diff
@@ -0,0 +1,9 @@
+from guppylang.decorator import guppy
+from guppylang.std.builtins import array
+
+@guppy
+def main() -> None:
```
nicolaassolini-qntm:
```suggestion
def main() -> int:
```

tests/error/array_errors/array_index_nested_array_outer.py
```diff
@@ -0,0 +1,9 @@
+from guppylang.decorator import guppy
+from guppylang.std.builtins import array
+
+@guppy
+def main() -> None:
```
nicolaassolini-qntm:
```suggestion
def main() -> int:
```

tests/error/array_errors/array_index_positive_oob.py
```diff
@@ -0,0 +1,9 @@
+from guppylang.decorator import guppy
+from guppylang.std.builtins import array
+
+@guppy
+def main() -> None:
```
nicolaassolini-qntm:
```suggestion
def main() -> int:
```

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
nicolaassolini-qntm: @maximilianruesch @mark-koch @rmfisher26 On my side I think that this solution is fine now!

tests/error/array_errors/array_index_equal_size.py
rmfisher26: Sounds good, this update has been applied accordingly to those files.

PR #1451, opened 2026-01-22
feat: use hugr used_extensions in engine.py

ss2165:
EDIT: the below is out of date, option 1 for addressing was taken and released.

Requires HUGR release, using to identify bugs upstream.
Summary of issues:
- HUGR `used_extensions` returns an `ExtensionRegistry`, which involves concrete Extension objects
- Opaque types and operations will therefore fail with "extension hasn't been resolved"
- HUGR Py doesn't really do extension resolution
- Guppy mostly doesn't use opaques, pointing instead to concrete extensions.
- The exceptions to this are:
  - WASM: loads the `tket_exts` object, signatures of ops refer to opaque types belonging to the same extension
  - pytket loading: the HUGR returned from `tket` opaquely refers to tket extensions

How to address?
1. Implement full resolution in hugr-py for both HUGR and extensions, call before calling `used_extensions`. Or compute both together.
2. WASM problem can probably be addressed by resolving just the extensions when loading them (i.e resolve the TKET_EXTENSIONS registry against itself + prelude + std)
3. Special case tket extension resolution for pytket loading, just find and replace the tket extension references
4. Don't return a full registry for `used_extensions`, just return names (or add an alternative function that does this, that won't fail on opaques), match those against the "known-used" registry
5. ??

### DIFF
--- a/guppylang-internals/src/guppylang_internals/engine.py
+++ b/guppylang-internals/src/guppylang_internals/engine.py
@@ -5,9 +5,8 @@ from typing import TYPE_CHECKING
 
 import hugr
 import hugr.build.function as hf
-from hugr import ext as hugr_ext
 from hugr import ops
-from hugr.ext import Extension
+from hugr.ext import Extension, ExtensionRegistry
 from hugr.package import ModulePointer, Package
 
 import guppylang_internals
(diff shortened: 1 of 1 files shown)


aborgna-q (approved):
🚀

Issue #924, opened 2025-04-25
Add function to swap two array elements that lowers to hugr swap op


ss2165:
Equivalent to using `mem_swap` but more efficient.

ss2165:
Would be good to also add an operation to swap an element out of the array

maximilianruesch:
Addressed by #1459

PR #1459, opened 2026-01-26
feat: Add function to swap two array elements that lowers to hugr swap op

rmfisher26:
Proposing code for issue #924 to see what others think and if there are other suggestions.


guppylang/src/guppylang/std/qsystem/random.py
```diff
@@ -98,7 +98,7 @@ def shuffle(self: "RNG", array: array[SHUFFLE_T, SHUFFLE_N]) -> None:
             # TODO use array swap once lowering implemented
             # https://github.com/quantinuum/guppylang/issues/924
```
croyzor: You can delete this comment!

guppylang/src/guppylang/std/array.py
```diff
@@ -250,6 +251,27 @@ def __next__(
 def _array_discard_all_used(xs: array[L, n] @ owned) -> None: ...
 
 
+@custom_function(ArraySwapCompiler())
+def array_swap(arr: array[L, n], idx: int, idx2: int) -> None:
+    """Swap two elements in an array at indices idx and idx2.
+    Exchanges the elements at two indices within the array. This operation
+    compiles directly to HUGR's native swap operation, making it more efficient
+    than using `mem_swap(arr[idx], arr[idx2])`.
```
croyzor: I don't buy the efficiency argument. It seems to me that `array_swap(array, i, j)` and `mem_swap(array[i], array[j])` amount to the same thing. I would drop this part of the docstring (maybe even say "equivalent to `mem_swap(arr[idx], arr[idx2])`" for clarity)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -414,3 +414,56 @@ def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
 
     def compile(self, args: list[Wire]) -> list[Wire]:
         raise InternalGuppyError("Call compile_with_inouts instead")
+
+
+class ArraySwapCompiler(ArrayCompiler):
+    """Compiler for the `array_swap` function.
+    Implements swap using borrow_array's borrow and return operations.
```
tatiana-s: I think the intention was to lower this to the swap op (https://github.com/Quantinuum/hugr/blob/14cd80a1af99609269ecb10b5decfb964bb6f677/hugr-core/src/std_extensions/collections/array/array_op.rs#L43) as opposed to borrow and return?

rmfisher26: Understood, I will try again to use the swap op as it was intended and I can share some of my results

guppylang/src/guppylang/std/array.py
rmfisher26: Thank you for your feedback, I have modified the description to using "equivalent" for clarity.

guppylang/src/guppylang/std/qsystem/random.py
rmfisher26: Sounds good, will do.

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
rmfisher26: @tatiana-s  I implemented a different approach to try to use the `swap ` op. 

After doing so, and updating the `ArraySwapCompiler`,  there a few failed tests, which they mention  a `borrow_array`. So there is anything that must be considered with this type of array? 


```
=========================== short test summary info ============================
FAILED tests/integration/test_qsystem.py::test_qsystem_random - selene_hugr_qis_compiler.HugrReadError: Error loading HUGR package.

Caused by:
    0: Error reading package payload in envelope.
    1: Error resolving opaque operation: Conflicting signature: resolved swap in extension collections.borrow_arr to a concrete implementation which computed [borrow_array(#1, #0), usize, usize] -> [[borrow_array(#1, #0)]+[borrow_array(#1, #0)]] but stored signature was [borrow_array(#1, #0), usize, usize] -> [borrow_array(#1, #0)]
    2: Conflicting signature: resolved swap in extension collections.borrow_arr to a concrete implementation which computed [borrow_array(#1, #0), usize, usize] -> [[borrow_array(#1, #0)]+[borrow_array(#1, #0)]] but stored signature was [borrow_array(#1, #0), usize, usize] -> [borrow_array(#1, #0)]
FAILED tests/integration/tracing/test_basic.py::test_inout_type_infer - selene_hugr_qis_compiler.HugrReadError: Error loading HUGR package.

Caused by:
    0: Error reading package payload in envelope.
    1: Error resolving opaque operation: Conflicting signature: resolved swap in extension collections.borrow_arr to a concrete implementation which computed [borrow_array(#1, #0), usize, usize] -> [[borrow_array(#1, #0)]+[borrow_array(#1, #0)]] but stored signature was [borrow_array(#1, #0), usize, usize] -> [borrow_array(#1, #0)]
    2: Conflicting signature: resolved swap in extension collections.borrow_arr to a concrete implementation which computed [borrow_array(#1, #0), usize, usize] -> [[borrow_array(#1, #0)]+[borrow_array(#1, #0)]] but stored signature was [borrow_array(#1, #0), usize, usize] -> [borrow_array(#1, #0)]
======= 2 failed, 1181 passed, 23 skipped, 4 xfailed in 80.09s (0:01:20) =======
`
```

rmfisher26: @tatiana-s @croyzor , the above test errors have been resolved.   Let me know if this is implemented as requested or needs further changes.

tests/integration/test_array_swap.py
```diff
@@ -0,0 +1,41 @@
...
+def test_basic_swap_compiles():
+    """Verify basic swap compiles without errors."""
+
+    @guppy
+    def swap_first_last() -> None:
+        arr = array(1, 2, 3, 4, 5)
+        array_swap(arr, 0, 4)
+
+    hugr = swap_first_last.compile()
+    assert hugr is not None
```
croyzor: Typically for hugr correctness tests we do
```python
def test_basic_swap_compiles(validate):
    @guppy
     def foo(): ...

    hugr = foo.compile()
    validate(hugr)
```
(`validate` is a test fixture)

guppylang/src/guppylang/std/array.py
```diff
@@ -250,6 +251,32 @@ def __next__(
 def _array_discard_all_used(xs: array[L, n] @ owned) -> None: ...
 
 
+@custom_function(ArraySwapCompiler())
+def array_swap(arr: array[L, n], idx: int, idx2: int) -> None:
+    """Swap two elements in an array at indices idx and idx2.
+
+    Exchanges the elements at two indices within the array. This operation
+    uses HUGR's native swap operation from the collections.array extension,
+    making it equivalent to using `mem_swap(arr[idx], arr[idx2])`.
```
croyzor:
```suggestion
 ```
Sorry, your previous implementation was equivalent to a `mem_swap`, but this one is not!

File: guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py [hunk 6]
```diff
@@ -414,3 +432,38 @@ def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
...
+    """Compiler for `array_swap`."""
+
+    def compile_with_inouts(self, args: list[Wire]) -> CallReturnWires:
+        [array, idx1, idx2] = args
+
+        idx1 = self.builder.add_op(convert_itousize(), idx1)
+        idx2 = self.builder.add_op(convert_itousize(), idx2)
+
+        # Swap returns Either(left=array, right=array)
+        # Left (case 0) is success, right (case 1) is error
```
croyzor: It's not well documented in hugr, but it should be the other way around. Our convention is usually `0 => failure; 1 => success`, as in the hugr bool and option types

croyzor (changes requested):

tests/integration/test_array_swap.py
rmfisher26: Thanks for for the explanation, I have added the `validate` part to those tests.

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
rmfisher26: I did modify the comment so that it looks like this now so it is the other way around.
```
    # Left (case 0) is failure, right (case 1) is success
```

croyzor: It's not just here, there's a similar comment a few lines below, where `unwrap_left` is used instead of `unwrap_right`

tests/integration/test_array_swap.py
croyzor: There's another fixture that can be used for testing that I'd recommend for this case: `run_int_fn` which can run a simple program and check that it gets the right answer.

We could have something like this:
```python
def test_basic_swap_compiles(validate, run_int_fn):
    """Verify basic swap compiles without errors."""

    @guppy
    def swap_first_last() -> array[int, 5]:
        arr = array(1, 2, 3, 4, 5)
        array_swap(arr, 0, 4)
        return arr

    @guppy
    def main() -> int:
        arr = swap_first_last()
        return arr[0]

    hugr = swap_first_last.compile()
    validate(hugr)
    run_int_fn(main, 5)
```

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
rmfisher26: Thank you. I replaced this with using `build_unwrap_right` and updated the comment above that, respectively. Let me know if this is what your are expecting to see.

tests/integration/test_array_swap.py
rmfisher26: Understood! I have updated the `test_basic_swap `and `test_multiple_swaps` to follow that structure.

croyzor (approved):
Looks good, thanks for your help!

PR #1460, opened 2026-01-27
fix: Add a copy before fallable `synthesize_call` call 

croyzor:
Another instance of the root issue in https://github.com/Quantinuum/guppylang/issues/1405. We make fallible calls to `synthesize_call` or `check_call` which can modify the AST even if they fail, causing the successive code to change.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -1190,7 +1190,8 @@ def check_call(
         subst = unify(ty, synth, {})
         if subst is None:
             raise GuppyTypeError(TypeMismatchError(node, ty, synth, kind))
-        return inputs, subst, inst
+        else:
+            return inputs, subst, inst
     except GuppyTypeInferenceError:
         pass
 


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1182,7 +1183,11 @@ def check_call(
     # First, try to synthesize
     res: tuple[Type, Inst] | None = None
     try:
-        inputs, synth, inst = synthesize_call(func_ty, inputs, node, ctx)
+        # check_call may modify args and node,
+        # thus we deepcopy them before passing in the function
+        node_copy = copy.deepcopy(node)
+        inputs_copy = copy.deepcopy(inputs)
```
nicolaassolini-qntm: This deepcopy cause the `TypeMismatchError` in most of the test

nicolaassolini-qntm:
The commit [a92844c](https://github.com/Quantinuum/guppylang/commit/a92844c98d0bb8bb83c78beb02e96ec53a867bc8) solves the `TypeMismatchError`. There is still one failing test.

tests/integration/std/test_either.py
```diff
@@ -71,3 +71,16 @@ def main() -> None:
 
     res = main.emulator(1).coinflip_sim().run().results[0].entries
     assert res == [("a", 3), ("b", 1)]
+
+
+def test_either_comprehension():
```
nicolaassolini-qntm:
```suggestion
def test_either_comprehension(validate):
```

tests/integration/std/test_either.py
```diff
@@ -71,3 +71,16 @@ def main() -> None:
...
+    from guppylang.std.array import array
+
+    @guppy
+    def main(b: bool) -> Either[array[int, 3], int]:
+        if b:
+            return left(array(42 for x in range(3)))
+        else:
+            return right(0)
+
+    main.check()
```
nicolaassolini-qntm:
```suggestion
    validate(main.compile_function())
```

nicolaassolini-qntm: I tested on my machine and should be fine

croyzor: Validation should fail for this because the function getting `compile()`d isn't meant to have arguments. It fails on my machine and in the CI for https://github.com/Quantinuum/guppylang/pull/1460/commits/a92844c98d0bb8bb83c78beb02e96ec53a867bc8

nicolaassolini-qntm: Yes, but we can use `compile_function()` for that, right?

croyzor: Yes, good point! I didn't notice that part

tests/integration/std/test_either.py
```diff
@@ -71,3 +71,16 @@ def main() -> None:
 
     res = main.emulator(1).coinflip_sim().run().results[0].entries
     assert res == [("a", 3), ("b", 1)]
+
+
+def test_either_comprehension(validate):
+    from guppylang.std.array import array
```
maximilianruesch: What is the reason you import at "runtime" here, instead of at the top of the file?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1182,10 +1183,15 @@ def check_call(
     # First, try to synthesize
     res: tuple[Type, Inst] | None = None
     try:
+        # check_call may modify args and node,
+        # thus we deepcopy them before passing in the function
+        node_copy = copy.deepcopy(node)
+        inputs_copy = copy.deepcopy(inputs)
```
maximilianruesch: Since this is not *actually* scoped anyway, you may want to consider moving this outside of the `try` block, to make the intention / success expectation clearer.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1182,10 +1183,15 @@ def check_call(
     # First, try to synthesize
     res: tuple[Type, Inst] | None = None
     try:
+        # check_call may modify args and node,
```
```suggestion
        # synthesize_call may modify args and node,
```
Alternatively, consider (any combination of):
```suggestion
        # Synthesis may modify args and node in place,
```

tests/integration/std/test_either.py
croyzor: No reason in particular...

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1179,19 +1180,25 @@ def check_call(
...
-    except GuppyTypeInferenceError:
-        pass
-    if res is not None:
-        synth, inst = res
         subst = unify(ty, synth, {})
         if subst is None:
             raise GuppyTypeError(TypeMismatchError(node, ty, synth, kind))
-        return inputs, subst, inst
+        else:
+            return inputs, subst, inst
```
maximilianruesch, line 1194: The actual ruff lint that failed was [`TRY300`](https://docs.astral.sh/ruff/rules/try-consider-else/), and the suggested fix is (confusingly) to move a return statement that was in the try into the last of the `try: ... except: ... else: ...` blocks, not inside an else inside the `try: ...` block. In this context it is highly ambiguous.

Approved, and up to you whether you use the suggested fix from Ruff or stick with your current solution 👍

PR #1461, opened 2026-01-27
feat!: Allow panicking with different signal values

### LINKED ISSUE #767: allow panicking with different signal values
to allow signalling e.g. "end shot and start next" or "give up whole job, error"

tatiana-s:
Closes #767

BREAKING CHANGE: (guppy-internals) `has_var_args` flag added to `RawCustomFunctionDef` and `CustomFunctionDef`
BREAKING CHANGE: (guppy-internals) `PanicExpr` renamed to `AbortExpr`, `ExitKind` renamed to `AbortKind`

Edit: Not a `guppylang` breaking change,
BEGIN_COMMIT_OVERRIDE
feat: Allow panicking with different signal values (#1461)
END_COMMIT_OVERRIDE

### DIFF
--- a/guppylang/src/guppylang/std/platform.py
+++ b/guppylang/src/guppylang/std/platform.py
@@ -93,7 +93,7 @@ def result(tag: str, value):
     # can't be represented by a standard Guppy signature.
     signature=FunctionType(
         [
-            FuncInput(string_type(), InputFlags.NoFlags, "message"),
+            FuncInput(string_type(), InputFlags.NoFlags, "msg"),
         ],
         NoneType(),
     ),
@@ -107,7 +107,7 @@ def _panic(msg: str, *args) -> None: ...
     higher_order_value=False,
     signature=FunctionType(
         [
-            FuncInput(string_type(), InputFlags.NoFlags, "message"),
+            FuncInput(string_type(), InputFlags.NoFlags, "msg"),
             FuncInput(int_type(), InputFlags.NoFlags, "signal"),
         ],
         NoneType(),
--- a/tests/error/misc_errors/panic_msg_not_str.err
+++ b/tests/error/misc_errors/panic_msg_not_str.err
@@ -7,7 +7,7 @@ Error: Invalid call of overloaded function (at $FILE:7:10)
   |                 `()`, `int`
 
 Note: Available overloads are:
-  def panic(message: str, ...) -> None
-  def panic(message: str, signal: int, ...) -> None
+  def panic(msg: str, ...) -> None
+  def panic(msg: str, signal: int, ...) -> None
 
 Guppy compilation failed due to 1 previous error
--- a/tests/integration/test_panic.py
+++ b/tests/integration/test_panic.py
@@ -84,6 +84,8 @@ def test_panic_with_signal(validate):
         panic("I panicked with signal!", 42, q)
 
     validate(main)
+    # The only integer constant in the HUGR should be the given signal (if the panic
+    # checker didn't use it, the default would be 1 instead).
     assert any(
         isinstance(node[1].op, Const)
         and isinstance(node[1].op.val, IntVal)
@@ -98,3 +100,8 @@ def test_panic_with_dynamic_signal(validate):
         panic("I panicked with dynamic signal!", s)
 
     validate(main)
+    # With a dynamic signal there should be no integer constants at all in the HUGR.
+    assert not any(
+        isinstance(node[1].op, Const) and isinstance(node[1].op.val, IntVal)
+        for node in main.modules[0].nodes()
+    )
--- a/tests/error/misc_errors/panic_msg_empty.err
+++ b/tests/error/misc_errors/panic_msg_empty.err
@@ -6,7 +6,7 @@ Error: Invalid call of overloaded function (at $FILE:7:4)
   |     ^^^^^^^ No variant of overloaded function `panic` takes 0 arguments
 
 Note: Available overloads are:
-  def panic(message: str, ...) -> None
-  def panic(message: str, signal: int, ...) -> None
+  def panic(msg: str, ...) -> None
+  def panic(msg: str, signal: int, ...) -> None
 
 Guppy compilation failed due to 1 previous error


guppylang-internals/src/guppylang_internals/std/_internal/checker.py
tatiana-s: Why does `exit` explicitly require a signal? If there isn't a particular reason we could do the same overloading to `exit` and then consequently combine both checkers to one

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -117,6 +117,10 @@ class RawCustomFunctionDef(ParsableDef):
 
     unitary_flags: UnitaryFlags = field(default=UnitaryFlags.NoFlags)
 
+    # Whether the custom function accepts a variable number of arguments (not supported
+    # in Guppy functions in general but some custom functions make use of them).
+    has_var_args: bool = field(default=False)
```
tatiana-s, line 122: I didn't expect this to be breaking (due to `RawWasmFunctionDef` extending `RawCustomFunctionDef`) - not sure how important not breaking vs. better error messages is or whether there is a non-breaking solution

ss2165: internal breaks are fine

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -203,6 +209,7 @@ class CustomFunctionDef(CompiledCallableDef):
     higher_order_value: bool
     higher_order_func_id: GlobalConstId
     has_signature: bool
+    has_var_args: bool = field(default=False)
```
ss2165, line 212: does this mean we just unlocked varargs for all custom functions?

guppylang-internals/src/guppylang_internals/definition/overloaded.py
```diff
@@ -52,12 +53,13 @@ def rendered_span_label(self) -> str:
 @dataclass(frozen=True)
 class AvailableOverloadsHint(Note):
     func_name: str
-    variants: list[FunctionType]
+    variants: list[tuple[FunctionType, bool]]
```
ss2165: suggest NamedTuple - not obvious what the bool is for here

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -306,29 +306,35 @@ class NoMessageError(Error):
...
                 # Check type of message and synthesize types for additional values.
                 msg, _ = ExprChecker(self.ctx).check(msg, string_type())
-                vals = [ExprSynthesizer(self.ctx).synthesize(val)[0] for val in rest]
-                # TODO variable signals once default arguments are available
-                # TODO this will also allow us to remove this manual AST node hack
-                signal_expr = with_type(
-                    int_type(), with_loc(self.node, ast.Constant(value=1))
-                )
+                vals = [ExprSynthesizer(self.ctx).synthesize(val) for val in rest]
+                # If the first value after msg is an int, we assume that this is the
```
ss2165, line 420: this is worth adding to the docstring for the function

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
ss2165: this is a good idea

tests/error/misc_errors/panic_msg_not_str.err
```diff
@@ -1,8 +1,13 @@
...
   | 
 5 | @compile_guppy
 6 | def foo(x: int) -> None:
 7 |     panic((), x)
-  |           ^^ Expected expression of type `str`, got `()`
+  |           ^^^^^ No variant of overloaded function `panic` takes arguments
+  |                 `()`, `int`
+
+Note: Available overloads are:
+  def panic(msg: str, ...) -> None
```
ss2165: changing the parameter name to "message" rather than "msg" will provide a more direct error

tests/integration/test_panic.py
```diff
@@ -72,3 +72,13 @@ def main(b: bool, s: str, i: int) -> None:
         exit(s, i + 1)
 
     validate(main)
+
+
+def test_panic_with_signal(validate):
+    @compile_guppy
+    def main(s: int) -> None:
+        q = qubit()
+        panic("I panicked with signal!", 42, q)
```
ss2165, line 84: Can we assert that this hits the signal pathway rather than *args pathway for the number?

guppylang-internals/src/guppylang_internals/definition/custom.py
tatiana-s: Not really - we already had varargs for custom functions with custom checkers as custom functions can be any python signature assuming the custom checker handles them. This just flags them for better error reporting

guppylang-internals/src/guppylang_internals/definition/custom.py
maximilianruesch, line 122: This feels like it should be part of the `FunctionType` / signature, since having a variable number of parameters is something inherent to the signature itself.

Is moving this into the signature breaking? Can we do it without allowing varargs in general Guppy functions (should be possible, since the user-entrypoints are fairly small).

guppylang-internals/src/guppylang_internals/definition/overloaded.py
```diff
@@ -26,6 +27,11 @@
 from guppylang_internals.tys.ty import FunctionType, Type
 
 
+class OverloadVariant(NamedTuple):
+    func_ty: FunctionType
+    has_var_args: bool
```
maximilianruesch, line 32: If you move it into the function type, this can be simplified back down / removed.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -400,99 +400,51 @@ def synthesize_array_comprehension(
         return with_loc(compr, array_compr), array_type(elt_ty, size)
 
 
-class PanicChecker(CustomCallChecker):
-    """Call checker for the `panic`  function."""
+class ExitChecker(CustomCallChecker):
```
maximilianruesch: Nit: Could be `AbortChecker` or similar to avoid leaning to one variant / usage of this checker. ExitChecker is also fine though.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -400,99 +400,51 @@ def synthesize_array_comprehension(
...
+    """Call checker for the `panic` and `exit` functions."""
+
+    def __init__(self, exit_kind: ExitKind):
+        self.exit_kind = exit_kind
 
     @dataclass(frozen=True)
     class NoMessageError(Error):
```
maximilianruesch: Is this ever hit? Can it be reproduced in a test? Perhaps discovering that no arguments are given should at this point cause a compiler panic itself, since the overloads should catch this before the synthesize call

tests/integration/test_panic.py
```diff
@@ -72,3 +75,17 @@ def main(b: bool, s: str, i: int) -> None:
...
+        panic("I panicked with dynamic signal!", s)
+
+    validate(main)
+    for node in main.modules[0].nodes():
+        if isinstance(node[1].op, Const) and isinstance(node[1].op.val, IntVal):
+            # Check that we are using custom signals, not the default 1.
+            assert node[1].op.val.v != 1
```
maximilianruesch: Generating assertions like this risks them being silently skipped, especially using the `if`. Instead, you could perhaps search for nodes with the name panic, assert that there are two of them, and then assert they carry custom signals. This gives the test more robustness against internal compiler changes.

guppylang-internals/src/guppylang_internals/definition/custom.py
tatiana-s: I think it would only be an internal break, but still more extensive than the current break. Given only custom functions can have this property at the moment and it's more of a flag for having to deal with this special case from the perspective of the compiler as opposed to being part of the type system, I would prefer to keep it to custom functions for now

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
tatiana-s: I initially named it `ExitChecker` to match with `ExitKind` but didn't notice `PanicExpr` being named the other way so naming them all abort seems sensible (breaking internally but that should be fine)

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
tatiana-s: Overload catches the error so this error is never shown, but it wouldn't work without some error being thrown here as the variable number of args can't be handled by the overloads themselves. It's also a bit awkward since you need a `GuppyError` as opposed to an internal guppy error in order for it to be caught. I've tried to make it a more generic error that can be thrown in these cases but it's still a bit awkward

guppylang-internals/src/guppylang_internals/definition/custom.py
maximilianruesch: I shall make a note to discuss this at some point then.

tests/integration/test_panic.py
```diff
@@ -72,3 +75,26 @@ def main(b: bool, s: str, i: int) -> None:
...
+        q = qubit()
+        panic("I panicked with signal!", 42, q)
+
+    validate(main)
+    assert any(
+        isinstance(node[1].op, Const)
+        and isinstance(node[1].op.val, IntVal)
+        and node[1].op.val.v == 42
+        for node in main.modules[0].nodes()
+    )
```
maximilianruesch, line 94: Why not add this to the other test as well, at least to assert that it is not the default value (since you cannot know the runtime value)? This would explicitly test that even just a single extra argument still picks the signal pathway.

guppylang/src/guppylang/std/platform.py
```diff
@@ -83,31 +85,109 @@ def result(tag: str, value):
...
+        ],
+        NoneType(),
+    ),
+    has_var_args=True,
+)
+def _panic_with_signal(msg: str, signal: int, *args) -> None: ...
+
+
+@guppy.overload(_panic, _panic_with_signal)
+def panic(msg: str, signal: int = 1, *args):
```
maximilianruesch, line 121: Overlooked this before (my bad), but is it okay that the function argument `msg` mismatches in name with the argument provided in the function type (which is `message`)? Renaming this outer argument would be a breaking change...

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
maximilianruesch: Hmm, that is indeed a bit awkward. I thought that perhaps the overload functionality would catch this before entering this function, but I might be mistaken on when exactly `check` and `synthesize` are used. For now, this should be fine as is.

guppylang/src/guppylang/std/platform.py
tatiana-s: Ah I initially changed it to message after Seyon's suggestion before realising that would be breaking and changing it back except here, but I think it's better for them to match

PR #1463, opened 2026-01-28
chore: upgrade ruff and mypy and fix up

ss2165:
Some big upgrades, so some big diffs

Mostly automated. Major manual change is adding init methods to ast nodes for mypy's satisfaction.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/nodes.py
+++ b/guppylang-internals/src/guppylang_internals/nodes.py
@@ -35,6 +35,10 @@ class PlaceNode(ast.expr):
         super().__init__()
         self.place = place
 
+    # See MakeIter for explanation
+    __reduce__ = object.__reduce__
+    __reduce_ex__ = object.__reduce_ex__
+
 
 class GlobalName(ast.Name):
     id: str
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -598,7 +598,7 @@ def visit_Compare(
                     comparators[:-1], node.ops, comparators[1:], strict=True
                 )
             ]
-            conj = ast.BoolOp(op=ast.And(), values=values)
+            conj = ast.BoolOp(op=ast.And(), values=values)  # type: ignore[arg-type]
```
aborgna-q, line 601: Are these required because of `ast` typing issues?

guppylang-internals/src/guppylang_internals/std/_internal/debug.py
```diff
@@ -63,7 +65,7 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
         from guppylang.std.quantum import qubit
 
         assert isinstance(qubit, GuppyDefinition)
-        qubit_ty = cast(TypeDef, qubit.wrapped).check_instantiate([])
+        qubit_ty = cast("TypeDef", qubit.wrapped).check_instantiate([])
```
aborgna-q, line 68: Not sure if applicable here, but would `from __future__ import annotations` let us use the type directly here?

ss2165: not sure, these were auto generated so inclined to leave as-is

PR #1466, opened 2026-02-04
chore(ci): Fix smoketest of building wheels

maximilianruesch:
The existing CI job failed every time there were breaking changes between `guppylang-internals` and `guppylang`, since their wheels were built independently, and subsequently the `guppylang` wheel was smoke-tested without using the `guppylang-internals` wheel. Thus, the latest release was fetched from pypi, which may not feature the latest requirements from the `guppylang` wheel. See e.g. https://github.com/Quantinuum/guppylang/actions/runs/21624672648/job/62321750068.

This PR changes the workflow such that
1. Both wheels are built and smoke tested together
2. The wheels are uploaded independently, and in a separate job

I am unsure how to test the release workflow itself pre-merge, but given that it follows [the recommendation from packaging.python.org](https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/#checking-out-the-project-and-building-distributions) rather closely, I think it should be fine.

> [!IMPORTANT]
> See https://github.com/Quantinuum/guppylang/actions/runs/21668492823?pr=1466 for a successful run of this

### DIFF
--- a/.github/workflows/python-wheels.yml
+++ b/.github/workflows/python-wheels.yml
@@ -14,9 +14,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
-    types:
-      - synchronize
   release:
     types:
       - published


.github/workflows/python-wheels.yml
```diff
@@ -75,27 +54,68 @@ jobs:
...
         run: |
-          echo "Testing the newly built ${{ matrix.target.package }} wheels..."
-          uv run -f dist --with ${{ matrix.target.package }} --refresh-package ${{ matrix.target.package }} --no-project -- python -c "import ${{ matrix.target.lib }}"
+          echo "Testing the newly built wheels..."
+          uv run -f dist --with guppylang --with guppylang-internals --refresh-package guppylang --refresh-package guppylang-internals --no-project -- python -c "import guppylang; import guppylang_internals; print('Smoke test done')"
           uvx twine check --strict dist/*
 
+  publish:
+    name: Publish wheels to PyPI
+    if: (github.event_name == 'release' || github.event_name == 'workflow_dispatch') && github.ref_type == 'tag'
```
aborgna-q, line 87: nit: A workflow_dispatch in a non-release tag will probably fail due to environment protection rules (We only allow specific release tags).

It's an edge-case though.

.github/workflows/python-wheels.yml
```diff
@@ -75,27 +54,68 @@ jobs:
...
+            dist/guppylang-internals-*.tar.gz
+            dist/guppylang-internals-*.whl
+
+      - name: Upload the built guppylang package as an artifact
+        if: ${{ steps.check-tag.outputs.run == 'true' }}
+        uses: actions/upload-artifact@v6
+        with:
+          name: build-guppylang-sdist
+          path: |
+            dist/guppylang-*.tar.gz
```
aborgna-q, line 75: This will upload the `-internals` wheels too, so the release will try to upload both

maximilianruesch: Good catch!

maximilianruesch: Actually, it does not, since the `-internals` wheels use `_` instead of `-`:
image

This was more a bug in the upload for the internals wheels then. I will rerun the wheel build to test this.

maximilianruesch: Works now: https://github.com/Quantinuum/guppylang/actions/runs/21668492823?pr=1466

aborgna-q (approved):

Issue #1443, opened 2026-01-14
Compile time array indexing linearity check for literal indices


ss2165:
> In general it is not possible to check this at runtime, the indices can be arbitrary runtime integers. Linearity cannot be checked through array indexing. In cases like yours where the indices are literal, it is possible, and we could add some special checks for this.  

 _Originally posted by @ss2165 in [#1442](https://github.com/Quantinuum/guppylang/issues/1442#issuecomment-3750791436)_

PR #1469, opened 2026-02-04
feat: Compile time array indexing linearity check for literal indices

rmfisher26:
Addressing  #1443 

**Summary:**
Added compile-time detection for duplicate literal subscripts in function call arguments. The linearity checker now tracks when the same array[literal_index] is used multiple times with non-copyable element types and raises an `AlreadyUsedError` at compile-time.

**Problem**
When the same array element with a literal index is used multiple times in a function call (e.g., `cx(qubits[0], qubits[0]))`, the compiler previously allowed this code to compile but would panic at runtime during simulation. For linear types like qubits, this violates linearity constraints and should be caught at compile-time.

Example that previously compiled but panicked at runtime:

```
@guppy
def main() -> None:
    qubits = array((qubit() for _ in range(2)))
    cx(qubits[0], qubits[0])  # Same qubit used twice - runtime panic!
```

**Core Implementation:**
linearity_checker.py:
- Added `get_literal_index()` helper to extract literal indices from subscript accesses
- Modified `_visit_call_args()` to track and detect duplicate literal subscripts
- Added `is_simple_literal_subscript`  used by `_visit_call_args() ` to check if an argument is a simple literal subscript access

**Tests:**
- Added `duplicate_subscript_literal.py `error test
- Added `duplicate_subscript_owned.py` error test
- Updated `test_array.py` and `test_list.py` to remove lines that now correctly fail at compile-time,
- Updated `test_array.py`  with 4 tests 
   - `test_nested_subscript_different_inner_indices`
   - `test_field_access_after_subscript` 
   - `test_dynamic_index_subscript ` 
   - `test_mixed_static_dynamic_index_subscript`
```
Error: Copy violation (at example.py:12:24)
   | 
10 | def main() -> None:
11 |     qubits = array((qubit() for _ in range(2)))
12 |     cx(qubits[0], qubits[0])
   |                   ^^^^^^^^^ Subscript `qubits[...]` with non-copyable type `qubit`
   |                             cannot be borrowed ...

Note:
   | 
12 |     cx(qubits[0], qubits[0])
   |        --------- Subscript `qubits[...]` already borrowed here
```


tests/integration/test_array.py
```diff
@@ -156,7 +156,7 @@ def foo(q1: qubit, q2: qubit) -> None: ...
     @guppy
     def main(qs: array[qubit, 42] @ owned) -> array[qubit, 42]:
         foo(qs[0], qs[1])
-        foo(qs[0], qs[0])  # Will panic at runtime
+        # Note: foo(qs[0], qs[0]) is now correctly rejected at compile time
```
qartik:
```suggestion
        # Note: `foo(qs[0], qs[0])` is rejected at compile time
        i = 0
        foo(qs[i], qs[i])  # Will panic at runtime
```

tests/integration/test_list.py
```diff
@@ -133,7 +133,7 @@ def foo(q1: qubit, q2: qubit) -> None: ...
     @guppy
     def main(qs: list[qubit] @ owned) -> list[qubit]:
         foo(qs[0], qs[1])
-        foo(qs[0], qs[0])  # Will panic at runtime
+        # Note: foo(qs[0], qs[0]) is now correctly rejected at compile time
```
qartik:
```suggestion
        # Note: `foo(qs[0], qs[0])` is rejected at compile time
        i = 0
        foo(qs[i], qs[i])  # Will panic at runtime
```

tests/error/linear_errors/duplicate_subscript_literal.py
```diff
@@ -0,0 +1,15 @@
...
+from guppylang.std.quantum import qubit, cx
+
+import guppylang
+guppylang.enable_experimental_features()
+
+
+@guppy
+def foo() -> None:
+    external_reg = array((qubit() for _ in range(2)))
+    cx(external_reg[0], external_reg[0])
```
qartik:
```suggestion
    qubits = array((qubit() for _ in range(2)))
    cx(qubits[0], qubits[0])
```

tests/error/linear_errors/duplicate_subscript_borrow.py
qartik, line 1: I believe this file should be called `duplicate_subscript_owned.py` as borrowed is the default behavior.

tests/error/linear_errors/duplicate_subscript_borrow.py
```diff
@@ -0,0 +1,23 @@
+from guppylang.decorator import guppy
+from guppylang.std.builtins import array, owned
+from guppylang.std.quantum import qubit, h
+
+import guppylang
+guppylang.enable_experimental_features()
+
+
+@guppy
+def swap_twice(q1: qubit, q2: qubit @owned) -> tuple[qubit, qubit]:
```
qartik: Why is this function named `swap_twice`?

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -326,7 +326,51 @@ def _visit_call_args(self, func_ty: FunctionType, call: AnyCall) -> None:
...
+            # This avoids false positives for:
+            # - Nested: qs[0][0] vs qs[0][1] (different inner index)
+            # - Fields: ss[0].q1 vs ss[0].q2 (different field)
+            if (
+                isinstance(arg, PlaceNode)
+                and (subscript := contains_subscript(arg.place))
+                and isinstance(subscript.parent, Variable)
+                and subscript == arg.place
+                and (literal_idx := get_literal_index(subscript)) is not None
+            ):
```
qartik: Given the condition is quite complex, it might be helpful to extract it into a helper function such as `is_simple_literal_subscript`.

tests/error/linear_errors/duplicate_subscript_borrow.py
rmfisher26: Sounds good,  I have renamed the file accordingly.

tests/error/linear_errors/duplicate_subscript_borrow.py
rmfisher26: I have renamed this as `mixed_ownership`  to pass the same qubit (borrowed and owned) and applying transformation it.  This demonstrates that it violates linearity where  the same qubit cannot be both borrowed and consumed simultaneously. 

If the name should be modified to something that is clearer to understand, please let me know.

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
rmfisher26: Sounds like a good idea to me. I have added `is_simple_literal_subscript` to `linearity_checker.py `

Before:
```
if (isinstance(arg, PlaceNode)
    and (subscript := contains_subscript(arg.place))
    and isinstance(subscript.parent, Variable)
    and subscript == arg.place
   and (literal_idx := get_literal_index(subscript)) is not None
):
```
After

```
 if result := is_simple_literal_subscript(arg):
       subscript, literal_idx = result
```

tests/error/linear_errors/duplicate_subscript_owned.py
```diff
@@ -0,0 +1,33 @@
...
+@guppy
+def mixed_ownership(q1: qubit, q2: qubit @owned) -> tuple[qubit, qubit]:
+    """Function with mixed ownership: first parameter is borrowed (inout),
+    second parameter is owned (consumed).
+
+    This demonstrates that the same qubit cannot be used with different
+    ownership modes in a single function call.
+    """
+    q1 = h(q1)  # Transform the borrowed qubit
+    return q1, q2  # Return both qubits
```
maximilianruesch:
```suggestion
    q1 = h(q1)
    return q1, q2
```
Several things:
1. The inline comments do not add clarity, since the code is pretty self-explanatory.
2. The bottom of the doc comment describes a function call, even though this function is the callee, not the caller. 
3. The top of the doc comment describes things that are either already documented by the code (i.e. which argument is owned and which is not), or things that are perhaps used in existing code but confusing (i.e. taking ownership "consuming" a qubit; it can be simply returned again as you do here).

I thus propose to just remove all of them.

tests/error/linear_errors/duplicate_subscript_owned.py
```diff
@@ -0,0 +1,33 @@
...
+
+@guppy
+def foo() -> None:
+    qubits = array((qubit() for _ in range(3)))
+
+    # ERROR: Cannot use qubits[0] twice with different ownership modes
+    # - First argument borrows qubits[0] (will be returned)
+    # - Second argument consumes qubits[0] (takes ownership)
+    # This violates linearity: the same qubit cannot be both borrowed
+    # and consumed simultaneously
```
maximilianruesch:
```suggestion
    # Compile time error: The same qubit cannot be used with different
    # ownership modes simultaneously due to required linearity.
```
More succinct, less confusing (see the comment above for "returning" and "ownership"), and less maintenance overhead.

tests/error/linear_errors/duplicate_subscript_literal.py
```diff
@@ -0,0 +1,15 @@
+from guppylang.decorator import guppy
+from guppylang.std.builtins import array
+from guppylang.std.quantum import qubit, cx
+
+import guppylang
+guppylang.enable_experimental_features()
```
maximilianruesch: These should not be required in the tests (for this file, and for the next). Update the snapshots accordingly.

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -710,6 +744,62 @@ def has_explicit_copy(ty: Type) -> bool:
...
+    - There are no projections after the subscript (e.g., field accesses)
+    - The index is a literal constant (not a variable)
+
+    Returns a tuple of (subscript, literal_index) if all conditions are met,
+    otherwise None.
+
+    This conservative check avoids false positives for:
+    - Nested subscripts: qs[0][0] vs qs[0][1] (different inner indices)
+    - Field accesses: ss[0].q1 vs ss[0].q2 (different fields)
+    - Dynamic indices: array[i] where i is a variable
```
maximilianruesch, line 770: Can you add tests for these please?

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -326,7 +326,41 @@ def _visit_call_args(self, func_ty: FunctionType, call: AnyCall) -> None:
...
+                            AlreadyUsedError.PrevUse(prev_arg, prev_use_kind)
+                        )
+                        if has_explicit_copy(subscript.ty):
+                            err.add_sub_diagnostic(AlreadyUsedError.MakeCopy(None))
+                        raise GuppyError(err)
+                else:
+                    # Track this literal subscript
+                    literal_subscripts[key] = (arg, use_kind)
+
+            # Visit the argument normally (existing behavior)
```
maximilianruesch:
```suggestion
            # Check for duplicate literal subscripts BEFORE visiting
            if result := is_simple_literal_subscript(arg):
                subscript, literal_idx = result
                key = (subscript.parent.id, literal_idx)
                use_kind = (
                    UseKind.BORROW if InputFlags.Inout in inp.flags else UseKind.CONSUME
                )

                if key in literal_subscripts and not subscript.ty.copyable:
                    prev_arg, prev_use_kind = literal_subscripts[key]
                    err = AlreadyUsedError(arg, subscript, use_kind)
                    err.add_sub_diagnostic(
                       AlreadyUsedError.PrevUse(prev_arg, prev_use_kind)
                    )
                    if has_explicit_copy(subscript.ty):
                       err.add_sub_diagnostic(AlreadyUsedError.MakeCopy(None))
                    raise GuppyError(err)
                
                literal_subscripts[key] = (arg, use_kind)

            # Visit the argument normally
```
Compression, throwing out comments that are not needed due to self-documenting code and perhaps confusing (e.g. why annotate something as "existing behaviour", when as soon as this PR is merged the entire function becomes existing behaviour?).

tests/error/linear_errors/duplicate_subscript_literal.err
```diff
@@ -0,0 +1,15 @@
+Error: Copy violation (at $FILE:12:18)
+   | 
+10 | def foo() -> None:
+11 |     qubits = array((qubit() for _ in range(2)))
+12 |     cx(qubits[0], qubits[0])
+   |                   ^^^^^^^^^ Subscript `qubits[...]` with non-copyable type `qubit`
+   |                             cannot be borrowed ...
```
maximilianruesch: Note: This is already a really helpful diagnostic. When https://github.com/Quantinuum/guppylang/pull/1448#discussion_r2720537995 is addressed, this should even automatically annotate the exact subscript that cannot be borrowed (even though that subscript is visible directly above the annotation).

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
rmfisher26: Sounds good, I will address that. Thanks for your feedback to improve this.

tests/error/linear_errors/duplicate_subscript_literal.py
rmfisher26: Got it. I have removed those imports and regenerated the snapshots as well.

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
rmfisher26: I have added 3 tests to test those scenarios:

`test_nested_subscript_different_inner_indices` — Verifies that `qs[0][0]` and `qs[0][1]` (same outer index, different inner index) are not flagged as duplicate subscripts.

`test_field_access_after_subscript `— Verifies that `ss[0].q1` and `ss[0].q2` (same subscript index, different struct fields) are not flagged as duplicate subscripts.

`test_dynamic_index_subscript `— Verifies that `qs[i]` and `qs[j] `(variable indices) are not flagged as duplicate subscripts, since the checker can't determine equality at compile time.

tests/integration/test_array.py
```diff
@@ -685,6 +687,40 @@ def main() -> int:
     validate(main.compile())
 
 
+def test_nested_subscript_different_inner_indices(validate):
```
maximilianruesch, line 690: The three tests added here should suffice (perhaps add another test where one of the indices is statically known, and one is dynamic), but from the context of the file it is not clear what they are trying to test (since the file contains a wide variety of tests). The easiest resolution would be to add short doc strings to the tests of the form e.g.:
```python
"""Smoketest for checking duplicate subscript accesses:
Asserts that nested subscript accesses with duplicate outer subscripts
are still allowed."""
```
with a similar body for the other tests and formatting to fit on as little lines as possible.

maximilianruesch (changes requested):
One minor change left. Looks great otherwise!

tests/integration/test_array.py
rmfisher26: Of course! I have added a fourth test `test_mixed_static_dynamic_index_subscript ` and added comments on each of the new tests with similar explanations.

maximilianruesch (approved):

PR #1470, opened 2026-02-05
refactor!: refactor: unify error messages for recursive definitions in enums and structs, added two error tests

nicolaassolini-qntm:
BREAKING CHANGES
removed default value for `class_type`, field of `DuplicateFieldError` in `guppylang_internals/definition/util.py`

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/struct.py
+++ b/guppylang-internals/src/guppylang_internals/definition/struct.py
@@ -245,7 +245,7 @@ def check_not_recursive(defn: ParsedStructDef, ctx: TypeParsingCtx) -> None:
         args: Sequence[Argument],
         loc: AstNode | None = None,
     ) -> Type:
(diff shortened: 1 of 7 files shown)


guppylang-internals/src/guppylang_internals/definition/enum.py
```diff
@@ -288,7 +287,7 @@ def dummy_check_instantiate(
         loc: AstNode | None = None,
     ) -> Type:
         """Dummy method that raises an error if called during type parsing."""
-        raise GuppyError(UnsupportedError(loc, "Recursive enums"))
+        raise GuppyError(UnsupportedError(loc, "Recursive definition"))
```
tatiana-s:
```suggestion
        raise GuppyError(UnsupportedError(loc, "Recursive definitions"))
```

Should be plural, same thing in the error in `struct.py`

Issue #1475, opened 2026-02-05
[Misc]: Updated labels


aborgna-q:
We updated the labels in [hugr](https://github.com/Quantinuum/hugr/labels) an [tket](https://github.com/Quantinuum/tket2/labels) to a more organised set that should help with triaging in the new project flow.

I'd like to update guppy's labels too following a similar schema.

Current labels:
```bash
$ gh label list --limit 100

NAME                  DESCRIPTION                                            COLOR  
bug                   Something isn't working                                #d73a4a
documentation         Improvements or additions to documentation             #0075ca
duplicate             This issue or pull request already exists              #cfd3d7
enhancement           New feature or request                                 #a2eeef
good first issue      Good for newcomers                                     #7057ff
invalid               This doesn't seem right                                #e4e669
question              Further information is requested                       #d876e3
wontfix               This will not be worked on                             #ffffff
autorelease: pending                                                         #ededed
autorelease: tagged                                                          #ededed
dependencies          Pull requests that update a dependency file            #0366d6
example                                                                      #74D61C
github_actions        Pull requests that update GitHub Actions code          #000000
python                Pull requests that update Python code                  #2b67c6
breaking-change       Changes that require semver bumps                      #B60205
ci                    CI / devops                                            #FBCA04
tracking-issue        An issue tracking progress on multiple sub-tasks.      #A71FA6
stdlib                                                                       #407408
perf                  Performance issue                                      #5319E7
docs-fail             Automated issue created when the docs workflow fails   #B60205
test/qa               Issues related to testing or QA                        #FEF2C0
wait to merge         This PR to be merged after all dependencies are ready  #1d76db
comptime              Issues to do with guppy comptime mode                  #611482
unitaryhack2025                                                              #5319E7
python:uv             Pull requests that update python:uv code               #2b67c6
```

Proposed:

```bash
NAME                  DESCRIPTION                                                                                    COLOR   
A-checker             Area: Type checker                                                                             #fbca04
A-ci                  Area: CI                                                                                       #fbca04
A-compiler            Area: Compiler                                                                                 #fbca04
A-documentation       Area: Docs                                                                                     #fbca04
A-emulator            Area: Emulator                                                                                 #fbca04
A-examples            Area: Examples                                                                                 #fbca04
A-python              Area: Python project                                                                           #fbca04
A-rust                Area: Rust project                                                                             #fbca04
A-stdlib              Area: Standard library                                                                         #fbca04
[...]

PR #1474, opened 2026-02-05
chore: Add issue templates

aborgna-q:
Adds issue templates with structured input forms. See them working on the [hugr repo](https://github.com/Quantinuum/hugr/issues/new/choose) (select one option to get the forms).

This should help with triaging of issues by ensuring we have some initial information and context.

The templates automatically adds labels from the updated set (see https://github.com/Quantinuum/guppylang/issues/1475).

PR #1479, opened 2026-02-06
chore: Update label names in configs

aborgna-q:
Update config files to follow https://github.com/Quantinuum/guppylang/issues/1475


release-please-config.json
```diff
@@ -38,4 +39,4 @@
       ]
     }
   }
-}
+}
```
maximilianruesch:
```suggestion
}

```

maximilianruesch (approved):
Before merging this PR (even perhaps before merging #1474 but ah well), the label names should be actually updated in this repo.

release-please-config.json
aborgna-q: My auto-formatter doesn't like having newlines at the end of `.json`s :P

maximilianruesch: Oop, will remove it again.

PR #1486, opened 2026-02-09
feat: Provide a more precise error message when missed return statement is inside branch

### LINKED ISSUE #1348: Confusing error message for missing return statement
```python
def foo(b: bool) -> int:
    if b:
        return 0
```

The error message just says "Expected return statement", but the span points to the existing return. A better message would be "Expected return statement in case where this expression is false" (pointing to the if condition)

nicolaassolini-qntm:
* The `build` function in `builder.py` now locates the specific statement and branch condition where a return statement is missing
* The `ExpectedError` class now includes a `Branch` note
* New test cases are added to ensure correct error reporting for missing returns in various branching scenarios

close #1348

### DIFF
--- a/guppylang-internals/src/guppylang_internals/cfg/builder.py
+++ b/guppylang-internals/src/guppylang_internals/cfg/builder.py
@@ -113,13 +113,13 @@ class CFGBuilder(AstVisitor[BB | None]):
         # If we're still in a basic block after compiling the whole body, we have to add
         # an implicit void return
         if final_bb is not None:
-            self.cfg.link(final_bb, self.cfg.exit_bb)
             if final_bb.reachable:
                 self.cfg.exit_bb.reachable = True
                 if not returns_none:
-                    if len(self.cfg.exit_bb.predecessors) <= 1:
-                        # If <= 1, the missing return is not in some branches,
-                        # thus we can point to the last statement of the function
+                    if len(self.cfg.exit_bb.predecessors) == 0:
+                        # If the exit BB has no predecessor, then the function contains
+                        # no return statements, so we can point to the last statement
+                        # in the function
                         err = ExpectedError(nodes[-1], "return statement")
                     else:
                         # otherwise, the missing return is in some branches,
@@ -138,6 +138,7 @@ class CFGBuilder(AstVisitor[BB | None]):
                             expr, idx = branch_cond
                             err.add_sub_diagnostic(Branch(expr, idx == 1))
                     raise GuppyError(err)
+            self.cfg.link(final_bb, self.cfg.exit_bb)
 
         # Prune the CFG such that there are no jumps from unreachable code back into
         # reachable code. Otherwise, unreachable code could lead to unnecessary type
@@ -780,10 +781,27 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
 
 
 def find_missing_return_point(
-    final_bb: BB, cfg: CFG, visited: set[BB] | None = None
+    final_bb: BB, cfg: CFG
 ) -> tuple[BBStatement | None, tuple[ast.expr, int] | None]:
-    """Finds the last statement from the final BB or its predecessors,
-    and the branch condition."""
+    """Finds the last statement from the final BB or its predecessors, and the branch
+    condition. It walks up the ancestors of the given final basic block to locate the
+    nearest block containing statements, and attempts to find the most relevant branch
+    condition that distinguishes the path without a return.
+
+    Returns:
+        A tuple where
+            - The first element is the last statement found in the nearest ancestor
+            - The second element is a tuple containing:
+                - The AST expression representing the branch condition that leads to
+                the missing return.
+                - 0 or 1 indicating which branch leads to the missing return, 0 means
+                the missing return is on the "false" branch, 1 means it is on the "true"
+                branch.
+
+    Returns None in the first element if no statement is found in any ancestor block.
+    Returns None in the second element if no relevant branch condition is found.
+
+    """
     final_statement = None
 
     # walk up the ancestors in the CFG
@@ -799,33 +817,38 @@ def find_missing_return_point(
             # of the branch without return.
             # However, there may be nested branches without returns.
             # We need to find which is the most significant condition.
-            # The heuristic (inspired by Rust error messages) finds
-            # the closest branch condition that distinguishes between
+            # The heuristic (inspired by Rust error messages, e.g
+            # https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=6b3fdc70d03ced70d3ce793b2be1c1a7)
+            # finds the closest branch condition that distinguishes between
             # the statement block and another branch.
             for cond_ancestor in itertools.islice(cfg.ancestors(fbb_ancestor), 1, None):
                 if cond_ancestor.branch_pred is not None:
-                    # Not the case now, but it may happen with the match case statement
-                    if len(cond_ancestor.successors) != 2:
-                        break
-                    # check if the final statement is in the left
-                    # or right branch of the condition and not in both
-                    in_false_branch = fbb_ancestor in set(
-                        cfg.successors(cond_ancestor.successors[0])
-                    )
-                    in_true_branch = fbb_ancestor in set(
-                        cfg.successors(cond_ancestor.successors[1])
-                    )
-                    if in_false_branch and not in_true_branch:
-                        return final_statement, (cond_ancestor.branch_pred, 0)
-                    elif in_true_branch and not in_false_branch:
-                        return final_statement, (cond_ancestor.branch_pred, 1)
+                    # Check in which branch the final statement is
+                    in_branches = [
+                        i
+                        for i, cond_succ in enumerate(cond_ancestor.successors)
+                        if any(
+                            succ == fbb_ancestor for succ in cfg.successors(cond_succ)
+                        )
+                    ]
+                    # Only return if the branch is unique
+                    if len(in_branches) == 1:
+                        return final_statement, (
+                            cond_ancestor.branch_pred,
+                            # Give us the index of the branch, so we can point
+                            # to the correct condition in the error message
+                            in_branches[0],
+                        )
             return final_statement, None
         if fbb_ancestor.branch_pred is not None:
             # We have found a branch condition before finding any statement,
-            # this happen with return inside loops.
+            # this happens with return inside loops.
             # Best solution here is give up on finding the missing return point,
             # considering node[-1] as error point
             # together with the help on the branch condition.
+            # Since the return statement is inside a loop, we suggest to add a return
+            # when the loop condition is false, thus we point to the false branch
+            # (returning 0)
             return None, (fbb_ancestor.branch_pred, 0)
 
     return None, None


tests/error/misc_errors/no_return_2.err
```diff
@@ -0,0 +1,10 @@
+Error: Expected return statement (at $FILE:6:4)
+  | 
+4 | @guppy
+5 | def test(b: bool) -> int:
+6 |     if b:
+  |     ^^^^^
+7 |         x = 2
+  | ^^^^^^^^^^^^^ Expected return statement
```
mark-koch, line 8: Why is the note missing in this case?

tests/error/misc_errors/no_return_2.err
mark-koch, line 1: These tests don't belong into `tracing_errors`, those are only for `comptime` functions. Could you move them somewhere else?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +770,52 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
+    if bb in visited:
+        return None
+    visited.add(bb)
+
+    if bb.statements:
+        return bb.statements[-1]
+
+    cfg = bb.containing_cfg
+    if bb == cfg.entry_bb:
+        return None
```
mark-koch: I don't think this check is necessary?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +770,52 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
     return with_loc(value, ast.Assign(targets=[target], value=value))  # type: ignore[list-item]
+
+
+def find_last_statement(bb: BB, visited: set[BB] | None = None) -> BBStatement | None:
+    """Finds the last statement from the final BB or its predecessors recursively."""
+    if visited is None:
+        visited = set()
+
+    if bb in visited:
+        return None
```
mark-koch: Why do we return `None` in case of cycles?

Can you add a test that involves a loop?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +770,52 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
             ast.Tuple(elts=lhs, ctx=ast.Store()),  # type: ignore[arg-type]
         )
     return with_loc(value, ast.Assign(targets=[target], value=value))  # type: ignore[list-item]
+
+
+def find_last_statement(bb: BB, visited: set[BB] | None = None) -> BBStatement | None:
+    """Finds the last statement from the final BB or its predecessors recursively."""
```
mark-koch: What does it mean if `None` is returned? Can you add this to the docstring?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +770,52 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
+            return result
+
+    return None
+
+
+def find_condition(
+    bb: BB, visited: set[BB] | None = None
+) -> tuple[ast.expr, int] | None:
+    """Finds the condition at the top of the wrong branches.
+    To be consistent in case of multiple missing returns, we use a DFS"""
```
mark-koch: Can't you extract the condition in `find_last_statement` to avoid doing DFS twice?

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
```diff
@@ -44,6 +44,13 @@ class ExpectedError(Error):
     def extra(self) -> str:
         return f", got {self.got}" if self.got else ""
 
+    @dataclass(frozen=True)
+    class Branch(Note):
+        span_label: ClassVar[str] = (
+            "Consider adding a return statement if this expression is `{truth_value}`"
+        )
+        truth_value: bool
+
```
mark-koch: I feel this should not be a nested class since it has nothing to do with a generic `ExpectedError`. Maybe move it to `builder.py`?

guppylang-internals/src/guppylang_internals/error.py
```diff
@@ -46,7 +46,7 @@ def exception_hook(hook: ExceptHook) -> Iterator[None]:
     try:
         # Check if we're inside a jupyter notebook since it uses its own exception
         # hook. If we're in a regular interpreter, this line will raise a `NameError`
-        ipython_shell = get_ipython()  # type: ignore[name-defined]
```
mark-koch, line 49:
```suggestion
```

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +770,52 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
             ast.Tuple(elts=lhs, ctx=ast.Store()),  # type: ignore[arg-type]
         )
     return with_loc(value, ast.Assign(targets=[target], value=value))  # type: ignore[list-item]
+
+
+def find_last_statement(bb: BB, visited: set[BB] | None = None) -> BBStatement | None:
```
mark-koch: Consider using `CFG.ancestors` here instead of rolling your own traversal

tests/error/misc_errors/no_return_2.err
nicolaassolini-qntm: Because we are missing the return in the whole function, not only one specific branch

guppylang-internals/src/guppylang_internals/cfg/builder.py
nicolaassolini-qntm: That's quite involving. Since I return when I find the first result, if I see a node twice, it means that it was None in the first place. This is done to avoid looping. When I find an already seen node, I can stop and ignore the result. Btw, adding a loop brings an unrelated funny little bug, I may refactor this function a bit.

guppylang-internals/src/guppylang_internals/cfg/builder.py
nicolaassolini-qntm: You're right

guppylang-internals/src/guppylang_internals/cfg/builder.py
nicolaassolini-qntm: Now, I cant, but, as said above, I have to refactor, I'll try to do all in one visit.

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +777,55 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
+            for cond_ancestor in itertools.islice(cfg.ancestors(fbb_ancestor), 1, None):
+                if cond_ancestor.branch_pred is not None:
+                    # Not the case now, but it may happen with the match case statement
+                    if len(cond_ancestor.successors) != 2:
+                        break
+                    # check if the final statement is in the left
+                    # or right branch of the condition and not in both
+                    in_false_branch = fbb_ancestor in set(
+                        cfg.successors(cond_ancestor.successors[0])
+                    )
```
mark-koch:
```suggestion
                    in_false_branch = any(
                        succ == fbb_ancestor
                        for succ in cfg.successors(cond_ancestor.successors[0])
                    )
```
No need to collect them in a set, `any` will stop early if it's found

Same for `in_true_branch` below

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +777,55 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
+            # we can stop the search and look for the condition in the branch
+            # that leads to final_bb.
+            final_statement = fbb_ancestor.statements[-1]
+            # To have a better error message, we also look for the condition
+            # of the branch without return.
+            # However, there may be nested branches without returns.
+            # We need to find which is the most significant condition.
+            # The heuristic (inspired by Rust error messages) finds
+            # the closest branch condition that distinguishes between
+            # the statement block and another branch.
```
mark-koch: Would it be better to pick the branch that distinguishes between this block and one that leads to a return?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -109,7 +117,27 @@ def build(
             if final_bb.reachable:
                 self.cfg.exit_bb.reachable = True
                 if not returns_none:
-                    raise GuppyError(ExpectedError(nodes[-1], "return statement"))
+                    if len(self.cfg.exit_bb.predecessors) <= 1:
+                        # If <= 1, the missing return is not in some branches,
+                        # thus we can point to the last statement of the function
```
mark-koch: The `<= 1` threw me off, I was expecting `== 0`.

But I see it's because we just added a link from `final_bb` to `self.cfg.exit_bb` a few lines above. Would be good to explain that in the comment. Or maybe even better: Only add the link afterwards if the error is not thrown and check for 0 here?

mark-koch: In fact, the better comment would be something like: "If the exit BB has no predecessor, then the function contains no return statements, so we can point to the last statement in the function"?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +777,55 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
+        if fbb_ancestor.statements:
+            # We have found the nearest block with statements,
+            # we can stop the search and look for the condition in the branch
+            # that leads to final_bb.
+            final_statement = fbb_ancestor.statements[-1]
+            # To have a better error message, we also look for the condition
+            # of the branch without return.
+            # However, there may be nested branches without returns.
+            # We need to find which is the most significant condition.
+            # The heuristic (inspired by Rust error messages) finds
```
mark-koch: Is Rust really doing is? If there is a reference for this, might be good to link it here?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +777,55 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
+            # the closest branch condition that distinguishes between
+            # the statement block and another branch.
+            for cond_ancestor in itertools.islice(cfg.ancestors(fbb_ancestor), 1, None):
+                if cond_ancestor.branch_pred is not None:
+                    # Not the case now, but it may happen with the match case statement
+                    if len(cond_ancestor.successors) != 2:
+                        break
```
mark-koch: Have you considered already generalising to the >2 case? Might even make the code more readable:

```python
# Check in which branch the final statement is
in_branches = [
    i
    for i, cond_succ in enumerate(cond_ancestor.successors)
    if any(succ == fbb_ancestor for succ in cfg.successors(cond_succ))
]
# Only return if the branch is unique
if len(in_branches) == 1:
    return final_statement, (cond_ancestor.branch_pred, in_branches[0])
```

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +777,55 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
+                    elif in_true_branch and not in_false_branch:
+                        return final_statement, (cond_ancestor.branch_pred, 1)
+            return final_statement, None
+        if fbb_ancestor.branch_pred is not None:
+            # We have found a branch condition before finding any statement,
+            # this happen with return inside loops.
+            # Best solution here is give up on finding the missing return point,
+            # considering node[-1] as error point
+            # together with the help on the branch condition.
+            return None, (fbb_ancestor.branch_pred, 0)
```
mark-koch, line 852: Why is it always 0 here?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +777,55 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
...
+                        cfg.successors(cond_ancestor.successors[1])
+                    )
+                    if in_false_branch and not in_true_branch:
+                        return final_statement, (cond_ancestor.branch_pred, 0)
+                    elif in_true_branch and not in_false_branch:
+                        return final_statement, (cond_ancestor.branch_pred, 1)
+            return final_statement, None
+        if fbb_ancestor.branch_pred is not None:
+            # We have found a branch condition before finding any statement,
+            # this happen with return inside loops.
```
mark-koch:
```suggestion
            # this happens with return inside loops.
```

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -749,3 +777,55 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
             ast.Tuple(elts=lhs, ctx=ast.Store()),  # type: ignore[arg-type]
         )
     return with_loc(value, ast.Assign(targets=[target], value=value))  # type: ignore[list-item]
+
+
+def find_missing_return_point(
+    final_bb: BB, cfg: CFG, visited: set[BB] | None = None
+) -> tuple[BBStatement | None, tuple[ast.expr, int] | None]:
+    """Finds the last statement from the final BB or its predecessors,
+    and the branch condition."""
```
mark-koch: Can you expand this docstring a bit?

When is `None` returned, respectively? What does the `int` mean?

mark-koch (commented):
Thanks, the new algorithm is looking a lot better! Also sorry for the long delay in getting back to you 😅

guppylang-internals/src/guppylang_internals/cfg/builder.py
nicolaassolini-qntm: Because if the return is inside the loop, the best suggestion we suggest to add a return when the loop condition is false. Added in the comment explanation.

guppylang-internals/src/guppylang_internals/cfg/builder.py
nicolaassolini-qntm: makes sense

PR #1489, opened 2026-02-11
refactor!: Experimental features should only be set globally

maximilianruesch:
Currently, experimental features are enabled and disabled via classes that simultaneously act as a function to enable/disable experimental features globally, or as a context manager that may enable experimental features within a given code range. This double usage, although with a somewhat clean usage API, already was a source of confusion / missing clarity in the code / suspected bugs (see #1321).

Depending on the range of experimental features added to Guppy, parts of code compiled against experimental APIs may break with parts of code compiled against non-experimental API. A context-manager API promotes usage of experimental features in the above way.

This PR changes the API of experimental features such that they are only usable as a function to globally set / unset experimental features. A context-manager-like behaviour can still be mimicked using `is_` and `set_` functions, but that is not considered intended usage anymore and to be used at own risk.

BREAKING CHANGE: Enabling / disabling experimental features is no longer possible via context managers. Features should either be enabled or disabled globally.

BEGIN_COMMIT_OVERRIDE
""
END_COMMIT_OVERRIDE

### DIFF
--- a/guppylang-internals/src/guppylang_internals/experimental.py
+++ b/guppylang-internals/src/guppylang_internals/experimental.py
@@ -1,6 +1,5 @@
 from ast import expr
 from dataclasses import dataclass
-from types import TracebackType
 from typing import ClassVar
 
 from guppylang_internals.ast_util import AstNode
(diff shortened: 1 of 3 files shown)


guppylang-internals/src/guppylang_internals/experimental.py
```diff
@@ -11,52 +10,25 @@
 EXPERIMENTAL_FEATURES_ENABLED = False
```
croyzor, line 10: I think it'd be nice to make this a property of the `CompilationEngine`, what do you think?

maximilianruesch: I do not necessarily see the need / benefit. Either it is a global variable, or a member of another global variable. The user should set the flag via functions in this file, and internal code asking for the flag value should do so via the remaining functions provided in this file to ensure consistent error messages / handling. At least that is the current standard, and I am not opposed to just keeping it.

PR #1494, opened 2026-02-12
feat!: Allow specifying link names for lowered functions and declarations

### LINKED ISSUE #1483: [Feature]: Annotating fully qualified names of HUGR functions in Guppy
### Problem Statement

Guppy functions should be able to annotate the (fully qualified) HUGR function name that they lower to. Furthermore, functions not explicitly annotated with such a name should auto-infer their lowered name based on their surrounding module, and their Python qualified name.

> [!NOTE]
> This is a preparatory task for the introduction of Guppy libraries.

### Proposed Solution

Similar to #1482, the `@guppy` decorator should receive a new keyword argument that allows explicitly specifying the HUGR name of the function, e.g. `@guppy(..., hugr_name="mod1.mod2.superb-function-1")`. Omission of this annotation leads to auto-derivation of the function name based on the result of `f"{f.__module__}.{f.__qualname__}"`, or similar qualification based on the containing struct name.

### Component

Guppy Compiler

### Effort Estimate

### Additional Context

Upon initial testing, there seems to be a small issue with invoking struct methods. This needs further investigation. See also changes made in https://github.com/Quantinuum/guppylang/pull/1452 for a contrast.

maximilianruesch:
Allows annotating function names for definitions and declarations lowered to HUGR. This also changes the default inferral of names to be fully qualified, instead of only partially qualified as in #1452.

See tests for more info on the details of how qualification works.

Closes #1483 

BREAKING CHANGE: ParsedFunctionDef and CheckedFunctionDecl have obtained a new required parameter that is set during parsing.

BREAKING CHANGE: The inheritance chain for function declarations (RawFunctionDecl, etc ...) was changed, behaving more similar to the inheritance chain for function definitions.

---

BEGIN_COMMIT_OVERRIDE
feat: Allow specifying link names for lowered functions and declarations (#1494)
END_COMMIT_OVERRIDE

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/declaration.py
+++ b/guppylang-internals/src/guppylang_internals/definition/declaration.py
@@ -23,10 +23,10 @@ from guppylang_internals.definition.common import (
 from guppylang_internals.definition.function import (
     PyFunc,
     compile_call,
+    default_func_link_name,
     load_with_args,
     parse_py_func,
 )
-from guppylang_internals.definition.struct import ParsedStructDef
 from guppylang_internals.definition.value import (
     CallableDef,
     CallReturnWires,
@@ -34,7 +34,6 @@ from guppylang_internals.definition.value import (
     CompiledHugrNodeDef,
 )
 from guppylang_internals.diagnostic import Error
-from guppylang_internals.engine import DEF_STORE, ENGINE
 from guppylang_internals.error import GuppyError
 from guppylang_internals.nodes import GlobalCall
 from guppylang_internals.span import SourceMap
@@ -87,13 +86,7 @@ class RawFunctionDecl(ParsableDef, UserProvidedLinkName):
         ty = check_signature(
             func_ast, globals, self.id, unitary_flags=self.unitary_flags
         )
-        link_name = f"{self.python_func.__module__}.{self.python_func.__qualname__}"
-        if self._user_set_link_name is not None:
-            link_name = self._user_set_link_name
-        elif (parent_ty_id := DEF_STORE.impl_parents.get(self.id)) is not None:
-            parent = ENGINE.get_parsed(parent_ty_id)
-            if isinstance(parent, ParsedStructDef):
-                link_name = f"{parent.link_name_prefix}.{self.python_func.__name__}"
+        link_name = self._user_set_link_name or default_func_link_name(self)
 
         if not has_empty_body(func_ast):
             raise GuppyError(BodyNotEmptyError(func_ast.body[0], self.name))
--- a/guppylang-internals/src/guppylang_internals/definition/function.py
+++ b/guppylang-internals/src/guppylang_internals/definition/function.py
@@ -56,9 +56,19 @@ from guppylang_internals.tys.subst import Inst, Subst
 from guppylang_internals.tys.ty import FunctionType, Type, UnitaryFlags, type_to_row
 
 if TYPE_CHECKING:
+    from guppylang_internals.definition.declaration import RawFunctionDecl
     from guppylang_internals.tys.param import Parameter
 
 PyFunc = Callable[..., Any]
+
+
+def default_func_link_name(raw_def: "RawFunctionDef | RawFunctionDecl") -> str:
+    if (parent_ty_id := DEF_STORE.impl_parents.get(raw_def.id)) is not None:
+        parent = ENGINE.get_parsed(parent_ty_id)
+        if isinstance(parent, ParsedStructDef):
+            return f"{parent.link_name_prefix}.{raw_def.python_func.__name__}"
+
+    return f"{raw_def.python_func.__module__}.{raw_def.python_func.__qualname__}"
 
 
 @dataclass(frozen=True)
@@ -93,14 +103,7 @@ class RawFunctionDef(ParsableDef, UserProvidedLinkName):
         ty = check_signature(
             func_ast, globals, self.id, unitary_flags=self.unitary_flags
         )
-
-        link_name = f"{self.python_func.__module__}.{self.python_func.__qualname__}"
-        if self._user_set_link_name is not None:
-            link_name = self._user_set_link_name
-        elif (parent_ty_id := DEF_STORE.impl_parents.get(self.id)) is not None:
-            parent = ENGINE.get_parsed(parent_ty_id)
-            if isinstance(parent, ParsedStructDef):
-                link_name = f"{parent.link_name_prefix}.{self.python_func.__name__}"
+        link_name = self._user_set_link_name or default_func_link_name(self)
 
         return ParsedFunctionDef(
             self.id,


guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -76,6 +84,23 @@ class RawFunctionDef(ParsableDef):
 
     metadata: GuppyMetadata | None = field(default=None, kw_only=True)
 
+    hugr_name: InitVar[str | None] = field(default=None, kw_only=True)
```
acl-cqc: So the fancy business with `InitVar`, `init=False`, and `post_init` is just so we can store the value passed in as `hugr_name="foo"` in a field with a different name (i.e. `_user_set_hugr_name`)?

Yeah, ok, that's nicer than what I'd probably have done (called it `hugr_name` and written a comment)!!

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -76,6 +84,23 @@ class RawFunctionDef(ParsableDef):
...
+    _user_set_hugr_name: str | None = field(default=None, init=False)
+
+    def __post_init__(self, hugr_name: str | None) -> None:
+        object.__setattr__(self, "_user_set_hugr_name", hugr_name)
+
+    @cached_property
+    def qualified_hugr_name(self) -> str:
```
acl-cqc: Nice, I like the use of `cached_property` too

acl-cqc: In hugr the field we are filling in is just `name` so this is ok (`hugr_qualified_name` would not be ok!!). I'd consider calling this just `hugr_name` as that might be more obvious (and avoid question of, is there a method returning the unqualified name), but this looks ok.

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -88,6 +113,7 @@ def parse(self, globals: Globals, sources: SourceMap) -> "ParsedFunctionDef":
             func_ast,
             ty,
             docstring,
+            self.qualified_hugr_name,
```
acl-cqc: It's a lot of parameters, we want might to start using more kwargs at some point

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -187,15 +214,10 @@ def monomorphize(
...
-        if parent_ty is None:
-            hugr_func_name = self.name
-        else:
-            hugr_func_name = f"{parent_ty.name}.{self.name}"
-
         mono_ty = self.ty.instantiate_partial(mono_args)
         hugr_ty = mono_ty.to_hugr_poly(ctx)
         func_def = module.module_root_builder().define_function(
-            hugr_func_name, hugr_ty.body.input, hugr_ty.body.output, hugr_ty.params
+            self.hugr_name, hugr_ty.body.input, hugr_ty.body.output, hugr_ty.params
```
acl-cqc: Time to put in a comment "TODO mangle the name with the monomorphized type arguments"?

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
```diff
@@ -80,6 +80,7 @@ def compile_local_func_def(
             mono_args,
             func.ty,
             None,
+            func.name,
```
acl-cqc, line 85: According to the docstring of `CompiledFunctionDef` (does that need updating?) this new `func.name` is the *docstring* ?!

guppylang-internals/src/guppylang_internals/definition/function.py
acl-cqc: Moreover, the same value is called `hugr_name` in the other classes

guppylang-internals/src/guppylang_internals/ast_util.py
```diff
@@ -395,3 +395,26 @@ def has_empty_body(func_ast: ast.FunctionDef) -> bool:
         and isinstance(n.value, ast.Constant)
         and n.value.value == Ellipsis
     )
+
+
+def parse_source(source_lines: list[str], line_offset: int) -> tuple[str, ast.AST, int]:
```
acl-cqc, line 400: Yeah good moving

guppylang/src/guppylang/decorator.py
```diff
@@ -93,6 +94,15 @@ class GuppyKwargs(TypedDict, total=False):
...
     max_qubits: int
+    hugr_name: str
+
+
+class GuppyStructKwargs(TypedDict, total=False):
+    """Typed dictionary specifying the optional keyword arguments for the
+    `@guppy.struct` decorator.
+    """
+
+    hugr_name: str
```
acl-cqc: No `metadata` or `flags`? Ah - those are not allowed for structs? Yeah, I guess they wouldn't be

guppylang/src/guppylang/decorator.py
```diff
@@ -93,6 +94,15 @@ class GuppyKwargs(TypedDict, total=False):
     dagger: bool
     power: bool
     max_qubits: int
+    hugr_name: str
+
+
+class GuppyStructKwargs(TypedDict, total=False):
```
acl-cqc, line 100: How about renaming `GuppyKwargs` to `GuppyFuncKwargs` to be clear that they are not general across structs too?

guppylang/src/guppylang/decorator.py
```diff
@@ -645,8 +673,14 @@ def _with_optional_kwargs(
...
 
+class ParsedGuppyKwargs(NamedTuple):
+    flags: UnitaryFlags
+    metadata: GuppyMetadata
+    hugr_name: str | None
+
+
 @hide_trace
-def _parse_kwargs(kwargs: GuppyKwargs) -> tuple[UnitaryFlags, GuppyMetadata]:
+def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
```
acl-cqc, line 691: Might be worth renaming to `parse_func_kwargs` (or that might be overkill)

guppylang/src/guppylang/decorator.py
```diff
@@ -645,8 +673,14 @@ def _with_optional_kwargs(
             raise TypeError(err)
 
 
+class ParsedGuppyKwargs(NamedTuple):
```
acl-cqc, line 684: ParsedGuppyFuncKwargs? Hmmm, maybe not, 'tis a bit of a mouthful :(

ParsedFuncKwargs, perhaps?

guppylang/src/guppylang/decorator.py
acl-cqc, line 210: maybe give example of passing parameter? (`@guppy.struct(....) \\ class MyStruct2`)

guppylang/src/guppylang/decorator.py
```diff
@@ -203,20 +216,30 @@ class MyStruct:
...
-        # Prior to Python 3.13, the `__firstlineno__` attribute on classes is not set.
-        # However, we need this information to precisely look up the source for the
-        # class later. If it's not there, we can set it from the calling frame:
-        if not hasattr(cls, "__firstlineno__"):
-            cls.__firstlineno__ = frame.f_lineno  # type: ignore[attr-defined]
-        # We're pretending to return the class unchanged, but in fact we return
-        # a `GuppyDefinition` that handles the comptime logic
-        return GuppyDefinition(defn)  # type: ignore[return-value]
+
+        def dec(cls: builtins.type[T], kwargs: GuppyStructKwargs) -> GuppyDefinition:
```
acl-cqc: Yeah, ok. I guess this amount of boilerplate/faff is unavoidable for a decorator that works both `@struct` and `@struct(....)` (?). It's a pain but I don't know of a better way myself :(

tests/integration/test_hugr_name.py
```diff
@@ -0,0 +1,164 @@
...
+def _func_names(package: Package) -> set[str]:
+    hugr = package.modules[0]
+
+    return {
+        n.op.f_name for n in hugr.values() if isinstance(n.op, (FuncDefn, FuncDecl))
+    }
+
+
+def test_func_hugr_name_annotated():
+    """Asserts that annotated HUGR func names are correctly passed to the HUGR nodes."""
```
acl-cqc:
```suggestion
    """Asserts that annotated function `hugr_name`s are correctly passed to the HUGR nodes."""
```

tests/integration/test_hugr_name.py
```diff
@@ -0,0 +1,164 @@
...
+
+    @guppy
+    def crazy_def() -> None:
+        pass
+
+    @guppy.declare
+    def crazy_dec() -> None: ...
+
+    assert _func_names(crazy_def.compile()) == {
+        "tests.integration.test_hugr_name.test_func_hugr_name_inferred.<locals>.crazy_def"
```
acl-cqc: It would be lovely to remove some of the `tests.integration.test_hugr_name.test_func_hugr_name_inferred.<locals>` stuff from the *test* even if not the hugr_name.

Is there anything we can do like `test_func_hugr_name_inferred.__qual_name__ + ".<locals>.crazy_def"`? Or maybe make a toplevel `file_name = ....` and then use `file_name + "test_func_hugr_name_inferred.<locals>.crazy_def"`?

Or `def some_local(): pass` then sthg like `prefix = some_local.__full_name__.remove_suffix("some_local")` (suspect that no such full-name exists but mentioning just in case it does)

**E.g.** to make the tests less fragile if we rename the file

tests/integration/test_link_name.py
```diff
@@ -0,0 +1,164 @@
+from hugr.ops import FuncDefn, FuncDecl
+from hugr.package import Package
+
+from guppylang import guppy
+
+
+def _func_names(package: Package) -> set[str]:
```
acl-cqc, line 18: Consider defining `func_names_excluding_main` that checks-existence+removes exactly one name ending in `main`

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
acl-cqc: It feels we should pass the hugr-name in somewhere rather than `func.name` a second time, even if only to support 'local aliases'?

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: Yes, it is such that you can supply a hugr name during init, but the only way to read it in the public interface is via the cached_property.

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: There is a problem with clashing names when we have the `hugr_name` init variable. We could consider renaming that to sth like `hugr_name_override`, and then rename this property to `hugr_name`. I would actually like this. What do you think?

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: At some point is a good point. I did have some issues with kwargs and type checkers, so I just avoided them for the moment.

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: I would rather say time to just open an issue. Or both.

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
maximilianruesch: The problem is that we create this freshly here. We would need to replicate the `qualified_hugr_name` functionality for this to be a proper hugr name. Is this what you had in mind?

guppylang-internals/src/guppylang_internals/ast_util.py
maximilianruesch: Was required to prevent a circular import.

guppylang/src/guppylang/decorator.py
maximilianruesch: Nope, this is the first time the kwargs REALLY differ from the ones you would pass to other decorators. I imagine that eventually we will need such a dict for all different decorators.

guppylang/src/guppylang/decorator.py
maximilianruesch: Would be possible. There is an argument as to leaving it as is since the main usage is for the `@guppy` decorator and not the `@guppy.func` decorator.

guppylang/src/guppylang/decorator.py
maximilianruesch: Will keep in sync with the name for `GuppyKwargs`, whatever the decision is there (see other comment).

guppylang/src/guppylang/decorator.py
maximilianruesch: It is simply replicating the same boilerplate from the other decorators, and yes it seems to be rather unavoidable.

tests/integration/test_hugr_name.py
maximilianruesch: For removing fragility it would be sufficient to add some top level `MODULE` string that dynamically determines the module, and prepending that to every expected name.

For shortening the expected strings, we can use the `request` fixture from pytest to react to the test name.

I will cook up a custom fixture for this file 👍

guppylang/src/guppylang/decorator.py
acl-cqc: Agreed there is no guppy.func; but GuppyKwargs feels to me like it could mean soooo many things

tests/integration/test_hugr_name.py
acl-cqc: I think the MODULE would probably have 90% of the benefit and be significantly simpler/more obvious. There is probably something to be said for making the `<locals>` explicit so people are less surprised when they see it in practice...

maximilianruesch: Then I would propose a middle ground `test_qualifier` fixture (which can be defined just as a simple function at the top of the file) that covers everything up to locals, and can be used to test a name as:
```python
<my_expr> == f"{test_qualifier}.<locals>.main"
```

guppylang-internals/src/guppylang_internals/definition/function.py
acl-cqc: I kinda like that, but it is a lot of typing for users, and I think "setting the hugr_name or leaving the default" (rather than "setting an override") is a reasonable mental model for them (i.e. they don't need to be concerned with quite how complex it is to calculate the default....)

I tried:
```
>>> @dataclass
... class Foo:
...     xx: InitVar[str | None] = None
...     def __post_init__(self, xx: str | None):
...         print("Post init called with xx=",xx)
...     @property
...     def xx(self):
...         return "world"
...         
>>> Foo().xx
Post init called with xx= <property object at 0x105e4e020>
'world'
>>> Foo(xx="hello").xx
Post init called with xx= hello
'world'
```
(same with `xx: InitVar[str | None] = field(default = None)`)

Not sure what that `property object` is but I don't see a problem with having two `xx`s...have you tried?

maximilianruesch: I have not actually tried, but some type checker did complain. I will give it a shot and try to identify the root cause 👍

maximilianruesch: We can do this, but we will have to disable ruffs `F811` rule for "redefined while unused" checks, since it does not seem to pick up on initvars yet.

Mypy also complains with "already defined", and later with "has no attribute" when trying to access it. I do not think this pattern is widely compatible with the tools that we use.

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
maximilianruesch: The docstring was indeed outdated, and included the "python_scope` which does not exist anymore.

guppylang/src/guppylang/decorator.py
maximilianruesch: As discussed, I will not apply this change since it would be breaking to the `guppylang` interface.

tests/integration/test_hugr_name.py
acl-cqc: sounds fine although IIUC you'll be defining that once per function....at least MODULE is only once per file

maximilianruesch: The trick is (using the fixture) I do not. See updated code, I can define
```
@pytest.fixture
def qualifier(request) -> str:
    """Provides the common qualifier for functions defined in the current test."""

    def tmp() -> None:
        pass

    return f"{tmp.__module__}.{request.node.originalname}"
```
which processes the test name I receive from `pytest`.

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -217,7 +217,14 @@ def check_nested_func_def(
...
             from guppylang_internals.definition.function import ParsedFunctionDef
 
-            func = ParsedFunctionDef(def_id, func_def.name, func_def, func_ty, None)
+            func = ParsedFunctionDef(
+                def_id,
+                func_def.name,
+                func_def,
+                func_ty,
+                None,
+                hugr_name=func_def.name,
```
acl-cqc: perhaps worth a comment that since nested functions are always private this doesn't really matter

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
acl-cqc: Ok fine. `hugr_name=func.name` might be slightly better, and/or a comment "hugr_name doesn't matter as nested functions are always private"

guppylang-internals/src/guppylang_internals/definition/declaration.py
```diff
@@ -67,12 +69,28 @@ class RawFunctionDecl(ParsableDef):
...
         """Parses and checks the user-provided signature of the function."""
         func_ast, docstring = parse_py_func(self.python_func, sources)
         ty = check_signature(
             func_ast, globals, self.id, unitary_flags=self.unitary_flags
         )
+        hugr_name = f"{self.python_func.__module__}.{self.python_func.__qualname__}"
+        if self._user_set_hugr_name is not None:
+            hugr_name = self._user_set_hugr_name
+        else:
+            parent_ty_id = DEF_STORE.impl_parents.get(self.id)
```
acl-cqc: style: I think `if (parent_ty_id := DEF_STORE.impl_parents.get(self.id)) is not None:` is fine, and allows reducing the indent via `elif`. Whether `elif parent_ty_id := DEF_STORE.impl_parents.get(self.id)` (dropping the explicit `is not None`) is good is a different question...

guppylang-internals/src/guppylang_internals/definition/declaration.py
```diff
@@ -83,20 +101,21 @@ def parse(self, globals: Globals, sources: SourceMap) -> "CheckedFunctionDecl":
...
             ty,
-            self.python_func,
             docstring,
+            hugr_name,
         )
 
 
 @dataclass(frozen=True)
-class CheckedFunctionDecl(RawFunctionDecl, CompilableDef, CallableDef):
+class CheckedFunctionDecl(CompilableDef, CallableDef):
```
acl-cqc, line 107: Dropping the `RawFunctionDecl` as a superclass seems quite a significant change, what happens to fields `python_func`, `description`, `unitary_flags`? Are they multiply-inherited?

guppylang/src/guppylang/decorator.py
```diff
@@ -202,21 +215,37 @@ class MyStruct:
             @guppy
             def add_fields(self: "MyStruct") -> int:
                 return self.field2 + self.field2
+
+            # Add optional parameters
+            @guppy.struct(...)
```
acl-cqc:
```suggestion
            @guppy.struct(hugr_name="my_struct")
```

guppylang-internals/src/guppylang_internals/definition/declaration.py
maximilianruesch: I dropped `python_func` on purpose, since similar to `ParsedFunctionDef` the original function object is not required anymore, and instead its parsed AST version is passed via `defined_at`.

`unitary_flags` is, again similar to `ParsedFunctionDef`, packaged into the function type passed via `ty` (and that is inherited from `CallableDef`).

`description` was not populated by the parsing process, and never used, so I dropped it without replacement.

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -217,7 +217,16 @@ def check_nested_func_def(
...
             from guppylang_internals.definition.function import ParsedFunctionDef
 
-            func = ParsedFunctionDef(def_id, func_def.name, func_def, func_ty, None)
+            func = ParsedFunctionDef(
+                def_id,
+                func_def.name,
+                func_def,
+                func_ty,
+                None,
+                # Even though global, this function will be private to the built hugr,
```
acl-cqc, line 226:
```suggestion
                # Even though lifted to the top level of the hugr, this function will be private,
```
(the key here is that nested in python is not nested in hugr!)

guppylang-internals/src/guppylang_internals/definition/declaration.py
acl-cqc: A significant (driveby?) improvement/cleanup, SGTM

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -76,18 +83,34 @@ class RawFunctionDef(ParsableDef):
 
     metadata: GuppyMetadata | None = field(default=None, kw_only=True)
 
+    link_name: InitVar[str | None] = field(default=None, kw_only=True)
```
acl-cqc: Is it worth commoning these up into a superclass (shared between `RawFunctionDef` and `RawFunctionDecl`, subclassing  ParsableDef)?

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -76,18 +83,34 @@ class RawFunctionDef(ParsableDef):
...
+        object.__setattr__(self, "_user_set_link_name", link_name)
+
     def parse(self, globals: Globals, sources: SourceMap) -> "ParsedFunctionDef":
         """Parses and checks the user-provided signature of the function."""
         func_ast, docstring = parse_py_func(self.python_func, sources)
         ty = check_signature(
             func_ast, globals, self.id, unitary_flags=self.unitary_flags
         )
+
+        link_name = f"{self.python_func.__module__}.{self.python_func.__qualname__}"
```
acl-cqc: if you made a superclass, you could give it a method `calculate_link_name` using early return to avoid this set-default-and-override pattern

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -161,8 +186,8 @@ class CheckedFunctionDef(ParsedFunctionDef, MonomorphizableDef):
         name: The name of the function.
         defined_at: The AST node where the function was defined.
         ty: The type of the function.
-        python_scope: The Python scope where the function was defined.
         docstring: The docstring of the function.
+        link_name: The name that the Hugr node for this function will receive.
```
acl-cqc:
```suggestion
        link_name: The external name for this function (applied to the Hugr node, and other representations, regardless of whether the function is actually visible for linking)
```

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -227,8 +247,8 @@ class CompiledFunctionDef(
         defined_at: The AST node where the function was defined.
         mono_args: Partial monomorphization of the generic type parameters.
         ty: The type of the function after partial monomorphization.
-        python_scope: The Python scope where the function was defined.
         docstring: The docstring of the function.
+        link_name: The name of the Hugr node corresponding to this function.
```
acl-cqc: similarly

guppylang-internals/src/guppylang_internals/definition/struct.py
```diff
@@ -117,6 +116,12 @@ class RawStructDef(TypeDef, ParsableDef):
     python_class: type
     params: None = field(default=None, init=False)  # Params not known yet
 
+    link_name: InitVar[str | None] = field(default=None, kw_only=True)
```
acl-cqc: Maybe a third subclass of `ParsableDef` with these same, is the critical threshold for commoning

guppylang/src/guppylang/decorator.py
```diff
@@ -202,21 +215,37 @@ class MyStruct:
...
-        # Prior to Python 3.13, the `__firstlineno__` attribute on classes is not set.
-        # However, we need this information to precisely look up the source for the
-        # class later. If it's not there, we can set it from the calling frame:
-        if not hasattr(cls, "__firstlineno__"):
-            cls.__firstlineno__ = frame.f_lineno  # type: ignore[attr-defined]
-        # We're pretending to return the class unchanged, but in fact we return
-        # a `GuppyDefinition` that handles the comptime logic
-        return GuppyDefinition(defn)  # type: ignore[return-value]
+
+        def dec(cls: builtins.type[T], kwargs: GuppyStructKwargs) -> GuppyDefinition:
```
acl-cqc: very minor, but can I suggest `deco` or `decor` to avoid possibly misinterpretation as `declare` / `declaration`, etc.

tests/integration/test_basic.py
```diff
@@ -72,7 +72,7 @@ def func_name() -> None: ...
         data.op for n, data in hugr.nodes() if isinstance(data.op, ops.FuncDecl)
     ]
     assert isinstance(def_op, ops.FuncDecl)
-    assert def_op.f_name == "func_name"
+    assert def_op.f_name.endswith("func_name")
```
acl-cqc: would `endswith(".func_name") work or is that too restrictive (as it would not match the string "func_name")

guppylang-internals/src/guppylang_internals/checker/func_checker.py
maximilianruesch: The function where this comment was added to is not nested in Python anymore, since it was elevated to be global (since it captured no local variables).

guppylang/src/guppylang/decorator.py
maximilianruesch: I have applied this to all things in this file now. It has been `dec` for a while, but I do kind of see the point.

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: With the fetching of the struct (which should be very specific and explicit for functions) it is imo rather awkward and not straight forward to raise into a superclass.

acl-cqc: Ah....that does reduce the value of raising the superclass. Hmmm, let me have another look

PR #1441, opened 2026-01-14
feat!: Monomorphise everything during type checking 

mark-koch:
This PR updates to compiler to monomorphize every generic definition. The resulting Hugr will contain no more generic functions.

Monomorphization happens during type checking, so every version is type checked separately.

This is a precursor for #1299.

Also closes #1032 as a drive-by.

In particular, note the breaking internal change that `guppylang_internals.tys.subst.Inst` is now a `tuple` instead of a `Sequence`.

BREAKING CHANGE: Removed `MonomorphizableDef` and `MonomorphizedDef`. Use `GenericCheckableDef` instead.
BREAKING CHANGE: `CompiledCallableDef.compile_call` no longer takes type args
BREAKING CHANGE: Removed `CompiledCallableDef.load_with_args`. Use `CompiledCallableDef.load` instead.
BREAKING CHANGE: Removed `ParameterBase.to_hugr` since we no longer need to be able to turn params into Hugr params
BREAKING CHANGE: Removed `PartiallyMonomorphizedArgs, partially_monomorphize_args`, `require_monomorphization`, `compile_variable_idx` `MonoArgs`, `CompilerContext.{current_mono_args, set_monomorphized_args, type_var_to_hugr, const_var_to_hugr}`, `instantiation_needs_unpacking`. 
BREAKING CHANGE: `Context.generic_params` renamed to `Context.generic_inst`, now storing an instantiation instead of parameters
BREAKING CHANGE: Removed `NoneType.preserve` and `TupleType.preserve` fields
BREAKING CHANGE: `Inst` is now a `tuple` instead of a `Sequence`
BREAKING CHANGE: `CompilerContext.compile` now requires type arguments
BREAKING CHANGE: `CompilerContext.build_compiled_def` no longer returns the remaining type args
BREAKING CHANGE: `CustomFunctionDef` is no longer a `CompiledCallableDef`. Added a new `CustomMonoFunctionDef` that is the result of checking a `CustomFunctionDef` for some concrete type args
BREAKING CHANGE: `RawFunctionDecl.parse` now returns an intermediate `ParsedFunctionDecl` that is turned into a monomorphic `CheckedFunctionDecl` after checking

BEGIN_COMMIT_OVERRIDE
feat: Monomorphise everything during type checking
END_COMMIT_OVERRIDE

PR #1495, opened 2026-02-12
feat!: Joint compilation of Guppy symbols to a library

### LINKED ISSUE #960: Add guppy libraries to enforce joint compilation of definitions

maximilianruesch:
Allows collecting symbols annotated with `@guppy` into a `GuppyLibrary`, that can be compiled together. So far supports only function definitions, declarations and structs (but this is not yet checked anywhere).

Also adds a `is_decl` function to `GuppyFunctionDefinition` to check whether a function is a declaration (i.e. has no body) or not. This will allow distinguishing between headers and implementations on a per-symbol level. This may be splittered into a separate PR.

This PR requires https://github.com/Quantinuum/hugr/pull/2947 to be released and pulled into Guppy.

Closes #960 

BREAKING CHANGE: Internal breaks to compilation engine interface. No user action is required.

BEGIN_COMMIT_OVERRIDE
feat: Joint compilation of Guppy symbols to a library
END_COMMIT_OVERRIDE

### DIFF
--- a/tests/integration/tket_opt/test_opt.py
+++ b/tests/integration/tket_opt/test_opt.py
@@ -32,18 +32,18 @@ def test_guppy_normalization() -> None:
         t(q1)
         cx(q0, q1)
 
-        unnormalized_hugr: Hugr = pauli_zz_rotation.compile_function().modules[0]
+    unnormalized_hugr: Hugr = pauli_zz_rotation.compile_function().modules[0]
 
-        # Count ops prior to normalization
-        assert _count_ops(unnormalized_hugr, "DataflowBlock") == 1
-        assert _count_ops(unnormalized_hugr, "MakeTuple") == 3
+    # Count ops prior to normalization
+    assert _count_ops(unnormalized_hugr, "DataflowBlock") == 1
+    assert _count_ops(unnormalized_hugr, "MakeTuple") == 3
 
-        normalized_hugr = normalize(unnormalized_hugr)
+    normalized_hugr = normalize(unnormalized_hugr)
 
-        # Test that the dataflow block is inlined by NormalizeGuppy
-        assert _count_ops(normalized_hugr, "DataflowBlock") == 0
-        # Test that MakeTuple nodes are removed by NormalizeGuppy
-        assert _count_ops(normalized_hugr, "MakeTuple") == 0
+    # Test that the dataflow block is inlined by NormalizeGuppy
+    assert _count_ops(normalized_hugr, "DataflowBlock") == 0
+    # Test that MakeTuple nodes are removed by NormalizeGuppy
+    assert _count_ops(normalized_hugr, "MakeTuple") == 0
 
 
 def test_redundant_cx_cancellation() -> None:


guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -283,18 +319,22 @@ def compile(self, id: DefId) -> ModulePointer:
...
-            # loosened after https://github.com/quantinuum/hugr/issues/2501 is fixed
-            graph.hugr.entrypoint = compiled_def.hugr_node
+        ctx = CompilerContext(graph, exported_defs)
+        compiled_defs: list[CompiledDef] = []
+        for def_id in def_ids:
+            compiled_defs.append(ctx.compile(self.checked[def_id]))
+            self.compiled = ctx.compiled
+
+            for member_id in DEF_STORE.impls[def_id].values():
+                ctx.compile(self.checked[member_id])
```
mark-koch: This will produce an error if the method requires monomorphization. Is that ok or do we just want to skip those methods?

guppylang/src/guppylang/defs.py
```diff
@@ -204,6 +220,26 @@ def compile_function(self) -> Package:
         """
         return super().compile()
 
+    def is_decl(self) -> bool:
```
mark-koch, line 253: Should this be a private method?

mark-koch: Nit: Also, maybe better as a property?

guppylang/src/guppylang/defs.py
```diff
@@ -204,6 +220,26 @@ def compile_function(self) -> Package:
...
+        """Whether this function definition is a declaration (i.e. has no body)."""
+        return isinstance(self.wrapped, RawFunctionDecl)
+
+
+@dataclass(frozen=True)
+class GuppyLibrary:
```
mark-koch, line 259: Can you add a docstring?

tests/integration/library/test_basic.py
```diff
@@ -0,0 +1,170 @@
...
+
+    adder_lib_1 = gen_adder_library(name="super_adder", value=5).compile()
+    adder_lib_2 = gen_adder_library(name="super_adder", value=20).compile()
+
+    @guppy
+    def main() -> None:
+        result("result", decl(5))
+
+    with pytest.raises(
+        Exception,  # For some reason we cannot import HugrLinkingError here
```
mark-koch: What happens if you try to import it?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -283,18 +319,22 @@ def compile(self, id: DefId) -> ModulePointer:
...
         from guppylang_internals.compiler.core import CompilerContext
 
-        ctx = CompilerContext(graph)
-        compiled_def = ctx.compile(self.checked[id])
-        self.compiled = ctx.compiled
+        exported_defs = set()
+        for def_id in def_ids:
+            exported_defs.add(def_id)
+            # TODO automatic member inclusion should be based on the automatic
+            # collection when available
```
mark-koch: Not sure I get what this means?

tests/integration/library/test_basic.py
mark-koch, line 1: Could you add an error test for trying to construct a library with generic functions? Also, (non-generic) structs with generic methods

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: For now, this is okay. In no case do I want to silently skip them. I am waiting for after #1441 is merged to explore what works and what does not, and to add a mechanism with which to specify the monomorphizations that should be included.

tests/integration/library/test_basic.py
maximilianruesch: I get an error stating that it is not a member of the module, as if it is not present from the binary. In `hugr-py` however, I can import it (and I do on the branch pointed to over there). I don't know why this is inconsistent.

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: Currently, including a struct in the library (by manually putting it into the `library(...)` call) adds all of its methods to be part of the library interface. However, in the future we want to have general functions specify whether they should be included in a library (or in which specific features / extras of a library) on their definition (similar to a visibility). Similar to how `library(...)` should then be auto-filled with these functions to begin with, when you add a struct (manually or automatically), this discovery should be applied to its members (so it should be recursive in a way).

tests/integration/library/test_basic.py
maximilianruesch: As stated in https://github.com/Quantinuum/guppylang/pull/1495#discussion_r2947049868, I am semi-waiting for the monomorphization overhaul for really rigorous testing here.

guppylang/src/guppylang/defs.py
maximilianruesch: This is supposed to be a helper for library users to help them determine whether they are using / importing headers or the actual implementations. As such, it is really useful to have this for libraries, but is not required for this core feature.

Do you think this should be moved into a separate PR, since it is fully separable?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -247,34 +247,70 @@ def get_checked(self, id: DefId) -> CheckedDef:
...
+                def_id, _ = self.types_to_check_worklist.popitem()
             else:
-                id, _ = self.to_check_worklist.popitem()
-            self.checked[id] = self.get_checked(id)
+                def_id, _ = self.to_check_worklist.popitem()
+            self.checked[def_id] = self.get_checked(def_id)
+
+            for member_id in DEF_STORE.impls[def_id].values():
+                if member_id not in self.checked:
+                    self.to_check_worklist[member_id] = self.get_parsed(member_id)
```
mark-koch: I think this is technically a breaking change: Programs that used to compile may no longer compile since now methods of any used types are checked as opposed to only the used methods

maximilianruesch: Technically yes, but imo we should only consider things breaking if they break "valid" Guppy programs. Is there a scenario where a valid Guppy program does not allow type checking all its (even unused) methods?

maximilianruesch: What worries me more is that this has performance implications, and the benchmark changes above are very likely to be real.

mark-koch: > we should only consider things breaking if they break "valid" Guppy programs

Yeah that depends on the definition of valid. If we consider using types with invalid methods as invalid, then the fact that we currently accept them would be a "bug", and fixing it maybe non-breaking

Performance is also a valid point. Can you compare how long it takes to run the entire test suite now vs before?

guppylang/src/guppylang/defs.py
mark-koch: Ah I see, if we expect user to call this method, then that's fine. Don't mind including it here

I would still prefer it as a property though :D

tests/integration/library/test_basic.py
mark-koch: Mhmm weird, don't know what could be going on there. Can you start an ipython session in your environment (`uv run python`) and try to import it there?

tests/integration/library/test_basic.py
mark-koch: Fair, I'd still prefer a test case though, just so we can track how the behaviour changes. And making sure that it doesn't crash the compiler.

Doesn't have to be rigorous

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: Running the entire test suite (skipping the library tests for fairness) takes `(-n auto | -n 0)`:
- Before the change: `(41.71s | 113.20s)`
- After the change: `(44.52s | 123.77s)`

Note that this is not statistically significant in any way since I am impatient, but I think the expected degradation is already visible here.

maximilianruesch: We might need a more sophisticated function to determine when things will get used. And I am also unsure how this will then interact with adding auto-generated methods (currently just `__new__` for structs, but it is the reason we put this in in the first place irrc).

tests/integration/library/test_basic.py
maximilianruesch: Something weird is going on. I can import `hugr` and `hugr._hugr` as usual, but importing `hugr._hugr.linking` gives me a module describing itself as `hugr._hugr.linking.linking`. The same thing occurs with `hugr._hugr.metadata` (which also has the `#[pymodule]` pyo3 attribute), but with `hugr._hugr.model` (which does not have this attribute for some reason), I just get `model`:
image

When inspecting `hugr._hugr.linking.linking` I also do not see the error class. 

Also, I was mistaken about the import on `hugr-py`. I only import `link_modules` from that module (which is present and works), but when I try to import `HugrLinkingError` as well, it fails with a similar error. I will have to do some more investigation.

maximilianruesch: Found it, I forgot to export in the module

maximilianruesch: Will be fixed after https://github.com/Quantinuum/hugr/pull/2964

guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -151,16 +151,22 @@ class CompilerContext(ToHugrContext):
 
     checked_globals: Globals
 
+    #: The definitions that should be exported (i.e. made public) in the currently
```
acl-cqc: "in the Hugr module currently *being* built"? And is this always true, or only if we are currently building a library rather than function/executable?

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -218,8 +220,13 @@ def monomorphize(
         """
         mono_ty = self.ty.instantiate_partial(mono_args)
         hugr_ty = mono_ty.to_hugr_poly(ctx)
+        visibility: Visibility = "Public" if self.id in ctx.exported_defs else "Private"
```
acl-cqc: Ok hang on do we still call this `monomorphize` if the function is not polymorphic? Maybe you've handled that case elsewhere

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -247,34 +247,69 @@ def get_checked(self, id: DefId) -> CheckedDef:
...
 
         This is the main driver behind `guppy.check()`.
         """
+        self.check([id])
+
+    @pretty_errors
+    def check(self, def_ids: list[DefId], *, reset: bool = True) -> None:
+        """Top-level function to kick of checking of multiple definitions.
+
+        This is the main driver behind `guppy.library(...).check()`.
```
acl-cqc, line 383: What's the `reset` arg? Yes, I can see it turns off the call to reset that we always used to do...when would I *not* want to do that?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -247,34 +247,69 @@ def get_checked(self, id: DefId) -> CheckedDef:
...
+                def_id, _ = self.types_to_check_worklist.popitem()
             else:
-                id, _ = self.to_check_worklist.popitem()
-            self.checked[id] = self.get_checked(id)
+                def_id, _ = self.to_check_worklist.popitem()
+            self.checked[def_id] = self.get_checked(def_id)
+
+    @pretty_errors
+    def compile_single(self, id: DefId) -> ModulePointer:
+        """Top-level function to kick of Hugr compilation of a definition.
```
acl-cqc:
```suggestion
        """Top-level function to kick off Hugr compilation of a definition.
```

Or maybe use plainer English and change "kick off" to "begin" or "start" !!

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -247,34 +247,69 @@ def get_checked(self, id: DefId) -> CheckedDef:
...
+        This is the function that is invoked by `guppy.compile`.
+        """
+        pointer, [compiled_def] = self._compile([id])
+
+        if (
+            isinstance(compiled_def, CompiledHugrNodeDef)
+            and isinstance(compiled_def, CompiledCallableDef)
+            and not isinstance(pointer.module[compiled_def.hugr_node].op, ops.FuncDecl)
+        ):
+            # if compiling a region set it as the HUGR entrypoint can be
```
acl-cqc, line 438: Hmmmm. So this is the case where you `@guppy.declare` and then call `compile_function` on that, yielding a two node Hugr (Module root plus FuncDecl).

Do we really want to allow that anyway??

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -247,34 +247,69 @@ def get_checked(self, id: DefId) -> CheckedDef:
...
+            # loosened after https://github.com/quantinuum/hugr/issues/2501 is fixed
+            pointer.module.entrypoint = compiled_def.hugr_node
+
+        return pointer
 
     @pretty_errors
-    def compile(self, id: DefId) -> ModulePointer:
+    def compile(self, def_ids: list[DefId], *, reset: bool = True) -> ModulePointer:
```
acl-cqc, line 445: Should this be `compile_library` and then keep `compile_single` as `compile` ?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -247,34 +247,69 @@ def get_checked(self, id: DefId) -> CheckedDef:
...
+        return self._compile(def_ids, reset=reset)[0]
+
+    def _compile(
+        self, def_ids: list[DefId], *, reset: bool = True
+    ) -> tuple[ModulePointer, list[CompiledDef]]:
         """Top-level function to kick of Hugr compilation of a definition.
```
acl-cqc: This doc is definitely not right now

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -283,18 +318,11 @@ def compile(self, id: DefId) -> ModulePointer:
...
-            and not isinstance(graph.hugr[compiled_def.hugr_node].op, ops.FuncDecl)
-        ):
-            # if compiling a region set it as the HUGR entrypoint can be
-            # loosened after https://github.com/quantinuum/hugr/issues/2501 is fixed
-            graph.hugr.entrypoint = compiled_def.hugr_node
+        ctx = CompilerContext(graph, set(def_ids))
+        compiled_defs: list[CompiledDef] = []
+        for def_id in def_ids:
+            compiled_defs.append(ctx.compile(self.checked[def_id]))
+            self.compiled = ctx.compiled
```
acl-cqc: I'm wondering what's the difference between `self.compiled` and `compiled_defs`....do we need to be assigning this in the loop? (It's not obvious that it's read until after the loop, unless `ctx.compile` calls back into the engine to get it, which seems a bit silly given there is `ctx.compiled`)

guppylang/src/guppylang/decorator.py
```diff
@@ -455,6 +456,12 @@ def compile(self, obj: Any) -> ModulePointer:
             raise TypeError(f"Object is not a Guppy definition: {obj}")
         return ModulePointer(obj.compile(), 0)
 
+    def library(self, *members: GuppyDefinition) -> GuppyLibrary:
+        """Defines a Guppy library, which is a collection of Guppy definitions that can
```
acl-cqc, line 514: Presumably this is not actually a decorator, but is here because the global `guppy` is an instance of this class? (In which case, mention that it's not a decorator in the doc - that's a lot to infer from the type signature!)

guppylang/src/guppylang/defs.py
```diff
@@ -109,12 +119,18 @@ def emulator(
...
+            function. This can be used to provide additional library functions that the
+            function being compiled depends on.
 
         Returns:
             An `EmulatorInstance` that can be used to run the function in an emulator.
         """
         mod = self.compile()
 
+        if libs is not None:
+            mod = mod.link(*libs)
```
acl-cqc, line 161: sweet!

guppylang/src/guppylang/defs.py
```diff
@@ -204,6 +220,44 @@ def compile_function(self) -> Package:
         """
         return super().compile()
 
+    @property
+    def is_decl(self) -> bool:
+        """Whether this function definition is a declaration (i.e. has no body)."""
```
acl-cqc, line 254:
```suggestion
        """Whether this function is a declaration (i.e. has no body), as opposed to a definition."""
```

Was gonna say "whether this function is a definition or declaration" but that's not a boolean ;)

guppylang/src/guppylang/defs.py
```diff
@@ -204,6 +220,44 @@ def compile_function(self) -> Package:
...
+
+    def _member_impls(self) -> list[DefId]:
+        """Any implementations registered for members of this library. Note that the
+        list is only guaranteed to be complete after calling `check()` on the library
+        members, since auto-generated implementations may be added during checking."""
+        impls: list[DefId] = []
+        for def_id in self.members:
+            # TODO automatic member inclusion should be based on the automatic
+            # collection when available
+            impls.extend(DEF_STORE.impls[def_id].values())
```
acl-cqc: how might there be more than one? Monomorphic instances? Should we allow that, or wait until we have an `instances` annotation?

guppylang/src/guppylang/defs.py
```diff
@@ -204,6 +220,44 @@ def compile_function(self) -> Package:
...
+    """A collection of Guppy definitions that can be compiled together into a linkable
+    unit exposing a public interface."""
+
+    members: list[DefId]
+
+    def _member_impls(self) -> list[DefId]:
+        """Any implementations registered for members of this library. Note that the
+        list is only guaranteed to be complete after calling `check()` on the library
+        members, since auto-generated implementations may be added during checking."""
+        impls: list[DefId] = []
```
acl-cqc: nit: consider making this into a list comprehension

guppylang/src/guppylang/defs.py
```diff
@@ -204,6 +220,44 @@ def compile_function(self) -> Package:
...
+            # TODO automatic member inclusion should be based on the automatic
+            # collection when available
+            impls.extend(DEF_STORE.impls[def_id].values())
+
+        return impls
+
+    def compile(self) -> Package:
+        """Compile this collection of definitions into a HUGR package."""
+        ENGINE.check(self.members)
+        pointer = ENGINE.compile(self.members + self._member_impls(), reset=False)
```
acl-cqc: Can you explain what's going on here / why we need to do this? `self._member_impls()` is gathered *from* `self.members`, right?

and `reset=False` is because we've done `check` first? Is this post- #1441 (PR is not "into" that branch, yet does not show those changes AFAICS...)

Does compile not include check? Do we not need to check the `_member_impls`?

guppylang/src/guppylang/defs.py
```diff
@@ -204,6 +220,44 @@ def compile_function(self) -> Package:
...
+        for def_id in self.members:
+            # TODO automatic member inclusion should be based on the automatic
+            # collection when available
+            impls.extend(DEF_STORE.impls[def_id].values())
+
+        return impls
+
+    def compile(self) -> Package:
+        """Compile this collection of definitions into a HUGR package."""
+        ENGINE.check(self.members)
```
acl-cqc, line 279: (or following on from next comment - should this be `self.check` ?)

guppylang-internals/src/guppylang_internals/compiler/core.py
acl-cqc: No, for executables, this will be just the function that is also made the entrypoint. I'd mention that here though.

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: This function will be changed with #1441 anyway (at least I expect it to). I am not doing any generics handling until that is merged.

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: You do not want to do it when you need to call `check` twice, for example when some things you definitely need to check only become apparent once you check for the first time.

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: I do not exactly get what you mean. This snippet is a remainder from the old `compile` call, and I kept it for consistency with the old compilation behaviour.

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: Imo no it should not. One could rename "def_ids" to "entrypoints" here, or take `compile_single` as an overload of this function, but since its internals anyway I do not really care.

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: `self.compiled` stores ALL compiled definitions, whereas `compiled_defs` stores just the compiled entrypoints (to be returned at the end of the function).

guppylang/src/guppylang/defs.py
maximilianruesch: It is also not a "definition" in the sense of a "FuncDefn" if it is not a declaration. It might be a pytket circuit, a custom function, etc.. Imo it would be confusing to say "definition", or we would have to add an "e.g.".

guppylang/src/guppylang/defs.py
maximilianruesch: These are not instances, these are member functions (e.g. of structs, enums, or opaque types). I do not grasp the meaning of your comment.

guppylang/src/guppylang/defs.py
maximilianruesch: Difficult since it would evaluate to a list of lists, and unpacking inside a list comprehension is not a thing in Python 3.10.

guppylang/src/guppylang/defs.py
maximilianruesch: We need to check first, since `self._member_impls()` will only return the full extent of what is known about a type after the type itself has been checked (currently the (only) case is that structs only generate a `__new__` method after checking, once they have their appropriate type parameters set).

So then, to avoid rechecking the type definition once we have all definitions we want to expose (`self.members + self._member_impls()`) from the library, we set `reset=False` so that the checking work is not repeated (and the reference to the auto-generated impls is still valid).

So yes, `self._member_impls()` is gathered from `self.members`, but it can only be gathered once `self.members` has been checked.

guppylang/src/guppylang/defs.py
maximilianruesch: It could be, it does not really matter performance wise, as long as the checking cache is not reset in between. It only matters that `self.members` is checked before compiling though.

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Ok...so `compiled_defs` is just the ones that were requested (i.e. public)? Might be worth renaming. And do we need to be reassiging `self.compiled` every time around the loop?

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Since `compile_single` calls straight to `_compile` not `compile`, this *is* only used for compiling a library?

guppylang/src/guppylang/defs.py
acl-cqc: So if true, it's a declaration.
What might it be if false? Any of those things? this is a documentation nit, but should this just say "whether this function is a declaration" and not mention definitions?

guppylang/src/guppylang/defs.py
acl-cqc: why is this `extend`? Is `DEF_STORE.impls[def_id].values()` really more than one thing? (ok so `DEF_STORE.impls[def_id]` must be a map! So maybe this is just "easiest")

guppylang/src/guppylang/defs.py
acl-cqc: Could just be `[v for def_id in self.members for v in DEF_STORE.impls[def_id].values()]` IIUC

guppylang/src/guppylang/defs.py
acl-cqc: Ok I suggest a comment inbetween ENGINE.check and ENGINE.compile that `Check fills _member_impls with additional members that have also been checked`

guppylang/src/guppylang/defs.py
acl-cqc: `DefinitionStore.impls` is a `dict[DefId, dict[str, DefId]]` which isn't documented :(. I infer from `register_impl` that (for structs) this is a map from struct-id to map from member-function-name to member-function-id. (It looks like `impls` works similarly for enums but not functions.)

I think renaming `impls` to `type_members` would be a huge improvement, other name suggestions welcome, but this could do with some explanation somehow.

guppylang/src/guppylang/defs.py
maximilianruesch: That is imo more verbose than an `extend` (and may even have worse performance)...

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: Outdated after merge with #1441

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: Currently, yes, but the engine does not need to know that it came from a library. Technically, every collection of entrypoints is a "library", so that includes the empty library, every compilation of a single entrypoint (where that library type is special in that it has the entrypoint marked in the Hugr, and could thus be seen as an executable "binary" instead of a "library" [to use Rust terms]), and joint compilation of more than one entrypoint.

I would like to keep the "library" terminology on the user interface only as far as possible.

guppylang/src/guppylang/defs.py
maximilianruesch: This is on the class `GuppyFunctionDefinition`, so I would think it is fair to say "this function definition".

tests/error/library_errors/generic_entrypoint.err
```diff
@@ -0,0 +1,11 @@
+Error: Invalid entry point (at $FILE:7:0)
+  | 
+5 | 
+6 | @guppy
+7 | def func(x: T) -> array[T, 3]:
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+8 |     return array(x for _ in range(3))
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Function `func` is not a valid compilation entry point since
+  |                                       the value of its generic parameter `T` is not known
```
maximilianruesch, line 9: Do we want to improve this message? It may not be obvious to the user that every member of the library essentially acts as a potential "entrypoint" (they know, they just do not use that naming).

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -426,21 +464,14 @@ def compile(self, id: DefId) -> ModulePointer:
...
-        ctx = CompilerContext(graph)
-        check_entry_point_non_generic(self.get_parsed(id))
-        entry_mono_args: Inst = ()
-        compiled_def = ctx.compile(self.checked[id, entry_mono_args], entry_mono_args)
+        ctx = CompilerContext(graph, set(def_ids))
+        requested_defs = []
+        for def_id in def_ids:
+            check_entry_point_non_generic(self.get_parsed(def_id))
+            requested_defs.append(ctx.build_compiled_def(def_id, type_args=None))
+        ctx.iterate_worklist()
```
maximilianruesch, line 470: This change is currently theoretically not necessary, but may improve performance, since we first register all known defs to be compiled, then iterate the worklist manually. Not sure if this is a good change (@mark-koch), and it also makes `CompilerContext.compile` unused.

EDIT: Iterating the worklist once also results in drops being inserted only once. There should be no harm in doing it multiple times (since it has an idempotency check)

guppylang-internals/src/guppylang_internals/engine.py
tatiana-s: I also prefer `compile` more than `compile_library`, if only to keep it consistent with the `check`/`check_single`

tests/error/library_errors/generic_entrypoint.err
tatiana-s: Would it be possible to add a hint / note that mentions that all library members need to be valid entrypoints?

maximilianruesch: The problem is that at this point, the compilation function does not know it is part of a library. I could raise an internal marker exception though, that is converted into a nicer message in `compile_single` and `GuppyLibrary.compile` (note the conversion happens once in internals, and once in the main language package). There we could add a hint. Thoughts?

maximilianruesch: We agreed to have this as an optional followup if it creates too much confusion.

guppylang-internals/src/guppylang_internals/engine.py
mark-koch: I don't mind going with this for now (also feel free to delete `CompilerContext.compile`).

The better long term change would be to get rid of the worklist in `CompilerContext` and handle everything in the `ENGINE`, see #990

guppylang/src/guppylang/defs.py
maximilianruesch: Pulled this change out to https://github.com/Quantinuum/guppylang/pull/1632.

PR #1496, opened 2026-02-12
fix: Resolve relative wasm file paths from calling file

### LINKED ISSUE #1407: Wasm file resolution bug
Following #1339 and subsequent release of guppy [v0.21.7](https://github.com/Quantinuum/guppylang/releases/tag/guppylang-v0.21.7), there's a regression in wasm file path resolution.

If I have my file structure as follows:
```sh
❯ tree
.
├── guppy_wasm.ipynb
└── wasm
    ├── arith.wasm
    └── wasm.py
```

and `wasm.py` has:

```py
@wasm_module("arith.wasm")
class MyWasm:
    @wasm
    def add(self: "MyWasm", a: int, b: int) -> int: ...
[...]
@guppy
def main() -> None:

    [mod1] = spawn_wasm_contexts(1, MyWasm)
    res5 = mod1.add(10, 32)
[...]
```

Then if I do `from wasm.wasm import main` from `guppy_wasm.ipynb`, I get:
```py
---------------------------------------------------------------------------
GuppyError                                Traceback (most recent call last)
Cell In[1], [line 3](vscode-notebook-cell:?execution_count=1&line=3)
      1 import qnexus as qnx
      2 from rich import print as rprint
----> [3](vscode-notebook-cell:?execution_count=1&line=3) from wasm.wasm import main

File ~/GH/qapi-submit/wasm/wasm.py:10
      6 from guppylang.std.qsystem.wasm import spawn_wasm_contexts
      7 from guppylang_internals.decorator import wasm, wasm_module
---> [10](~/GH/qapi-submit/wasm/wasm.py:10) @wasm_module("arith.wasm")
     11 class MyWasm:
     12     @wasm
     13     def add(self: "MyWasm", a: int, b: int) -> int: ...

File ~/GH/qapi-submit/.venv/lib/python3.13/site-packages/guppylang_internals/decorator.py:217, in wasm_module(filename)
    215     wasm_sigs = decode_wasm_functions(filename)
    216 else:
--> [217](~/GH/qapi-submit/.venv/lib/python3.13/site-packages/guppylang_internals/decorator.py:217)     raise GuppyError(WasmFileNotFound(None, filename))
    219 def type_def_wrapper(
    220     id: DefId,
    221     name: str,
   (...)    224     config: str | None,
    225 ) -> OpaqueTypeDef:
    226     assert config is None

GuppyError: WasmFileNotFound(span=None, _parent=None, children=[], file='arith.wasm')
```

Changing the file path in the decorate to `@wasm_module("wasm/arith.wasm")` works.

We should be looking at the relative path from the guppy source, not the execution site.

rmfisher26:
**Summary**

- Fix `@wasm_module `resolving relative WASM file paths against the current working directory instead of the source file's directory
- Walk the call stack past all guppylang-internal frames to find the user's source file, then resolve the path relative to its directory
- Fall back to CWD resolution when `__file__` is unavailable (e.g. interactive/notebook sessions without a backing file)

**Motivation**
Closes #1407

**Details**

- When a module using `@wasm_module("arith.wasm")` is imported from a different directory (e.g. a Jupyter notebook in a parent folder), the decorator failed to find the WASM file because `pathlib.Path(filename)` resolves relative to the process CWD, not the directory containing the Python source file.

- The existing `get_calling_frame()` helper couldn't be reused here because it only skips frames from decorator.py itself. Since `wasm_module` is wrapped by `@pretty_errors `(defined in error.py), `get_calling_frame()` stops at the error.py frame rather than reaching the user's code. The fix instead walks past all frames belonging to `guppylang*` modules.

**Test plan**

-  Existing WASM integration tests pass (`pytest tests/integration/test_wasm.py`)
-  Manual test: create a subdirectory with a `.wasm` file and a module using `@wasm_module("file.wasm")`, import it from a script in the parent directory, and verify it loads successfully

`test_relative_path_from_different_cwd`

- The core fix — imports a module that uses @wasm_module("test.wasm") with a relative path while cwd is set to a completely different directory (tmp_path). This would have failed before the fix

`test_absolute_path_unaffected`

- Absolute paths continue to work as before (no regression from the resolution 

`test_absolute_path_unaffected_by_cwd `

- Absolute paths work even when cwd is changed — the resolution logic doesn't break them.

`test_nonexistent_relative_path_error`

- A relative path to a non-existent file still raises GuppyError with the

### DIFF
--- a/guppylang-internals/src/guppylang_internals/decorator.py
+++ b/guppylang-internals/src/guppylang_internals/decorator.py
@@ -234,11 +234,8 @@ def wasm_module(
             caller_dir = pathlib.Path(caller_file).resolve().parent
             wasm_path = caller_dir / filename
 
-    wasm_file = wasm_path
-
-    resolved = str(wasm_file)
-    if wasm_file.is_file():
-        wasm_sigs = decode_wasm_functions(resolved)
+    if wasm_path.is_file():
+        wasm_sigs = decode_wasm_functions(str(wasm_path))
     else:
         raise GuppyError(WasmFileNotFound(None, filename))
 
@@ -261,7 +258,7 @@ def wasm_module(
     )
 
     def inner_fun(ty: builtins.type[T]) -> GuppyDefinition:
-        decorator_inner = decorator(resolved, None)
+        decorator_inner = decorator(str(wasm_path), None)
         return decorator_inner(ty)
 
     return inner_fun
--- a/tests/integration/test_wasm_path_resolution.py
+++ b/tests/integration/test_wasm_path_resolution.py
@@ -1,10 +1,4 @@
-"""Tests for WASM file path resolution.
-
-See https://github.com/Quantinuum/guppylang/issues/1407.
-
-Verifies that @wasm_module resolves relative paths relative to the source file
-containing the decorator, not the current working directory.
-"""
+"""Tests for WASM file path resolution."""
 
 from pathlib import Path
 
@@ -21,7 +15,9 @@ WASM_FILE = str(RESOURCES_DIR / "test.wasm")
 
 
 def test_relative_path_from_different_cwd(validate, monkeypatch, tmp_path):
-    """Relative paths resolve from the source file's directory, not the cwd."""
+    """A module using @wasm_module with a relative path should work even when
+    the process cwd differs from the module's directory (See https://github.com/Quantinuum/guppylang/issues/1407)."""
+
     # Change cwd to a directory that does NOT contain the wasm file
     monkeypatch.chdir(tmp_path)
 
@@ -42,7 +38,8 @@ def test_relative_path_from_different_cwd(validate, monkeypatch, tmp_path):
 
 
 def test_relative_path_without_cwd_change(validate):
-    """Relative paths resolve correctly without any cwd manipulation."""
+    """A module using @wasm_module with a relative path should work when
+    the process cwd is the project root (See https://github.com/Quantinuum/guppylang/issues/1407)."""
     from tests.resources.test_wasm_def import RelativeWasm
 
     @guppy
@@ -97,7 +94,8 @@ def test_absolute_path_from_different_cwd(validate, monkeypatch, tmp_path):
 
 
 def test_nonexistent_relative_path_error():
-    """A relative path that doesn't exist raises WasmFileNotFound."""
+    """A module using @wasm_module with a relative path that doesn't exist should
+    raise WasmFileNotFound (See https://github.com/Quantinuum/guppylang/issues/1407)."""
     with pytest.raises(GuppyError, match="WasmFileNotFound"):
 
         @wasm_module("no_such_file.wasm")


maximilianruesch (changes requested):
This does not yet constitute a full / in detail review, I will come back to this later. However, you state that the test plan includes
> Manual test: create a subdirectory with a .wasm file and a module using @wasm_module("file.wasm"), import it from a script in the parent directory, and verify it loads successfully

Can you add specific tests for this, to assert that:
1. Your fix works, and it fixes something that did not work before
2. Your fix will continue to work alongside all other behaviour in the future
3. Your fix does not break resolution of given absolute paths (the resolution should only change for relative paths)

tests/integration/test_wasm_path_resolution.py
```diff
@@ -0,0 +1,88 @@
...
+        [mod] = spawn_wasm_contexts(1, AbsWasm)
+        x = mod.two()
+        mod.discard()
+        return x
+
+    mod = main.compile_function()
+    validate(mod)
+
+
+def test_nonexistent_relative_path_error():
```
maximilianruesch, line 96: Please add a test that asserts that absolute paths that do not exist also raise an error

tests/resources/test_wasm_def.py
maximilianruesch, line 1: I don't think you need an entire different wasm file here. You should be able to move this file to the directory above. It still asserts relative paths since the test is not right next to it.

Additionally, please rename the file to `test_wasm.py` or `test_wasm_def.py`.

tests/resources/wasm_subdir/__init__.py
maximilianruesch, line 1: Subdir can be removed after the changes to the Python file.

tests/integration/test_wasm_path_resolution.py
```diff
@@ -0,0 +1,88 @@
...
+        b = mod.add(a, a)
+        mod.discard()
+        return b
+
+    mod = main.compile_function()
+    validate(mod)
+
+
+def test_absolute_path_unaffected(validate):
+    """Absolute paths should continue to work regardless of resolution logic."""
```
maximilianruesch:
```suggestion
def test_absolute_path(validate):
```
The `unaffected` refers to temporals that are only valid when this PR is still open, and thus become confusing when it is merged.

tests/integration/test_wasm_path_resolution.py
```diff
@@ -0,0 +1,88 @@
...
+from guppylang.std.qsystem.wasm import spawn_wasm_contexts
+
+
+RESOURCES_DIR = Path(__file__).resolve().parent.parent / "resources"
+WASM_FILE = str(RESOURCES_DIR / "test.wasm")
+
+
+def test_relative_path_from_different_cwd(validate, monkeypatch, tmp_path):
+    """A module using @wasm_module with a relative path should work even when
+    the process cwd differs from the module's directory (issue #1407)."""
```
maximilianruesch: Nice test! Please add a full URL to the issue instead of its number. This makes jumping to the issue easier, and clarifies the repository in which the issue can be found.

tests/integration/test_wasm_path_resolution.py
```diff
@@ -0,0 +1,88 @@
+"""Tests for WASM file path resolution (issue #1407).
+
+Verifies that @wasm_module resolves relative paths relative to the source file
+containing the decorator, not the current working directory.
+"""
```
maximilianruesch:
```suggestion
"""Tests for WASM file path resolution."""
```
Issue #1407 only refers to relative paths, but this file contains absolute path tests as well, so both the issue reference and the remainder of the doc comment are somewhat misplaced here.

tests/integration/test_wasm_path_resolution.py
```diff
@@ -0,0 +1,88 @@
...
+        x = mod.two()
+        mod.discard()
+        return x
+
+    mod = main.compile_function()
+    validate(mod)
+
+
+def test_absolute_path_unaffected_by_cwd(validate, monkeypatch, tmp_path):
+    """Absolute paths should work even when cwd is changed."""
```
maximilianruesch:
```suggestion
def test_absolute_path_changed_dir(validate, monkeypatch, tmp_path):
```

tests/integration/test_wasm_path_resolution.py
```diff
@@ -0,0 +1,88 @@
...
+        x = mod.two()
+        mod.discard()
+        return x
+
+    mod = main.compile_function()
+    validate(mod)
+
+
+def test_nonexistent_relative_path_error():
+    """A relative path that doesn't exist should still raise WasmFileNotFound."""
```
maximilianruesch:
```suggestion
```
The comment contains temporal language (see other comments), and its content is already (almost) clear from the test.

Instead, please strengthen the assertion of the error from a `GuppyError` to a `WasmFileNotFound` error. You could also just move this to a file that produces an error snapshot, that asserts the entire user-visible diagnostic.

tests/integration/test_wasm_path_resolution.py
```diff
@@ -0,0 +1,88 @@
...
+
+import pytest
+
+from guppylang import guppy
+from guppylang_internals.decorator import wasm, wasm_module
+from guppylang_internals.error import GuppyError
+from guppylang.std.qsystem.wasm import spawn_wasm_contexts
+
+
+RESOURCES_DIR = Path(__file__).resolve().parent.parent / "resources"
```
maximilianruesch:
```suggestion
RESOURCES_DIR = Path(__file__).resolve().parent.parent / "resources"
```
Add a `.absolute()` to ensure (and implicitly document) that this is an absolute path.

tests/integration/test_wasm_path_resolution.py
```diff
@@ -0,0 +1,88 @@
...
+WASM_FILE = str(RESOURCES_DIR / "test.wasm")
+
+
+def test_relative_path_from_different_cwd(validate, monkeypatch, tmp_path):
```
maximilianruesch, line 17: Can you add a test that tests that relative resolution works without changing the directory as well? Ideally, it should be entirely unaffected.

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -213,9 +213,27 @@ def dec(c: type[T]) -> type[T]:
...
+    # not the current working directory.
+    caller_file = None
+    frame = inspect.currentframe()
+    while frame:
+        module = inspect.getmodule(frame)
+        if module is None or not getattr(module, "__name__", "").startswith(
+            "guppylang"
+        ):
+            caller_file = frame.f_globals.get("__file__")
+            break
```
maximilianruesch: Can you get away with just doing `module.__name__`? A module should always have a name imo, unless there is some real world example where it does not (then please state one in this thread). If the type checker complains, it would be clearer to raise an internal error when a non-None module does not have `__name__`.

In any case, this statement (or perhaps the while loop) deserves a comment shortly explaining what it does (and why you include the magic string "guppylang").

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -213,9 +213,27 @@ def dec(c: type[T]) -> type[T]:
...
+        frame = frame.f_back
+    if caller_file is not None:
+        caller_dir = pathlib.Path(caller_file).resolve().parent
+        wasm_file = caller_dir / filename
+    else:
+        wasm_file = pathlib.Path(filename)
```
maximilianruesch: It is not clear to me how absolute paths work here. Could you clarify in this thread? What happens if I have a caller file (and thus a caller dir), but pass an absolute filename?

tests/integration/test_wasm_path_resolution.py
rmfisher26: Done!

tests/resources/wasm_subdir/__init__.py
rmfisher26: Completed.

tests/resources/test_wasm_def.py
rmfisher26: Will do!

tests/integration/test_wasm_path_resolution.py
rmfisher26: Added `test_nonexistent_absolute_path_error `

tests/integration/test_wasm_path_resolution.py
rmfisher26: Updated to be 
`RESOURCES_DIR = Path(__file__).resolve().absolute().parent.parent / "resources"
`

tests/integration/test_wasm_path_resolution.py
rmfisher26: Added test: `test_relative_path_without_cwd_change `

guppylang-internals/src/guppylang_internals/decorator.py
rmfisher26: Sure,  I updated this to using `module.__name__` and added a comment to explain what it does.

guppylang-internals/src/guppylang_internals/decorator.py
rmfisher26: This was changed a little bit now, so let me know if that helps.  When the caller file passes an absolute file name it will use the path as-is. 

When filename is absolute (e.g. "`/home/user/project/test.wasm`"):

1. `wasm_path = pathlib.Path(filename)` — creates an absolute Path
2. `wasm_path.is_absolute()` returns True
3. The entire if block (lines 220-235) is skipped
4. `wasm_file = wasm_path` — the absolute path is used as-is

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -213,9 +213,32 @@ def dec(c: type[T]) -> type[T]:
...
+            module = inspect.getmodule(frame)
+            if module is None or not module.__name__.startswith("guppylang"):
+                caller_file = frame.f_globals.get("__file__")
+                break
+            frame = frame.f_back
+        if caller_file is not None:
+            caller_dir = pathlib.Path(caller_file).resolve().parent
+            wasm_path = caller_dir / filename
+
+    wasm_file = wasm_path
```
maximilianruesch: This variable seems to be superfluous. Especially when you are testing `.is_file` later, just use `wasm_path` and delete this aliasing.

guppylang-internals/src/guppylang_internals/decorator.py
maximilianruesch: Thats much more helpful, thanks!

tests/integration/test_wasm_path_resolution.py
maximilianruesch: You did not follow this suggestion in the intended manner. The issue reference should be on one (or both) of the relative resolution tests. This file contains tests for which this comment is not relevant (i.e. the absolute ones), hence why the comment should not be on the file.

maximilianruesch (changes requested):
This has improved quite a lot, especially adding clarity for ensuring that absolute paths still work as intended. Only a few small changes / comments left!

guppylang-internals/src/guppylang_internals/decorator.py
rmfisher26: Understood, thanks for the note. I have updated this accordingly to just use the wasm_path

maximilianruesch (approved):

PR #1497, opened 2026-02-13
fix: Misleading error message for leaked qubits

### LINKED ISSUE #909: Misleading error message for leaked qubits
```python
from guppylang import GuppyModule
from guppylang.std import quantum
from guppylang.std.quantum import measure_array

def build_phase_gadget_prog(n_qb: int, theta: float):
    my_module = GuppyModule(name="phase_gadget")
    my_module.load_all(quantum)
    my_module.load(angle)
    
    @guppy(my_module)
    def phase_gadget() -> None:
        qs = array(qubit() for _ in range(comptime(n_qb)))
        # Allocated array of qubits is not consumed.
    
        for i in range(comptime(n_qb) - 1):
            cx(qs[i], qs[i + 1])

        rz(qs[-1], angle(comptime(theta)))

        for i in range(comptime(n_qb) - 1):
            cx(qs[i], qs[i + 1])


    return phase_gadget


phase_gadget1 = build_phase_gadget_prog(5, 0.5)
phase_gadget2 = build_phase_gadget_prog(10, 0.25)


@guppy
def main() -> None:
    qs1 = array(qubit() for _ in range(5))
    qs2 = array(qubit() for _ in range(10))

    phase_gadget1(qs1)
    phase_gadget2(qs2)


    b1 = measure_array(qs1)
    b2 = measure_array(qs2)

guppy.compile_module()
```

Using commit (51b1b81c06257e88583d68c245223c891dcc4e32) GIves the compiler error 

```
Error: Drop violation (at <In [?]>:3:8)
   | 
 1 |     @guppy(my_module)
 2 |     def phase_gadget() -> None:
 3 |         qs = array(qubit() for _ in range(comptime(n_qb)))
   |         ^^ Variable `qs` with non-droppable type `array[qubit, 5]` may
   |            be leaked ...
   | 
11 |         for i in range(comptime(n_qb) - 1):
   |                  ------------------------- if this expression is `False`

Help: Make sure that `qs` is consumed or returned to avoid the leak

```

The trace is misleading as I think 
```
if this expression is `False` 
```
refers to some compiler internals.

rmfisher26:
**Summary**
Resolves the misleading error message incorrectly blaming a branch condition when a non-droppable variable (e.g. a qubit) leaks unconditionally.

Fixes #909

**Root Cause**
The linearity checker uses liveness analysis to determine whether a variable is used on successor code paths. When a variable is live on some paths but not all, a `Branch `sub-diagnostic is attached pointing at the branch condition (e.g. `if this expression is 'False'`).

However, liveness analysis does not distinguish between borrowing and consuming a variable. A gate like `h(q)` borrows the qubit — only borrows the qubit, so it leaks on both paths. A borrow made the variable appear "used" on that path, so the checker assumed the leak was conditional.

**Fix:**

Before attaching the branch note, check the `UseKind `on each path. Only show the branch note when the variable is actually consumed (not just borrowed) on at least one path.

Before (misleading):

```
Variable `q` with non-droppable type `qubit` may be leaked ...
  if this expression is `False`
```
After (correct):

`Variable `q` with non-droppable type `qubit` is leaked`

When the variable is consumed on one path (e.g. via measure(q)), the Branch note still appears as before.

**Tests**
|Test | Scenario | Diagnostic|
| :--- | :---: | ---: |
|`borrow_branch_leak` | `h(q)` borrows in one branch |  `is leaked` — no Branch note |
|`borrow_both_branches_leak `| `h(q) `/` t(q)` borrow in both branches |  `is leaked` — no Branch note |
|`borrow_array_loop_leak `| From reproducer sample: borrow-only gates on array in a loop |  `is leaked` — no Branch note |
|`consume_branch_leak `| `measure(q) `consumes in one branch | `may be leaked ... if this expression is 'False'` — Branch note correctly shown |

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
@@ -898,15 +898,24 @@ def check_cfg_linearity(
                 used_later = all(x in live_before[succ] for succ in bb.successors)
                 if not leaf.ty.droppable and not scope.used(x) and not used_later:
                     err = PlaceNotUsedError(scope[x].defined_at, leaf)
-                    # If there are some paths that lead to a consumption, we can give
-                    # a nicer error message by highlighting the branch that leads to
-                    # the leak
-                    if any(x in live_before[succ] for succ in bb.successors):
+                    # If the variable is consumed (not just borrowed) on some
+                    # paths, highlight the branch that leads to the leak.
+                    consumed_succs = set()
+                    for succ in bb.successors:
+                        if x not in live_before[succ]:
+                            continue
+                        use_bb = live_before[succ][x]
+                        if x in scopes[use_bb].used_parent:
+                            use = scopes[use_bb].used_parent[x]
+                            if use.kind not in (UseKind.BORROW, UseKind.COPY):
+                                consumed_succs.add(succ)
+                    if consumed_succs:
                         assert bb.branch_pred is not None
                         [left_succ, _] = bb.successors
                         err.add_sub_diagnostic(
                             PlaceNotUsedError.Branch(
-                                bb.branch_pred, x in live_before[left_succ]
+                                bb.branch_pred,
+                                left_succ in consumed_succs,
                             )
                         )
                     err.add_sub_diagnostic(PlaceNotUsedError.Fix(None))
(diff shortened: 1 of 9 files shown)


maximilianruesch:
Hi @rmfisher26, whenever you create a PR for an issue, please use GitHub's [closing keywords](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword) in the PR body (instead of e.g. "Addresses") to auto-link the issue. This makes it easier for us to discover that work has already begun!

PR #1499, opened 2026-02-16
fix: improve docstring formatting array methods

CalMacCQ:
Previously this file used "```" instead of the `.. code-block::` sphinx directive meaning the code snippet did not render properly in the docstrings

Before 
Screenshot 2026-02-16 at 10 49 00

After

Screenshot 2026-02-16 at 10 46 49

### DIFF
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -94,9 +94,6 @@ class array(builtins.list[_T], Generic[_T, _n]):
         This is the case whenever a non-copyable element is borrowed, or when an element
         is manually taken out of the array via the `take` method.
 
-        Example
-        -------
-
         .. code-block:: python
 
             qs = array(qubit() for _ in range(10))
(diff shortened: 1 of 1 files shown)


guppylang/src/guppylang/std/array.py
```diff
@@ -96,15 +96,15 @@ def is_borrowed(self: array[L, n], idx: int) -> bool:
 
         # Example
```
aborgna-q: In your screenshot this isn't being formatted either.

Is there a subheader incantation for sphinxs?

CalMacCQ: Screenshot 2026-02-16 at 11 19 06

I can make them headings by doing


```
Example
======
```

or 

```
Example
^^^^^^
```

however I think this doesn't look so nice.

Screenshot 2026-02-16 at 11 19 06


I think its cleaner without these headings so I've removed them for now.

PR #1501, opened 2026-02-16
feat: Improve error message for failed borrows due to failed leaf borrows 

### LINKED ISSUE #1446: Improve error message for failed borrows due to failed leaf borrows
Consider the below example:
```python
from guppylang import guppy
from guppylang.std.builtins import array, qubit, comptime
from guppylang.std.quantum import discard_array
from typing import Generic

n_qubits = guppy.nat_var("n_qubits")

@guppy.struct
class Register(Generic[n_qubits]):
    register: array[qubit, n_qubits]

    @guppy
    def anything(self) -> bool:
        return True

@guppy
def my_foo(main_register: Register[4]) -> None:
    register = main_register.register
    if main_register.anything():
       pass

b = 4

@guppy
def main() -> None:
    qs = array(qubit() for _ in range(comptime(b)))
    reg = Register(qs)
    my_foo(reg)
    qs = reg.register
    discard_array(qs)

main.check();
```
The statement `main_register.anything()` will not compile, since `main_register` has been blocked from borrows due to leaf nodes (here `.register`) being borrowed and not being returned yet:
```
Error: Copy violation (at <In[2]>:19:7)
   | 
17 | def my_foo(main_register: Register[4]) -> None:
18 |     register = main_register.register
19 |     if main_register.anything():
   |        ^^^^^^^^^^^^^ Field `main_register.register` with non-copyable type
   |                      `array[qubit, 4]` cannot be borrowed ...

Note:
   | 
17 | def my_foo(main_register: Register[4]) -> None:
18 |     register = main_register.register
   |                ---------------------- Field `main_register.register` already moved here

```

However, the error message is quite confusing, since we are not trying to borrow `main_register.register` here. The message should be improved to be clearer. Restating @mark-koch:
> Re error message, I think it would be better to say “Value main_register with non-copyable type Register cannot be borrowed … since its field register was already moved here”?

rmfisher26:
Closes #1446 

Summary

- When a parent struct/variable is used but a child field was already moved, the error now reports the violation on the parent place (e.g. `Variable 's'`) instead of the leaf (e.g. `Field 's.q'`), and adds a note explaining which specific member caused the conflict
- In `linearity.py`, added a new `FieldPrevUse `sub-diagnostic to `AlreadyUsedError `that displays the conflicting member's name, its non-copyable type, and how it was previously used (e.g. "`... since the member 's.q' with non-copyable type 'qubit' was already returned here`")
- In `linearity_checker.py`, updated `visit_PlaceNode` to distinguish between two cases when a leaf place has already been used:
- **Parent differs from leaf** (``node.place != place``): reports the error on the user's original expression (`node.place`) and attaches the new `FieldPrevUse `note pointing to the specific child that was already moved
- **Parent is the leaf** (direct re-use): preserves the existing behavior using `PrevUse`
- Adds a new test case (`field_moved_parent_borrow`) covering the scenario where a field is moved and then the parent is borrowed

Tests

-  Updated expected `.err` output for `field_copy3`, `field_copy_nested1`, `field_copy_nested2`, `field_copy_nested4`, `tuple_use_index2`
-  Added new test `field_moved_parent_borrow`

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/errors/linearity.py
+++ b/guppylang-internals/src/guppylang_internals/checker/errors/linearity.py
@@ -39,7 +39,7 @@ class AlreadyUsedError(Error):
         prev_kind: UseKind
 
     @dataclass(frozen=True)
-    class FieldPrevUse(Note):
+    class MemberPrevUse(Note):
         span_label: ClassVar[str] = (
             "... since the member `{child_place}` with non-copyable type "
             "`{child_place.ty}` was already {prev_kind.subjunctive} here"
--- a/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
@@ -287,7 +287,7 @@ class BBLinearityChecker(ast.NodeVisitor):
                     if node.place != place:
                         err = AlreadyUsedError(node, node.place, use_kind)
                         err.add_sub_diagnostic(
-                            AlreadyUsedError.FieldPrevUse(
+                            AlreadyUsedError.MemberPrevUse(
                                 prev_use.node, place, prev_use.kind
                             )
                         )
--- a/tests/error/linear_errors/tuple_moved_parent_borrow.err
+++ b/tests/error/linear_errors/tuple_moved_parent_borrow.err
@@ -1,0 +1,16 @@
+Error: Copy violation (at $FILE:21:7)
+   | 
+19 |     s = S((qubit(), 100))
+20 |     use(s.t[0])
+21 |     if s.check():
+   |        ^ Variable `s` with non-copyable type `S` cannot be borrowed
+   |          ...
+
+Note:
+   | 
+19 |     s = S((qubit(), 100))
+20 |     use(s.t[0])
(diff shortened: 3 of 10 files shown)


guppylang-internals/src/guppylang_internals/checker/errors/linearity.py
```diff
@@ -38,6 +38,15 @@ class PrevUse(Note):
         )
         prev_kind: UseKind
 
+    @dataclass(frozen=True)
+    class FieldPrevUse(Note):
```
maximilianruesch: Its not necessarily a `Field`, it may also be a subscript / tuple access (see `leaf_places` function definition). Could you add a test to explore this, and rename this note to `MemberPrevUse`?

maximilianruesch (changes requested):
Very nice PR! One comment regarding a rename and a missing test 👍

guppylang-internals/src/guppylang_internals/checker/errors/linearity.py
rmfisher26: Sure, i have renamed to `MemberPrevUse` and added some additional tests as well

PR #1510, opened 2026-02-18
fix!: Revert experimental features break

maximilianruesch:
This reverts commit de0a3188653ca47c74301bfdfd8f623137ac43b2, to allow a way easier patch release of `guppylang`.

An issue will be created to reapply this update once we want to include a breaking change for `guppylang`.

BREAKING CHANGE: Revert the experimental features breaking change from the previous release.

BEGIN_COMMIT_OVERRIDE
""
END_COMMIT_OVERRIDE


aborgna-q (approved):
Should we do the `-internals` release before adding the commit override directive?

PR #1511, opened 2026-02-18
ci: Check lowest py dependencies

aborgna-q:
Checks that the library works with `uv --resolution=lowest-direct` in the release checks.

Requires some fixes/version bumps for tests to pass.

drive-by: Remove test enabling the `tket` extra, since it's now a no-op (#1440).

### DIFF
--- a/.github/workflows/release-checks.yml
+++ b/.github/workflows/release-checks.yml
@@ -25,10 +25,9 @@ jobs:
     # Check if we are running on a release PR, targeting some branch (sits in the middle of the release branch name)
     #
     # release-please always uses this branch name.
-    # TODO: Removed if for testing, re-enable before merging
-    # if: |
-    #  github.event_name == 'workflow_dispatch' ||
-    #  (startsWith(github.head_ref, 'release-please--') && endsWith(github.head_ref, '--components--guppylang-internals'))
+    if: |
+     github.event_name == 'workflow_dispatch' ||
+     (startsWith(github.head_ref, 'release-please--') && endsWith(github.head_ref, '--components--guppylang-internals'))
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/release-checks.yml
```diff
@@ -54,18 +62,3 @@ jobs:
         run: uv run --no-sync ruff check guppylang
       - name: Run tests
         run: uv run --no-sync pytest
```
maximilianruesch: If there is a `--no-sync` env variable, can we enable parallelism for tests as a drive by here?

.github/workflows/release-checks.yml
```diff
@@ -14,18 +14,24 @@ on:
...
   py-release:
-    name: Check `guppylang` release compatibility
+    name: Check `guppylang` release compatibility with ${{ matrix.target.resolution }} dependencies
     runs-on: ubuntu-latest
     # Check if we are running on a release PR.
     #
     # release-please always uses this branch name.
-    if: ${{ github.event_name == 'workflow_dispatch' || github.head_ref == 'release-please--branches--main--components--guppylang' }}
+    # TODO: Removed if for testing, re-enable before merging
+    #if: ${{ github.event_name == 'workflow_dispatch' || github.head_ref == 'release-please--branches--main--components--guppylang' }}
```
maximilianruesch: I think this should also be run when release please targets non-main branches. The check could maybe be: starts with `release-please` and ends with `components--guppylang`?

maximilianruesch:
Issues found
---
Some of the current checks rightfully fail (tests even fail to execute), since there is an API mismatch between the latest released `guppylang-internals` and the local `guppylang` (which is already adapted to the unreleased one). 

 -> Perhaps we should include release checks for `guppylang-internals` as well, where both packages are installed from local source with lowest-direct and highest dependencies, to catch dependency issues earlier?

--- 
A mypy check fails because there are missing generic parameters for a `numpy` type. Our compatible versions range from `numpy==2.0.0` to `numpy==2.4.2`, and we would have to raise the lower bound to `numpy==2.2.0` to satisfy mypy here. Adding a `type: ignore` is infeasible since it will be flagged by any other mypy check that is running in an environment with a newer numpy version.

-> Unclear what the best strategy is. Bumping numpy by two minors just because of a mypy complaint that does nothing at runtime seems a bit overkill.

.github/workflows/release-checks.yml
```diff
@@ -14,18 +14,26 @@ on:
...
-    name: Check `guppylang` release compatibility
+  check-release-guppylang-internals:
+    name: Check `guppylang-internals` release compatibility with ${{ matrix.target.resolution }} dependencies
     runs-on: ubuntu-latest
-    # Check if we are running on a release PR.
+    # Check if we are running on a release PR, targeting some branch (sits in the middle of the release branch name)
     #
     # release-please always uses this branch name.
-    if: ${{ github.event_name == 'workflow_dispatch' || github.head_ref == 'release-please--branches--main--components--guppylang' }}
+    # TODO: Removed if for testing, re-enable before merging
```
aborgna-q: Don't forget to undo this

PR #1513, opened 2026-02-18
feat: (WIP) add metadata and guppy kwarg for inline="always"

### LINKED ISSUE #1054: Inlining annotations
For example, `@guppy(inline=True)`. Or maybe `@guppy(inline="always")` vs `@guppy(inline="never")`?

Could also consider an extra decorator:

```python
@guppy
@always_inline
def foo(): ...
```

Releated: #567

Implementation:

* The most general way to do this in the Guppy compiler could be a new `annotations` field on `Definition`?
* Should lower to Hugr metatdata

acl-cqc:
* We should migrate to the new hugr typed metadata; this might be a good opportunity
* tket2 change to use the metadata not done yet and may be more controversial, so should probably be merged first

closes #1054


aborgna-q:
https://github.com/Quantinuum/tket2/pull/1535 defines a `hugr.metadata.Metadata` for inline hints. We should use that here

acl-cqc:
Believe the right way to do this is now #1976

PR #1518, opened 2026-02-19
feat!: Add enum type

nicolaassolini-qntm:
This pull request introduces enums in guppy, with all the compilation pipeline required, as well as a refactor to avoid code duplication.

* Replaced references to `StructField` with a new, more general `CheckedField` type across the codebase, allowing for more flexible handling of both struct fields and enum variants in diagnostics and error messages. 

* Improved error messages for attribute access, especially for enums, by distinguishing between "field or method" and "variant or method" in the `AttributeNotFoundError`.

* Added support for the instantiation of Enum and the use of Enum methods.

* Added test cases for Enum 

BREAKING CHANGES: renamed `guppylang-internals/src/guppylang_internals/tys/printing.py:_visit_OpaqueType_StructType` in `visit_OpaqueType_StructType_EnumType`

BREAKING CHANGES: removed default value for `class_type`, field of `DuplicateFieldError` in `guppylang_internals/definition/util.py`

BREAKING CHANGE: Moved shared parsing logic for classes (`parse_py_class`, `extract_generic_params`, `try_parse_generic_base`, `extract_generic_params`, `params_from_ast`) into a new utility module, `definition/util.py`, and updated both `enum.py`, `struct.py` and `function.py` to use these utilities. 

BREAKING CHANGE: Moved error classes and parsing functions from `struct.py` and `function.py` (`RepeatedTypeParamError`, `DuplicateFieldError`, `RedundantParamsError`)

BREAKING CHANGE: Defined UncheckedField and CheckedField instead of UncheckedStructField and StructField

Not a `guppylang` breaking change

---

BEGIN_COMMIT_OVERRIDE
feat: Add enum type (#1518)
END_COMMIT_OVERRIDE

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/struct.py
+++ b/guppylang-internals/src/guppylang_internals/definition/struct.py
@@ -6,7 +6,7 @@ from typing import ClassVar
 
 from hugr import Wire, ops
 
-from guppylang_internals.ast_util import AstNode, annotate_location, parse_source
+from guppylang_internals.ast_util import AstNode
 from guppylang_internals.checker.core import Globals
 from guppylang_internals.checker.errors.generic import (
     UnexpectedError,
@@ -18,7 +18,6 @@ from guppylang_internals.definition.common import (
     CompiledDef,
     DefId,
     ParsableDef,
-    UnknownSourceError,
     UserProvidedLinkName,
 )
 from guppylang_internals.definition.custom import (
@@ -26,7 +25,6 @@ from guppylang_internals.definition.custom import (
     CustomFunctionDef,
     DefaultCallChecker,
 )
-from guppylang_internals.definition.parameter import ParamDef
 from guppylang_internals.definition.ty import TypeDef
 from guppylang_internals.diagnostic import Help
 from guppylang_internals.engine import DEF_STORE
--- a/guppylang-internals/src/guppylang_internals/definition/function.py
+++ b/guppylang-internals/src/guppylang_internals/definition/function.py
@@ -41,7 +41,6 @@ from guppylang_internals.definition.common import (
     UserProvidedLinkName,
 )
 from guppylang_internals.definition.metadata import GuppyMetadata, add_metadata
-from guppylang_internals.definition.util import parse_source
 from guppylang_internals.definition.struct import ParsedStructDef
 from guppylang_internals.definition.value import (
     CallableDef,
--- a/guppylang-internals/src/guppylang_internals/definition/util.py
+++ b/guppylang-internals/src/guppylang_internals/definition/util.py
@@ -7,7 +7,7 @@ from dataclasses import dataclass
 from types import FrameType
 from typing import ClassVar
 
-from guppylang_internals.ast_util import annotate_location
+from guppylang_internals.ast_util import annotate_location, parse_source
 from guppylang_internals.checker.core import Globals
 from guppylang_internals.checker.errors.generic import (
     ExpectedError,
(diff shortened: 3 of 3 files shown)


guppylang-internals/src/guppylang_internals/definition/enum.py
```diff
@@ -0,0 +1,398 @@
...
+                # TODO: do we allow `pass` statements to define empty enum?
+                case _, ast.Pass():
+                    pass
+                # Docstrings are also fine if they occur at the start
+                case 0, ast.Expr(value=ast.Constant(value=v)) if isinstance(v, str):
+                    pass
+                case _, ast.FunctionDef(name=name) as node:
+                    used_func_names[name] = node
+                # Enum variant are declared via dictionary, where key are the variant
+                # fields and values are types;
```
tatiana-s: nit:
```suggestion
                # Enum variants are declared via a dictionary, where keys are the variant
                # fields and values are types:
```

guppylang-internals/src/guppylang_internals/definition/enum.py
```diff
@@ -0,0 +1,398 @@
...
+                            DuplicateVariantError(
+                                node.targets[0], self.name, variant_name
+                            )
+                        )
+                    assert isinstance(node.value, ast.Dict)  # for mypy
+                    variants[variant_name] = parse_enum_variant(
+                        variant_index, variant_name, node.value
+                    )
+                    variant_index += 1
+                # if unexpected statement are found
```
tatiana-s: nit:
```suggestion
                # If unexpected statements are found
```

guppylang-internals/src/guppylang_internals/definition/struct.py
```diff
@@ -172,30 +94,43 @@ def parse(self, globals: Globals, sources: SourceMap) -> "ParsedStructDef":
 
                     v = getattr(self.python_class, name)
                     if not isinstance(v, GuppyDefinition):
-                        raise GuppyError(NonGuppyMethodError(node, self.name, name))
+                        raise GuppyError(
+                            NonGuppyMethodError(node, self.name, name, "struct")
```
tatiana-s, line 99: I know this is also the case with some other existing errors, but looking at this it doesn't seem that nice to have to know whether something you pass to the error needs to be capitalised or not, I wonder whether that could be adjusted in the error maybe, but this is an optional suggestion

guppylang/src/guppylang/decorator.py
```diff
@@ -218,6 +219,39 @@ def add_fields(self: "MyStruct") -> int:
...
+            @guppy.enum
+            class MyEnum:
+                Variant1 = {"a": int, "b": qubit}
+                Variant2 = {"a": int}
+
+                @guppy
+                def method_on_enum(e: MyEnum) -> int:
+                    return e.a + 1
+        ..
+        Enum supposed to have variants as class attributes.
```
tatiana-s:
```suggestion
        Enums have variants as class attributes.
```
But also seems a bit redundant after the example

guppylang/src/guppylang/decorator.py
```diff
@@ -218,6 +219,39 @@ def add_fields(self: "MyStruct") -> int:
         # a `GuppyDefinition` that handles the comptime logic
         return GuppyDefinition(defn)  # type: ignore[return-value]
 
+    @dataclass_transform()
+    def enum(self, cls: builtins.type[T]) -> builtins.type[T]:
```
tatiana-s, line 254: Unfortunately we currently have to manually maintain the decorator docs in the guppylang API due to some issues with sphinx, so as a follow up you should also update the docs with the same text as the comment https://github.com/Quantinuum/guppy-docs/blob/main/sphinx/api/decorator.md

tests/error/enum_errors/bad_variant_def1.err
```diff
@@ -0,0 +1,11 @@
+Error: Unexpected statement (at $FILE:6:4)
+  | 
+4 | @guppy.enum
+5 | class MyEnum:
+6 |     v1 = 20
+  |     ^^^^^^^ Unexpected statement in enum definition
+
+Help: Enum can contain only variants of the form `VariantName = {"var1": Type1,
```
tatiana-s:
```suggestion
Help: Enums can only contain variants of the form `VariantName = {"var1": Type1,
```

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -723,8 +725,72 @@ def transform(self, transformer: Transformer) -> "Type":
...
+            EnumVariant(
+                variant.index,
+                variant.name,
+                [CheckedField(f.name, f.ty.transform(inst)) for f in variant.fields],
+            )
+            for variant in variants_list
+        ]
+
+    @cached_property
+    def variant_as_dict(self) -> Mapping[str, "EnumVariant[CheckedField]"]:
```
tatiana-s: Forgot one:

```suggestion
    def variants_as_dict(self) -> Mapping[str, "EnumVariant[CheckedField]"]:
```

for consistency with the other method

guppylang-internals/src/guppylang_internals/definition/struct.py
nicolaassolini-qntm: I can force the capitalisation in the error

PR #1525, opened 2026-02-23
chore!: Update `hugr-py` dependency incl. switching to core metadata 

tatiana-s:
BREAKING CHANGE: (guppylang-internals) `CoreMetadataKeys` removed, use keys from `hugr.metadata` instead

---

BEGIN_COMMIT_OVERRIDE
chore: Update hugr-py dependency incl. switching to core metadata (#1525)
END_COMMIT_OVERRIDE

### DIFF
--- a/guppylang-internals/src/guppylang_internals/compiler/core.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/core.py
@@ -7,7 +7,6 @@ from typing import TYPE_CHECKING, Any, ClassVar, cast
 
 import tket_exts
 from hugr import Hugr, Node, Wire, ops
-from hugr import ext as he
 from hugr import tys as ht
 from hugr.build import function as hf
 from hugr.build.dfg import DP, DefinitionBuilder, DfBase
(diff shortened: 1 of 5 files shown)


guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -322,26 +317,20 @@ def compile(self, id: DefId) -> ModulePointer:
...
             for ext in used_extensions_result.used_extensions.extensions.values()
         ]
         # Add unresolved extensions as well, but we only have the names
         used_exts_meta.extend(
-            {
-                "name": ext,
-            }
-            for ext in used_extensions_result.unresolved_extensions
+            [
+                ExtensionDesc(name=ext_name, version=Version.parse("0.0.0"))
```
tatiana-s: I am not sure if there is something better to set it to / whether this will cause issues, but it is not possible to just leave it empty like it was previously

aborgna-q: Uhm, that's annoying.

We could go with `Version(major=0, prerelease="unknown")` or something for now, and see if we can make the field optional in `hugr`.

PR #1527, opened 2026-02-23
feat: Improve error message for passing non-comptime values into comptime

### LINKED ISSUE #908: Improve error message for passing non-comptime values into `comptime`
In the following code `comptime` and `angle` are composed in the wrong order. This leads to an error as the `angle` type cannot be passed into comptime.

```python
from guppylang import guppy
from guppylang.std.builtins import comptime
from guppylang.std.quantum import cx, rz, qubit


@guppy
def phase_gadget(q0: qubit, q1: qubit) -> None:
    cx(q0, q1)
    rz(q1, comptime(angle(0.5)))
    cx(q0, q1)

phase_gadget.compile()
```

I think the error should state explicitly which types are valid inputs to the `comptime` function.

Current error...

```
---------------------------------------------------------------------------
GuppyComptimeError                        Traceback (most recent call last)
File ~/quantinuum/guppylang/guppylang-internals/src/guppylang_internals/checker/expr_checker.py:1368, in eval_comptime_expr(node, ctx)
   1367 try:
-> 1368     python_val = eval(ast.unparse(node.value), DummyEvalDict(ctx, node.value))  # noqa: S307
   1369 except DummyEvalDict.GuppyVarUsedError as e:

File <string>:1

File ~/quantinuum/guppylang/guppylang-internals/src/guppylang_internals/tracing/util.py:41, in hide_trace.<locals>.wrapped(*args, **kwargs)
     40 with exception_hook(tracing_except_hook):
---> 41     return f(*args, **kwargs)

File ~/quantinuum/guppylang/guppylang-internals/src/guppylang_internals/tracing/object.py:504, in TracingDefMixin.__call__(self, *args)
    503 if not tracing_active():
--> 504     raise GuppyComptimeError(
    505         f"{self.wrapped.description.capitalize()} `{self.wrapped.name}` may "
    506         "only be called in a Guppy context"
    507     )
    509 defn = ENGINE.get_checked(self.wrapped.id)

GuppyComptimeError: Type `angle` may only be called in a Guppy context

The above exception was the direct cause of the following exception:

GuppyError                                Traceback (most recent call last)
    [... skipping hidden 1 frame]

Cell In[8], line 13
     10     rz(q1, comptime(angle(0.5)))
     11     cx(q0, q1)
---> 13 phase_gadget.compile()

GuppyError: ComptimeExprEvalError(span=<ast.Call object at 0x11078c130>, _parent=None, children=[], err='Traceback (most recent call last):\n  File "<string>", line 1, in <module>\n  File "/Users/maximilian.ruesch/quantinuum/guppylang/guppylang-internals/src/guppylang_internals/tracing/util.py", line 41, in wrapped\n    return f(*args, **kwargs)\n  File "/Users/maximilian.ruesch/quantinuum/guppylang/guppylang-internals/src/guppylang_internals/tracing/object.py", line 504, in __call__\n    raise GuppyComptimeError(\nguppylang_internals.error.GuppyComptimeError: Type `angle` may only be called in a Guppy context\n')
```

rmfisher26:
## Summary

- Add a new `ComptimeGuppyObjectError` diagnostic that fires when a Guppy type or function is called inside a `comptime()` expression
- Fix `sys.excepthook `being overridden by `@hide_trace` during comptime eval, which previously caused the new error to render as a raw Python traceback instead of a formatted Guppy error
- Add two snapshot error tests covering both cases

**Motivation**
Closes #908 

## Details

When a user writes `comptime(angle(0.5))`, Python eagerly evaluates `angle(0.5)` before passing the result to `comptime`. Since `angle `is a Guppy struct wrapped in `TracingDefMixin`, calling it outside a tracing context raises a `GuppyComptimeError`. Previously this surfaced as a generic "Python error" with a raw traceback — the actual message ("Type `angle `may only be called in a Guppy context") provided no guidance on what types are valid.   

The except `GuppyComptimeError` handler now restores `sys.excepthook` before re-raising as a `GuppyError`, so the pretty-printing hook handles it correctly.

The fix adds a dedicated `except GuppyComptimeError` handler in `eval_comptime_expr `that produces a clearer diagnostic:

**Tests**

- `tests/error/comptime_expr_errors/guppy_type_in_comptime.py` — calling a Guppy struct type (`angle`) inside `comptime()`

```
Error: Invalid comptime expression (at $FILE:8:20)
  | 
6 | @compile_guppy
7 | def foo() -> int:
8 |     return comptime(angle(0.5))
  |                     ^^^^^^^^^^ Cannot use a Guppy object in a comptime expression

Type `angle` may only be called in a Guppy context

Comptime expressions are evaluated as plain Python. Guppy functions and types
require a Guppy context and cannot be called inside `comptime`.

```

- ` tests/error/comptime_expr_errors/guppy_func_in_comptime.py` — calling a `@guppy` function inside` comptime()`

```
Error: Invalid comptime expression (at $FILE:13:20)
   | 
11 | @compile_guppy
12 | def foo() -> int:
13 |     return comptime(my_func())
   |                     ^^^^^^^^^ Cannot use a Guppy object in a comptime expression

Function `my_func` may only be called in a Guppy context

Comptime expressions are evaluated as plain Python. Guppy functions and types
require a Guppy context and cannot be called inside `comptime`.

```

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -1367,24 +1367,23 @@ def eval_comptime_expr(node: ComptimeExpr, ctx: Context) -> Any:
     if sys.implementation.name != "cpython":
         raise GuppyError(ComptimeExprNotCPythonError(node))
 
-    # Ensure that any modifications to sys.excepthook performed in the `eval` call (e.g. through 
-    # `@hide_trace`-decorated Guppy callables) are rolled back to the current exception hook
-    # when `eval` exits. Without this, exceptions raised during `eval` may cause the sys.excepthook
-    # to render raw tracebacks instead of formatted diagnostics.
-    with saved_exception_hook():
-        try:
+    # Ensure that any modifications to sys.excepthook performed in the `eval` call
+    # (e.g. through `@hide_trace`-decorated Guppy callables) are rolled back to the
+    # current exception hook when `eval` exits. Without this, exceptions raised
+    # during `eval` may cause the sys.excepthook to render raw tracebacks instead
+    # of formatted diagnostics.
+    try:
+        with saved_exception_hook():
             python_val = eval(ast.unparse(node.value), DummyEvalDict(ctx, node.value))  # noqa: S307
-        except DummyEvalDict.GuppyVarUsedError as e:
-            raise GuppyError(
-                ComptimeExprNotStaticError(e.node or node, e.var)
-            ) from None
-        except GuppyComptimeError as e:
-            raise GuppyError(ComptimeGuppyObjectError(node.value, str(e))) from e
-        except Exception as e:
-            # Remove the top frame pointing to the `eval` call from the stack trace
-            tb = e.__traceback__.tb_next if e.__traceback__ else None
-            tb_formatted = "".join(traceback.format_exception(type(e), e, tb))
-            raise GuppyError(ComptimeExprEvalError(node.value, tb_formatted)) from e
+    except DummyEvalDict.GuppyVarUsedError as e:
+        raise GuppyError(ComptimeExprNotStaticError(e.node or node, e.var)) from None
+    except GuppyComptimeError as e:
+        raise GuppyError(ComptimeGuppyObjectError(node.value, str(e))) from e
+    except Exception as e:
+        # Remove the top frame pointing to the `eval` call from the stack trace
+        tb = e.__traceback__.tb_next if e.__traceback__ else None
+        tb_formatted = "".join(traceback.format_exception(type(e), e, tb))
+        raise GuppyError(ComptimeExprEvalError(node.value, tb_formatted)) from e
     return python_val
 
 


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1364,10 +1366,18 @@ def eval_comptime_expr(node: ComptimeExpr, ctx: Context) -> Any:
...
+    # and functions) may override it with `tracing_except_hook` if they raise a
+    # GuppyComptimeError during eval. We need to restore it so that the GuppyError
+    # we raise below is handled by the correct (pretty-printing) hook.
+    saved_excepthook = sys.excepthook
     try:
         python_val = eval(ast.unparse(node.value), DummyEvalDict(ctx, node.value))  # noqa: S307
     except DummyEvalDict.GuppyVarUsedError as e:
         raise GuppyError(ComptimeExprNotStaticError(e.node or node, e.var)) from None
+    except GuppyComptimeError as e:
+        sys.excepthook = saved_excepthook
```
maximilianruesch: How was this not an issue in the previous except branch (for used variables)? Does `str(e)` utilize this hook under the hood? I am struggling to understand where it is used.

guppylang-internals/src/guppylang_internals/checker/errors/comptime_errors.py
```diff
@@ -45,6 +45,19 @@ class ComptimeExprEvalError(Error):
...
 
+@dataclass(frozen=True)
+class ComptimeGuppyObjectError(Error):
+    title: ClassVar[str] = "Invalid comptime expression"
+    span_label: ClassVar[str] = "Cannot use a Guppy object in a comptime expression"
+    message: ClassVar[str] = (
+        "{err_message}\n\n"
+        "Valid comptime expressions must evaluate to a plain Python value. "
+        "Supported types are: `bool`, `str`, `int`, `float`, `tuple`, `list`, "
+        "and `None`."
```
maximilianruesch: Is there any reference for this list? Anything one can reduce to a single source of truth, or at least point to it for documentation here?

guppylang-internals/src/guppylang_internals/checker/errors/comptime_errors.py
```diff
@@ -45,6 +45,19 @@ class ComptimeExprEvalError(Error):
     err: str
 
 
+@dataclass(frozen=True)
+class ComptimeGuppyObjectError(Error):
+    title: ClassVar[str] = "Invalid comptime expression"
+    span_label: ClassVar[str] = "Cannot use a Guppy object in a comptime expression"
+    message: ClassVar[str] = (
+        "{err_message}\n\n"
+        "Valid comptime expressions must evaluate to a plain Python value. "
```
maximilianruesch: Imo the main point is that the value must be evaluable in Python, not that it evaluates to a plain Python value. The problem raising the error is that the function that is called is not callable from pure Python, so the expression cannot be evaluated there.

Similar to the other comment for this object, we should keep this consistent with other sources about what comptime supports. In particular, we should be wary about adding an error message that we need to update and maintain when increasing support in Guppy for more involved data structures, e.g. when adding (limited) support for numpy arrays.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
rmfisher26: ### Summary:

The reason it's only needed in the `GuppyComptimeError `branch (not `GuppyVarUsedError `or the generic Exception branch) is that only `TracingDefMixin.__call__` (via `@hide_trace`) modifies `sys.excepthook`. `DummyEvalDict.GuppyVarUsedError` and ordinary Python exceptions don't go through `@hide_trace`, so the hook is never touched in those paths.

#### Observation 1:  GuppyVarUsedError  does not use @hide_trace
`GuppyVarUsedError ` is raised directly inside `DummyEvalDict.__getitem__`  method with no Guppy decorators. `@hide_trace` is never entered, so `sys.excepthook` is never touched.

_`guppylang-internals/src/guppylang_internals/checker/core.py`_
```
 def __getitem__(self, key: str) -> Any:
        self._check_item(key)
        return super().__getitem__(key)
```
#### Observation 2:  GuppyComptimeError  uses @hide_trace 
`GuppyComptimeError `passes through `@hide_trace` on `TracingDefMixin.__call__`, which wraps with `exception_hook(tracing_except_hook)`. When the exception propagates, the hook is left overwritten.

When `eval `is called, this invokes `GuppyDefinition.__call__ -->  TracingDefMixin.__call__` (which has the `@hide_trace` decorator) as seen below:

_`guppylang-internals/src/guppylang_internals/tracing/object.py`_

```
@dataclass(frozen=True)
class TracingDefMixin(DunderMixin):
    """Mixin to provide tracing semantics for definitions."""

    wrapped: Definition

    @property
    def id(self) -> DefId:
        return self.wrapped.id

    @hide_trace
    def __call__(self, *args: Any) -> Any:
        from guppylang_internals.tracing.function import trace_call

        if not tracing_active():
            raise GuppyComptimeError(
                f"{self.wrapped.description.capitalize()} `{self.wrapped.name}` may "
                "only be called in a Guppy context"
            )

        defn = ENGINE.get_checked(self.wrapped.id)
        if isinstance(defn, CallableDef):
            return trace_call(defn, *args)
        elif (
            isinstance(defn, TypeDef)
            and defn.id in DEF_STORE.impls
            and "__new__" in DEF_STORE.impls[defn.id]
        ):
            constructor = DEF_STORE.raw_defs[DEF_STORE.impls[defn.id]["__new__"]]
            return TracingDefMixin(constructor)(*args)
        err = f"{defn.description.capitalize()} `{defn.name}` is not callable"
        raise GuppyComptimeError(err)
```

#### Observation 3: : The code for the @hide_trace is here: 
From `_guppylang-internals/src/guppylang_internals/tracing/util.py_`

```
def hide_trace(f: Callable[P, T]) -> Callable[P, T]:
    """Function decorator that hides compiler-internal frames from the traceback of any
    exception thrown by the decorated function."""

    @functools.wraps(f)
    def wrapped(*args: P.args, **kwargs: P.kwargs) -> T:
        with exception_hook(tracing_except_hook):
            return f(*args, **kwargs)

    return wrapped
```

guppylang-internals/src/guppylang_internals/checker/errors/comptime_errors.py
rmfisher26: This was based on from `python_value_to_guppy_type` from `expr_checker.py`  which mentions those types. So I did update that  `ComptimeGuppyObjectError ` to remove those specific types now and modified the error message as mentioned in the below comment. If I should modify this further just let me know!

guppylang-internals/src/guppylang_internals/checker/errors/comptime_errors.py
rmfisher26: I did update the `guppylang-internals/src/guppylang_internals/checker/errors/comptime_errors.py` to remove the specific types and rephrased the message. If this message is not necessary, it can also be removed.

```
class ComptimeGuppyObjectError(Error):
    title: ClassVar[str] = "Invalid comptime expression"
    span_label: ClassVar[str] = "Cannot use a Guppy object in a comptime expression"
    message: ClassVar[str] = (
        "{err_message}\n\n"
        "Comptime expressions are evaluated as plain Python. Guppy functions "
        "and types require a Guppy context and cannot be called inside `comptime`."
    )
    err_message: str

```

maximilianruesch: There is some duplication with the inner error message (see your tests for example) with the inner thing requiring a Guppy context. I think its still fine though, and now the message is more on point while requiring less maintenance. Good job!

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
doug-q: I admit I don't entirely understand, but I do think this is pretty gross.

It couples implementation details here with implementation details in hide_trace.

Is there another way to factor this that isn't so fragile?

maximilianruesch: There is something fishy going on here. The hook is modified by `@hide_trace`, but usually its reset to its old value. It seems like now the hook change inside the `eval` persists for some reason.

If we find out why and fix it, we should get away without manually resetting it here.

maximilianruesch: Ah, so the context manager from `@hide_trace` simply does not reset the exception hook when an exception is raised inside it, which makes sense since you want that exception to be handled using the exception hook. However, we want to reset the exception hook here since we do not want the changes from inside an `eval` context to persist.

I propose a more general solution. Add a context manager that simply saves the current exception hook, yields inside a try-finally block and resets it to the previous. Then wrap the eval call in `expr_checker.py` in this context manager. So its still manual, just more maintainable.

Something like
```
@contextmanager
def saved_exception_hook():
    old_hook = sys.excepthook
    try:
        yield
    finally:
        sys.excepthook = old_hook
```
with some documentation around when it is applicable. This should be clearer than what is going on, and still be usable for branches / exceptions that do not need to go through `@hide_trace`.

rmfisher26: Thank you both for your review and recommendations! I will let you know when I have an update on this solution.

rmfisher26: I have added the `contextmanager `and added some documentation.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1364,15 +1367,26 @@ def eval_comptime_expr(node: ComptimeExpr, ctx: Context) -> Any:
...
-        tb = e.__traceback__.tb_next if e.__traceback__ else None
-        tb_formatted = "".join(traceback.format_exception(type(e), e, tb))
-        raise GuppyError(ComptimeExprEvalError(node.value, tb_formatted)) from e
+    # Wrap the eval in saved_exception_hook so that any changes to sys.excepthook made
+    # by `@hide_trace`-decorated Guppy callables (e.g. types and functions) are rolled
+    # back when the block exits. Without this, a GuppyComptimeError propagating through
+    # @hide_trace would leave `tracing_except_hook` as sys.excepthook, causing the
+    # GuppyError raised below to render as a raw Python traceback instead of a formatted
+    # Guppy error.
+    with saved_exception_hook():
```
maximilianruesch: The intention was to wrap literally just the eval statement, not the whole block. It should still work out as expected. This is to reduce indentation and clarify the impact of exception hook saving.

guppylang-internals/src/guppylang_internals/error.py
```diff
@@ -76,6 +76,21 @@ def ipython_excepthook(
     sys.excepthook = old_hook
 
 
+@contextmanager
+def saved_exception_hook() -> Iterator[None]:
```
maximilianruesch, line 80: Nice function, and nice docs!

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1364,15 +1367,26 @@ def eval_comptime_expr(node: ComptimeExpr, ctx: Context) -> Any:
...
-        # Remove the top frame pointing to the `eval` call from the stack trace
-        tb = e.__traceback__.tb_next if e.__traceback__ else None
-        tb_formatted = "".join(traceback.format_exception(type(e), e, tb))
-        raise GuppyError(ComptimeExprEvalError(node.value, tb_formatted)) from e
+    # Wrap the eval in saved_exception_hook so that any changes to sys.excepthook made
+    # by `@hide_trace`-decorated Guppy callables (e.g. types and functions) are rolled
+    # back when the block exits. Without this, a GuppyComptimeError propagating through
+    # @hide_trace would leave `tracing_except_hook` as sys.excepthook, causing the
+    # GuppyError raised below to render as a raw Python traceback instead of a formatted
+    # Guppy error.
```
maximilianruesch: Move this comment with the `with` block inside the `try:`, and it can be shortened to (up to moving the words to satisfy ruff):
```suggestion
    # Ensure that any modifications to sys.excepthook performed in the `eval` call (e.g. through 
    # `@hide_trace`-decorated Guppy callables) are rolled back to the current exception hook
    # when `eval` exits. Without this, exceptions raised during `eval` may cause the sys.excepthook
    # to render raw tracebacks instead of formatted diagnostics.
```

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
maximilianruesch: This is much cleaner, thanks!

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
rmfisher26: Thanks for the clarification! I have updated this as directed.

maximilianruesch (approved):

PR #1532, opened 2026-02-24
feat!: Allow validation of wasm for h2 platform

trvto:
- Adds paramter to `@wasm_module` used to toggle which platform wasm should be validated for
- Signature validation errors mention which platform is used

BREAKING CHANGE: Added wasm platform parameter to `decode_wasm_functions`, `decode_sig` and `decode_type` functions in `guppylang_internals/wasm_util.py`

BREAKING CHANGE: Added wasm platform information to `WasmSignatureError`

BREAKING CHANGE: Added wasm platform parameter to `WasmModuleTypeDef.__init__`

Not a `guppylang` breaking change

### DIFF
--- a/guppylang-internals/src/guppylang_internals/wasm_util.py
+++ b/guppylang-internals/src/guppylang_internals/wasm_util.py
@@ -82,7 +82,7 @@ class WasmSigMismatchError(Error):
         actual: str
 
 
-def decode_type_default(ty: wt.ValType) -> Type | None:
+def decode_type_helios(ty: wt.ValType) -> Type | None:
     if ty == wt.ValType.i64():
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/wasm_util.py
```diff
@@ -74,7 +82,7 @@ class Actual(Note):
         actual: str
 
 
-def decode_type(ty: wt.ValType) -> Type | None:
+def decode_type_default(ty: wt.ValType) -> Type | None:
```
croyzor: This could be `decode_type_helios` now, it's restricted in its own way...

PR #1534, opened 2026-02-26
feat: Allow discarding arrays with taken elements

### LINKED ISSUE #1533: [Feature]: Add method to discard array where elements have been borrowed
### Problem Statement

There is currently no way to discard an array where elements have been borrowed. Consider the following code:
```python
from guppylang import guppy
from guppylang.std.quantum import qubit, discard_array

@guppy
def main() -> None:
    arr = array(qubit())
    q = arr.take(0)
    discard_array(arr)
    q.discard()

main.emulator(n_qubits=1).run()
```

The compiler will throw an error if we do not include `discard_array` but with it included the program will fail at runtime with the following error:
```
guppylang.emulator.exceptions.EmulatorError: Panic (#1002): Array element is already borrowed
```

### Proposed Solution

This can be solved by using `guppylang.std.array._array_discard_all_used` in place of `discard_array`. This functionality should be made public. 

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

hsemenenko:
Closes #1533

### DIFF
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -231,6 +231,25 @@ class array(builtins.list[_T], Generic[_T, _n]):
             qs.discard_all_taken() # Succeeds since all qubits have been taken out
         """
 
+    @guppy
+    @no_type_check
+    def try_discard_all_taken(self: array[L, n] @ owned) -> Result[None, array[L, n]]:
+        """Similar to `discard_all_taken`, but instead of panicking returns the
+        unmodified array when not all elements have been taken out.
+
+        .. code-block:: python
+
+            qs = array(qubit() for _ in range(2))
+            discard(qs.take(0))
+            # Does not panic, even though not all elements have been taken out.
+            qs = qs.try_discard_all_taken()
+        """
+        for i in range(n):
+            if not self.is_borrowed(i):
+                return err(self)
+        self.discard_all_taken()
+        return ok(None)
+
     def __new__(cls, *args: _T) -> builtins.list[_T]:  # type: ignore[no-redef]
         # Runtime array constructor that is used for comptime. We return an actual list
         # in line with the comptime unpacking logic that turns arrays into lists.
--- a/tests/integration/test_array.py
+++ b/tests/integration/test_array.py
@@ -1,7 +1,9 @@
+import pytest
 from hugr import ops
 
 from guppylang import comptime
 from guppylang.decorator import guppy
+from guppylang.emulator import EmulatorError
 from guppylang.std.builtins import array, owned
 from guppylang.std.mem import mem_swap
 from guppylang.std.num import nat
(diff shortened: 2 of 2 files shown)


guppylang/src/guppylang/std/array.py
```diff
@@ -217,6 +217,11 @@ def try_put(self: array[L, n], elem: L @ owned, idx: int) -> Result[None, L]:
         self.put(elem, idx)
         return ok(None)
 
+    @guppy
+    def discard_all_taken(self: array[L, n] @ owned) -> None:
+        """Discards array assuming that all elements been taken out."""
```
mark-koch: Can you add a code example to the docstring similar to the other methods

Also, please document that it will panic if not everything is taken out

guppylang/src/guppylang/std/array.py
```diff
@@ -217,6 +217,11 @@ def try_put(self: array[L, n], elem: L @ owned, idx: int) -> Result[None, L]:
...
+        _array_discard_all_used(self)
```
mark-koch: I think we should delete `_array_discard_all_used` now (non-breaking since private) and just use this method

guppylang/src/guppylang/std/array.py
```diff
@@ -217,6 +217,11 @@ def try_put(self: array[L, n], elem: L @ owned, idx: int) -> Result[None, L]:
         self.put(elem, idx)
         return ok(None)
 
+    @guppy
+    def discard_all_taken(self: array[L, n] @ owned) -> None:
```
mark-koch, line 221: Do we also want a `try_discard_all_taken` that returns a `Result[None, array[L, n]]`?

tests/integration/test_array.py
```diff
@@ -687,6 +687,26 @@ def main() -> int:
...
+        q.discard()
+        discard_array(qubits)  # Here, not all qubits in the array are taken out yet
+
+    res = main.emulator(2).coinflip_sim().run().results[0].entries
+    assert res == [
+        ("before_take", 0),
+        ("after_take", 1),
+    ]
+
+    validate(main.compile())
```
mark-koch: Don't think we need a separate validate here when we're already building the emulator above

tests/integration/test_array.py
```diff
@@ -687,6 +689,72 @@ def main() -> int:
...
+
+def test_discard_all_taken(validate):
+
+    @guppy
+    def main() -> None:
+        qubits = array(qubit(), qubit())
+        qubits.take(0).discard()
+        qubits.take(1).discard()
+        qubits.discard_all_taken()
+        result("after_discard", True)
```
mark-koch, line 718: Are these results with constant values and later assertions needed? The fact that the emulator runs without panic should already tell you that this point is reached?

maximilianruesch: Theoretically yes, but I like to avoid unit tests without any assertions (so called smoke tests). Using the assertion, you not only make sure nothing goes wrong, but you actually also ran the function, so something would have gone wrong if it could.

See also https://stackoverflow.com/questions/137399/unit-testing-without-assertions.

PR #1544, opened 2026-03-03
fix: Resolve very deep recursion inside replaced Jupyter traceback renderer

### LINKED ISSUE #1404: Some errors aren't being displayed in jupyter
It seems that in some cases, when guppy is used in a notebook and raises an error, the bug isn't rendered and the user sees a lot of 
```
Custom TB Handler failed, unregistering
Custom TB Handler failed, unregistering
Custom TB Handler failed, unregistering
```
coming from pygments. Maybe pygments is trying to lex the error message as python?

maximilianruesch:
Fixes the underlying issue of #1404 which was caused by very deep recursion inside the `exception_hook` handler. Shortly, in the Jupyter case, the handler did not reset the hook after successfully yielding, but instead added a lambda wrapper to ignore the first argument. This caused the lambda wrappers to build up, and sufficient calls to functions decorated with `exception_hook` caused the chain to become so deep it exceeded the call stack size.

This also adds an integration notebook to test this, and as a driveby fixes:
1. The display of notebook names in integration tests
2. Automatic collection of integration tested notebooks that are not examples
3. The `errorsdemo.ipynb` which was neglected

Fixes #1404


guppylang-internals/src/guppylang_internals/error.py
```diff
@@ -61,9 +61,8 @@ def ipython_excepthook(
         old_exc_tuple = ipython_shell.custom_exceptions
         ipython_shell.set_custom_exc((Exception,), ipython_excepthook)
         yield
-        ipython_shell.set_custom_exc(
-            old_exc_tuple, lambda shell, *args, **kwargs: old_hook(*args, **kwargs)
-        )
+        ipython_shell.CustomTB = old_hook
```
croyzor, line 64: should this not come before the `yield`?

maximilianruesch: Short answer: No

Long answer: It is consistent with how non-jupyter tracebacks are handled, and makes does make sense because the `yield` actually invokes the thing you are context managing (and we want to invoke it with the changed hook, not the reset one). When an exception is raised, we want the changed hook to get invoked (not the reset one). For different behaviour on raising exceptions, `saved_exception_hook` was created in #1527.

PR #1547, opened 2026-03-03
docs: minor improvements to QAOA notebook

CalMacCQ:
1. Improvements to the networkx display
2. renamed some variables for clarity
3. Used `pi` instead of `angle`. This is more consistent with the matrix defintions of `rx` and `zz_phase` that I give
4. minor code simplifications with comptime.
5. Fixed some minor issues with equations
6. added a comment about the slow notebook https://github.com/Quantinuum/guppylang/pull/1547/changes#diff-174de0355490d4aa5f52b7265fcaf9aa5c921737fb6256662acf5c7635dbe918R10

### DIFF
--- a/examples/qaoa_maxcut_example.ipynb
+++ b/examples/qaoa_maxcut_example.ipynb
@@ -10,7 +10,7 @@
     "**Download this notebook - {nb-download}`qaoa_maxcut_example.ipynb`**\n",
     "\n",
     "\n",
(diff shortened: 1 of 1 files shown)


maximilianruesch:
> I think it probably makes sense to not put it on the docs site until we have resolved the performance issues and can test the example in C.I. as we test the others.

E.g. via an exclusion in `guppy-docs` when upgrading when this is released? Do we have the means to do that?

maximilianruesch (approved):
Love it

PR #1551, opened 2026-03-05
fix!: Nested modifiers completely override each other

### LINKED ISSUE #1324: Nested modifier bug
See the fixme comment in #1292

https://github.com/CQCL/guppylang/blob/26e00e2a33dabe70aacc6348111b4b18f5de3fe2/guppylang-internals/src/guppylang_internals/cfg/builder.py#L311-L316

https://github.com/CQCL/guppylang/blob/26e00e2a33dabe70aacc6348111b4b18f5de3fe2/tests/error/modifier_errors/flags_nested.py

rmfisher26:
## Motivation
Closes #1324 

`ModifiedBlock`

- `__init__ `now requires a `modifiers: Modifiers` argument — existing call sites passing only `cfg ` will fail
- `push_modifier() `has been removed — callers that built up modifier lists incrementally via this method must migrate to constructing a `Modifiers `object and passing it at construction time
- The `dagger`, `control`, and `power `attributes are now set from the `Modifiers` instance rather than initialized as empty lists, so any code that relied on constructing a `ModifiedBlock `and then mutating these lists directly will break

`CheckedModifiedBlock`

- `__init__` now requires a `modifiers: Modifiers` argument in place of the three separate `dagger`, `control`, `power `list parameters — existing call sites must be updated to pass a `Modifiers `instance instead

**Migration**: construct a `Modifiers()` instance, call `push()` for each modifier, and pass it to both constructors. The `dagger`, `control`, and `power `attributes remain available on both classes with the same types — only construction changes.

## Problem
When `with` modifier blocks are nested, the unitary flag checker failed to enforce constraints from outer blocks on calls made in inner blocks.

For example:

```
@guppy(power=True)   # supports power, but NOT dagger
def foo(q: qubit) -> None: ...

@guppy
def test() -> None:
    q = qubit()
    with dagger:
        with power(2):
            foo(q)   # should be a compile error — foo doesn't support dagger
```
This compiled silently instead of raising a `Unitary constraint violation error`. The same issue affected any combination of nested `with dagger, with control(...)`, and `with power(...)` blocks.

## Root cause
`CFGBuilder.visit_With` builds the body of each `with` block as its own inner CFG by calling `CFGBuilder().build(node.body, ...)`. The `build()` method initializes the new builder with the given `unitary_flags`, which are later used by `check_cfg_unitary` to validate all calls inside that CFG.

The bug was that `visit_With` passed no flags to `build()`, so the inner CFG was built with `unitary_flags = NoFlags` regardless of what outer `with` blocks were active. A FIXME comment attempted a partial patch by writing `cfg.unitary_flags = new_node.flags()` after the fact, but this only set the flags for the immediately enclosing block — it did not propagate outer context inward, and it ran after the CFG was already constructed so any recursive inner `visit_With` calls had already used the wrong flags.

## Fix
Two changes to `cfg/builder.py` and a new Modifiers class in `nodes.py`:
1. `Modifiers` class (nodes.py) — collects the Dagger, Control, and Power modifiers from a with block's items into typed lists, and exposes a flags() method that derives the corresponding UnitaryFlags (including dagger cancellation via % 2). Both ModifiedBlock and CheckedModifiedBlock now take a Modifiers reference instead of separate lists.

2. `visit_With `— the loop that builds context expressions and extracts modifiers now runs before `CFGBuilder().build()`. This means `Modifier` objects are available before the inner CFG is constructed, so `modifiers.flags()` can be called to derive the block's flags without re-parsing the AST. The accumulated flags — `self.cfg.unitary_flags | modifiers.flags()` — are passed into `build()`, which sets `self.cfg.unitary_flags` at the start of every new `CFGBuilder`. Any recursive `visit_With` call inside the body therefore sees the full outer context through `self.cfg.unitary_flags`, making propagation correct to arbitrary nesting depth.

3. The FIXME line `cfg.unitary_flags = new_node.flags()` is removed — it was both too late (after recursive builds had already used wrong flags) and incomplete (it only set the immediate block's own flags, not the accumulated outer context).

## Tests
**Error tests** —  test cases that assert the correct `Unitary constraint violation` error is raised for nested combinations that were previously missed:

File | Scenario | Reported missing flag
-- | -- | --
`flags_nested_control_dagger.py` | `with control: with dagger:`, `foo` only supports `control` | `dagger`
`flags_nested_power_control.py` | `with power: with control:`, `foo` only supports `power` | `control`
`flags_nested_triple.py` | `with dagger: with control: with power:`, `foo` has no flags | `unitary` (all three)
`flags_nested_missing_power.py` | `foo(dagger=True)` under `with dagger: with power(2)`: |  inverse of flags_nested.py; confirms power flag is propagated from inner `with`, not just outer
`flags_nested_combined_outer.py` | `foo(power=True)` under `with control(ctrl), dagger: with power(2)`: | Multiple items in a single outer `with` are all accumulated via `Modifiers` and propagated inward
`flags_triple_dagger.py` | `with dagger, dagger, dagger: foo(q)`, `foo` has no flags | `Modifiers.flags() % 2` logic with odd counts greater than 1 — three daggers leave the dagger context active

**Integration tests** — new tests in `test_modifier.py` that assert valid nested-modifier code compiles successfully end-to-end:

Test | Scenario
-- | -- 
`test_nested_dagger_power` | `foo(dagger=True, power=True)` called under `with dagger: with power(2)`
`test_nested_control_dagger` | `foo(control=True, dagger=True)` called under `with control: with dagger`
`test_nested_power_control `| `foo(power=True, control=True)` called under `with power: with control`
`test_nested_triple_all_flags `| `foo(dagger=True, control=True, power=True)` under all three nested levels
`test_nested_same_modifier `| `foo(dagger=True)` under `with dagger`: `with dagger:` (repeated modifier)
`test_double_dagger_cancellation` | `with dagger, dagger: foo(q)` where `foo` has no dagger flag 
`test_combined_with_items_nested` | `foo(all flags)` under `with control(ctrl), dagger: with power(2)`
`test_triple_dagger` | `foo(dagger=True)` called under `with dagger, dagger, dagger`: | 
`test_double_dagger_cancel_nested_power` | `foo(power=True)` called under `with dagger, dagger: with power(2)`

BREAKING CHANGE: In `guppylang-internals`, the refactor of `ModifiedBlock `and `CheckedModifiedBlock` to accept a `Modifiers `object constitutes a breaking API change for any code that constructs or subclasses these nodes directly.

### DIFF
--- a/tests/integration/test_modifier.py
+++ b/tests/integration/test_modifier.py
@@ -1,10 +1,8 @@
-import ast
 from guppylang.decorator import guppy
 from guppylang.std.quantum import qubit
 from guppylang.std.num import nat
 from guppylang.std.builtins import owned
 from guppylang.std.array import array
-from guppylang_internals.nodes import Control, Modifiers
 
 # Dummy variables to suppress Undefined name
 # TODO: `ruff` fails when without these, which need to be fixed
@@ -274,28 +272,3 @@ def test_double_dagger_cancel_nested_power(validate):
                 foo(q)
 
     validate(bar.compile_function())
-
-
-def _make_control(ctrl_args: list[ast.expr]) -> Control:
-    call = ast.Call(func=ast.Name(id="control", ctx=ast.Load()), args=[], keywords=[])
-    ast.fix_missing_locations(call)
-    return Control(call, ctrl_args)
-
-
-def test_has_control_empty_ctrl():
-    """A Control node with no ctrl args should not count as having control."""
-    modifiers = Modifiers()
-    modifiers.push(_make_control([]))
-
-    assert not modifiers.has_control()
-
-
-def test_has_control_with_ctrl():
-    """A Control node with at least one ctrl arg counts as having control."""
-    q = ast.Name(id="q", ctx=ast.Load())
-    ast.fix_missing_locations(q)
-
-    modifiers = Modifiers()
-    modifiers.push(_make_control([q]))
-
-    assert modifiers.has_control()


guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -327,7 +327,13 @@ def visit_With(self, node: ast.With, bb: BB, jumps: Jumps) -> BB | None:
         check_modifiers_enabled(node)
         self._validate_modified_block(node)
 
-        cfg = CFGBuilder().build(node.body, True, self.globals)
+        # Pre-compute this block's modifier flags and combine them with any flags
+        # already accumulated from enclosing `with` blocks (stored in
+        # `self.cfg.unitary_flags`).  Passing the combined value into `build()` lets
+        # every inner CFG see the full modifier context it is compiled under, which is
+        # necessary for correct unitary checking of deeply nested `with` blocks.
+        accumulated_flags = self.cfg.unitary_flags | self._peek_flags(node.items)
```
maximilianruesch: It seems odd that you need the `_peek_flags` functionality. The `ExprBuilder` builds those context items further down, so at the very least you would also need to build those expressions when peaking.

A possible proposal: Extract the modifier lists from `ModifiedBlock` into their own class (e.g. `Modifiers`), and make the block (and its checked variant) take a reference to that class instead. They may define forwarding functions if need be (e.g. to preserve current API, but remember that in internals we can break at any time if we want to) at your discretion. Then, modifiers are built and pushed onto this list *before* building the CFG (they do not reference that CFG anyway, so might as well), and flags are extracted to be passed into the builder together with the current flags.

I took the liberty to sketch this out. It is now clear enough that the comment is not required anymore, and `_peek_flags` is not required:
```python
modifiers = Modifiers()
for item in node.items:
    item.context_expr, bb = ExprBuilder.build(item.context_expr, self.cfg, bb)
    modifiers.push(self._handle_withitem(item))

accumulated_flags = self.cfg.unitary_flags | modifiers.flags()
cfg = CFGBuilder().build(node.body, True, self.globals, accumulated_flags)
new_node = ModifiedBlock(
    cfg=cfg,
    modifiers=modifiers,
    **dict(ast.iter_fields(node)),
)

set_location_from(new_node, node)
bb.statements.append(new_node)
...
```
This will have some rippling effects across internals depending on how much you commit to changing the API, but this is a chance to consistently define modifiers (e.g. I noticed that the definition of `is_power` is different in `ModifiedBlock` and `CheckedModifiedBlock`, where we probably want the latter).

tests/integration/test_modifier.py
```diff
@@ -130,3 +130,84 @@ def bar(q: array[qubit, 3]) -> None:
             use(a)
 
     validate(bar.compile_function())
+
+
+def test_nested_dagger_power_valid(validate):
```
maximilianruesch: You can omit `_valid` from all of these, since these are not error test cases anyway (otherwise they would be in `error/modifier_errors`.

rmfisher26: Thanks for catching this. I have removed the `_valid` part of these tests.

guppylang-internals/src/guppylang_internals/cfg/builder.py
rmfisher26: Thank you for the example! I have create a `Modifiers `class and updated that flow accordingly. I have also refactored the `ModifiedBlock` and `CheckedModifiedBlock` to take a `Modifiers` reference. Please let me know if that is what you had envisioned or anything I may have missed or can clean up.

rmfisher26: I see that is also failed the check for breaking changes as you may have anticipated. 

For example:
```
Installed 2 packages in 4ms
semver-checks diagnostic:

guppylang-internals/src/guppylang_internals/nodes.py:793: ModifiedBlock.dagger:
Attribute value was changed:
  Old: []
  New: modifiers.dagger


...
```

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -740,18 +740,59 @@ def __init__(self, node: ast.expr, iter: ast.expr) -> None:
...
+    ) -> None:
         super().__init__(*args, **kwargs)
         self.cfg = cfg
-        self.dagger = []
-        self.control = []
-        self.power = []
+        self.modifiers = modifiers
+        self.dagger = modifiers.dagger
+        self.control = modifiers.control
+        self.power = modifiers.power
```
maximilianruesch: Instead of copying the reference, you can make this more explicit by adding a read-only getter via a `@property` method of the same name, that returns `modifiers.dagger`. Most usages of this should be read-only anyway.

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -740,18 +740,59 @@ def __init__(self, node: ast.expr, iter: ast.expr) -> None:
...
+    _all: list[Modifier]
+
+    def __init__(self) -> None:
+        self.dagger = []
+        self.control = []
+        self.power = []
+        self._all = []
+
+    def push(self, modifier: Modifier) -> None:
+        self._all.append(modifier)
```
maximilianruesch: I do not understand why we need `_all`. Where is iteration over all modifiers really required? If it is, please provide a link and, if those locations do not require iteration over the modifiers in insertion order, you can make the iterator a concatenation of the three lists using `itertools.chain`. If it is not, we can omit the `__iter__` and the `_all` alltogether.

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -823,9 +851,10 @@ def __init__(
         self.cfg = cfg
         self.ty = ty
         self.captured = captured
-        self.dagger = dagger
-        self.control = control
-        self.power = power
+        self.modifiers = modifiers
+        self.dagger = modifiers.dagger
+        self.control = modifiers.control
+        self.power = modifiers.power
```
maximilianruesch: Same here for read-only getters, as with the non-checked modified block.

guppylang-internals/src/guppylang_internals/nodes.py
rmfisher26: Thanks for the feedback, and you're right this really is not necessary so I have removed that.

guppylang-internals/src/guppylang_internals/nodes.py
rmfisher26: Thanks for the review and I have applied that update to use the `@property` that way.

guppylang-internals/src/guppylang_internals/nodes.py
rmfisher26: Similar to above.  I have applied that update to use the `@property` that way.

guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -836,10 +880,10 @@ def __str__(self) -> str:
         return f"__WithBlock__({self.def_id})"
 
     def has_dagger(self) -> bool:
-        return len(self.dagger) % 2 == 1
+        return self.modifiers.has_dagger()
 
     def has_control(self) -> bool:
-        return any(len(c.ctrl) > 0 for c in self.control)
+        return self.modifiers.has_control()
```
maximilianruesch, line 886: I think this implementation (the deleted line with the `any`) should likely be the implementation used in `Modifier`, as we only want to count control blocks that actually control on *something*.

Please use that implementation, and add a negative test case for it (asserting that it does not trigger when invoked without controls, e.g. by including code that would produce an error if the control would count). If you write the test first, you can also use this to verify that your change is detected by the test (following test driven development).

rmfisher26: Thank you for clarifying this. I was questioning if I made the incorrect choice here when I was trying to refactor this. Also, i have added 2 tests  `test_has_control_empty_ctrl` (the negative test) and `test_has_control_with_ctrl` (positive test).  I tried testing the my first version of `has_control  `which failed as you expected, then changed it back to using the `any` line which passed. 

I did make a helper function `_make_control` which is used by both tests. I wasn't sure if I should keep it that way, or even relocate these new tests outside of  `test_modifier.py` . I'd be happy to apply any suggestion you may have.

tests/integration/test_modifier.py
```diff
@@ -130,3 +132,170 @@ def bar(q: array[qubit, 3]) -> None:
...
+    ast.fix_missing_locations(call)
+    return Control(call, ctrl_args)
+
+
+def test_has_control_empty_ctrl():
+    """A Control node with no ctrl args should not count as having control."""
+    modifiers = Modifiers()
+    modifiers.push(_make_control([]))
+
+    assert not modifiers.has_control()
```
maximilianruesch: This is not at all what I meant to include a negative test. I meant to create something like
```
def test_control_no_expr(validate):
    """Cancelled daggers in outer block don't impose dagger constraint on nested."""

    @guppy
    def foo(q: qubit) -> None:
        pass

    @guppy
    def bar(q: qubit) -> None:
        with control():
            foo(q)

    validate(bar.compile_function())
```
which should not fail even though `foo` does not allow control modifiers.

What you created is a unit test for the `Modifier` class, and we want integration tests (as that is what is more relevant to the language user).

guppylang-internals/src/guppylang_internals/nodes.py
maximilianruesch: See the next comment, I think you misunderstood what I meant by creating a test.

tests/integration/test_modifier.py
rmfisher26: Thanks for your patience, @maximilianruesch. To understand the requirement better, since your example of `test_control_no_expr`  should not fail, when I test this it will generate a `WrongNumberOfArgsError`. 

 Since  `tests/error/modifier_errors/ctrl_no_arg.err` already exists, I'm assuming that is not your intent to repeat this scenario.

I was looking further and found at least 2 places that check for the argument length to be > 0, so are you expecting impacts to these checks?

1. **cfg/builder.py**  `_handle_withitem` :  raises `WrongNumberOfArgsError `if `len(e.args) == 0`

2. **checker/stmt_checker.py** `visit_ModifiedBlock` has `assert len(ctrl) > 0`

Perhaps I'm just going down the wrong path but I wanted to check for committing that new test.

maximilianruesch: Hmm, I did not realize that there were existing tests, because I assumed there were none if the implementations for `ModifiedBlock` and `CheckedModifiedBlock` were different without failing. In this instance, if you just remove the unit tests (so `_make_control` and further down in the file; they are testing implementation details that are irrelevant to the user) then this will be fine.

maximilianruesch: Also, thank you for your patience @rmfisher26, I know this has been dragging on for a while.

PR #1552, opened 2026-03-06
docs: Document Guppy libraries

maximilianruesch:
Checkout, build with `just build-docs` and then open `docs/build/library-docs/index.html` in your favourite HTML renderer.

Closes Quantinuum/guppy-docs#103


docs/library-docs/overview.md
```diff
@@ -0,0 +1,160 @@
+# Overview
+
+## Defining a library
+
+A Guppy library is, at its simplest, a collection of functions that are compiled together into a Hugr package.
+In contrast to the usual executable Hugr packages (like those produced from compiling single entrypoint functions), such
+library Hugr packages feature (usually multiple) *public* functions, possibly with arguments and non-``None`` return
+types.
+
+A developer may define a bunch of functions beyond this collection, for example for internally sharing functionality
```
ss2165:
```suggestion
A developer may define several functions beyond this collection, for example for internally sharing functionality
```

docs/library-docs/overview.md
```diff
@@ -0,0 +1,160 @@
...
+@guppy
+def main() -> None:
+    consumer_func()
+```
+
+In this case, the consumer aims to create an executable Hugr package (e.g. by calling `main.compile()` and creating a
+package with a single, argument-less entrypoint). However, the created Hugr package is incomplete: It lacks the function
+bodies of the library functions, and thus cannot be executed.
+
+Thus, `hugr-py` MUST provide means to link the library Hugr package into the consumer Hugr package. For convenience, the
```
ss2165: given this is a minimal requirement, please include the proposed hugr-py interface for linking packages together

docs/library-docs/overview.md
```diff
@@ -0,0 +1,160 @@
...
+library Hugr package may be provided to the consumer program executor, so that it can be automatically linked in before
+lowering to QIR. For example, using selene, this may look like:
```
ss2165:
```suggestion
compiling to binary. For example, using selene, this may look like:
```

we do not go via QIR

docs/library-docs/proposals/collection.md
```diff
@@ -0,0 +1,82 @@
+# (Proposal) More ways to define a library / auto-collection
+
+Currently, functions must be manually fed into `guppy.library(...)` to be included in the library. This proposal
+concerns additional ways to define (members of) a library, and automatically collect functions to be included in it.
+
+A potential addition would be a `@guppy(export=...)` argument, that registers the function to be included in *the*
+library in some global Guppy store (like the `DEF_STORE`). A function could then be included simply by adding that
```
ss2165, line 7: a "global library" is a bit confusing - this still would require some notion of a specific library the function is exported in. The string method below addresses this, but passing in library objects rather than strings seems preferable

docs/library-docs/proposals/discovery.md
```diff
@@ -0,0 +1,79 @@
...
+
+```python
+from guppylang.library import discover_hugr_packages
+
+main = ...  # Define some consuming entry point
+
+main.emulator(n_qubits=..., libraries=discover_hugr_packages()).with_shots(100).run()
+```
+
+Optionally, omission of `libraries` may lead to internal calls to automatic discovery, and other means of configuring
```
ss2165, line 73: what about instead just refer to library by name, which is then looked up in the discovered set, seems more familiar to the search path/`-L mylib` compiler CLI pattern

docs/library-docs/proposals/generic.md
```diff
@@ -0,0 +1,27 @@
...
+T = guppy.type_var("T")
+
+
+# Choice 1, includes at least the monomorphisations for int and float, and any privately required ones
+@guppy.instances([{T: int}, {T: float}, ...])  # may be unpacked, as QOL for manual specification
+def my_generic_func(x: T) -> T: ...
+
+
+# Choice 2, different syntax, same effect
+@guppy(instances=[{T: int}, {T: float}, ...])  # may be unpacked, as QOL for manual specification
```
ss2165, line 20: alternative, specify at point of library specification, and use existing type parameter syntax:

```python
library = guppy.library(my_generic_func[int], my_generic_func[float])
```

docs/library-docs/proposals/stubs.md
```diff
@@ -0,0 +1,50 @@
...
+`pyright`](https://github.com/microsoft/pyright/blob/main/docs/type-stubs.md#generating-type-stubs-from-command-line),
+and more), but these usually do not fit the needs of Guppy libraries, since we not only need to transform signatures,
+but also make sure that we use `@guppy.declare` instead of `@guppy`. Link names should be explicitly specified in the
+stubs, to make them independent of changes in the default link name inference mechanisms.
+
+These stub generators are often meant to provide a start for writing stubs for the library, but often require manual
+labor to clean, fix, and clarify generated code.
+
+```{note}
+Nonetheless, it might be worthwhile testing existing stub generators on real-world examples for Guppy libraries. Perhaps
```
ss2165: 👍

docs/library-docs/proposals/collection.md
maximilianruesch: This seems very similar to how GuppyModules were handled.

docs/library-docs/proposals/generic.md
maximilianruesch: Potentially a lot more verbose, but would allow for different instances in different libraries, which is a win. Perhaps we can do / allow both?

docs/library-docs/proposals/discovery.md
maximilianruesch: I think all three should be allowed (pass hugr, pass name (which is looked up), and passing nothing which means all known hugrs get thrown in).

docs/library-docs/overview.md
```diff
@@ -0,0 +1,199 @@
...
+
+package_a: Package = ...
+package_b: Package = ...
+package_c: Package = ...
+
+# (a) add modules and extensions from other packages
+package_a.add_from(package_b).add_from(package_c)
+
+# (b) combine multiple modules into a single module inside the package
+package_a.link_modules()
```
acl-cqc: This seems to me a lot like `link` with no other packages

docs/library-docs/overview.md
```diff
@@ -0,0 +1,199 @@
...
+# Now package_a contains a single module with all the functions
+# from package_a, package_b, and package_c
+```
+
+There are no guarantees about the order of modules being linked together, or whether that is pairwise or all at once.
```
acl-cqc, line 166: If we reject duplicate definitions then the order generally only affects *what* error is reported, rather than whether there is an error or what the resulting Hugr looks like in case of success, so fine :)

docs/library-docs/overview.md
```diff
@@ -0,0 +1,199 @@
...
+entire packages, steering the reduction ordering:
+
+```python
+# Extensions and modules are added from package_b,
+# and its module is linked into the module of package_a
+package_a.link(package_b)
+
+# In the future, keywords may be added to allow e.g. privatization of all
+# functions from package_b (i.e. swallowing), so for example:
+package_a.link(package_b, privatize=True)
```
acl-cqc: Or possibly something funky like `package_a.link(Privatize(package_b), ReplaceFuncsWithExisting(package_c))` where Privatize/... have a type of `LinkingDirective` (i.e. are not funcs producing a Hugr/Package)

docs/library-docs/overview.md
```diff
@@ -0,0 +1,199 @@
...
+package_b: Package = ...
+package_c: Package = ...
+
+# (a) add modules and extensions from other packages
+package_a.add_from(package_b).add_from(package_c)
+
+# (b) combine multiple modules into a single module inside the package
+package_a.link_modules()
+# Now package_a contains a single module with all the functions
+# from package_a, package_b, and package_c
```
acl-cqc, line 159: Let's also mention the entrypoint (error if more than one Hugr has a non-module-entrypoint; keep that entrypoint if there is exactly one such)

docs/library-docs/overview.md
maximilianruesch: I will leave that to be figured out by future implementors

maximilianruesch:
Moved to and superseded by https://github.com/Quantinuum/guppy-docs/pull/104.

Issue #1493, opened 2026-02-12
[Feature]: Annotate HUGR with debug info metadata during compilation


tatiana-s:
### Problem Statement

In order to provide debug support, we need to attach information to the HUGR generated by Guppy.

### Proposed Solution

Metadata should include source code location, functions, and variables to start with (according to the HUGR debug info specification).

### Component

Guppy Compiler

### Effort Estimate

None

Issue #1393, opened 2025-12-16
Unify decorator keyword arg namespacing and naming


maximilianruesch:
## Prelude

Currently, we have a number of keywords to the guppy decorator, which control things like `UnitaryFlags` (enforced by the compiler) or a hint on the number of qubits needed to run the function with `max_qubits`. It has been voiced that perhaps `max_qubits` is not a good naming for this, as it does not correctly qualify the intent of the argument, which is to only provide a _hint_ that might or might not be true.

Keyword arguments to `@guppy(*, ...)` are currently layed out flat. We plan to introduce a `meta` keyword, which holds a dictionary that gets passed down to the HUGR node corresponding to the function. We additionally thought about introducing a `hint` keyword that holds a separate namespace for hints, but are currently considering to consider all keyword arguments that are not `meta` a "hint" of sorts.

## Proposal

_Metadata_ is a key-value collection that is passed throughout the Guppy compiler for internal use and to pass onto HUGR nodes for lower-level processing. Keys explicitly passed down should be imported from lower-level modules to ensure cross-module consistency.

A _hint_ is a value that the compiler may use for a specified purpose like allowing omission of arguments somewhere else, strictly enforce during type-checking and/or pass onto HUGR using metadata. Which of these are true for a given hint depends on the hint itself, i.e. the semantics of "hints" are hint-specific.

The `@guppy(*, ...)` decorator is able to receive a number of defined and typed keyword arguments, all of which are considered _hints_, except for the `meta=` argument, which is a `dict[str, Any]` that is not type or key checked and gets passed onto HUGR as is.

Finally, the `max_qubits` may receive a rename to improve visibility of its intent. The proposed new name for this is `expected_qubits`.

maximilianruesch:
The `expected_qubits` hint should specify the number of qubits required to run the function safely, i.e.: "if you allocate at least this number of qubits for your emulator, there should be no code path that runs out of qubits".

Thus, the value in the above case should be 2.

maximilianruesch:
We concluded the discussion by aiming to remove all non-essential / non-heavly-function-specific (like modifiers) kwargs from the decorator, and instead have a follow up #1922 to handle all other cases.

PR #1554, opened 2026-03-09
feat!: Support for compilation unit, subprogram and location debug metadata

### LINKED ISSUE #1507: [Feature]: Support for compilation unit and source locations
### Problem Statement

Add annotations on the following HUGR node types: `Module` (`DICompileUnit`), `FuncDefn` (`DICompileUnit`), and `Call`/` ExtensionOp` (`DILocation`). We need to define typed metadata definitions for the debug information described above, as this is the first step of adding debug information we also need to figure out the best way of adding the data during compilation.

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

tatiana-s:
Closes #1507

Main changes with some points for discussion (+ see comments for further issues in specific files)
- ~~Specification for debug records using the Hugr metadata protocol is in `metadata.debug_info.py` for now - *would it make sense to have this in `hugr-py` instead given the key is `core.debug_info` (should at least the key be defined there?)*~~ now moved to a hugr PR: https://github.com/Quantinuum/hugr/pull/2971
- `turn_on_debug_mode()` / `turn_off_debug_mode()` to toggle global flag for adding debug metadata during compilation (off by default)
- Debug records attached to module root, function definitions, calls and most extension ops - you should now use the `add_op` methods defined in the Guppy compiler as opposed to the ones from Hugr directly, as the Guppy ones now include a check for adding debug metadata 
	-  there are however some extension ops used in various custom compilation where metadata is currently missing (still need to figure out which are really necessary and how to fix those), mainly where it is harder to propagate the needed AST node to the location (~~for example, in modifiers, comptime unpacking, bool conversion, option unwrapping~~ see #1493 sub- issues for specific limitations) - *what would currently happen under the current plan of using this data further down if it is missing?*
	-  on the other hand, some ops which aren't extension ops like `MakeTuple` are getting annotations when used in struct constructors for example

BREAKING CHANGE: (guppylang-internals) Refactored metadata-related code structure and renamed `GuppyMetadata` to `FunctionMetadata`
BREAKING CHANGE: (guppylang-internals) `DFContainer` now uses Guppy-specific `DFBuilder` instead of the Hugr `DfBase` directly
BREAKING CHANGE: (guppylang-internals) `ParsedPytketDef` and `CompiledPytketDef` now have an optional `source_span` field

BEGIN_COMMIT_OVERRIDE
feat: Support for compilation unit, subprogram and location debug metadata
END_COMMIT_OVERRIDE


guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -0,0 +1,88 @@
...
+    title: ClassVar[str] = "Metadata key is reserved"
+    message: ClassVar[str] = (
+        "The following metadata keys are reserved by Guppy but also provided in "
+        "additional metadata: `{keys}`"
+    )
+    keys: set[str]
+
+
+@dataclass
+class FunctionMetadata:
```
tatiana-s, line 32: This likely isn't the best way for storing this data, but after trying a couple of solutions that were all not ideal, my opinion is it doesn't make sense to overthink the design too much at the moment because this https://github.com/Quantinuum/guppylang/issues/1393, modifiers, and future debug metadata could all influence the design so for now something that is working is all that's needed

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -330,3 +348,26 @@ def parse_source(source_lines: list[str], line_offset: int) -> tuple[str, ast.AS
...
+# due to using `CompilerContext` (not an issue for `make_location_record`).
+def make_subprogram_record(
+    node: ast.FunctionDef, ctx: CompilerContext, is_decl: bool = False
+) -> DISubprogram:
+    """Create a DISubprogram debug record for `node`, which should be a function
+    definition or declaration."""
+    filename = get_file(node)
+    # If we can't fine a file for a node, we default to 0 which corresponds to the
+    # entrypoint file.
+    file_idx = ctx.metadata_file_table.get_index(filename) if filename else 0
```
tatiana-s: In the case that we are unable to find a file for the node (shouldn't be the case in theory but in practice you could get issues with custom node) I don't think it would be good to error during compilation but rather through whatever consumes the debug data - just setting it to the root file 0 probably isn't the best solution in that case, should it be something like -1 to indicate the issue?

cgh-qtnm: Yup, it's best to have a specific null value so we can tell the user "unknown filename" rather than give them an incorrect file name. -1 is fine, could also just not include the `filename` member in the record if it's not found.

cgh-qtnm: More generally I think this is also the answer to your question about what to do when it's harder to acquire the correct annotation for a node. I would suggest just marking it with some kind of null placeholder, which the backend can report as "debug info missing", and then we can revisit it when the missing information becomes an issue.

tests/metadata/test_debug_info.py
```diff
@@ -0,0 +1,182 @@
...
+
+    hugr = foo.compile().modules[0]
+    meta = hugr.module_root.metadata
+    assert HugrDebugInfo in meta
+    debug_info = DICompileUnit.from_json(meta[HugrDebugInfo.KEY])
+    assert [get_last_uri_part(uri) for uri in debug_info.file_table] == [
+        "test_debug_info.py",
+        "metadata_example.py",
+    ]
+    funcs = hugr.children(hugr.module_root)
```
acl-cqc, line 71: nit: I prefer patterns where you also check that `funcs` has the elements you expect. (E.g. if `funcs` was empty, you'd trivially pass the for loop + `match` test.)

You could assert names of funcs first, but that duplicates the names between the assert and all the individual cases; better still would be to build a dict from `f_name` to `func`, and then pull the functions out one-by-one to make the checks on them. (So, no `match`.) At the end you can assert there are no functions left in the dict.

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -129,6 +131,12 @@ def compile_row(self, expr: ast.expr, dfg: DFContainer) -> list[Wire]:
         """
         return [self.compile(e, dfg) for e in expr_to_row(expr)]
 
+    def add_op(
+        self, op: ops.DataflowOp, /, *args: Wire, ast_node: AstNode | None = None
+    ) -> Node:
+        """Adds an op to the builder, with optional debug info."""
```
ss2165: is the ast_node the debug info? seems more correct to say "with optional AST node related to the op"

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -748,6 +771,23 @@ def visit_Compare(self, node: ast.Compare) -> Wire:
...
+
+
+def add_op(
+    builder: DfBase[P],
+    op: ops.DataflowOp,
+    /,
+    *args: Wire,
+    ast_node: AstNode | None = None,
+) -> Node:
+    """Adds an op to the builder, with optional debug info."""
```
ss2165: same comment as above about "debug info"

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
```diff
@@ -249,6 +284,16 @@ def compile_outer(
                 call_node = outer_func.call(
                     hugr_func, *(input_list + bool_wires + param_wires)
                 )
+                if debug_mode_enabled():
+                    if self.defined_at is not None:
```
ss2165: good to comment the two cases

guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -0,0 +1,88 @@
...
+    _RESERVED_KEYS: ClassVar[set[str]] = {
+        HugrDebugInfo.KEY,
+        MetadataMaxQubits.KEY,
+    }
+
+    def as_dict(self) -> dict[str, JsonType]:
+        return self._node_metadata
+
+    def set_debug_info(self, debug_info: DebugRecord) -> None:
+        self._node_metadata[HugrDebugInfo.KEY] = debug_info.to_json()
```
ss2165: do we need to `to_json()` at this stage? Why not keep structured and call this at the serialization stage?

guppylang-internals/src/guppylang_internals/metadata/debug_info.py
```diff
@@ -0,0 +1,170 @@
...
+        return value.to_json()
+
+    @classmethod
+    def from_json(cls, value: JsonType) -> DebugRecord:
+        return DebugRecord.from_json(value)
+
+
+@dataclass
+class DICompileUnit(DebugRecord):
+    """Debug information for a compilation unit, corresponds to a module node."""
```
ss2165:
```suggestion
    """Debug information for a compilation unit, corresponds to a HUGR module node."""
```

guppylang-internals/src/guppylang_internals/metadata/debug_info.py
```diff
@@ -0,0 +1,170 @@
...
+
+    @classmethod
+    def from_json(cls, value: JsonType) -> "DICompileUnit":
+        if not isinstance(value, dict):
+            msg = f"Expected a dictionary for DICompileUnit, but got {type(value)}"
+            raise TypeError(msg)
+        for key in ("directory", "filename", "file_table"):
+            if key not in value:
+                msg = f"Expected DICompileUnit to have a '{key}' key but got {value}"
+                raise TypeError(msg)
```
ss2165: is this better as a valueerror?

guppylang-internals/src/guppylang_internals/metadata/debug_info.py
```diff
@@ -0,0 +1,170 @@
...
+class DISubprogram(DebugRecord):
+    """Debug information for a subprogram, corresponds to a function definition or
+    declaration node."""
+
+    file: int  # Index into the string table for filenames.
+    line_no: int  # First line of the function definition.
+    scope_line: int | None = None  # First line of the function body.
+
+    def to_json(self) -> dict[str, str]:
+        return (
```
ss2165: bit more readable to create the shared dict once then conditionally add the `scope_line` key

guppylang-internals/src/guppylang_internals/metadata/debug_info.py
```diff
@@ -0,0 +1,170 @@
...
+
+@dataclass
+class DILocation(DebugRecord):
+    """Debug information for a location, corresponds to call or extension operation
+    node."""
+
+    column: int
+    line_no: int
+
+    def to_json(self) -> dict[str, str]:
```
ss2165: should we have json schema defined for these?

guppylang-internals/src/guppylang_internals/metadata/debug_info.py
```diff
@@ -0,0 +1,170 @@
...
+@dataclass
+class StringTable:
+    """Utility class for managing a string table for debug info serialization."""
+
+    table: list[str]
+
+    def get_index(self, s: str) -> int:
+        """Returns the index of `s` in the string table, adding it if it's not already
+        present."""
+        for idx, entry in enumerate(self.table):
```
ss2165: use list.index() and catch the ValueError if missing

guppylang-internals/src/guppylang_internals/std/_internal/compiler/either.py
```diff
@@ -113,7 +113,8 @@ def compile(self, args: list[Wire]) -> list[Wire]:
             with cond.add_case(i) as case:
                 if i == self.tag:
                     out = case.add_op(
-                        ops.Tag(1, ht.Option(*target_tys)), *case.inputs()
+                        ops.Tag(1, ht.Option(*target_tys)),
```
ss2165: is this just formatting?

guppylang-internals/src/guppylang_internals/debug_mode.py
```diff
@@ -0,0 +1,18 @@
+"""Global state for determining whether to attach debug information to Hugr nodes
+during compilation."""
+
+DEBUG_MODE_ENABLED = False
```
ss2165: should this be private? 
```suggestion
_DEBUG_MODE_ENABLED = False
```

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -296,6 +311,16 @@ def compile(self, id: DefId) -> ModulePointer:
             # loosened after https://github.com/quantinuum/hugr/issues/2501 is fixed
             graph.hugr.entrypoint = compiled_def.hugr_node
 
+        # Add debug info about the module to the root node
+        if debug_mode_enabled():
+            module_info = DICompileUnit(
+                directory=Path.cwd().as_uri(),
```
ss2165, line 489: should this be the cwd or the directory where the file is?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -296,6 +311,16 @@ def compile(self, id: DefId) -> ModulePointer:
...
+                # We know this file is always the first entry in the file table.
+                filename=ctx.metadata_file_table.get_index(filename),
```
ss2165, line 491: get appends if not present - do you need to initialise with the file up top?

tests/metadata/test_debug_info.py
```diff
@@ -0,0 +1,182 @@
...
+def test_compile_unit():
+    @guppy
+    def foo() -> None:
+        pass
+
+    hugr = foo.compile().modules[0]
+    meta = hugr.module_root.metadata
+    assert HugrDebugInfo in meta
+    debug_info = DICompileUnit.from_json(meta[HugrDebugInfo.KEY])
+    assert get_last_uri_part(debug_info.directory) == "guppylang"
```
ss2165: this wouldn't be true if the tests were run from somewhere else?

tests/metadata/test_debug_info.py
```diff
@@ -0,0 +1,182 @@
...
+        debug_info = DILocation.from_json(call.metadata[HugrDebugInfo.KEY])
+        if debug_info.line_no == 28:
+            assert debug_info.column == 0
+        else:
+            assert debug_info.column == 8
+        lines.append(debug_info.line_no)
+    assert lines == [113, 114, 28, 116]
+
+
+# TODO: Improve this test.
```
ss2165: expand on what needs to be improved/raise an issue and link it

guppylang-internals/src/guppylang_internals/metadata/debug_info.py
tatiana-s: The other core metadata classes seem to all use `TypeError` in `from_json`

guppylang-internals/src/guppylang_internals/metadata/debug_info.py
tatiana-s: could you elaborate on what you mean by json schema here?

ss2165: e.g. https://github.com/Quantinuum/hugr/tree/main/resources/json-schema

guppylang-internals/src/guppylang_internals/engine.py
tatiana-s: It's meant to be the working directory of the guppy -> hugr compiler

tests/metadata/test_debug_info.py
tatiana-s: I couldn't really think of a better way of doing this that works universally, so this is just for working locally and on CI as the compiler will always be the repository itself in those two cases

ss2165: would it? I often run tests from subdirectories locally

guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -448,7 +453,71 @@ def __contains__(self, place: Place) -> bool:
...
+            and get_file(self.current_ast_node) is not None
+        ):
+            op_node.metadata[HugrDebugInfo] = make_location_record(
+                self.current_ast_node
+            )
+        return op_node
+
+    @contextmanager
+    def set_ast_context(self, ast_node: AstNode) -> Iterator[None]:
+        """Sets the current AST node context for debug information attachment."""
```
ss2165: need more docs to explain the iterator behaviour (temporary override?)

guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -448,7 +453,71 @@ def __contains__(self, place: Place) -> bool:
...
+    *args: Wire,
+    ast_node: AstNode | None,
+) -> Node:
+    """Adds an op to the given builder, attaching debug information to it if the related
+    `ast_node` is not None. To be used in contexts where want to add operations to
+    specific builders as opposed to the default `DFBuilder`, e.g. `ConditionalBuilder`,
+    `CaseBuilder`, etc.
+    """
+    op_node = other_builder.add_op(op, *args)
+    if debug_mode_enabled() and ast_node is not None and get_file(ast_node) is not None:
```
ss2165: this compound condition is used in multiple places, might be worth centralising so it doesn't go out of sync

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -248,22 +249,41 @@ def array_swap(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
...
-            res = add_op(
-                builder, array_unpack(elem_ty, length), array, ast_node=ast_node
-            )
-            return [res[i] for i in range(length)]
-        case _:
-            raise InternalGuppyError("Invalid array type args")
+    pass
+
+
+def unpack_array(
```
ss2165, line 250: docstring welcome

guppylang-internals/src/guppylang_internals/std/_internal/compiler/array.py
```diff
@@ -248,22 +249,41 @@ def array_swap(elem_ty: ht.Type, length: ht.TypeArg) -> ops.ExtOp:
...
+    builder: DFBuilder | DfBase[P], array: Wire, ast_node: AstNode | None = None
+) -> list[Wire]:
+    if isinstance(builder, DFBuilder):
+        array_ty = builder.raw_builder.hugr.port_type(array.out_port())
+        assert isinstance(array_ty, ht.ExtType)
+        match array_ty.args:
```
ss2165: consider commoning up the match logic between the two cases

tests/metadata/test_debug_info.py
ss2165: ping

guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
```diff
@@ -162,53 +171,87 @@ def build_unwrap_either(
...
+        left=False,
+        error_msg=error_msg,
+        error_signal=error_signal,
+        ast_node=ast_node,
     )
 
 
 def build_unwrap(
-    builder: DFBuilder, option: Wire, error_msg: str, error_signal: int = 1
+    builder: DFBuilder | DfBase[P],
```
ss2165: should both of these implement a shared Protocl that saves all of these union types and isinstance checks?

tests/metadata/test_debug_info.py
tatiana-s: I changed it to also do `Path.cwd().as_uri()`?

guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
tatiana-s: I agree all the union types/instance checks don't look to great, but the difficulty is that they don't actually behave the same (being able to call `.add_op` vs having to use `add_op_to`)

guppylang-internals/src/guppylang_internals/compiler/core.py
ss2165: nit at this point but I still think this check can be centralised with the others by doing 

```python
def _debug_conditions_fulfilled(ast_node: AstNode | None) -> bool:
    return debug_mode_enabled() and ast_node is not None and get_file(ast_node) is not None

```

PR #1559, opened 2026-03-11
feat: Improve type printing for error messages from match operators

### LINKED ISSUE #1550: [Feature]: Add a special to string method to types
### Problem Statement

Would be useful for some error rendering to add a way to print the 'kind' of the type instead of the type itself. E.g.
```
Error: Unsupported (at <In[1]>:16:10)
   | 
14 | @guppy
15 | def main(p: Point @owned) -> None:
16 |     match (p,2):
   |           ^^^^^ Pattern matching on (Point, int) is not supported

```
the error would be better if we print:
```
  |           ^^^^^ Pattern matching on Tuple is not supported
```
since here we are discarding all tuples, not just (Point, int)

### Proposed Solution

_No response_

### Component

Guppy Compiler

### Effort Estimate

### Additional Context

_No response_

kkoci:
## Add `kind_str()` to types for cleaner error messages

  Resolves #1550. Pattern match errors on tuple types now show `Tuple` instead of the full type like `(int, int)`.    
  ### Changes                                                                                                             
  - Add `type_kind_str()` to `tys/printing.py` and `kind_str()` method to `TypeBase` (with `OpaqueType` override
  preserving full string)                                                                                                 
  - Use `kind_str()` in `expr_checker.py` for unsupported pattern match subject errors
  - Update snapshot `unsupported_subject_tuple.err` to reflect new message                                              
  ### Windows test fixes                                                                                                  
  - `tests/error/util.py`: replace `re.sub` with `str.replace` for WASM path substitution — backslashes in Windows paths   (e.g. `\U`) are invalid regex escapes                                                                                  
  - `tests/error/comptime_expr_errors/unsupported_in_list.py`: pin numpy dtype to `int64` via `.astype(np.int64)` —
  `np.random.choice` returns `int32` on Windows, causing snapshot mismatch

### DIFF
--- a/tests/test_type_printing.py
+++ b/tests/test_type_printing.py
@@ -36,34 +36,38 @@ def test_comptime_function_type() -> None:
 
 
 def test_kind_str() -> None:
-    # NoneType -> str(self) = "None"
-    assert NoneType().kind_str() == "None"
+    none_ty = NoneType()
+    assert none_ty.kind_str() == "None"
+    assert none_ty.kind_str() == str(none_ty)
 
-    # NumericType -> str(self) = "int" / "float" / "nat"
-    assert NumericType(NumericType.Kind.Int).kind_str() == "int"
-    assert NumericType(NumericType.Kind.Float).kind_str() == "float"
-    assert NumericType(NumericType.Kind.Nat).kind_str() == "nat"
+    int_ty = NumericType(NumericType.Kind.Int)
+    assert int_ty.kind_str() == "int"
+    assert int_ty.kind_str() == str(int_ty)
 
-    # TupleType -> "Tuple"
+    float_ty = NumericType(NumericType.Kind.Float)
+    assert float_ty.kind_str() == "float"
+    assert float_ty.kind_str() == str(float_ty)
+
+    nat_ty = NumericType(NumericType.Kind.Nat)
+    assert nat_ty.kind_str() == "nat"
+    assert nat_ty.kind_str() == str(nat_ty)
+
     assert TupleType([NumericType(NumericType.Kind.Int)]).kind_str() == "Tuple"
 
-    # FunctionType -> "Function"
-    ty = FunctionType(
+    func_ty = FunctionType(
         inputs=[FuncInput(NumericType(NumericType.Kind.Int), InputFlags.NoFlags)],
         output=NumericType(NumericType.Kind.Int),
     )
-    assert ty.kind_str() == "Function"
+    assert func_ty.kind_str() == "Function"
 
-    # OpaqueType -> str(self), e.g. "array[int, 1]"
     ty_param = TypeParam(0, "T", must_be_copyable=False, must_be_droppable=False)
     len_param = ConstParam(1, "n", NumericType(NumericType.Kind.Nat))
     array_ty = OpaqueType([ty_param.to_bound(0), len_param.to_bound(1)], array_type_def)
     assert array_ty.kind_str() == str(array_ty)
 
-    # BoundTypeVar -> display_name
     bound_var = BoundTypeVar("T", 0, copyable=True, droppable=True)
     assert bound_var.kind_str() == "T"
+    assert bound_var.kind_str() == str(bound_var)
 
-    # ExistentialTypeVar -> "?T"
     ex_var = ExistentialTypeVar.fresh("T", copyable=True, droppable=True)
     assert ex_var.kind_str() == str(ex_var)


guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -668,6 +679,14 @@ def hugr_bound(self) -> ht.TypeBound:
     def cast(self) -> "Type":
         """Casts an implementor of `TypeBase` into a `Type`."""
         return self
+    
+    def kind_str(self) -> str:
+        """Returns the full string representation, including type arguments.
```
maximilianruesch: This docstring is not compatible with the docstring of the base function. The base contract cannot say it is returned without type arguments, and here they are included.

guppylang-internals/src/guppylang_internals/tys/printing.py
```diff
@@ -169,3 +169,25 @@ def signature_to_str(name: str, sig: FunctionType, has_var_args: bool = False) -
...
+    """Returns the kind name of a type, without type arguments.
+
+    Unlike ``str(ty)``, which includes full type parameters, this returns only
+    the constructor name (e.g. ``'Tuple'`` for ``(Point, int)``, ``'list'`` for
+    ``list[int]``, ``'Callable'`` for a function type).
+    """
+    match ty:
+        case TupleType():
+            return "Tuple"
+        case OpaqueType() | StructType() | EnumType():
```
maximilianruesch: This function should imo never be called with `OpaqueType`, since that provides an override to the `kind_str` function. It should raise an error here.

guppylang-internals/src/guppylang_internals/tys/printing.py
```diff
@@ -169,3 +169,25 @@ def signature_to_str(name: str, sig: FunctionType, has_var_args: bool = False) -
...
+            return ty.defn.name
+        case FunctionType():
+            return "Callable"
+        case NumericType():
+            return ty.kind.name.lower()
+        case NoneType():
+            return "None"
+        case _:
+            # BoundTypeVar / ExistentialTypeVar — fall back to full str
```
maximilianruesch: This comment is superfluous

tests/error/match_errors/unsupported_subject_tuple.err
```diff
@@ -3,6 +3,6 @@ Error: Unsupported (at $FILE:7:10)
 5 | def main(x: int, y: int) -> None:
 6 |     t = (x, y)
 7 |     match t:
-  |           ^ Pattern matching on (int, int) is not supported
+  |           ^ Pattern matching on Tuple is not supported
```
maximilianruesch, line 6: Are there more tests like these where you can test the new pretty printing capabilities?

nicolaassolini-qntm: At the moment, no

guppylang-internals/src/guppylang_internals/tys/printing.py
```diff
@@ -169,3 +169,25 @@ def signature_to_str(name: str, sig: FunctionType, has_var_args: bool = False) -
...
+            return "Tuple"
+        case OpaqueType() | StructType() | EnumType():
+            return ty.defn.name
+        case FunctionType():
+            return "Callable"
+        case NumericType():
+            return ty.kind.name.lower()
+        case NoneType():
+            return "None"
+        case _:
```
nicolaassolini-qntm: A suggestion, we should match Tuple, Enum, Struct, and Function, then the `case _` can simply return str(ty). For now, I think it's fine.

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -668,6 +679,14 @@ def hugr_bound(self) -> ht.TypeBound:
...
+
+        For ``OpaqueType``, ``__str__`` already gives a concise kind description
+        (e.g. ``'list[int]'``), so we use it directly.
+        """
+        return str(self)
```
nicolaassolini-qntm, line 701: This method is not needed if we change type_kind_str I think

guppylang-internals/src/guppylang_internals/tys/printing.py
nicolaassolini-qntm: Also, for Struct and Enum we should return 'Enum' or 'Struct', to get the name there is the standard printing method

guppylang-internals/src/guppylang_internals/tys/ty.py
kkoci: @maximilianruesch @nicolaassolini-qntm — quick question on OpaqueType.kind_str():

  @nicolaassolini-qntm suggests removing the override if we change type_kind_str, but @maximilianruesch suggests          type_kind_str should raise an error if called with OpaqueType. These two seem to conflict — removing the override       would cause type_kind_str to be called with OpaqueType, which would then raise.                                                                                                                                                                 Should we:
  1. Keep the override and raise in type_kind_str (current approach), or
  2. Remove the override and handle OpaqueType in type_kind_str (e.g. return str(ty))?

nicolaassolini-qntm: Hi @kkoci, 
We had an internal discussion, and we think that we should implement this method in the following way:
1. Add an abstract method `kind_str` to TypeBase
2. Define the method in all Type classes such that:
      - TupleType.kind_str() return "Tuple"
      - StructType() return 'Struct'
      - EnumType return Enum'
      - FunctionType return 'Function'
      - all the others return the default `__str__ `

Also, if you can add the following tests:
```
match None:
match (,):
match [1.2]:
match array[int, 1]
match "aaa"
def fn(): ....
match fn:
T = guppy.type_var("T")
match T
```
So we can test the method with most of the types.

kkoci: Hi @nicolaassolini-qntm, understood, apologies for the delay, will work in this asap.

kkoci: Hi @nicolaassolini-qntm , so just pushed the changes, these make `kind_str()` abstract on `TypeBase` so every concrete type class owns its implementation directly (no more
  delegation through printing.py). 
  Removes the now-dead `type_kind_str()` free function. Adds a `test_kind_str()` unit test  and two integration error-snapshot tests (unsupported_subject_array, unsupported_subject_function) covering `OpaqueType`
   and `FunctionType` (which now returns "Function" instead of "Callable").

Also **mypy** and **ruff** tests pass, have a look whenever possible, Thank you!

tests/test_type_printing.py
```diff
@@ -21,11 +25,45 @@ def test_generic_function_type():
...
+    array_ty = OpaqueType([ty_param.to_bound(0), len_param.to_bound(1)], array_type_def)
+    assert array_ty.kind_str() == str(array_ty)
+
+    # BoundTypeVar -> display_name
+    bound_var = BoundTypeVar("T", 0, copyable=True, droppable=True)
+    assert bound_var.kind_str() == "T"
+
+    # ExistentialTypeVar -> "?T"
+    ex_var = ExistentialTypeVar.fresh("T", copyable=True, droppable=True)
+    assert ex_var.kind_str() == str(ex_var)
```
nicolaassolini-qntm, line 73: cuold you test if type.kind_str() == str(ty) for all the types this is true (e.g. inf, float, ...)

tests/test_type_printing.py
```diff
@@ -21,11 +25,45 @@ def test_generic_function_type():
...
     ty = FunctionType(
         inputs=[FuncInput(NumericType(NumericType.Kind.Nat), InputFlags.Comptime)],
         output=ty_param.to_bound(0).ty,
         params=[ty_param],
     )
     assert str(ty) == "forall T. nat @comptime -> T"
+
+
+def test_kind_str() -> None:
+    # NoneType -> str(self) = "None"
```
nicolaassolini-qntm: Could you remove these comments?

tests/test_type_printing.py
kkoci: Hi @nicolaassolini-qntm just pushed the change, have a look whenver possible, thank you!

nicolaassolini-qntm: > Hi @nicolaassolini-qntm just pushed the change, have a look whenver possible, thank you!

All good, thanks!

Issue #1560, opened 2026-03-12
[Feature]: Add `std.qsystem.lazy_measure`


tatiana-s:
### Problem Statement

As a first step towards #1485

### Effort Estimate

None

PR #1562, opened 2026-03-12
feat: Add `lazy_measure` to `std.qsystem`

tatiana-s:
Closes https://github.com/Quantinuum/guppylang/issues/1560

This required exposing the sum bool type in the guppy type system alongside the opaque bool, technically it is hidden inside the `Measurement` type however I am not convinced we should actually merge this PR into main because of this.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -1352,8 +1352,8 @@ def to_bool(node: ast.expr, node_ty: Type, ctx: Context) -> tuple[ast.expr, Type
         )
     except GuppyError:
         if not node_ty.copyable:
-            # Linear types may implement a `__into_bool__` method that consumes the
-            # value, instead of borrowing it.
+            # Linear types may implement a `__consume_as_bool__` method that consumes
+            # the value, instead of borrowing it.
             exp_sig = FunctionType([FuncInput(node_ty, InputFlags.Owned)], bool_type())
             return synth.synthesize_instance_func(
                 node, [], "__consume_as_bool__", "truthy", exp_sig, True
--- a/guppylang/src/guppylang/std/qsystem/__init__.py
+++ b/guppylang/src/guppylang/std/qsystem/__init__.py
@@ -201,17 +201,19 @@ class MaybeLeaked:
 @no_type_check
 def lazy_measure(q: qubit @ owned) -> "Measurement":
     """Request a destructive lazy measurement of a qubit, returning a `Measurement`
-    value. Call `.read()` on the value to get the result.
+    value. Call `.read()` on the value to block until the result is available.
     """
 
 
 N = guppy.nat_var("N")
 
 
+@guppy
 @no_type_check
 def lazy_measure_array(qubits: array[qubit, N] @ owned) -> array["Measurement", N]:
     """Request a destructive lazy measurement of an array of qubits, returning an array
-    of `Measurement` values. Call `.read()` on each value to get the results.
+    of `Measurement` values. Call `.read()` on each value to block until results are
+    available.
     """
     return array(lazy_measure(q) for q in qubits)
 
@@ -236,6 +238,15 @@ class Measurement:
     @no_type_check
     def __consume_as_bool__(self: "Measurement" @ owned) -> bool:
         return self.read()
+
+
+@guppy
+@no_type_check
+def collect_measurements(measurements: array[Measurement, N] @ owned) -> array[bool, N]:
+    """Block on each measurement until it is available and collect results into an
+    array of bools.
+    """
+    return array(m.read() for m in measurements)
 
 
 # ------------------------------------------------------
--- a/tests/integration/test_qsystem.py
+++ b/tests/integration/test_qsystem.py
@@ -3,7 +3,13 @@ from guppylang.std.angles import angle
 from guppylang.std.builtins import owned, array
 from guppylang.std.qsystem.random import make_discrete_distribution, RNG
 
-from guppylang.std.qsystem import MaybeLeaked, lazy_measure, measure_leaked
+from guppylang.std.qsystem import (
+    MaybeLeaked,
+    collect_measurements,
+    lazy_measure,
+    lazy_measure_array,
+    measure_leaked,
+)
 from guppylang.std.qsystem.utils import get_current_shot
 from guppylang.std.quantum import qubit, measure_array, x
 from guppylang.std.qsystem.functional import (
@@ -16,6 +22,7 @@ from guppylang.std.qsystem.functional import (
     measure,
     qfree,
 )
+from guppylang.std.lang import comptime
 
 
 def test_qsystem(validate):  # type: ignore[no-untyped-def]
@@ -94,8 +101,6 @@ def test_measure_leaked(validate):  # type: ignore[no-untyped-def]
 
 
 def test_lazy_measure(validate):  # type: ignore[no-untyped-def]
-    """Validate the behaviour of a measurement function returning a future."""
-
     @guppy
     def test(q: qubit @ owned) -> bool:
         f = lazy_measure(q)
@@ -105,8 +110,6 @@ def test_lazy_measure(validate):  # type: ignore[no-untyped-def]
 
 
 def test_lazy_measure_conditional(validate, run_int_fn):  # type: ignore[no-untyped-def]
-    """Validate the behaviour of a measurement function returning a future."""
-
     @guppy
     def test() -> int:
         q = qubit()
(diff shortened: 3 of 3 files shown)


tests/integration/test_qsystem.py
```diff
@@ -91,3 +91,29 @@ def test(q: qubit @ owned) -> bool:
...
+    """Validate the behaviour of a measurement function returning a future."""
+
+    @guppy
+    def test() -> int:
+        q = qubit()
+        x(q)
+        if lazy_measure(q):
+            return 1
+        return 0
+
```
nicolaassolini-qntm, line 120: Is the lazy_measure possible with loops? If yes, I would add an example, if yes we need to add the error test

guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -192,6 +191,40 @@ def discard(self: "MaybeLeaked @ owned") -> None:
...
+@guppy
+@no_type_check
+def lazy_measure(q: qubit @ owned) -> "Measurement":
+    """Measure a qubit destructively, returning a Measurement for the result."""
+    return Measurement(_lazy_measure(q))
+
+
+@guppy.struct
+@no_type_check
+class Measurement:
```
nicolaassolini-qntm, line 226: Not sure about the name, maybe 'MeaureResult' would be better?

if we write a program like:
```python
@guppy
def test(q: qubit @ owned) -> Measurement:
    f = lazy_measure(q)
    return f
 ```
the type annotation is a bit strange, isn't it?

**Review by nicolaassolini-qntm — CHANGES_REQUESTED** [review]
Just a couple of comments

File: tests/integration/test_qsystem.py (hunk 1 above)
Comment (tatiana-s) [reply tests/integration/test_qsystem.py:120]: I am not sure what would be so special about using it inside of loops? The conditional test is there to test the syntactic desugaring of automatically forcing a read in positions that can be coerced to bool

File: guppylang/src/guppylang/std/qsystem/__init__.py (hunk 2 above)
Comment (tatiana-s) [reply guppylang/src/guppylang/std/qsystem/__init__.py:226]: There is potentially a better name that would hint towards the future nature of the result but I am not sure there is a significant difference between Measurement and MeasureResult (they seem almost equivalent to me?)

Comment (nicolaassolini-qntm) [reply guppylang/src/guppylang/std/qsystem/__init__.py:226]: I feel that Measurement is more about the operation, while MeasureResult would be more about the result.
I think it depends if we want to give the idea that we are returning the operation or the result. Just a personal thought, bot names are fine.

File: examples/deferred_measurement.ipynb [hunk 3]
```diff
@@ -0,0 +1,258 @@
...
+    "N = guppy.nat_var(\"N\")\n",
+    "\n",
+    "@guppy\n",
+    "@no_type_check\n",
+    "def eager_read(qs: array[qubit, N] @ owned) -> None:\n",
+    "    for q in qs:\n",
+    "        result(\"t\", measure(q)) # forces immediate measurement of q\n",
+    "    # [... more quantum operations  ...]\n",
+    "\n",
+    "eager_read.compile();"
```
ss2165: use .check rather than compile (don't need the semicolon then either)

examples/deferred_measurement.ipynb
```diff
@@ -0,0 +1,258 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "3a33772a",
+   "metadata": {},
+   "source": [
+    "It is important to understand that while `std.quantum.measure` returns a `bool` on the surface, on Quantinuum systems the physical measurement doesn't always happen immediately. Instead it can be deferred until it is actually needed, which improves ion transport on the machine.\n",
```
ss2165:
```suggestion
    "It is important to understand that while `std.quantum.measure` returns a `bool` on the surface, on Quantinuum systems the physical measurement doesn't always happen immediately. Instead it can be deferred until it is actually needed, which improves parallelism on the machine.\n",
```

more accurate

examples/deferred_measurement.ipynb
```diff
@@ -0,0 +1,258 @@
...
+    "\n",
+    "lazy_read.compile();"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "11043eea",
+   "metadata": {},
+   "source": [
+    "The program now type-checks and ends up compiling to the exact same operation order as `eager_read`. We can now see where exactly the `read()` happens instead of it being hidden, so we can try to move it further down the program."
```
ss2165:
```suggestion
    "The program now type-checks and ends up compiling to the exact same operation order as `eager_read`. We can now see where exactly the `read()` happens, rather than do it implicitly, so we can try to move it further down the program."
```

examples/deferred_measurement.ipynb
ss2165, line 1: Not sure if this is a spot change but a key thing to convey is:

The call to `lazy_measure` _requests_ a measurement. A subsequent call to `.read()` on the returned `Measurement` _blocks_ any further execution until the measurement resolves, i.e. we won't send any more gates until we get the measurement result back. By sending more operations between the request and the block we give the runtime more opportunities to parallelise.

guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -192,6 +197,41 @@ def discard(self: "MaybeLeaked @ owned") -> None:
         self._measurement.discard()
 
 
+@hugr_op(quantum_op("LazyMeasure", ext=QSYSTEM_EXTENSION))
+@no_type_check
+def lazy_measure(q: qubit @ owned) -> "Measurement":
+    """Measure a qubit destructively, returning a Measurement for the result."""
```
ss2165: docstring should explain lazy behaviour

guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -192,6 +197,41 @@ def discard(self: "MaybeLeaked @ owned") -> None:
...
+    droppable=False,
+)
+class Measurement:
+    """Represents the result of a lazy measurement which needs to be explicitly read
+    before being used."""
+
+    @custom_function(compiler=ReadFutureBoolCompiler())
+    @no_type_check
+    def read(self: "Measurement" @ owned) -> bool:
+        """Read the result of the measurement, consuming it."""
```
ss2165: explain blocking-if-necessary-to-wait behaviour

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1302,6 +1302,11 @@ def to_bool(node: ast.expr, node_ty: Type, ctx: Context) -> tuple[ast.expr, Type
         return node, node_ty
     synth = ExprSynthesizer(ctx)
     exp_sig = FunctionType([FuncInput(node_ty, InputFlags.Inout)], bool_type())
+    # TODO: This handles the special case of Measurement types and is fine while we
```
ss2165: this seems extremely fragile, either we should avoid the special case or actually special case on the specific type rather than a class of types we may accidentally add to. If the user defines a non-copyable type with a __bool__ method, does something break?

tatiana-s: Nothing would break, it would just expect that the `__bool__` method consumes the linear value, which may be a reasonable assumption to make to be fair - but for now I added an alternative `__consume_as_bool__` that can be used in that case, such that `__bool__` remains borrowing

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1346,7 +1346,19 @@ def to_bool(node: ast.expr, node_ty: Type, ctx: Context) -> tuple[ast.expr, Type
...
     synth = ExprSynthesizer(ctx)
     exp_sig = FunctionType([FuncInput(node_ty, InputFlags.Inout)], bool_type())
-    return synth.synthesize_instance_func(node, [], "__bool__", "truthy", exp_sig, True)
+    try:
+        return synth.synthesize_instance_func(
+            node, [], "__bool__", "truthy", exp_sig, True
+        )
+    except GuppyError:
+        if not node_ty.copyable:
+            # Linear types may implement a `__into_bool__` method that consumes the
```
ss2165:
```suggestion
            # Linear types may implement a `__consume_as_bool__` method that consumes the
```

guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -192,6 +197,47 @@ def discard(self: "MaybeLeaked @ owned") -> None:
...
+    """
+
+
+N = guppy.nat_var("N")
+
+
+@no_type_check
+def lazy_measure_array(qubits: array[qubit, N] @ owned) -> array["Measurement", N]:
+    """Request a destructive lazy measurement of an array of qubits, returning an array
+    of `Measurement` values. Call `.read()` on each value to get the results.
```
ss2165:
```suggestion
    of `Measurement` values. Call `.read()` on each value to block until results are available.
```

guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -192,6 +197,47 @@ def discard(self: "MaybeLeaked @ owned") -> None:
...
+    """Request a destructive lazy measurement of a qubit, returning a `Measurement`
+    value. Call `.read()` on the value to get the result.
+    """
+
+
+N = guppy.nat_var("N")
+
+
+@no_type_check
+def lazy_measure_array(qubits: array[qubit, N] @ owned) -> array["Measurement", N]:
```
ss2165, line 213: I think worth adding a `collect_measurements` utility function that blocks on all of them and returns an array of bools

guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -192,6 +197,47 @@ def discard(self: "MaybeLeaked @ owned") -> None:
         self._measurement.discard()
 
 
+@hugr_op(quantum_op("LazyMeasure", ext=QSYSTEM_EXTENSION))
+@no_type_check
+def lazy_measure(q: qubit @ owned) -> "Measurement":
+    """Request a destructive lazy measurement of a qubit, returning a `Measurement`
+    value. Call `.read()` on the value to get the result.
```
ss2165:
```suggestion
    value. Call `.read()` on the value to block until the result is available.
```

Issue #1569, opened 2026-03-18
[Feature]: Handling special keywords in Guppy


nicolaassolini-qntm:
### Problem Statement

We have a certain number of keywords in Guppy, such as `comptime` or the modifiers, that can be imported to make the lint happy, but since they are handled in a hardcoded way during compilation, they are not defined inside Guppy. 
This causes instruction like `a = dagger` or `b = comptime` to fail with: 
```
a = dagger
    ^^^^^^ `dagger` is not defined
b = comptime
    ^^^^^^^^ `comptime` is not defined
```
while the proper error should be something like:
`Invalid expression: {{keyword}} cannot be used`.

At the same time programs like:
```
comptime = 2
a = comptime(math.pi / 3)```
```
or 
```
@guppy
def control():....

with control(q1)
```
Should fail since we have overwritten the keyword

### Proposed Solution

Add in the `Context`, or in a different place accessible during the checking, a list of reserved keywords.
In this way, during checking, if a name is not in `Context.local` or `Context.global`, we can check if it is a keyword and detect invalid uses of keywords.
Check while we are checking an expression involving a keyword that that keyword has not been overwritten somewhere (i.e. is not a global or local definition).

### Component

Guppy Compiler

### Effort Estimate

None

PR #1571, opened 2026-03-18
feat: Define modifiers as Python functions

### LINKED ISSUE #1290: Modifiers are undefined as python functions
Curretly, modifiers are parsed by pattern matching over the name as in
https://github.com/CQCL/guppylang/blob/fe8501854507c3c43cec2f26bba75198766a4a17/guppylang-internals/src/guppylang_internals/cfg/builder.py#L317-L319
and `dagger`/`control`/`power` are undefined as python program, so `ruff-check` results in an error like:
```
tests/integration/test_modifier.py:17:14: F821 Undefined name `dagger`
   |
15 |     @guppy
16 |     def bar() -> None:
17 |         with dagger:
   |              ^^^^^^ F821
18 |             pass
   |
```

The reason of why such an ad hoc pattern matching is chosen is because we needed to have strong restriction on the way how modifiers are used. They cannot be used anywhere else than `with` notation, and for example, the following should not compile:
```python
if b:
  modifier = dagger()
else:
  modifier = power(5)
with modifier:
  ...
```
However, this is not ideal and should be fixed. These modifiers should be defined as context managers in std library, and if used in any other place, Guppy CFG builder should say something like `Invalid expression: {{modifier}} cannot be used here.`

nicolaassolini-qntm:
close #1290 

Added new dummy context managers—`control`, `dagger`, and `power`—to the Guppy language standard library. These additions allow importing the names when running a guppy code that uses `with control(...)`, `with dagger`, and `with power(...)`, avoiding linter errors.

The new imports have been added to all test files.

New tests about using modifiers with `comptime` are added

Some open issues are left regarding the error generated from an invalid use of modifier keywords. They are collected in #1569

### DIFF
--- a/guppylang/src/guppylang/std/lang.py
+++ b/guppylang/src/guppylang/std/lang.py
@@ -1,7 +1,6 @@
 """Provides Python objects for builtin language keywords."""
 
 from collections.abc import Generator
-from contextlib import contextmanager
 from typing import Any, Protocol, TypeVar
 
 from guppylang_internals.error import GuppyComptimeError


guppylang/src/guppylang/std/lang.py
```diff
@@ -42,3 +44,21 @@ class Copy(Protocol):
 
 class Drop(Protocol):
     """Bound to mark generic type parameters as being implicitly droppable."""
+
+
+@contextmanager
+def control() -> Generator[None]:
+    """Dummy context manager to support `with control(...):` blocks in Guppy code."""
+    yield
```
mark-koch: Can you raise a `GuppyComptimeError` here explaining that modifiers are not supported in comptime functions yet?

Same for the other modifiers. And then add some tests that try to use them inside a `@guppy.comptime` function

PR #1574, opened 2026-03-18
fix: Cleanup all imports of pytket

maximilianruesch:
Cleans up the complicated import structure of the pytket module (since pytket is already a required dependency of `guppylang-internals`) and adds pytket as a dependency to `guppylang` (which is a noop in all environments since it is also a dependency of guppylang-internals, this is done for completeness).

This should make errors that occur during the import of `pytket` or `tket` modules much more visible, since previously some errors were swallowed / shadowed by the import boilerplate.

### DIFF
--- a/guppylang/pyproject.toml
+++ b/guppylang/pyproject.toml
@@ -35,6 +35,7 @@ dependencies = [
     "selene-hugr-qis-compiler~=0.2.9",
     "selene-sim~=0.2.7",
     "tqdm>=4.67.1",
+    "pytket>=1.34",
     "types-tqdm>=4.67.0.20250809",
 ]
 
--- a/guppylang/src/guppylang/decorator.py
+++ b/guppylang/src/guppylang/decorator.py
@@ -57,6 +57,7 @@ from hugr import ops
 from hugr import tys as ht
 from hugr import val as hv
 from hugr.package import ModulePointer
+from pytket.circuit import Circuit as TketCircuit
 from typing_extensions import Unpack, dataclass_transform, deprecated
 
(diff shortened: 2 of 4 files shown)


guppylang/src/guppylang/decorator.py
```diff
@@ -58,6 +57,7 @@
 from hugr import tys as ht
 from hugr import val as hv
 from hugr.package import ModulePointer
+from pytket.circuit import Circuit as TketCircuit
```
tatiana-s, line 60: Any particular reason for renaming it here but not in the `pytket_circuits` file?

maximilianruesch: In the `pytket_circuits` file, the file name provides the hint that this is a tKet thing. In this file, I felt there was not enough surrounding hints (if one looked at names available in the file outside of the `pytket` functions on the `Guppy` class.

Is what I would say if this was not just a coincidence of me inserting this import vs the other one preexisting.

maximilianruesch: I'd say this is not particularly important, and the rename adds a bit of extra context in this file, so its fine. Will merge.

guppylang/pyproject.toml
```diff
@@ -35,6 +35,7 @@ dependencies = [
     "selene-hugr-qis-compiler~=0.2.9",
     "selene-sim~=0.2.7",
     "tqdm>=4.67.1",
+    "pytket>=1.34",
```
cqc-melf, line 38: @maximilianruesch can we shortly talk about this?

PR #1575, opened 2026-03-19
fix: Raise error when using an enum class in places where a variant is expected

### LINKED ISSUE #1572: [Bug]: AssertionError when returning a wrong type with Enum
### Description

When we wrongly return the Enum class, instead of having a guppy error the compilation fail due to an AssertionError

### Steps to Reproduce

```python
@guppy
def main() -> TupleEnumWithFields:
    
    return TupleEnumWithFields
```

### Expected Behaviour

We should have an error like:
```
Error: Type mismatch (at <In[13]>:21:11)
   | 
19 | def main() -> TupleEnumWithFields:
20 | 
21 |     return TupleEnumWithFields.Left
   |            ^^^^^^^^^^^^^^^^^^^^^^^^ Expected return value of type `TupleEnumWithFields`, got
   |                                     `int -> TupleEnumWithFields`

```

### Actual Behaviour

```
File "/Users/nicola.assolini/Documents/GitHub/guppylang/guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py", line 275, in visit_GlobalName
    assert isinstance(defn, CompiledValueDef)
           ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError
```

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
closes #1572 

When an enum class is used as a value (e.g., `x = MyEnum`), the checker now raises an `ExpectedError` (Added a new `EnumHelp` class to provide a help message in error diagnostics)

To spot the error I moved the logic in the function `ExprSynthesizer.visit_Name` in `expr_checker.py` to a helper `ExprSynthesizer._check_name_id`. In this way, when visiting a Name, `_check_name_id` raise an error if it finds an Enum Def (we are in `x = MyEnum` case),  while calling `_check_name_id` from `visit_Attribute` does not cause the error, since `x = MyEnum.Variant` is valid.

Added 3 new tests (`use_enum_class1.py`, `use_enum_class2.py`, `use_enum_class3.py`) to test these cases.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/errors/generic.py
+++ b/guppylang-internals/src/guppylang_internals/checker/errors/generic.py
@@ -47,7 +47,8 @@ class ExpectedError(Error):
     @dataclass(frozen=True)
     class EnumHelp(Help):
         message: ClassVar[str] = "You might use an enum variant here instead"
-          
+
+    @dataclass(frozen=True)
     class NotInstantiable(Note):
         name: str
         message: ClassVar[str] = (
--- a/uv.lock
+++ b/uv.lock
@@ -966,7 +966,7 @@ wheels = [
 
 [[package]]
 name = "guppylang"
-version = "0.21.9"
+version = "0.21.10"
 source = { editable = "guppylang" }
 dependencies = [
     { name = "guppylang-internals" },


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -432,16 +432,41 @@ def _check_generic_param(self, name: str, node: ast.expr) -> tuple[ast.expr, Typ
...
+            match self.ctx.globals[name_id]:
                 case Definition() as defn:
-                    return self._check_global(defn, x, node)
+                    if not allow_enum:
+                        from guppylang_internals.definition.enum import (
+                            CheckedEnumDef,
+                            ParsedEnumDef,
+                        )
+
+                        if isinstance(defn, ParsedEnumDef | CheckedEnumDef):
```
doug-q, line 457: This is pretty fragile, is there any way to write this without isinstance?

nicolaassolini-qntm: We need to check if the defn is an Enum definition, the isinstace on the class is the only way I think.

PR #1577, opened 2026-03-19
refactor: Common up build_unwrap_ for either wires

maximilianruesch:
Addresses a TODO I stumbled upon, since there referenced issue https://github.com/quantinuum/hugr/issues/1596 has a solution (commit https://github.com/Quantinuum/hugr/commit/583d21d371320851f8608daa295ef8b723d31326) that has been part of `hugr-py` since at least `v0.10.0` (and we pull in `v0.15.x`).

Related to #575

### DIFF
--- a/guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
+++ b/guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
@@ -108,8 +108,8 @@ P = TypeVar("P", bound=ops.DfParentOp)
 def build_unwrap_either(
     builder: DfBase[P], either: Wire, left: bool, error_msg: str, error_signal: int = 1
 ) -> Node:
-    """Unwraps the value indicated by `left` from a `hugr.tys.Either` value, panicking
-    with the given message if the result is not in the place indicated by `left`.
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
```diff
@@ -104,47 +102,50 @@ def build_static_error(builder: DfBase[P], signal: int, msg: str) -> Wire:
...
 
 
-def build_unwrap_right(
-    builder: DfBase[P], either: Wire, error_msg: str, error_signal: int = 1
+def build_unwrap_either(
+    builder: DfBase[P], either: Wire, left: bool, error_msg: str, error_signal: int = 1
 ) -> Node:
-    """Unwraps the right value from a `hugr.tys.Either` value, panicking with the given
-    message if the result is left.
+    """Unwraps the value indicated by `left` from a `hugr.tys.Either` value, panicking
```
aborgna-q: This is a bit confusing imho.

Maybe something like "Unwraps the left value or right value from a `...Either` according to the `left` flag." ?

aborgna-q (approved):

PR #1579, opened 2026-03-19
fix: Improve error message for type definitions without a constructor

### LINKED ISSUE #1564: [Bug]: Improve error message when creating a `frozenarray`
### Description

If I try to create an instance of `frozenarray` as I would a normal `array` I get the error 

```
Error: Expected a value (at <In[15]>:5:10)
  | 
3 | @guppy
4 | def main() -> None:
5 |     arr = frozenarray(1, 11, 21)
  |           ^^^^^^^^^^^ Expected a value, got type `frozenarray`
```

I gather `frozenarray` is only intended to be used in a `comptime` or `py` context. I.e. the user is only expected to encounter the `frozenarray` type if they are pass a Python list into a `comptime` expression. If this is the case maybe it could be clearer from the error message that creating an array outside of comptime is not the intended use.

### Steps to Reproduce

```python
from guppylang.std.array import frozenarray

@guppy
def main() -> None:
    arr = frozenarray(1, 11, 21)

main.check()
```

nicolaassolini-qntm:
closes #1564 

Partial close issue #1564 by improving the error message. The problem of initialising a frozen array directly should be addressed in a different PR.

I added a `Note` to the `ExpectedError` class to give the user a better error when a type with no constructor is found.

I only tested the new feature on `frozenarray` since I'm not sure if there are other cases like this one.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/errors/generic.py
+++ b/guppylang-internals/src/guppylang_internals/checker/errors/generic.py
@@ -45,16 +45,12 @@ class ExpectedError(Error):
         return f", got {self.got}" if self.got else ""
 
     @dataclass(frozen=True)
-    class NotInstanziable(Note):
-        description: str
+    class NotInstantiable(Note):
         name: str
         message: ClassVar[str] = (
-            "Cannot construct an instance of `{name}`, as it is missing a `__new__` method."
+            "Cannot construct an instance of `{name}`, as it is missing a `__new__` "
+            "method."
         )
-
-        @property
-        def capitalized_description(self) -> str:
-            return self.description.capitalize()
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


guppylang-internals/src/guppylang_internals/checker/errors/generic.py
```diff
@@ -44,6 +44,18 @@ class ExpectedError(Error):
     def extra(self) -> str:
         return f", got {self.got}" if self.got else ""
 
+    @dataclass(frozen=True)
+    class NotInstanziable(Note):
```
maximilianruesch: Use either `Instantiatable` or even better `Constructable`, grep throughout the file(s) and file names

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
```diff
@@ -44,6 +44,18 @@ class ExpectedError(Error):
...
+        description: str
+        name: str
+        message: ClassVar[str] = (
+            "{capitalized_description} {name} is not an instaziable {description}"
```
maximilianruesch:
```suggestion
            "Cannot construct an instance of `{name}`, as it is missing a `__new__` method."
```
and remove `capitalized_description`.

PR #1585, opened 2026-03-23
feat: Automatically package user-defined extensions used in the guppy program

### LINKED ISSUE #1556: [Feature]: Auto-include user declared extension in compiled packages
### Problem Statement

`CompilationEngine.compile` decides which extensions to bundle in a package via a whitelist
https://github.com/Quantinuum/guppylang/blob/f6543f607a7229a08ba65114ce3f016843711e69/guppylang-internals/src/guppylang_internals/engine.py#L299-L302
, which gets intersected with the extensions actually used in the hugr
https://github.com/Quantinuum/guppylang/blob/f6543f607a7229a08ba65114ce3f016843711e69/guppylang-internals/src/guppylang_internals/engine.py#L338-L343

The user is able to add packages to the whitelist via `CompilationEngine.register_extension`.
This method is undocumented and not referenced from anywhere in the code or docs.

Couldn't we just bundle any non-std extension being used by the Hugr instead?
The user is able to clean up that set afterwards, but leaving extensions missing is a terrible default that causes missing extension errors for most workflows that pass the program to the rust backend.

### Proposed Solution

_No response_

### Component

Guppy Compiler

### Effort Estimate

### Additional Context

_No response_

aborgna-q:
Closes #1556 

Deprecates `CompilationEngine.register_extension`, as it's no longer needed.
(`additional_extensions` are now only used to resolve non-packaged extensions when loading a Hugr. We shouldn't need that anymore)

### DIFF
--- a/guppylang-internals/src/guppylang_internals/engine.py
+++ b/guppylang-internals/src/guppylang_internals/engine.py
@@ -157,6 +157,7 @@ class CompilationEngine:
         Cached at class level.
         """
         if CompilationEngine._base_resolve_registry is None:
+            from guppylang_internals.compiler import hugr_extension
             from guppylang_internals.std._internal.compiler.tket_exts import (
                 TKET_EXTENSIONS,
             )
@@ -165,7 +166,7 @@ class CompilationEngine:
             for ext in [
                 *hugr.std._std_extensions().extensions.values(),
                 *TKET_EXTENSIONS,
-                guppylang_internals.compiler.hugr_extension.EXTENSION,
+                hugr_extension.EXTENSION,
             ]:
                 registry.register_updated(ext)
             CompilationEngine._base_resolve_registry = registry


guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -142,46 +143,29 @@ class CompilationEngine:
...
-                hugr.std.prelude.PRELUDE_EXTENSION,
-                hugr.std.collections.array.EXTENSION,
-                hugr.std.float.FLOAT_OPS_EXTENSION,
-                hugr.std.float.FLOAT_TYPES_EXTENSION,
-                hugr.std.int.INT_OPS_EXTENSION,
-                hugr.std.int.INT_TYPES_EXTENSION,
-                hugr.std.logic.EXTENSION,
+                *hugr.std._std_extensions().extensions.values(),
+                *TKET_EXTENSIONS,
+                guppylang_internals.compiler.hugr_extension.EXTENSION,
```
croyzor: need to `import guppylang_internals.compiler.hugr_extension`?

croyzor (approved):

PR #1594, opened 2026-03-24
refactor: Clarify indexing of RHS elements while checking unpacking assignments

### LINKED ISSUE #780: Suspicious index of rhs element in `_check_unpack_assign`
https://github.com/CQCL/guppylang/blob/634e241511c7b5a90bea279b0c847230b2cf9a98/guppylang/checker/stmt_checker.py#L222

uses `rhs_elts[0]`, when the indices that correspond to the `starred` section are `rhs_elts[le:rs]`. Not sure whether we should pass the whole range or one element; in *many* cases all `rhs_elts` are the same but not if the RHS is a literal tuple.

I played around with many test cases here and couldn't make a program that produced the wrong answer, so I suspect the value here is not used *per se* - perhaps only for debugging information or something like that?

maximilianruesch:
This is indeed not a bug since `starred` is always an `ast.Name` when this code is encountered, and the dynamic dispatch method for `ast.Name` (`_check_variable_assign`) does not use its RHS element. I have clarified this in the code and used the dispatch method directly to make this clearer.

Closes #780

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
@@ -282,7 +282,9 @@ class StmtChecker(AstVisitor[BBStatement]):
             assert isinstance(starred, ast.Name), "Python grammar"
             # We can use any value for `rhs` as it is ignored for variable assignments.
             unpack.pattern.starred = self._check_variable_assign(
-                starred, rhs_elts[0], array_ty
+                starred,
+                rhs_elts[0],  # ignored
+                array_ty,
             )
 
         return with_type(rhs_ty, with_loc(lhs, unpack))


guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -279,7 +279,11 @@ def _check_unpack_assign(
                 unsolved = array_type(ExistentialTypeVar.fresh("T", True, True), 0)
                 raise GuppyError(TypeInferenceError(starred, unsolved))
             array_ty = array_type(starred_ty, len(starred_tys))
-            unpack.pattern.starred = self._check_assign(starred, rhs_elts[0], array_ty)
+            assert isinstance(starred, ast.Name), "Python grammar"
+            # We can use any value for `rhs` as it is ignored for variable assignments.
```
acl-cqc, line 283: Another possibility:
```
unpack.pattern.starred = self._check_variable_assign(
                starred,
                rhs_elts[0],  # ignored
                array_ty,
             )
```
and yet another is to make a `_check_variable_assign_helper` that omits the unused argument altogether, but I don't think there's a sufficiently-easy way to register that with the `@singledispatchmethod` to make that worthwhile.

Or yet another:
```
unpack.pattern.starred = self._check_variable_assign(
                starred,
                None,  # type: ignore # unused
                array_ty,
             )
```
Even doing the last in a call to `self._check_assign` wouldn't be too bad, although would be better with some equivalent of haskell's `undefined: forall a. a` ;-)

Any of these looks fine to me, including yours

PR #1602, opened 2026-03-25
fix!: Better scoping for existentials

croyzor:
## I. Unquantified
This fixes a bug in `unquantified`, which we use to abstract all of the variables in a `FunctionType` before unifying them with the arguments to the function in a call. Previously, it was not abstracting nested variables, so it would do
```python
(S -> T[S]) |-> (?S -> ?T[S])
```
instead of
```python
(S -> T[S]) |-> (?S -> ?T[?S])
```

This is fixed in 4de6964277077771d64aa610fe956ffa03674989

## II. Const types
Fixing the first bug reveals a second: when we unify an `ExistentialConstVar` with a concrete const given as an argument, we don't unify the types!

If we have a program:
```python
def foo(x: T, y: T @comptime) -> T: ...

def main(x: int) -> int:
    foo(x, 42)
``` 
then unifying the signature of const with the args will solve `?T := int`, but leaves the type of `?y` unsolved: `?y := Const(42, ty=?T`).

This didn't matter before because of issue `I`: we would have `?y := Const(42, ty=T)` because `unquantified` didn't abstract the bound variable `T`! 

It took trying to implement protocols to run into cases where these problems showed themselves.

## III. Substitution keys
Our substitutions are maps with `ExistentialVar` keys,  and that `ExistentialConstVar`s contain their types, meaning that after we unify the type of a const, it maps to a different entry in the substitution. 

Thus, I've set `compare=False` in the type field of `Const`. This means that two const variables in a substitution will be identified even when their types don't match. This has been done under the assumption that the times where the types don't match will only be because one side has an existential which is solved in the substitution, which _would_ make it the same. This way we don't have to keep doing this update before looking up keys.

--- 

BREAKING CHANGE: `post_init` method removed from `ConstBase` and its children

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/const.py
+++ b/guppylang-internals/src/guppylang_internals/tys/const.py
@@ -132,7 +132,7 @@ class ExistentialConstVar(ExistentialVar, ConstBase):
     def transform(self, transformer: Transformer, /) -> "Const":
         """Accepts a transformer on this constant."""
         return transformer.transform(self) or ExistentialConstVar(
-            transformer.transform(self.ty) or self.ty, self.display_name, self.id
+            self.ty.transform(transformer) or self.ty, self.display_name, self.id
         )
 
 
--- a/guppylang-internals/src/guppylang_internals/tys/subst.py
+++ b/guppylang-internals/src/guppylang_internals/tys/subst.py
@@ -1,5 +1,6 @@
 import functools
 from collections.abc import Sequence
+from dataclasses import replace
 from typing import Any
 
 from guppylang_internals.error import InternalGuppyError
@@ -43,9 +44,15 @@ class Substituter(Transformer):
 
     @transform.register
     def _transform_ExistentialConstVar(self, c: ExistentialConstVar) -> Const | None:
-        s = self.subst.get(c, None)
-        assert not isinstance(s, TypeBase)
-        return s
+        if s := self.subst.get(c, None):
+            assert not isinstance(s, TypeBase)
+            return s
+        c = replace(c, ty=c.ty.transform(self))
+        if s := self.subst.get(c, None):
+            assert not isinstance(s, TypeBase)
+            return s
+
+        return c
 
 
 class Instantiator(Transformer):


guppylang-internals/src/guppylang_internals/tys/var.py
```diff
@@ -47,3 +49,6 @@ class ExistentialVar(Var, ABC):
 
     # Generator of fresh unique ids
     _fresh_id: ClassVar[Iterator[UniqueId]] = itertools.count()
+
+    @abstractmethod
+    def transform(self, t: Transformer) -> Any: ...
```
mark-koch, line 54: Better to inherit this from `Transformable`

guppylang-internals/src/guppylang_internals/tys/const.py
```diff
@@ -131,7 +131,7 @@ def cast(self) -> "Const":
     def transform(self, transformer: Transformer, /) -> "Const":
         """Accepts a transformer on this constant."""
         return transformer.transform(self) or ExistentialConstVar(
-            transformer.transform(self.ty) or self.ty, self.display_name, self.id
+            self.ty.transform(transformer) or self.ty, self.display_name, self.id
```
mark-koch, line 135: That looks fine 👍

guppylang-internals/src/guppylang_internals/tys/var.py
croyzor: I think it's worse, because this way we can specify `ExistentialConstVar.transform(t): Const`, `ExistentialTypeVar.transform(t): Type`, I couldn't work out how to do that nicely while inheriting from `Transformable[T]`

guppylang-internals/src/guppylang_internals/tys/subst.py
```diff
@@ -43,9 +44,15 @@ def _transform_ExistentialTypeVar(self, ty: ExistentialTypeVar) -> Type | None:
...
-        s = self.subst.get(c, None)
-        assert not isinstance(s, TypeBase)
-        return s
+        if s := self.subst.get(c, None):
+            assert not isinstance(s, TypeBase)
+            return s
+        c = replace(c, ty=c.ty.transform(self))
+        if s := self.subst.get(c, None):
+            assert not isinstance(s, TypeBase)
+            return s
```
mark-koch, line 53: I don't think this branch is reachable: the type of `c` is not taken into account for equality, so if the original `c` is not in the substitution, then the one with substituted type also won't be?

croyzor: yes!

PR #1604, opened 2026-03-26
feat: Add support for enum type in `comptime`

### LINKED ISSUE #1590: [Feature]: Enums are not supported in comptime
### Problem Statement

We need to add support for enums in comptime functions

nicolaassolini-qntm:
close #1590 

* Added the `GuppyEnumObject` class to represent enum values at runtime during tracing
* Updated `unpacking.py::unpack_guppy_object` to recognise and correctly handle `EnumType` and `unpacking.py::guppy_object_from_py` to handle `GuppyEnumObject`
* Added `GuppyFunctionDefinition` in `guppylang/defs` that implements a custom `__getattr__` to allow only valid variant constructors 
* Updated `len` in `tracing/builtins_mock.py` to support `GuppyEnumObject`
* added test for enum use in comptime

Regarding Enum usage, currently, attribute access is not possible at all, regardless of context (comptime or standard guppy), and will become possible with the introduction of pattern matching.

### DIFF
--- a/tests/error/tracing_errors/enum_get_attr.err
+++ b/tests/error/tracing_errors/enum_get_attr.err
@@ -3,4 +3,4 @@ Traceback (most recent call last):
     test.compile()
   File "$FILE", line 13, in test
     e.x
-guppylang_internals.error.GuppyComptimeError:  Expression of enum type `E` has no method `x`, and variant fields are not accessible
+guppylang_internals.error.GuppyComptimeError:  Expression of enum type `E` has no method `x`. It may be a variant field, but variant fields are inaccessible
--- a/tests/error/tracing_errors/enum_set_method.err
+++ b/tests/error/tracing_errors/enum_set_method.err
@@ -3,4 +3,4 @@ Traceback (most recent call last):
     test.compile()
   File "$FILE", line 21, in test
     e.foo = bar
-AttributeError: Expression of enum type do not support attribute assignment
+AttributeError: Enum variants do not support attribute assignment


tests/integration/tracing/test_enum.py
```diff
@@ -0,0 +1,126 @@
...
+
+    run_int_fn(main, 42)
+
+
+def test_generic(validate):
+    T = guppy.type_var("T")
+    S = guppy.type_var("S")
+
+    @guppy.enum
+    class EnumA(Generic[T]):  # pyright: ignore[reportInvalidTypeForm]
```
maximilianruesch: Afaik we do not use `pyright`. If you need to ignore something here, should you instead be using `# type:` ignores?

tests/integration/tracing/test_mocked_builtins.py
```diff
@@ -106,10 +114,14 @@ def test(xs: array[int, 10]) -> int:
         # In particular, `builtins.len` doesn't work for GuppyObjects, but our mocked
         # version does
         s = S(100)
+        e = E.x(1)
```
maximilianruesch: nit: Move below the test from the `GuppyStructObject` for better grouping.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -439,7 +439,8 @@ def _check_name_id(
     ) -> tuple[ast.expr, Type]:
         """Helper method to check a name by its identifier, used for both `ast.Name` and
         `ast.Attribute` nodes. If allow_enum is False, we raise an error if the name
-        corresponds to an enum definition, since enums cannot be used as values."""
+        corresponds to an guppy.enum class, since class name cannot be used as values,
```
```suggestion
        corresponds to an guppy.enum class, since enum classes cannot be used as values,
```

tests/error/tracing_errors/not_value2.py
```diff
@@ -0,0 +1,14 @@
...
+
+@guppy.enum
+class E:
+    A = {}  # noqa: RUF012
+
+
+@guppy.comptime
+def test() -> E:
+
+    return E
```
maximilianruesch, line 11: nit: too much space (`ruff` should have caught this imo)

tests/error/tracing_errors/input_mutate4.py
```diff
@@ -8,7 +8,7 @@ class S:
 
 
 @guppy.comptime
-def foo(s: S @ owned) -> None:
+def foo(s: S @ owned) -> None: # pyright: ignore[reportInvalidTypeForm]
```
maximilianruesch: I am confused as to why these are needed. Please check your local configuration.

tests/error/tracing_errors/enum_set_method.py
```diff
@@ -0,0 +1,19 @@
...
+
+    @guppy
+    def foo(self: "E") -> None:
+        pass
+
+
+@guppy.comptime
+def test() -> None:
+    e = E.A()
+    e.foo = 0
```
maximilianruesch: Try to create a value that theoretically "could be assigned", i.e. a Guppy callable with the right signature. May not be super easy.

tests/error/tracing_errors/enum_get_attr.err
```diff
@@ -0,0 +1,6 @@
+Traceback (most recent call last):
+  File "$FILE", line 16, in <module>
+    test.compile()
+  File "$FILE", line 13, in test
+    e.x
+guppylang_internals.error.GuppyComptimeError: Expression of type `E` has no attribute `x`
```
maximilianruesch: It is unclear to the user why this is the case, e.g. that this should only work in a Guppy context. Can we do something about that?

guppylang-internals/src/guppylang_internals/tracing/unpacking.py
```diff
@@ -60,6 +61,8 @@ def unpack_guppy_object(
                 for field, wire in zip(ty.fields, unpack.outputs(), strict=True)
             ]
             return GuppyStructObject(ty, field_values, frozen)
+        case EnumType() as ty:
+            return GuppyEnumObject(ty, obj._use_wire(None))
```
maximilianruesch, line 65: Can we not emit `UnpackTuple()` to support variant fields?

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -519,6 +562,32 @@ def __call__(self, *args: Any) -> Any:
         err = f"{defn.description.capitalize()} `{defn.name}` is not callable"
         raise GuppyComptimeError(err)
 
+    @hide_trace
+    def __getattr__(self, name: str) -> Any:
+        # Handle attribute access when defining  an enum variant constructors, like
```
maximilianruesch:
```suggestion
        # Handle attribute access when calling an enum variant constructor, like
```

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -519,6 +562,32 @@ def __call__(self, *args: Any) -> Any:
...
+                impl_def = DEF_STORE.raw_defs[DEF_STORE.impls[defn.id][name]]
+                return TracingDefMixin(impl_def)
+            raise AttributeError(
+                f"{defn.description.capitalize()} `{defn.name}` "
+                f"has no attribute `{name}`"
+            )
+        raise AttributeError(
+            f"{self.wrapped.description.capitalize()} `{self.wrapped.name}`"
+            f" has no attribute `{name}`"
+        )
```
maximilianruesch: Should we call `super.__getattr__` instead?

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -519,6 +562,32 @@ def __call__(self, *args: Any) -> Any:
         err = f"{defn.description.capitalize()} `{defn.name}` is not callable"
         raise GuppyComptimeError(err)
 
+    @hide_trace
+    def __getattr__(self, name: str) -> Any:
```
maximilianruesch: Why do these changes have to be in here, rather than somewhere specific for enums? And should we not have a barrier like for `__call__` that prevents access from outside a Guppy context? This should make attribute accesses a lot clearer, i.e. we can differentiate between "this object does not have this attribute" and "this object has this attribute, but you can only access it inside a Guppy context".

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -486,6 +487,48 @@ def __iter__(self) -> Any:
         raise GuppyComptimeError(f"Expression of type `{self._ty}` is not iterable")
 
 
+class GuppyEnumObject(DunderMixin):
+    """The runtime representation of Guppy enum objects during tracing.
+
+    Enum values are tagged unions: the active variant is not known at comptime, so
+    variant fields are not accessible. This class is backed by a single Hugr wire
+    (like `GuppyObject`) and only exposes methods defined on the enum type.
```
maximilianruesch, line 495: In some cases, the variant *is* known at compile time, especially when it was just created in that function. In that case the tagged union could be narrowed to just contain one variant (or in general any subset), and then one can safely access the attributes that all the variants have. What do you think?

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -486,6 +487,48 @@ def __iter__(self) -> Any:
...
+        object.__setattr__(self, "_ty", ty)
+        object.__setattr__(self, "_wire", wire)
+
+    @hide_trace
+    def __getattr__(self, key: str) -> Any:
+        # We can only access methods
+        func = get_tracing_state().globals.get_instance_func(self._ty, key)
+        if func is None:
+            raise GuppyComptimeError(
+                f"Expression of type `{self._ty}` has no attribute `{key}`"
```
maximilianruesch: Perhaps it is better to talk about enum variants here, and expose that we do not support attribute access on variants whatsoever (except for functions). Perhaps also with the underlying reason. "Expression of type `E`" is technically correct, but confusing.

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -486,6 +487,48 @@ def __iter__(self) -> Any:
...
+            )
+        return lambda *xs: TracingDefMixin(func)(self, *xs)
+
+    @hide_trace
+    def __setattr__(self, key: str, value: Any) -> None:
+        raise AttributeError("Enum objects do not support attribute assignment")
```
maximilianruesch:
```suggestion
        raise AttributeError("Enum variants do not support attribute assignment")
```

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -486,6 +487,48 @@ def __iter__(self) -> Any:
...
+
+    @hide_trace
+    def __iter__(self) -> Any:
+        # Abstract Guppy objects are not iterable from Python since our iterator
+        # protocol doesn't work during tracing.
+        raise GuppyComptimeError(f"Expression of type `{self._ty}` is not iterable")
```
maximilianruesch: Similar variant wording changes here

guppylang-internals/src/guppylang_internals/tracing/object.py
nicolaassolini-qntm: Interesting, but this will create a discrepancy with respect to the not compile time version, where attribute access is never possible (will be with pattern matching). Also, since given enum: E,  e = enum.var1(): E and e = enum.var2(): E, we don't have an easy way to know the variant given an enum typed variable.

guppylang-internals/src/guppylang_internals/tracing/object.py
nicolaassolini-qntm: Saying "Enum of type `{self._ty}` has no method `{key}`" should suggest that only methods are accesible, what do you think?

tests/error/tracing_errors/enum_get_attr.err
nicolaassolini-qntm: what do you mean?

guppylang-internals/src/guppylang_internals/tracing/unpacking.py
nicolaassolini-qntm: An EnunType() is a sum of type in hugr (the sum of all variant types). To access field we should select the variant type, using Tag() and the unpack, this is why we need to the pattern match

tests/error/tracing_errors/input_mutate4.py
nicolaassolini-qntm: pylance 😅

guppylang-internals/src/guppylang_internals/tracing/object.py
nicolaassolini-qntm: `super.__getattr__` is not defined

guppylang-internals/src/guppylang_internals/tracing/object.py
nicolaassolini-qntm: Regarding the Guppy context check, it should not be needed (I added 2 test cases)
Regarding defining this method here, the alternative would be to create a GuppyEnumDefinition(GuppyDefinition) that implements this method only, and then make the guppy.enum decorator defines a new GuppyEnumDefinition. I'm not sure, implementing the method here seems to me a bit more maintainable since all the future changes about comptime can be done here. For sure having the a GuppyEnumDefinition is more elengant than a isinstance check

tests/error/tracing_errors/enum_set_method.py
```diff
@@ -0,0 +1,24 @@
...
+        return 42
+
+@guppy
+def bar(e: E) -> int:
+    return 42
+
+@guppy.comptime
+def test() -> None:
+    e = E.A()
+    e.foo = bar(e)
```
maximilianruesch: Can you not just assign `bar` as a function value? Then it would at least have the right signature.

tests/error/tracing_errors/enum_set_method.py
maximilianruesch: See [new comment](https://github.com/Quantinuum/guppylang/pull/1604/changes#r3008633002), this is not resolved.

guppylang-internals/src/guppylang_internals/tracing/object.py
maximilianruesch: This can work, but we should add a hint that explains that attributes other than methods are not accessible.

tests/error/tracing_errors/enum_get_attr.err
maximilianruesch: Will be resolved by the wording changes.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -439,7 +439,8 @@ def _check_name_id(
     ) -> tuple[ast.expr, Type]:
         """Helper method to check a name by its identifier, used for both `ast.Name` and
         `ast.Attribute` nodes. If allow_enum is False, we raise an error if the name
-        corresponds to an enum definition, since enums cannot be used as values."""
+        corresponds to an guppy.enum class, since enum classes cannot be used as values,
+        e.g: `x = Color` raise an error, while `x = Color.Red()` is fine."""
```
```suggestion
        e.g: `x = Color` raises an error, while `x = Color.Red()` is fine."""
```

guppylang-internals/src/guppylang_internals/tracing/object.py
maximilianruesch: > this will create a discrepancy with respect to the not compile time version

Well, it should be implemented for the Guppy context as well. It will be essentially the same, it is just "do we know the exact variant / restriction of variants at compile time?".

> Also, since given enum: E, e = enum.var1(): E and e = enum.var2(): E, we don't have an easy way to know the variant given an enum typed variable.

One can overapproximate the variant union, that is not a problem. And we can narrow them with dataflow analysis in the future.

Of course, this is a proposal for a future PR, and does not have to be done right now.

guppylang-internals/src/guppylang_internals/tracing/object.py
maximilianruesch: Yeah, I would go with the `GuppyEnumDefinition` because I do not like `isinstance` in superclasses. Imo it would be fair enough to create a `GuppyEnumMixin` that lives in this file and inherits from this mixin, and then you create a `GuppyEnumDefinition` that inherits from the new mixin. But open to discussion on this.

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -461,7 +462,7 @@ def __getattr__(self, key: str) -> Any:
         # Or a method
         func = get_tracing_state().globals.get_instance_func(self._ty, key)
         if func is None:
-            err = f"Expression of type `{self._ty}` has no attribute `{key}`"
+            err = f"Struct of type `{self._ty}` has no attribute `{key}`"
```
maximilianruesch: For structs it was fine. You could say `Expression of struct type ... has no attribute ...` for consistency (and similar for the iterables).

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -483,7 +484,49 @@ def __setattr__(self, key: str, value: Any) -> None:
...
+        object.__setattr__(self, "_ty", ty)
+        object.__setattr__(self, "_wire", wire)
+
+    @hide_trace
+    def __getattr__(self, key: str) -> Any:
+        # We can only access methods
+        func = get_tracing_state().globals.get_instance_func(self._ty, key)
+        if func is None:
+            raise GuppyComptimeError(
+                f"Enum variant of type `{self._ty}` has no method `{key}`"
```
maximilianruesch: On second thought (sorry), the "expression" part is pretty important. I propose Expression of enum type ... has no method ... to have a mix of both worlds (see other comment for structs; similar for iterables).

guppylang-internals/src/guppylang_internals/tracing/object.py
nicolaassolini-qntm: I added “and variant fields are not accessible,” but I’m not entirely happy with it. The issue is that we can’t tell whether the user is accessing an existing field or just a random attribute.

guppylang-internals/src/guppylang_internals/tracing/object.py
nicolaassolini-qntm: I see the point of a `GuppyEnumMixin` class but this wuold imply an empty `GuppyEnumDefinition`. I think that just creating a `GuppyEnumDefinition` wuold be better

maximilianruesch: Note that the `GuppyEnumDefinition` would live in `guppylang`, while the mixin would live in `guppylang_internals`

nicolaassolini-qntm: you're right!!

nicolaassolini-qntm: Since also GuppyTypeVarDefinition has a __getatrr__, as well as GuppyFunctionDefinition has a __call__ method, both in guppylang/defs, I defined a GuppyEnumDefinition there with the _get_atrr_ method.

guppylang-internals/src/guppylang_internals/tracing/object.py
maximilianruesch: Perhaps something like "it may be a variant field, but variant fields are inaccessible" as a `Note` subdiagnostic?

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -514,19 +514,24 @@ def __getattr__(self, key: str) -> Any:
...
+                f" Expression of enum type `{self._ty}` has no method `{key}`, "
+                "and variant fields are not accessible"
             )
         return lambda *xs: TracingDefMixin(func)(self, *xs)
 
     @hide_trace
     def __setattr__(self, key: str, value: Any) -> None:
-        raise AttributeError("Enum objects do not support attribute assignment")
+        raise AttributeError(
+            "Expression of enum type do not support attribute assignment"
```
maximilianruesch: Please leave this as `Enum variants do not support attribute assignment` (a slight reword from before)

PR #1605, opened 2026-03-26
chore: add snapshot test for disallowed shadowing of extern variables

ferbetanzo:
The `tests/integration/test_extern.py::test_extern_conditional_assign` test case can now be enabled and is expected to fail.

### DIFF
--- a/tests/integration/test_extern.py
+++ b/tests/integration/test_extern.py
@@ -1,10 +1,6 @@
-import pytest
-
 from hugr import ops, val
 
 from guppylang.decorator import guppy
-from guppylang_internals.error import GuppyError
-from guppylang_internals.checker.cfg_checker import VarMaybeNotDefinedError
 
 
 def test_extern_float(validate):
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -157,3 +157,4 @@ devenv.local.nix
 
 **/.benchmarks/
 .vscode/settings.json
+**/tests.integration*.hugr
```
maximilianruesch: This may be a change unrelated to your PR

tests/integration/test_extern.py
```diff
@@ -48,8 +48,9 @@ def main() -> float:
     validate(main.compile_function())
 
 
-@pytest.mark.skip("See https://github.com/quantinuum/guppylang/issues/827")
-def test_extern_conditional_assign(validate):
+# See https://github.com/quantinuum/guppylang/issues/827
+@pytest.mark.xfail(reason="Shadowing of extern variables is not allowed")
```
maximilianruesch: Instead of an `xfaill`, why not move this to an actual error test? The discussion in the linked issue is done and it was decided to keep the error (and disallow shadowing for the time being), so we should assert the whole message rendering here as well.

.gitignore
ferbetanzo: Nice catch

tests/integration/test_extern.py
ferbetanzo: I've added some assertions on the error type and fields

maximilianruesch: What I meant is: Why not make this a snapshot test like those in `tests/error`?

PR #1615, opened 2026-03-31
fix: Decouple from tket and pytket dependencies

### LINKED ISSUE #1595: [Maintenance]: Decouple Guppy from tket and pytket packages
### Description

Decouple `guppylang` and `guppylang-internals` from `tket` and `pytket` by only importing from those packages on a "need-to-have" basis.

Tests should be added for this decoupling.

### Motivation

`tket` and `pytket` packages have recently (in #1440 and #1574) become non-optional dependencies of `guppylang`. In the midst of those changes, the imports from those packages have risen to top-level imports in the respective files.

This made the `guppylang` ecosystem dependent on `tket` and `pytket` always being importable, which they recently had some issues with (for example because of non-existent glibc versions that the compiled Rust bindings require). However, this dependence is also present when Guppy users are not even using `pytket` features from Guppy!

### Scope

_No response_

### Component

Dependencies

### Priority

Medium - Should be done soon

### Additional Context

Once we raise support to at least Python 3.15, we can use https://peps.python.org/pep-0810/ to introduce lazy imports and make the whole thing more maintainable. Until then, we have to go a more verbose route.

maximilianruesch:
Decouples Guppy from its dependencies `tket` and `pytket`, by virtue of dynamically importing. This ensures that dependencies are only imported when absolutely necessary, i.e. when related features are used.

A test framework with a sprinkle of importlib magic is added to cement the decoupling.

Closes #1595

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
+++ b/guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
@@ -162,6 +162,7 @@ class ParsedPytketDef(CallableDef, CompilableDef):
         from pytket.circuit import Circuit  # Decoupled import
         from tket.circuit import Tk2Circuit  # Decoupled import
 
+        # Type mismatch should have been raised in decorator
         assert isinstance(self.input_circuit, Circuit)
         # TODO extract the correct entry point from the module
         circ = envelope.read_envelope(
@@ -362,6 +363,7 @@ def _signature_from_circuit(
     from guppylang.std.quantum import qubit
     from pytket.circuit import Circuit  # Decoupled import
 
+    # Type mismatch should have been raised in decorator
     assert isinstance(input_circuit, Circuit)
 
     assert isinstance(qubit, GuppyDefinition)


guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
```diff
@@ -72,7 +70,7 @@ class RawPytketDef(ParsableDef):
     """
 
     python_func: PyFunc
-    input_circuit: Circuit
+    input_circuit: Any
```
ss2165, line 73: no hope in using an `if TYPE_CHECKING: ` block to keep this?

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
```diff
@@ -161,6 +159,10 @@ def compile_outer(
         self, module: DefinitionBuilder[OpVar], ctx: CompilerContext
     ) -> "CompiledPytketDef":
         """Adds a Hugr `FuncDefn` node for this function to the Hugr."""
+        from pytket.circuit import Circuit  # Decoupled import
+        from tket.circuit import Tk2Circuit  # Decoupled import
+
+        assert isinstance(self.input_circuit, Circuit)
```
ss2165, line 166: better to raise a `TypeError` with useful message

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
maximilianruesch: May be possible, I shall give it a try

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
maximilianruesch: One should never reach this point with "not a circuit", as the decorator should have thrown earlier. May be able to omit this one if I can use `TYPE_CHECKING`.

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
maximilianruesch: Turns out for classes this is not possible, since it appears to not be purely type checking. I.e. this fails (but succeeds when `Iterable` is imported outside):
```python
from typing import TYPE_CHECKING

if TYPE_CHECKING:
    from typing import Iterable

class A:
    b: Iterable[int]
```

ss2165: shame

PR #1628, opened 2026-04-02
refactor!: Make frames mandatory and reduce usage of Globals

mark-koch:
Some refactors around the usage of `Globals` instances, pulled out from #1441:
* Reduces uses of the `Globals` object: `get_instance_func` moved into the engine and `python_value_to_guppy_type` no longer requires globals
* Also ensures that all definition now have a corresponding frame in the `DEF_STORE`

BREAKING CHANGE: Removed `Globals.get_instance_func`. Use `ENGINE.get_instance_func` instead.
BREAKING CHANGE: Removed `TracingState.globals`
BREAKING CHANGE: Removed `CompilerContext.checked_globals`
BREAKING CHANGE: `DEF_STORE.register_def` now requires a mandatory stack frame
BREAKING CHANGE: Removed `globals` argument from `python_value_to_guppy_type`

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/core.py
+++ b/guppylang-internals/src/guppylang_internals/checker/core.py
@@ -12,7 +12,6 @@ from typing import (
     NamedTuple,
     TypeAlias,
     TypeVar,
-    cast,
     overload,
 )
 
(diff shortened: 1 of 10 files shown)


mark-koch:
> Please review the coverage changes and adjust if necessary.

As far as I can see, the uncovered changes are all from moving stuff around, i.e. they were uncovered before. Also, many of them are defensive `assert_never` calls, which we never expect to hit

PR #1630, opened 2026-04-02
feat!: Support generic comptime functions

### LINKED ISSUE #896: Support generic comptime functions

mark-koch:
Closes #896 

BREAKING CHANGE: `TracedFunctionDef` is no longer a `CompilableDef`, but now a `CheckableGenericDef` that returns a monomorphic `TracedMonoFunctionDef` after checking
BREAKING CHANGE: `trace_function` now requires the generic instantiation as an extra argument


tests/integration/tracing/test_generic.py
```diff
@@ -0,0 +1,72 @@
...
+        return foo(1.5, True) + foo(1.0, False)
+
+    run_float_fn_approx(main, 0.5)
+
+
+def test_dependent(run_float_fn_approx):
+    T = guppy.type_var("T", copyable=True, droppable=True)
+
+    @guppy.comptime
+    def foo(x: T @ comptime) -> T:
```
acl-cqc, line 44: I admit I'm a bit confused here as to whether the `@ comptime` applied to the `T` here or the `x` or indeed whether it's necessary at all?

tests/integration/tracing/test_generic.py
```diff
@@ -0,0 +1,72 @@
...
+    T = guppy.type_var("T")
+    n = guppy.nat_var("n")
+
+    @guppy.struct
+    class S(Generic[T, n]):
+        x: T
+
+        @guppy.comptime
+        def foo(self) -> int:
+            assert isinstance(n, int)
```
acl-cqc, line 64: Presumably this is not in any sense "required" but we are just checking that it works? (Which is good to check - i.e. that we are able to access `n` during comptime execution)

tests/integration/tracing/test_generic_py312.py
```diff
@@ -0,0 +1,78 @@
...
+
+def test_mixed(run_int_fn):
+    n = guppy.nat_var("n")
+
+    @guppy.comptime
+    def foo[m: nat](xs: array[int, n], ys: array[int, m], k: int @ comptime) -> int:
+        assert isinstance(n, int)
+        assert isinstance(m, int)
+        assert isinstance(k, int)
+        return n + m + k
```
acl-cqc:
```suggestion
        return xs[k] + ys[k]
```

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -75,19 +84,37 @@ def trace_function(
...
     state = TracingState(ctx, DFContainer(builder, ctx, {}), node, func_def)
     with set_tracing_state(state):
-        inputs = [
-            unpack_guppy_object(
-                GuppyObject(inp.ty, wire),
-                builder,
+        generic_values = {
+            x: const_argument_to_python_value(arg)
+            for x, arg in generic_args.items()
+            # TODO: We don't have a comptime representation of types yet, so we can only
```
acl-cqc, line 89: This "representation of types" that we need is not a Hugr representation, right, just a python representation to provide *at comptime*?

In which case, we do have one: the same representation as in the rest of the compiler, namely `guppylang_internals.tys.....`. So we *could* expose that. I'm not saying we *should***, but maybe clarify the comment ("public representation" perhaps, or say something about not wanting to expose guppylang-internals stuff)

** I mean, how much guppylang-internals to expose, is a subjective question, but I'm not going to push you one way or the other. We do in a few places I think but not many.

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -76,9 +75,32 @@ def parse(self, globals: Globals, sources: SourceMap) -> "TracedFunctionDef":
...
+class TracedFunctionDef(RawTracedFunctionDef, CallableDef, CheckableGenericDef):
     defined_at: ast.FunctionDef
 
+    @property
+    def params(self) -> Sequence[Parameter]:
+        """Generic parameters of this function."""
+        return self.ty.params
+
+    def check(self, type_args: Inst, globals: Globals) -> "TracedMonoFunctionDef":
+        """Type checks the body of the function."""
```
acl-cqc:
```suggestion
        """Type checks (and thus traces/executes) the body of the function."""
```

acl-cqc: Actually I'm not sure it does trace/execute?? That only happens in compile? In which case are we really doing "check the body" at all?

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -162,14 +190,9 @@ def compile_call(
         node: AstNode,
     ) -> CallReturnWires:
         """Compiles a call to the function."""
-        type_args: Inst = ()  # Comptime functions are not generic
-        func_ty: ht.FunctionType = self.ty.instantiate(type_args).to_hugr(ctx)
-        hugr_type_args: list[ht.TypeArg] = [arg.to_hugr(ctx) for arg in type_args]
```
acl-cqc, line 168: This has become the new `generic_args: Mapping[str, Argument]`, right?

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -162,14 +190,9 @@ def compile_call(
...
         num_returns = len(type_to_row(self.ty.output))
         with dfg.builder.set_ast_context(node):
-            call = dfg.builder.call(
-                self.func_def, *args, instantiation=func_ty, type_args=hugr_type_args
-            )
+            call = dfg.builder.call(self.func_def, *args)
```
acl-cqc, line 199: So...just checking...the `ht.FunctionType` has been compiled into `self.func_def`, right?

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -75,19 +84,37 @@ def trace_function(
...
+
+        input_wires = iter(builder.inputs())
+        inputs = []
+        for inp in ty.inputs:
+            if InputFlags.Comptime in inp.flags:
+                assert inp.name is not None
+                val = generic_values.pop(inp.name)
+            else:
                 # Function inputs are only allowed to be mutable if they are borrowed.
                 # For owned arguments, mutation wouldn't be observable by the caller,
```
acl-cqc, line 103: (driveby?) Not sure I follow this. The mutation wouldn't be observable by the caller but only because the caller has lost/handed over the object that was mutated, so that respects python semantics (in that we have statically disallowed programs in which they'd be able to see the mutations!)

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -218,3 +245,49 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
...
+    # The values of Python 3.12+ style generics are generated using an annotation scope
+    # (see https://docs.python.org/3/reference/compound_stmts.html#generic-functions).
+    # Thus, their values will be bound in the `__closure__` table of the function and
+    # referenced in the `co_freevars` of the functions `__code__`.
+    if f.__closure__ is not None:
+        for i, x in enumerate(f.__code__.co_freevars):
+            if x in generic_values:
+                f.__closure__[i].cell_contents = generic_values.pop(x)
+
+    # If all values have been assigned, we can stop here
```
acl-cqc: nit: I think all the code below is fine (does nothing) if `generic_values` are empty... is it worth this? (It's nice to have only one `yield`)

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -218,3 +245,49 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
...
+    match arg.const:
+        case ConstValue(value=v):
+            return v
+        case BoundConstVar() | ExistentialConstVar():
+            # By this point, everything should be monomorphized!
+            raise InternalGuppyError("Unexpected const variable")
+
+
+@contextmanager
+def insert_generic_args(
```
acl-cqc: how about `add_generic_args_to_globals` or `...to_function_globals`? That'd make the callsite a lot clearer (I was thinking it was gonna prepend them to an existing list of generic args or something)

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -218,3 +245,49 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
...
+    f: Callable[..., Any], generic_values: dict[str, Any]
+) -> Iterator[None]:
+    """Context manager that updates the given function to allow access to the
+    instantiation of generic parameters."""
+    # The values of Python 3.12+ style generics are generated using an annotation scope
+    # (see https://docs.python.org/3/reference/compound_stmts.html#generic-functions).
+    # Thus, their values will be bound in the `__closure__` table of the function and
```
acl-cqc: Argh. This is by far the nastiest bit. I note there is no python version check, either, which might make me feel slightly happier. I didn't see this from the link you give just above, is that the best reference?

acl-cqc (approved):
Looks great, Mark, lots of nice examples in tests :). Comments relate only to comments/docs in the code really although I am nervous about the `__closure__`, `__code__` and `co_freevars` bit....is there any alternative there?

tests/integration/tracing/test_generic.py
mark-koch: It applies to the `x`. It says that `foo` takes an input `x` of type `T` whose value must be known at compile-time. This in turn is syntactic sugar for a generic function

```python
@guppy.comptime
def foo[T: (Copy, Drop), x: T]() -> T: ...
```

guppylang-internals/src/guppylang_internals/tracing/function.py
mark-koch: In my mind we should not expose the internal `Type` to users. Ideally they should be able to write types like they're used to from Python/Guppy and we turn them into our internal representation later.

What I mean by "comptime representation of types", is that users can use regular type syntax (e.g. `array[int, 42]`) in comptime expressions, and it evaluates to something that we can turn into our internal representation

guppylang-internals/src/guppylang_internals/definition/traced.py
mark-koch: Good point, that docstring was a copy paste mistake

guppylang-internals/src/guppylang_internals/definition/traced.py
mark-koch: Yes, `generic_args` remembers the instantiation 👍

guppylang-internals/src/guppylang_internals/definition/traced.py
mark-koch: Yes 👍

guppylang-internals/src/guppylang_internals/tracing/function.py
mark-koch: > in that we have statically disallowed programs in which they'd be able to see the mutations!

True, by only allowing mutation of values that are affine. I just realised that this is actually something I overlooked with the mutable struct change: at the moment comptime structs are always mutable, unless they are owned inputs. Now we can just respect the `frozen` flag on structs and everything should be fine!

Will make another PR to address this

guppylang-internals/src/guppylang_internals/tracing/function.py
mark-koch: This was the best I could come up with, but would be very happy if you could find something less hacky 😅

I'll add a version check, but don't know of a better reference. But will add some more explanation

mark-koch: Ooops turns out the previous explanation was wrong. We also have to check the closure vars for pre 3.12 generics. Updated the comments

PR #1645, opened 2026-04-07
feat!: Add type aliases

### LINKED ISSUE #1066: Type aliases

ss2165:
## Summary

Adds `guppy.type_alias()` to create named type aliases. Closes #1066.

### Basic alias
```python
MyInt = guppy.type_alias("MyInt", "int")

@guppy
def f(x: MyInt) -> MyInt:
    return x + 1
```

The name is always passed explicitly as the first argument — consistent with `guppy.type_var("T")`.

### Generic aliases

Free type variables in the body are collected implicitly, in order of first appearance:
```python
T = guppy.type_var("T")
Pair = guppy.type_alias("Pair", "tuple[T, T]")  # T inferred from body
```

Or pass `params=` to fix the order explicitly (required when order matters):
```python
A = guppy.type_var("A")
B = guppy.type_var("B")
Map = guppy.type_alias("Map", "tuple[B, A]", params=[A, B])
```

### Python 3.12+ `type` statement

On Python 3.12+ the PEP 695 `type` statement is also supported. The alias value must be a quoted string; type parameters are read from the `[...]` list:

```python
# type Foo[T] = "..."  maps TypeVar bounds to guppy constraints:
#   no bound       → linear (copyable=False, droppable=False)
#   T: Copy        → copyable only
#   T: Drop        → droppable only
#   T: (Copy, Drop) → classical (copyable and droppable)
type Pair[T: (Copy, Drop), U: (Copy, Drop)] = "tuple[T, U]"
Pair = guppy.type_alias(Pair)
```

BREAKING CHANGE: ParsedRecursiveTypeDef extended to include ParsedTypeAliasDef


tests/error/test_alias_errors.py
```diff
@@ -0,0 +1,26 @@
...
+files = [str(f) for f in files]
+
+
+@pytest.mark.parametrize("file", files)
+def test_alias_errors(file, capsys, snapshot):
+    run_error_test(file, capsys, snapshot)
+
+
+def test_type_alias_bad_type_syntax():
+    with pytest.raises(SyntaxError, match="Not a valid Guppy type: `foo bar`"):
```
mark-koch, line 25: Can you add an error tests for invalid aliases that fail at a later point, i.e. via a GuppyError instead of a SyntaxError?

tests/error/alias_errors/mutual_recursive.err
```diff
@@ -0,0 +1,20 @@
...
+Notes:
+  | 
+3 | 
+4 | Alias1 = guppy.type_alias("Alias2")
+  | ----------------------------------- Alias `Alias1` is part of this cycle
+5 | Alias2 = guppy.type_alias("Alias1")
+  | ----------------------------------- Alias `Alias2` is part of this cycle
+
+Help: Type aliases must eventually resolve to a non-alias type. Break the cycle
+by inlining one alias or introducing a struct or enum wrapper.
```
mark-koch, line 18: Inlining can never break cycles? Similarly, I don't see how structs or enums would help?

tests/error/alias_errors/recursive.err
```diff
@@ -0,0 +1,17 @@
...
+2 | 
+3 | 
+4 | MyAlias = guppy.type_alias("MyAlias")
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Type alias `MyAlias` expands to itself
+
+Note:
+  | 
+3 | 
+4 | MyAlias = guppy.type_alias("MyAlias")
+  | ------------------------------------- Alias `MyAlias` is part of this cycle
```
mark-koch, line 12: This note doesn't seem useful if it's a self cycle

guppylang/src/guppylang/decorator.py
```diff
@@ -353,6 +354,20 @@ def const_var(self, name: str, ty: str) -> TypeVar:
...
         return GuppyTypeVarDefinition(defn, TypeVar(name))  # type: ignore[return-value]
 
+    def type_alias(self, ty: str) -> Any:
+        """Creates a new type alias."""
+        type_ast = _parse_expr_string(
+            ty, f"Not a valid Guppy type: `{ty}`", DEF_STORE.sources
+        )
+        defn = RawTypeAliasDef(
+            DefId.fresh(),
+            ty,
```
mark-koch, line 364: This sets the name of the definition as the type string which seems odd.

I see that you use a separate `_alias_name` function to resolve the name later by inspecting the Python scope. Note that this is quite different from how we handle naming of definition in other parts of the compiler. Usually, `guppy.type_var` etc take the definition name as an argument.

Checking the scope seems reasonable though, but it should happen here in the decorator! Also, the fallback probably shouldn't be the type string. Maybe, instead raise an error and force users to pass it as an argument?

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,204 @@
...
+    type_ast: ast.expr
+    description: str = field(default="type alias", init=False)
+
+    def check(self, globals: Globals) -> "CheckedTypeAliasDef":
+        recursion_ctx = TypeParsingCtx(globals, allow_free_vars=True)
+        check_not_recursive(self, recursion_ctx)
+
+        ctx = TypeParsingCtx(globals, allow_free_vars=True)
+        ty = type_from_ast(self.type_ast, ctx)
+        params = tuple(ctx.param_var_mapping.values())
```
mark-koch, line 97: This looks like an attempt to support generic aliases? I think those would be useful, but we should either add tests for generic aliases or just set `params = []` here

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,204 @@
...
+    # cycle, so avoid attaching the same note more than once.
+    if any(
+        isinstance(child, RecursiveTypeAliasError.AliasNote)
+        and child.alias_name == alias_name
+        for child in err.children
+    ):
+        return
+    err.add_sub_diagnostic(
+        RecursiveTypeAliasError.AliasNote(defn.defined_at, alias_name)
+    )
```
mark-koch, line 194: This will fail to render if the aliases are defined in separate files

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,204 @@
...
+
+def _add_alias_note(
+    err: RecursiveTypeAliasError, defn: ParsedTypeAliasDef, globals: Globals
+) -> None:
+    alias_name = _alias_name(defn, globals)
+    # The same recursive error is re-raised while unwinding through each alias in the
+    # cycle, so avoid attaching the same note more than once.
+    if any(
+        isinstance(child, RecursiveTypeAliasError.AliasNote)
+        and child.alias_name == alias_name
```
mark-koch, line 188: Going by name feels a bit fragile, can we store `DefId`s instead?

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,204 @@
...
+        err = RecursiveTypeAliasError(loc, cycle)
+        _add_alias_note(err, defn, ctx.globals)
+        raise GuppyError(err)
+
+    try:
+        with _patched_check_instantiate(defn, dummy_check_instantiate):
+            type_from_ast(defn.type_ast, ctx)
+    except GuppyError as err:
+        if isinstance(err.error, RecursiveTypeAliasError):
+            _add_alias_note(err.error, defn, ctx.globals)
```
mark-koch, line 174: I don't understand why we need to add the notes while unwinding the exception. We already have the cycle, can't we just add all notes in one go?

guppylang-internals/src/guppylang_internals/definition/alias.py
mark-koch: While testing, I also found some other rendering errors. For example

```python
Alias1 = guppy.type_alias("Alias2")
Alias2 = guppy.type_alias("Alias3")
Alias3 = guppy.type_alias("Alias2")

@guppy
def main(x: Alias1) -> Alias1:
    return x

main.check()
```

fails with

```
Error in sys.excepthook:
  File "guppylang-internals/src/guppylang_internals/error.py", line 120, in hook
    renderer.render_diagnostic(err.error)
  File "guppylang-internals/src/guppylang_internals/diagnostic.py", line 323, in render_diagnostic
    self.render_snippet(
  File "guppylang-internals/src/guppylang_internals/diagnostic.py", line 419, in render_snippet
    leading_whitespace = min(len(line) - len(line.lstrip()) for line in all_lines)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: min() iterable argument is empty
```

Not sure what's going on there...

tests/integration/test_type_alias.py
mark-koch, line 1: Can you add some tests for how the alias acyclicity checking interacts with structs and enums? I.e. add some integration and error tests that involve struct fields/enum variants whose types contain aliases.

tests/error/test_alias_errors.py
ss2165: yes see undefined_type.py

tests/error/alias_errors/mutual_recursive.err
ss2165: yes, removed

tests/error/alias_errors/recursive.err
ss2165: yes, removed

guppylang/src/guppylang/decorator.py
ss2165: I've gone with requiring the name in this PR to match type_var

guppylang-internals/src/guppylang_internals/definition/alias.py
ss2165: generic support added and tested, along with 3.12 `type ...` syntax

guppylang-internals/src/guppylang_internals/definition/alias.py
ss2165: fixed, and your example added as a test

tests/integration/test_type_alias.py
ss2165: yes, see struct_cycle.py

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,236 @@
...
+    def check(self, globals: Globals) -> "CheckedTypeAliasDef":
+        if self.params is not None:
+            # Explicit params: re-index them and pre-load into the context so that
+            # type vars in the body are resolved to these parameters in order.
+            reindexed = [p.with_idx(i) for i, p in enumerate(self.params)]
+            param_var_mapping = {p.name: p for p in reindexed}
+            check_not_recursive(
+                self, TypeParsingCtx(globals, param_var_mapping=dict(param_var_mapping))
+            )
+            ctx = TypeParsingCtx(globals, param_var_mapping=param_var_mapping)
```
mark-koch: Why do we need two different `TypeParsingCtx`s here and why duplicate the `param_var_mapping`? The context should have `allow_free_vars=False` so nothing should be mutated

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,236 @@
...
+
+    Notes are only emitted when the alias definition has a valid, same-file span — i.e.
+    when the AST node was annotated with file information by `_parse_expr_string`.
+    Cross-file or un-annotated spans are silently skipped; the cycle chain in the main
+    error's span label is still fully informative on its own.
+    """
+    import ast as _ast
+
+    from guppylang_internals.ast_util import get_file
+    from guppylang_internals.span import Span
```
mark-koch: Why local imports?

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,236 @@
...
+        if isinstance(node, Span):
+            return node.file
+        return get_file(node)
+
+    unique_defs = cycle_defs[:-1]  # drop the repeated last element
+    if len(unique_defs) <= 1:
+        return
+
+    # Determine the file that the main error is anchored to (may be None if unset)
+    err_file: str | None = _span_file(err.span)
```
mark-koch:
```suggestion
    err_file = to_span(err.span).file if err.span else None
````

File: guppylang-internals/src/guppylang_internals/definition/alias.py [hunk 12]
```diff
@@ -0,0 +1,236 @@
...
+
+    # Use DefId for deduplication so that aliases with identical names don't collide.
+    seen_ids: set[DefId] = {
+        child.defn_id
+        for child in err.children
+        if isinstance(child, RecursiveTypeAliasError.AliasNote)
+    }
```
mark-koch: The error starts without any children so this will always be empty?

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,236 @@
...
+    # The last element of unique_defs is the alias whose definition the error span
+    # already underlines — skip it to avoid a redundant note on the same line.
+    defs_to_annotate = unique_defs[:-1]
+    for defn in defs_to_annotate:
+        if defn.id not in seen_ids and defn.defined_at is not None:
+            # Skip if the AST node lacks file annotation or is from a different file
+            note_file = get_file(defn.defined_at)
+            if note_file is None or note_file != err_file:
```
mark-koch: Can `note_file` be None if `defn.defined_at` is not None?

guppylang/src/guppylang/decorator.py
```diff
@@ -406,6 +421,127 @@ def const_var(self, name: str, ty: str) -> TypeVar:
...
+        )
+        explicit_params = _params_from_type_alias_params(ta.__type_params__)
+        defn = RawTypeAliasDef(
+            DefId.fresh(),
+            ta.__name__,
+            type_ast,
+            type_ast,
+            explicit_params,
+        )
+        assert frame is not None, "Could not determine calling frame for type alias"
```
mark-koch: Why accept `frame | None` when you assert that it's not None here?

guppylang/src/guppylang/decorator.py
```diff
@@ -708,14 +844,28 @@ def _parse_expr_string(ty_str: str, parse_err: str, sources: SourceMap) -> ast.e
...
 
     # Try to annotate the type AST with source information. This requires us to
-    # inspect the stack frame of the caller
+    # inspect the stack frame of the caller.
     if caller_frame := get_calling_frame():
         info = inspect.getframeinfo(caller_frame)
+        filename = info.filename
+        # Prefer getsourcelines (works for real files and importable modules), but
+        # fall back to linecache for interactive contexts like `python -c` or REPLs
+        # where the module may not be importable via inspect.getmodule().
```
mark-koch: These changes seem unrelated, why is this needed?

guppylang/src/guppylang/decorator.py
```diff
@@ -863,3 +1013,116 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
...
+
+    result = []
+    for i, p in enumerate(params):
+        if not isinstance(p, GuppyDefinition):
+            raise TypeError(
+                "type_alias params must be type variables created with "
+                f"guppy.type_var() or guppy.nat_var(), got {p!r}"
+            )
+        defn = p.wrapped
+        if not isinstance(defn, ParamDef) or isinstance(defn, RawConstVarDef):
```
mark-koch: `RawConstVarDef` is a subclass of `ParamDef` so this `or` doesn't do anything

guppylang/src/guppylang/decorator.py
```diff
@@ -863,3 +1013,116 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
...
+    from guppylang_internals.definition.parameter import ParamDef, RawConstVarDef
+
+    from guppylang.defs import GuppyDefinition
+
+    result = []
+    for i, p in enumerate(params):
+        if not isinstance(p, GuppyDefinition):
+            raise TypeError(
+                "type_alias params must be type variables created with "
+                f"guppy.type_var() or guppy.nat_var(), got {p!r}"
```
mark-koch: Also `guppy.const_var`

guppylang/src/guppylang/decorator.py
```diff
@@ -863,3 +1013,116 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
...
+def _params_from_list(params: list[Any]) -> list[Parameter]:
+    """Extract :class:`~guppylang_internals.tys.param.Parameter` objects from a list of
+    Guppy type-variable definitions (e.g. results of :func:`guppy.type_var`).
+
+    The index of each parameter is set to its position in the list so that
+    ``Alias[int, bool]`` binds the first param to ``int`` and the second to ``bool``.
+    """
+    from guppylang_internals.definition.parameter import ParamDef, RawConstVarDef
+
+    from guppylang.defs import GuppyDefinition
```
mark-koch: Why local imports?

guppylang/src/guppylang/decorator.py
```diff
@@ -863,3 +1013,116 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
...
+    ``nat``-const params — use ``params=[N]`` for those).
+    """
+    import typing
+
+    from guppylang_internals.tys.param import TypeParam
+
+    from guppylang.std.lang import Copy, Drop
+
+    # TypeVarTuple was added in Python 3.11; ParamSpec in 3.10.
+    _TypeVarTuple: type | None = getattr(typing, "TypeVarTuple", None)
```
mark-koch: Could use `typing_extensions`?

guppylang/src/guppylang/decorator.py
```diff
@@ -863,3 +1013,116 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
...
+                "in Guppy type aliases."
+            )
+        if not isinstance(tp, typing.TypeVar):
+            raise TypeError(
+                f"Unsupported type parameter {tp!r} in Guppy type alias. "
+                "Only TypeVar is supported."
+            )
+        # Python 3.14+ uses __constraints__ for `T: (A, B)` style;
+        # Python 3.12/3.13 may use __bound__ for `T: A`.
+        constraints: tuple[Any, ...] = getattr(tp, "__constraints__", ())
```
mark-koch: Imo better to use `sys.version_info` here

guppylang/src/guppylang/decorator.py
```diff
@@ -863,3 +1013,116 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
...
+                i, tp.__name__, must_be_copyable=False, must_be_droppable=False
+            )
+        elif bound is Copy:
+            param = TypeParam(
+                i, tp.__name__, must_be_copyable=True, must_be_droppable=False
+            )
+        elif bound is Drop:
+            param = TypeParam(
+                i, tp.__name__, must_be_copyable=False, must_be_droppable=True
+            )
```
mark-koch: A lot of duplication here. Consider something like 

```python
if bound is None:
    bounds = ()
elif isinstance(bound, tuple):
    bounds = bound
else:
    bounds = (bound,)
```

and then construct the `TypeParam` based on that

guppylang/src/guppylang/decorator.py
```diff
@@ -863,3 +1013,116 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
...
+    return result
+
+
+def _params_from_type_alias_params(type_params: tuple[Any, ...]) -> list[Parameter]:
+    """Convert Python 3.12+ ``type`` statement type params to guppy Parameters.
+
+    Handles ``TypeVar`` with optional ``Copy``/``Drop`` bounds. Raises
+    ``TypeError`` for unsupported parameter kinds (``TypeVarTuple``,
+    ``ParamSpec``) and for bounds that require ``globals`` to resolve (e.g.
+    ``nat``-const params — use ``params=[N]`` for those).
```
mark-koch: Mhmm the fact that this doesn't work really makes me question if we should even support this syntax?

tests/error/test_alias_errors.py
```diff
@@ -0,0 +1,26 @@
...
+# Turn paths into strings, otherwise pytest doesn't display the names
+files = [str(f) for f in files]
+
+
+@pytest.mark.parametrize("file", files)
+def test_alias_errors(file, capsys, snapshot):
+    run_error_test(file, capsys, snapshot)
+
+
+def test_type_alias_bad_type_syntax():
```
mark-koch, line 24: Can you add tests for all the different errors that you raise from `guppy.type_alias`?

tests/integration/test_type_alias_py312.py
```diff
@@ -0,0 +1,103 @@
...
+from typing import Generic
+
+from guppylang import guppy
+from guppylang.std.lang import Copy, Drop
+
+
+def test_simple_alias_from_type_stmt(run_int_fn):
+    """A plain `type X = "..."` alias works just like the string form."""
+    type MyInt = "int"
+    MyInt = guppy.type_alias(MyInt)
```
mark-koch: Imo this doesn't really bring many benefits over the other syntax

tests/integration/test_type_alias_py312.py
```diff
@@ -0,0 +1,103 @@
...
+    ``guppy.type_var("T")`` (copyable=True, droppable=True).
+    """
+    T = guppy.type_var("T")  # copyable=True, droppable=True by default
+
+    @guppy.struct
+    class Box(Generic[T]):
+        value: T
+
+    # T: (Copy, Drop) matches the default guppy.type_var("T") constraints
+    type Boxed[T: (Copy, Drop)] = "Box[T]"
```
mark-koch: Can you add an error test for what happens if the bound doesn't match?

mark-koch (commented):
Thanks, generic aliases make this feature a lot better!

The biggest question imo is whether we actually want to support the 3.12 syntax? I don't see much benefit since we have to wrap it in `guppy.type_var` anyways. Also, const vars are not supported so users have to fall back to the other syntax anyways...

ss2165:
Agree that the python 312 syntax doesn't add enough, reverting

guppylang/src/guppylang/decorator.py
ss2165: some kind of merge-with-vestigial error sorry

tests/error/alias_errors/mutual_recursive.err
maximilianruesch, line 1: Perhaps add a test that tests cycles containing more than two members? Or two independent cycles? Or a cycle that contains another smaller cycle (the output of the error message could be either, but has to be deterministic).

tests/error/alias_errors/too_many_args.py
```diff
@@ -0,0 +1,23 @@
...
+
+T = guppy.type_var("T")
+
+
+@guppy.struct
+class Box(Generic[T]):
+    value: T
+
+
+# Too many type args for generic alias (Box[T] takes 1, given 2)
```
maximilianruesch: I believe this comment to be on the wrong line (it should be above the instantiation) and is rather confusing. Either move or remove it.

tests/error/alias_errors/struct_cycle.py
maximilianruesch, line 1: Since structs and enums are not processed in _exactly_ the same manner, I would prefer if an actual enum test was added. My bad if I overlooked one.

guppylang/src/guppylang/decorator.py
```diff
@@ -365,6 +375,50 @@ def const_var(self, name: str, ty: str) -> TypeVar:
...
+        .. code-block:: python
+
+            T = guppy.type_var("T")
+            U = guppy.type_var("U")
+            Pair = guppy.type_alias("Pair", "tuple[T, U]", params=[T, U])
+
+        When ``params`` is omitted, free type variables are collected from the body
+        in order of first appearance.
+        """
+        frame = get_calling_frame()
```
maximilianruesch: By inlining this call, we can defer calling frame detection until after the string is parsed and potential errors are thrown, giving a tiny nudge to performance (while also reducing the cognitive lifetime of the variable / value, so win-win).

tests/error/test_alias_errors.py
```diff
@@ -0,0 +1,47 @@
...
+    with pytest.raises(SyntaxError, match="Not a valid Guppy type: `foo bar`"):
+        guppy.type_alias("MyAlias", "foo bar")
+
+
+def test_type_alias_invalid_param():
+    with pytest.raises(
+        TypeError,
+        match="type_alias params must be type variables created with",
+    ):
+        guppy.type_alias("MyAlias", "int", params=["not a type var"])
```
maximilianruesch, line 34: Is
```
T = guppy.type_var("T")
guppy.type_alias("MyAlias", "int", params=["T"])
```
rejected?

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,238 @@
...
+    object.__setattr__(defn, "check_instantiate", replacement)
+    try:
+        yield
+    finally:
+        # Remove the instance attribute so method resolution falls back to the
+        # class descriptor, restoring the original behaviour cleanly.
+        object.__delattr__(defn, "check_instantiate")
+
+
+def check_not_recursive(defn: ParsedTypeAliasDef, ctx: TypeParsingCtx) -> None:
```
maximilianruesch: This seems like it could benefit from changes such as https://github.com/Quantinuum/guppylang/pull/1809. Fair enough if you want to stick with this for now, but I suggest that you open a follow up since we do not want to stick with such brittle patches long term.

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,238 @@
...
+    We do not have a separate alias-expansion pass, so we detect recursion by
+    temporarily swapping out this alias's `check_instantiate` method while parsing its
+    target type. If parsing the alias body reaches this same alias again, the patched
+    method fires and turns that recursive re-entry into a user-facing cycle diagnostic.
+
+    All cycle notes are attached at once inside `dummy_check_instantiate` so that only
+    aliases that are actually part of the cycle receive notes (not outer aliases that
+    merely lead to a cycle).
+    """
+    token = _active_alias_checks.set((*_active_alias_checks.get(), defn))
```
maximilianruesch: Cheeky trick with the `ContextVar`. Consider moving it to just above this function to signal that it is not used in the functions above this one.

guppylang-internals/src/guppylang_internals/definition/alias.py
```diff
@@ -0,0 +1,238 @@
...
+    seen_ids: set[DefId] = set()
+    for defn in unique_defs[:-1]:
+        if defn.id in seen_ids or defn.defined_at is None:
+            continue
+        note_file = get_file(defn.defined_at)
+        assert note_file is not None, (
+            f"defined_at node for alias `{defn.name}` has no file annotation"
+        )
+        if note_file != err_file:
+            continue
```
maximilianruesch: Perhaps we can instead add a note (once, and at the end) that some parts of the cycle are in other files, rather than silently skipping?

tests/integration/test_type_alias.py
```diff
@@ -0,0 +1,265 @@
...
+def test_implicit_generic_alias(run_int_fn):
+    """When params is omitted, free vars are collected from body in appearance order."""
+    T = guppy.type_var("T")
+
+    @guppy.struct
+    class Box(Generic[T]):
+        value: T
+
+    # No params= → T is a free var, collected automatically
+    BoxAlias = guppy.type_alias("BoxAlias", "Box[T]")
```
maximilianruesch, line 126: Is there some scenario in which the alias can be defined first, without explicitly passing parameters? Does it still resolve? Perhaps add this as a test too?

tests/integration/test_type_alias.py
maximilianruesch, line 1: How about a test where you have an alias of a generic struct (expanding the generic value into a field), and either pass it another unrelated alias, or pass it to itself, so something like
```python
MyAlias = guppy.type_alias("MyAlias", "MyStruct[T]")
OtherAlias = guppy.type_alias("OtherAlias", "int")
# ... then somewhere annotate as `MyAlias[OtherAlias]`?
# ... or somewhere annotate as `MyAlias[MyAlias]`?
```

guppylang/src/guppylang/decorator.py
```diff
@@ -365,6 +375,50 @@ def const_var(self, name: str, ty: str) -> TypeVar:
...
+
+        type_ast = _parse_expr_string(
+            ty, f"Not a valid Guppy type: `{ty}`", DEF_STORE.sources
+        )
+        explicit_params: Sequence[ParamDef] | None = (
```
```suggestion
        explicit_params = (
```
Thou shall not need thy type annotation, far as we knew! (Allows you to collapse this to one line hopefully).

maximilianruesch (changes requested):
🚀 

I have left one comment unresolved where the reviewer and reviewee do not seem to quite do the expected / the same, but I think that case is okay.

tests/integration/test_type_alias.py
ss2165: not sure I follow this, can you give an example

tests/error/test_alias_errors.py
ss2165: yes, test added

tests/integration/test_type_alias.py
maximilianruesch: For example replacing this excerpt with
```python
# No params= → T is a free var, collected automatically
BoxAlias = guppy.type_alias("BoxAlias", "Box[T]")

T = guppy.type_var("T")

@guppy.struct
class Box(Generic[T]):
    value: T
```
simply moving the alias defn to be at the top.

ss2165:
Thanks for pointing me to #1809, I have chosen to extend that directly for alias handling which saves a lot of brittle code for aliases. Have raised a follow up issue for adding richer cycle diagnostics for all three definition types: https://github.com/Quantinuum/guppylang/issues/1908

tests/integration/test_type_alias.py
ss2165: yes that works, test added

tests/error/alias_errors/mutual_recursive.err
ss2165: I think this test has limited value now that the rich messages are gone

guppylang-internals/src/guppylang_internals/definition/util.py
```diff
@@ -178,17 +186,18 @@ def _type_parsing_ctx(defn: ParsedRecursiveTypeDef) -> TypeParsingCtx:
     """
     from guppylang_internals.engine import DEF_STORE
 
-    param_var_mapping = {p.name: p for p in defn.params}
+    param_var_mapping = {p.name: p for p in (defn.params or [])}
```
maximilianruesch:
```suggestion
    param_var_mapping = {p.name: p for p in defn.params)} if defn.params else set()
```
As an alternative, I find that to be slightly clearer

tests/error/alias_errors/mutual_recursive.err
maximilianruesch: Sad but true

tests/integration/test_type_alias.py
maximilianruesch: Did I miss where you added the second test? May add some value, but I can also be convinced it is not needed.

maximilianruesch (approved):
🚀

tests/integration/test_type_alias.py
ss2165: i thought the new test with alias into enum field covered it, you think a struct one is needed on top?

PR #1658, opened 2026-04-13
chore(lint): mark hugr as third party to stabilise import sorting

ss2165:
Locally `hugr` was being picked up as a local package for me (for reasons unknown) so just mark it as known third party to avoid future mess.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -82,6 +82,9 @@ ignore = [
 "tests/{hugr,integration}/*" = ["B", "FBT", "SIM", "I"]
 "__init__.py" = ["F401"]                                              # module imported but unused
 
+[lint.isort]
+known-third-party = ["hugr"]
+
 [per-file-target-version]
 "tests/*/*_py312.py" = "py312"
 


guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
```diff
@@ -25,6 +23,8 @@
 from guppylang_internals.compiler.stmt_compiler import StmtCompiler
 from guppylang_internals.std._internal.compiler.tket_bool import OpaqueBool, read_bool
 from guppylang_internals.tys.ty import type_to_row
+from hugr import Wire, ops
+from hugr import tys as ht
```
croyzor: what could the rule for this possibly be??

ss2165: see updated description ☝️

croyzor: 🙇

croyzor (approved):

PR #1659, opened 2026-04-13
feat: Add compiler warning framework with opt-in rich warning rendering

### LINKED ISSUE #1654: [Tracking]:[Feature]: Add a general compiler warning framework
## Problem Statement

Guppylang currently lacks a general mechanism for compiler warnings as first-class, non-fatal diagnostics. This makes it harder to surface useful feedback to users in cases where compilation should continue, but the compiler should still report something actionable.

This issue should cover the shared warning framework only. Specific warning-producing use cases should be tracked separately.

## Proposed Solution

Add a general warning framework to the compiler for reporting non-fatal diagnostics during checking and compilation using the Python warnings framework.

This issue should cover shared machinery such as:
- representation and reporting of warnings;
- where warnings are emitted in the pipeline;
- user-facing behavior such as formatting and configurability;
- test support for warnings.

## Component

- Guppy Compiler

## Effort Estimate

Medium - Moderate complexity

## Additional Context

Relevant existing infrastructure:
- `DiagnosticLevel.WARNING` already exists in `guppylang-internals/src/guppylang_internals/diagnostic.py`.
- The current reporting path in `guppylang-internals/src/guppylang_internals/error.py` is centered around exception-based error reporting.

Related child / follow-up issues:
- #739 Emit warning for unreachable code
- #1643 Remove deprecated decorators, functions from v0
- #1019 Remove deprecated standard library reexports

ss2165:
Closes #1654

This adds the base compiler warning framework for guppylang. Warnings are now first-class non-fatal diagnostics that are buffered across a top-level check() or compile() operation, deduplicated within that operation, and emitted through Python’s warnings system as guppylang.GuppyWarning.

The change also adds an opt-in rich warning mode via with guppylang.rich_warnings(): .... In that mode, Guppy still emits the normal Python warning for filtering/tooling compatibility, and additionally renders the full warning diagnostic with source context to stderr.

What’s included:

- warning session plumbing in the diagnostics/error layer;
- public guppylang.GuppyWarning;
- public guppylang.rich_warnings() context manager;
- single-flush behaviour across nested compiler entrypoints and GuppyLibrary multi-call flows;
- framework, integration, renderer, and notebook coverage for the new warning surfaces.

Unreachable-code warning production in follow up: #1660 

Developed iteratively with Codex, see individual commits for steps.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/warning.py
+++ b/guppylang-internals/src/guppylang_internals/warning.py
@@ -16,12 +16,16 @@ class GuppyWarning(UserWarning):
     """Warning category for non-fatal compiler diagnostics."""
 
 
-class WarningKey(NamedTuple):
+class _WarningKey(NamedTuple):
     """Stable identity for deduplicating warnings within one operation."""
 
+    # File path passed through to Python's warning machinery, if available.
     filename: str | None
+    # 1-based source line passed through to Python's warning machinery, if available.
     lineno: int | None
+    # 0-based source column used only for deduplicating distinct warnings on one line.
     column: int | None
+    # Concise warning text emitted through Python's warning machinery.
     message: str
 
 
@@ -29,11 +33,30 @@ class WarningKey(NamedTuple):
 class PendingWarning:
     """Buffered warning waiting to be emitted at the end of a top-level operation."""
 
+    # Original structured diagnostic used for rich rendering.
     diagnostic: "Diagnostic"
-    message: str
-    filename: str | None
-    lineno: int | None
-    key: WarningKey
+    # Stable warning identity and Python-warning payload.
+    _key: _WarningKey
+
+    @property
+    def message(self) -> str:
+        """Concise warning text emitted through Python's warning machinery."""
+        return self._key.message
+
+    @property
+    def filename(self) -> str | None:
+        """Source file reported to Python's warning machinery, if available."""
+        return self._key.filename
+
+    @property
+    def lineno(self) -> int | None:
+        """1-based source line reported to Python's warning machinery, if available."""
+        return self._key.lineno
+
+    @property
+    def column(self) -> int | None:
+        """0-based source column used for deduplication within one operation."""
+        return self._key.column
 
 
 @dataclass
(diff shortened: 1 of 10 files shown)


guppylang-internals/src/guppylang_internals/warning.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+@dataclass(frozen=True)
+class PendingWarning:
+    """Buffered warning waiting to be emitted at the end of a top-level operation."""
+
+    diagnostic: "Diagnostic"
+    message: str
+    filename: str | None
+    lineno: int | None
```
tatiana-s: Given `message`, `filename` and `lineno` are also stored in the `WarningKey`, can they just be retrieved from there when needed instead of storing the information twice?

guppylang-internals/src/guppylang_internals/warning.py
```diff
@@ -0,0 +1,221 @@
...
+
+class _WarningKey(NamedTuple):
+    """Stable identity for deduplicating warnings within one operation."""
+
+    # File path passed through to Python's warning machinery, if available.
+    filename: str | None
+    # 1-based source line passed through to Python's warning machinery, if available.
+    lineno: int | None
+    # 0-based source column used only for deduplicating distinct warnings on one line.
+    column: int | None
```
mark-koch, line 27: Consider using the existing `Span` class instead

guppylang-internals/src/guppylang_internals/warning.py
```diff
@@ -0,0 +1,221 @@
...
+        return self._key.column
+
+
+@dataclass
+class DiagnosticSession:
+    """Per-operation diagnostic state shared across nested compiler calls."""
+
+    rich_warnings: bool = False
+    pending_warnings: list[PendingWarning] = field(default_factory=list)
+    seen_warnings: set[_WarningKey] = field(default_factory=set)
```
mark-koch, line 68: It would be very cool if we could use `set[Warning]` here, it already includes the span and the message

This would require hashability, but maybe we can just force the new Warning subclass to be hashable instead of all Diagnostics?

PR #1661, opened 2026-04-13
chore: Add env var for enabling debug info

cgh-qtnm:
When testing debug info generation in hugr-llvm, it is useful to be able to take an arbitrary Guppy source file and compile it with debug info without modifying the file. This is somewhat difficult because typical Guppy programs drive compilation from their own source code.

I have tried to figure something out with importlib and/or runpy, but I haven't found an approach that works for all the Guppy programs I want to compile. I don't love config via env var but I think this is the cleanest way to do it.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/debug_mode.py
+++ b/guppylang-internals/src/guppylang_internals/debug_mode.py
@@ -1,7 +1,9 @@
 """Global state for determining whether to attach debug information to Hugr nodes
 during compilation."""
 
-_DEBUG_MODE_ENABLED = False
+import os
+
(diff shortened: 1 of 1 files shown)


mark-koch (approved):
Seems reasonable to me, but we need to remember to also put this into the docs somewhere (cc @CalMacCQ)

tatiana-s:
Debug info in general will have to be documented, but probably only once there is a way for users to use it and we have things like https://github.com/Quantinuum/guppylang/issues/1607

tatiana-s (approved):

PR #1666, opened 2026-04-16
fix: Allow check on generic functions that do not require monomorphisation

mark-koch:


guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -576,6 +598,29 @@ def params_str(self) -> str:
...
 
+@dataclass(frozen=True)
+class EntryCheckMonomorphizeError(Error):
+    title: ClassVar[str] = "Invalid check point"
+    span_label: ClassVar[str] = (
+        "{thing} can only be checked if the value{plural_s} of its generic "
+        "parameter{plural_s} {params_str} {is_are} known"
+    )
+    thing: str
+    params: Sequence[Parameter]
```
maximilianruesch: Can we just not store `CheckableGenericDef` here and to the capitalisation and parameter access as an implementation detail of this error? That would have the benefit of reducing the size of the `check` function above, which is already fairly complicated to wrap your head around if you are not used to it.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -405,12 +405,34 @@ def check(self, def_ids: list[DefId], *, reset: bool = True) -> None:
...
+                # checking without having the monomorphization available. In that case,
+                # we give up and prompt the user to specify the generic arguments.
+                assert isinstance(entry_defn, CheckableGenericDef)
+                description = (
+                    f"{entry_defn.description.capitalize()} `{entry_defn.name}`"
+                )
+                err = EntryCheckMonomorphizeError(
+                    entry_defn.defined_at, description, entry_defn.params
+                )
+                raise GuppyError(err) from None
```
maximilianruesch: We are raising from `None`, while we could be raising from the caught error. This should not impact the rendering, but raising from the previous error may provide help for people catching `GuppyError`.

tests/integration/test_comptime_args.py
maximilianruesch, line 1: It took me a while to realise that you are checking the generic functions, and not the main function. Perhaps you can move the check to "just below the generic function definition" to better visualise that you are just checking that unit. Or you can also decide that you want all the assertions grouped at the end (following arrange-act-assert patterns) and keep it like it is.

Either way, it is probably a good idea to add a comment like `# Smoke test` to each call to signal that this is indeed not a random statement.

maximilianruesch (changes requested):

maximilianruesch (approved):
🚀

PR #1667, opened 2026-04-16
feat: Add array helpers for qubit measurements for qsystem measurement variants

dunc-go:
Hi all, 

This PR adds measure_and_reset_array, an array helper built on top of the existing qsystem measure_and_reset operation. It mirrors measure_array, but preserves the reset semantics for each qubit.

A test prepares a non-symmetric bitstring in a register, verifies the measured bits match the pattern, and then checks that the same register reads back as all zeros afterwards.

I’ve placed this in std.qsystem to keep the qsystem-specific measurement/reset operations together, but I’d like feedback on whether this should instead live in std.quantum.

### DIFF
--- a/guppylang/src/guppylang/std/qsystem/__init__.py
+++ b/guppylang/src/guppylang/std/qsystem/__init__.py
@@ -216,6 +216,13 @@ N = guppy.nat_var("N")
 
 @guppy
 @no_type_check
+def measure_and_reset_array(qubits: array[qubit, N]) -> array[bool, N]:
+    """Measure and reset an array of qubits, returning an array of bools."""
+    return array(measure_and_reset(qubits[i]) for i in range(N))
+
+
+@guppy
+@no_type_check
 def lazy_measure_array(qubits: array[qubit, N] @ owned) -> array["Measurement", N]:
     """Request a destructive lazy measurement of an array of qubits, returning an array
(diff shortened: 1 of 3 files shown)


guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -214,6 +214,13 @@ def lazy_measure_and_reset(q: qubit) -> "Measurement":
 N = guppy.nat_var("N")
 
 
+@guppy
+@no_type_check
+def measure_and_reset_array(qubits: array[qubit, N]) -> array[bool, N]:
+    """Measure and reset an array of qubits, returning an array of bools."""
+    return array(measure_and_reset(qubits[i]) for i in range(N))
```
maximilianruesch, line 228: While at it, let us add `measure_array` and `lazy_measure_and_reset_array` as well (above and below respectively). There should be enough information in this file to find appropriate definitions and doc-strings.

maximilianruesch (approved):
This is the correct place to put it, since the operations are lowered to the `tket.qsystem` extension, whereas the functions in `std.quantum` are lowered to `tket.quantum`.

Just one comment to add some missing functions (also add tests please). Optional, since you originally only meant to add one. Still feel like its valid to add those in the same swoop though.

maximilianruesch (approved):

PR #1672, opened 2026-04-17
fix(internals): misleading error for unknown generic bounds

### LINKED ISSUE #1653: [Bug]: Misleading error for function type args
### Description

If I write the following (note: this is a valid program apart from the `Bogus` type bound which doesn't exist)
```python
from guppylang import guppy
from guppylang.std.lang import Copy

@guppy
def foo[T: (Copy, Bogus)](t: T) -> T:
    return t

@guppy
def main() -> None:
    foo(42)

main.compile()
```
the error message is
```
Error: Variable not defined (at /Users/craig.roy/work/guppy/jinimal.py:5:12)
  | 
3 | 
4 | @guppy
5 | def foo[T: (Copy, Bogus)](t: T) -> T:
  |             ^^^^ `Copy` is not defined

```
which points to the wrong type bound!

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

ss2165:
## Summary
- report unknown names in generic type bounds at the correct bound span
- add a regression snapshot for an unresolved name in a (Copy, Bogus) type parameter bound

## Root cause
- tuple bounds mixing known modifiers and unknown names fell through the generic const-bound parsing path, which attributed the error to the wrong AST span

Fixed by Codex, cleaned up by me

Closes #1653

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/parsing.py
+++ b/guppylang-internals/src/guppylang_internals/tys/parsing.py
@@ -302,6 +302,45 @@ def check_function_arg(
 
 if sys.version_info >= (3, 12):
 
+    def _parse_type_param_bound_tuple(
+        node: ast.TypeVar, idx: int, bound: ast.Tuple
+    ) -> TypeParam | None:
+        """Parses a `Copy`/`Drop` tuple used as a type parameter bound."""
+        from guppylang_internals.checker.cfg_checker import VarNotDefinedError
+
+        must_be_copyable = False
+        must_be_droppable = False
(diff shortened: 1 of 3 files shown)


croyzor:
My inclination with this is to just leave it for now and fix it in the protocols branch. This is less likely to come up before we have bounds other then `(Copy, Drop)`, and all of this code is going to have to be rewritten to handle protocol bounds anyway

PR #1677, opened 2026-04-20
refactor: Redundant subst case

croyzor:
A just review comment from #1602 that wasn't merged to `main` because I pushed it to the wrong branch


maximilianruesch:
Note: This would currently show up in changelog, which is probably unwanted

mark-koch (approved):

PR #1687, opened 2026-04-21
Security: Global experimental-feature flag is process-wide and not thread-safe

tomaioo:
## Summary

Security: Global experimental-feature flag is process-wide and not thread-safe

## Problem

**Severity**: `Medium` | **File**: `guppylang-internals/src/guppylang_internals/experimental.py:L11`

`EXPERIMENTAL_FEATURES_ENABLED` is a mutable module-level global that is toggled by `enable_experimental_features` / `disable_experimental_features`. In concurrent or multi-tenant execution, one request/thread can temporarily enable experimental behavior for another, potentially bypassing feature-gating expectations and causing policy isolation issues.

## Solution

Replace the process-global boolean with a context-local mechanism (e.g., `contextvars.ContextVar`) so feature flags are scoped to the current execution context. Also provide explicit APIs to query/set state per compilation session object rather than globally.

## Changes

- `guppylang-internals/src/guppylang_internals/experimental.py` (modified)

### DIFF
--- a/guppylang-internals/src/guppylang_internals/experimental.py
+++ b/guppylang-internals/src/guppylang_internals/experimental.py
@@ -1,4 +1,5 @@
 from ast import expr
+from contextvars import ContextVar
 from dataclasses import dataclass
 from types import TracebackType
 from typing import ClassVar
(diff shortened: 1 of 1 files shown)


nicolaassolini-qntm:
Thank you for the pull request. We’re not currently focusing on thread-safety in the compilation stage, so we’re avoiding changes that add complexity to the compiler, especially without tests.
The ability to set a global variable from a thread is also a fairly limited attack vector, so the security impact here isn’t something we’re prioritising.

For these reasons, we’ll be closing this PR. Thanks for your understanding.

PR #1690, opened 2026-04-24
chore: use codspeed for one-shot continuous benchmarking

### LINKED ISSUE #1599: [Maintenance]: Wall-clock time benchmarks are too unstable for CI
### Description

We're currently benchmarking two metrics in CI:

- Hugr artifact binary sizes
- Benchmark test runtimes

The first one is stable across runs, but doing timing benchmarks in CI leads to wild variations.
This causes PRs to be [spammed with benchmark warnings](https://github.com/Quantinuum/guppylang/pull/1580#issuecomment-4099810740) for untouched codepaths.

The best alternative, that we currently use in hugr and tket, is to do one-shot benchmarking with CPU instrumentation to count instructions rather than time.
This still has some variability due to cache and whatnot, but is _much_ more stable than timing things in a github runner.

I propose keeping the current bencher configuration for non-timing metrics, but [swapping to codspeed](https://codspeed.io/docs/benchmarks/python) for [one-shot benchmarking](https://codspeed.io/docs/instruments/cpu). Codspeed already has a github bot that reports regressions in PRs, so the change should be quick and simple.

### Motivation

_No response_

### Scope

_No response_

### Component

Testing Infrastructure, CI/Build System

### Priority

Low - Nice to have

### Additional Context

_No response_

ferbetanzo:
Closes #1599. Note that we are now using `--benchmark-disable` so benchmark tests are run when doing `just test` or `uv run pytest ...`, but no metrics are saved/uploaded. This means the only place where we can run the benchmarks is in the CI.

I have also enabled tracking of hugr bytes and nodes to the `test_queue_push_pop` benchmarks.


pyproject.toml
```diff
@@ -55,11 +54,10 @@ exclude_also = [
   "@abstractmethod",
 ]
 
-
 [tool.pytest.ini_options]
-addopts = "--benchmark-skip -m 'not test_exported_hugrs'" # benchmarks run explicitly with `just bench`
+addopts = "--benchmark-disable -m 'not test_exported_hugrs'"  # benchmarks run explicitly with `just bench`
```
ferbetanzo, line 58: Using `--benchmark-disable` instead of `--benchmark-skip`. The tests are only executed once and no metrics are collected.

justfile
```diff
@@ -56,27 +56,18 @@ clean-docs:
...
 bench_upload *BENCHER_FLAGS:
-    uv run pytest --benchmark-only --benchmark-json="{{NOW}}-pytest-benchmark.json"
-    bencher run \
-            --adapter python_pytest \
-            --file "{{NOW}}-pytest-benchmark.json" \
-            --project {{BENCHER_PROJECT}} \
-            --quiet \
-            {{BENCHER_FLAGS}}
+    uv run pytest \
+           --override-ini addopts="" \
```
ferbetanzo, line 66: This is needed because we are using `--benchmark-disable` in our `[tool.pytest.ini-options]`.

DEVELOPMENT.md
```diff
@@ -58,35 +58,25 @@ just export-integration-tests
 
 Run `just` to see all available commands.
 
-## 🐰 Running the benchmarks
+## 🐰 Continuous Benchmarking
 
-To run the benchmarks using `pytest-benchmark`:
+The benchmarks are disabled (not skipped) in local test runs (using `--benchmark-disable`), which means that
+the corresponding benchmark tests are executed only once and no metrics are collected when you
```
nicolaassolini-qntm: > are executed only once

Maybe clarify this a bit?

DEVELOPMENT.md
```diff
@@ -58,35 +58,25 @@ just export-integration-tests
...
-[guppylang-benchmarks](https://bencher.dev/console/projects/guppylang-benchmarks)
-bencher project with:
+### bencher.dev benchmarks
 
-```bash
-just bench_upload
-```
+We also use [bencher.dev](https://bencher.dev/) in the CI for tracking custom metrics like number of nodes or
+size in bytes of the hugrs generated by our benchmarks. This is also done for every PR
+and main.
```
nicolaassolini-qntm:
```suggestion
We also use [bencher.dev](https://bencher.dev/) in CI to track custom metrics, such as the number of nodes and the byte size of the HUGRs generated by our benchmarks. These metrics are recorded for every pull request as well as for the main branch.
```

nicolaassolini-qntm (approved):
just 2 small suggestions

ferbetanzo:
@nicolaassolini-qntm I have improved the DEVELOPMENT.md as per your comments.

I am also now skipping the benchmarks for building executables as those were taking too long. With the current set up, the benchmarks take around 15 mins to run in the CI, where almost all of the time is dominated by the codspeed single-shot simulations. If that's too long we can revisit/discuss and disable other benchmarks.

PR #1691, opened 2026-04-24
fix!: Metadata Generation with Nested Modifiers

nicolaassolini-qntm:
Previously, when multiple modifiers were nested, inner modifiers did not inherit metadata from the outer ones. This caused a bug during pass resolution, which has now been fixed.

* The `ModifiedBlock` class now has an `accumulated_flags` attribute, which tracks all modifier flags applied to the block, including those inherited from nesting. This replaces the old `flags()` method. [[1]](diffhunk://#diff-6cd5c729fdf52a3e6a9dd9a468582fc9f49f7a68995845093f88d3e89d499dabR832-R847) [[2]](diffhunk://#diff-6cd5c729fdf52a3e6a9dd9a468582fc9f49f7a68995845093f88d3e89d499dabL877-L886)
* The control flow graph (CFG) builder and modifier checker are updated to use and propagate `accumulated_flags`, ensuring that metadata and type signatures reflect all active modifiers, even in nested contexts. [[1]](diffhunk://#diff-eb6972436fbc83bbc8d18700eab1d1d066aa62a6b32dc4487b5912d431c5f97fL363-R367) [[2]](diffhunk://#diff-dc0b997af6db2c778960b29570cd008671b1f161dde7f3159f023a5796185086L105-R107)
* Added  (`test_modifier_metadata.py`)  to verify that modifier blocks have the correct `unitary` metadata attached.

BREAKING CHANGE: deleted `guppylang_internals/nodes.py::def flags(self)` replaced by `accumulated_flags`

### DIFF
--- a/guppylang-internals/src/guppylang_internals/cfg/builder.py
+++ b/guppylang-internals/src/guppylang_internals/cfg/builder.py
@@ -360,10 +360,11 @@ class CFGBuilder(AstVisitor[BB | None]):
         accumulated_flags = self.cfg.unitary_flags | modifiers.flags()
         cfg = CFGBuilder().build(node.body, True, self.globals, accumulated_flags)
         new_node = ModifiedBlock(
-            cfg=cfg,
-            modifiers=modifiers,
+            cfg,
+            modifiers,
             # we save the first modifier node for a better error rendering
-            first_modifier_node=node.items[0].context_expr,
+            node.items[0].context_expr,
+            accumulated_flags,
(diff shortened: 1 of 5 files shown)


qartik:
Will review one copilot comments are resolved. Most of them seem reasonable.

nicolaassolini-qntm:
> Will review one copilot comments are resolved. Most of them seem reasonable.

@qartik I solved the reasonable Copilot suggestions

qartik (approved):

PR #1699, opened 2026-04-28
feat: Support static methods on types

### LINKED ISSUE #694: Support static methods for types

GBoydQ:
Adds staticmethods to guppy, supporting structs, enums.

Functions are checked for staticness during parsing, and `ParsedDefinition`s now have an `is_static` flag.

`ExprSynthesizer.visit_Attribute` contains logic for calling staticmethods on instantiated types (we need to not partially apply self) and `Name`s both for globals and generic parameters.

Comptime staticmethods are not yet supported, and work still needs to be done to support `Self` output for staticmethods generally.

Closes #694


guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -109,6 +102,12 @@
 MonoDefId = tuple[DefId, Inst]
 
 
+@dataclass(frozen=True)
+class ImplDefinition:
+    id: DefId
+    is_static: bool
```
maximilianruesch: For this it is useful to inherit from `NamedTuple` (and remove the decorator), since you do not need the full blown features of a dataclass. It gives you lower runtime overhead and virtually the same attribute access.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -109,6 +102,12 @@
 MonoDefId = tuple[DefId, Inst]
 
 
+@dataclass(frozen=True)
+class ImplDefinition:
```
maximilianruesch: Let us call this `TypeMember` or `TypeMemberDef`. `Impl...` was the previous wording in this file, which was confusing.

guppylang/src/guppylang/decorator.py
```diff
@@ -254,8 +246,14 @@ def decorator(
             frame = get_calling_frame()
             DEF_STORE.register_def(defn, frame)
             for val in cls.__dict__.values():
-                if isinstance(val, GuppyDefinition):
-                    DEF_STORE.register_type_member(defn.id, val.wrapped.name, val.id)
+                if isinstance(val, GuppyDefinition) and hasattr(
+                    val.wrapped, "python_func"
+                ):
```
maximilianruesch: Two things:
Instead of silently failing to register if the wrapped definition has no python_func, let us fail in case it is a GuppyDefinition but we cannot determine whether the function is static or not. If the user can run into this case, it should be a proper error with a diagnostic (and an error test), if not, it can be a InternalGuppyError. This will help us debug / discover new cases where we introduce members that are GuppyDefinition, but have no function attached (for whatever reason).
We discourage hasattr checks, since they are dependent on "implementation-ish" details (characterising a class by which attribute is present), and are not hardened against refactoring changes. Furthermore, from here it is completely invisible which classes may have python_func. Instead, you should use isinstance checks to check all classes that you want to cover explicitly.

guppylang/src/guppylang/decorator.py
```diff
@@ -298,8 +296,13 @@ def decorator(
...
```
maximilianruesch: See https://github.com/Quantinuum/guppylang/pull/1699/changes#r3159470912

guppylang/src/guppylang/defs.py
```diff
@@ -107,13 +104,14 @@ def __getattr__(self, name: str) -> Any:
         defn = ENGINE.get_checked(self.wrapped.id, mono_args=())
         assert isinstance(defn, CheckedEnumDef)
         if (
-            # We can only access the variants of the enum from the enum class,
-            # not methods
-            name in defn.variants
-            and defn.id in DEF_STORE.type_members
+            # We can only access variants of the enum from the enum class
+            # and staticmethods
```
maximilianruesch: For one, this comment should be on the line with the `name in variants OR is a static member` clause, and for another it is worded confusingly: Right now it sounds like we can access variants from [either the enum class or static methods], which does not make sense.

Instead, let us say something like
```python
# We can only access enum variants or static methods from the enum class
```

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -135,9 +134,11 @@ def register_def(self, defn: RawDef, frame: FrameType) -> None:
         self.raw_defs[defn.id] = defn
         self.frames[defn.id] = frame
 
-    def register_type_member(self, ty_id: DefId, name: str, member_id: DefId) -> None:
+    def register_type_member(
+        self, ty_id: DefId, name: str, member_id: DefId, is_static: bool = False
```
maximilianruesch:
```suggestion
        self, ty_id: DefId, name: str, member_id: DefId, *, is_static: bool = False
```
Make boolean parameters keyword-only, this gives them more semantics on the caller side.

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -567,9 +559,9 @@ def __call__(self, *args: Any) -> Any:
             # Definition is non-generic, so we can use `mono_args=()` here
             defn = ENGINE.get_checked(self.wrapped.id, mono_args=())
             if isinstance(defn, TypeDef) and (
-                constructor_id := DEF_STORE.type_members[defn.id].get("__new__")
+                constructor_impl := DEF_STORE.type_members[defn.id].get("__new__")
```
maximilianruesch: Let us call this `constructor_member_def` (or `constructor_def` if that is too long).

Or alternatively use the below approach of
```python
if (
    defn.id in DEF_STORE.type_members
    and "__new__" in DEF_STORE.type_members[defn.id]
):
    constructor = DEF_STORE.raw_defs[
        DEF_STORE.type_members[defn.id]["__new__"].id
    ]
    return TracingDefMixin(constructor).to_guppy_object(
```

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -157,8 +157,11 @@ def extend_type(defn: TypeDef, return_class: bool = False) -> Callable[[type], t
 
     def dec(c: type) -> type:
         for val in c.__dict__.values():
-            if isinstance(val, GuppyDefinition):
-                DEF_STORE.register_type_member(defn.id, val.wrapped.name, val.id)
+            if isinstance(val, GuppyDefinition) and hasattr(val.wrapped, "python_func"):
+                is_static = isinstance(val.wrapped.python_func, staticmethod)
```
maximilianruesch: Same same for these checks (three in this file). At this point you probably want a function `determine_static` that takes the wrapped definition and can raise appropriate errors. This gives you a bit more consistency.

uv.lock
maximilianruesch, line 1: We should really have a CI check for uv.lock synchronisation.

tests/integration/test_staticmethod.py
```diff
@@ -0,0 +1,100 @@
...
+    class Test(Generic[T]):
+        @guppy
+        @staticmethod
+        def default(a: T) -> "Test[T]":
+            return Test()
+
+    @guppy
+    def main() -> None:
+        t = Test.default(1.0)
+        # can call Test.default[int] from Test[float] instance
```
maximilianruesch, line 23: A bit counterintuitive, but reasonable in this case. Are there cases where this is not allowed?

tests/integration/test_staticmethod.py
maximilianruesch, line 1: Can we test static `guppy.declare`s? Perhaps with library linking?

On that note, you should check whether link name inference still works with static methods (in the `link_name` tests).

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -324,19 +314,20 @@ def compile_call(
...
     return CallReturnWires(
         regular_returns=list(call[:num_returns]),
         inout_returns=list(call[num_returns:]),
     )
 
 
 def parse_py_func(f: PyFunc, sources: SourceMap) -> tuple[ast.FunctionDef, str | None]:
+    # get the wrapped function if a staticmethod
+    if isinstance(f, staticmethod):
+        f = f.__func__
```
maximilianruesch: The fact that the saved `self.python_func` is wrapped may be breaking as well (depending on if the wrapper copies or forwards key attributes such as `__module__`.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -580,7 +580,23 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                    ty_def = ENGINE.parsed[defn.id]
+                    if (
+                        node.attr in DEF_STORE.type_members[ty_def.id]
+                        and isinstance(ty_def, TypeDef)
+                        and (func := ENGINE.get_instance_func(ty_def, node.attr))
+                        and DEF_STORE.type_members[ty_def.id][node.attr].is_static
+                    ):
+                        return with_loc(
+                            node, GlobalName(id=node.attr, def_id=func.id)
+                        ), func.ty
```
maximilianruesch: Can this fail when the thing that we are accessing on is not in globals? Can this even happen? I am confused what this does, and what all the if clauses mean. Perhaps @mark-koch has thoughts on this.

maximilianruesch (changes requested):
Thank you for the PR, this looks really interesting. I have a bunch of suggestions. I will also request @mark-koch to provide his comments on this PR!

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -580,7 +580,23 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
             # Name can be a EnumDef only if it is in a attribute access, thus we
             # manually need to call the helper instead of relying on the standard
             # visit_Name (that is called through synthesize)
+
+            # visit node for case of staticmethods on a non-instantiated type
             ty = get_type_opt(node.value)
+            if node.value.id in self.ctx.globals:
+                defn = cast("ParsedDef", self.ctx.globals[node.value.id])
+                if not isinstance(defn, PythonObject):
```
mark-koch:
```suggestion
                defn_or_python_object = self.ctx.globals[node.value.id]
                if isinstance(defn_or_python_object, TypeDef):
```

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -580,7 +580,23 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                    ty_def = ENGINE.parsed[defn.id]
+                    if (
+                        node.attr in DEF_STORE.type_members[ty_def.id]
+                        and isinstance(ty_def, TypeDef)
+                        and (func := ENGINE.get_instance_func(ty_def, node.attr))
+                        and DEF_STORE.type_members[ty_def.id][node.attr].is_static
```
mark-koch: Do we actually need this static check? Imo `type.func` would also be fine if `func` is not static, similar to Python:

```python
x = int.__add__(1, 2)
```

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -580,7 +580,23 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
             # Name can be a EnumDef only if it is in a attribute access, thus we
             # manually need to call the helper instead of relying on the standard
             # visit_Name (that is called through synthesize)
+
+            # visit node for case of staticmethods on a non-instantiated type
             ty = get_type_opt(node.value)
+            if node.value.id in self.ctx.globals:
+                defn = cast("ParsedDef", self.ctx.globals[node.value.id])
+                if not isinstance(defn, PythonObject):
+                    ty_def = ENGINE.parsed[defn.id]
```
mark-koch: It should already be `ParsedDef`?

guppylang/src/guppylang/decorator.py
GBoydQ: Yes the hasattr is not ideal, the problem I was facing here was that I think I originally had `isinstance(..., GuppyFunctionDefinition)` but mypy was not able to determine that the wrapped `Definition` actually contained a `python_func` in that case. Is there a way to ensure a `GuppyFunctionDefinition` contains a function?

tests/integration/test_staticmethod.py
GBoydQ: This does seem odd but I think it makes sense, as I would expect the behaviour of staticmethods to be to ignore the instance.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
GBoydQ: This could definitely be cleaner / wrapped up into a function.
Also the cast / isinstances were also a result of my fight with mypy due to my lack of experience with the compiler types.
Essentially I was going for, "if this is a type that has has a type_member of this name and it's static, then return the staticmethod"

guppylang-internals/src/guppylang_internals/definition/function.py
GBoydQ: @mark-koch and I briefly discussed this but I would appreciate any more thoughts / discussion to make sure this is ok

guppylang/src/guppylang/decorator.py
maximilianruesch: `GuppyFunctionDefinition` is a subclass of `GuppyDefinition`, and is not the wrapped member. What you are searching for is `RawFunctionDef` and other similar classes.

tests/integration/test_staticmethod.py
maximilianruesch: We should extend the comment by saying "as static methods ignore instance types".

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: Currently there is no guarantees made by the `python_func` about its type (because we do not have a proper type for it). However we do access things like `__module__` on it, which is defined with regular functions. I guess if you add the link_name tests (and any other paths that may trigger attribute accesses) we should be fine.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
maximilianruesch: I am confused in particular about the `if not isinstance(defn, PythonObject):`. What does it do?

Also, I guess `ENGINE.get_instance_func(...)` is not necessarily the right name anymore.

Let us wait for a comment from Mark on this one.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -580,7 +580,23 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                    if (
+                        node.attr in DEF_STORE.type_members[ty_def.id]
+                        and isinstance(ty_def, TypeDef)
```
mark-koch: `isinstance(ty_def, TypeDef)` could be removed if you go with the refactor above

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -642,16 +658,28 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
             )
-            # Make a closure by partially applying the `self` argument
-            # TODO: Try to infer some type args based on `self`
-            result_ty = FunctionType(func.ty.inputs[1:], func.ty.output, func.ty.params)
-            return with_loc(node, PartialApply(func=name, args=[node.value])), result_ty
+            ty_id = DEF_STORE.type_member_parents[func.id]
+
+            if (
+                impl_def := DEF_STORE.type_members[ty_id].get(node.attr)
+            ) and impl_def.is_static:
```
mark-koch: I think this would be clearer if we introduced a new `ENGINE.get_type_member` function that returns the `TypeMember`:

```python
if member := ENGINE.get_type_member(ty, node.attr):
    if member.is_static:
        ...
    else:
        ...
```

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -317,19 +317,22 @@ def check_signature(
...
 
     inputs = []
     ctx = TypeParsingCtx(globals, param_var_mapping, allow_free_vars=True)
     for i, inp in enumerate(func_def.args.args):
-        # Special handling for `self` arguments. Note that `__new__` is excluded here
-        # since it's not a method so doesn't take `self`.
-        if self_defn and i == 0 and func_def.name != "__new__":
+        # Special handling for `self` arguments. Note that `__new__` and staticmethods
+        # are excluded here since they do not take `self`.
+        if self_defn and i == 0 and func_def.name != "__new__" and not is_static:
```
mark-koch: Once this PR lands, we should turn `__new__` into a static method, then we won't need this hack anymore 🎉

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -158,12 +162,47 @@ def extend_type(defn: TypeDef, return_class: bool = False) -> Callable[[type], t
...
+    match defn.wrapped:
+        case RawFunctionDef() | RawCustomFunctionDef() | RawFunctionDecl():
+            return isinstance(defn.wrapped.python_func, staticmethod)
+        # comptime methods not yet supported
+        case RawTracedFunctionDef():
+            if isinstance(defn.wrapped.python_func, staticmethod):
+                # TODO guppy error handling
+                raise TypeError("static comptime func")
+            else:
+                return False
```
mark-koch: What's the issue with comptime functions, I'd assume everything should just work?

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -158,12 +162,47 @@ def extend_type(defn: TypeDef, return_class: bool = False) -> Callable[[type], t
...
+        case OverloadedFunctionDef():
+            # check all the methods in the overload are also static
+            for func_id in defn.wrapped.func_ids:
+                func_def = DEF_STORE.raw_defs[func_id]
+                if not isinstance(func_def, RawFunctionDef):
+                    return False
```
mark-koch: It could also be a `RawCustomFunctionDef`, `RawFunctionDecl`, `RawTracedFunctionDef`, or even another `OverloadedFunctionDef`

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -158,12 +162,47 @@ def extend_type(defn: TypeDef, return_class: bool = False) -> Callable[[type], t
...
+                if not isinstance(func_def.python_func, staticmethod):
+                    # TODO guppy error handling
+                    raise TypeError(
+                        "one of the functions in this overload is not static"
+                    )
```
mark-koch: I think a vanilla Python error is ok here since it raised while decorating, not during compilation.

However, the message could be improved a bit, for example mentioning which overload is static and which isn't. Also could you add an error test for this case?

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -158,12 +162,47 @@ def extend_type(defn: TypeDef, return_class: bool = False) -> Callable[[type], t
...
+            return isinstance(defn.wrapped.python_func, staticmethod)
+        # comptime methods not yet supported
+        case RawTracedFunctionDef():
+            if isinstance(defn.wrapped.python_func, staticmethod):
+                # TODO guppy error handling
+                raise TypeError("static comptime func")
+            else:
+                return False
+        case OverloadedFunctionDef():
+            # check all the methods in the overload are also static
```
mark-koch: I think the current implementation doesn't do that. You just check if any overload is static and then raise an error? I would expect something like

```python
is_static = [determine_static(overload_id) for overload_id in defn.wrapped.func_ids]
if all(is_static):
    return True
elif not any(is_static):
    return False:
else:
    raise TypeError("Some static, some not")
```

Note that this would refquire refactoring `determine_static` to take ids instead of `GuppyDefinition`s

guppylang-internals/src/guppylang_internals/definition/function.py
mark-koch: We could consider storing the wrapped function directly instead of unwrapping here. That might be cleaner?

guppylang-internals/src/guppylang_internals/decorator.py
mark-koch: Or, maybe we should just look for `@staticmethod` on the `@guppy.overloaded` function instead?

guppylang-internals/src/guppylang_internals/decorator.py
GBoydQ: I have not looked into this, but I get `'staticmethod' object has no attribute '__globals__'` in `mock_builtins`

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
GBoydQ: I was able to make most of the changes in this area but this one didn't work

tests/integration/test_staticmethod.py
GBoydQ: I have added a test for library linking and another one to the `link_name` tests.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
GBoydQ: I have introduced this although I still need to use `ENGINE.get_instance_func` or `ENGINE.get_parsed` and then cast / check the type is a `CallableDef`.
I could wrap up `get_instance_func` and `get_type_member` into one function that returns both if that would be helpful.

guppylang-internals/src/guppylang_internals/definition/function.py
GBoydQ: I am not 100% sure what you mean here.
Are you saying that on decoration we should instead store `.__func__` so that the stored `py_func` is always a function?
I can change this, although it would also require reworking `determine_static` as I'm not sure there would still be a way to check if a function is static without the decorator being stored in the Definition.

guppylang/src/guppylang/decorator.py
GBoydQ: Thank you this was very helpful. Logic for different FunctionDef types is now wrapped up in `determine_static`

GBoydQ:
Apologies for this becoming stale, I have reworked this for v1 and also made some small changes to support protocols.
Other than that, I had to move some things round due to restructuring of the codebase (in particular I could not find a good place for `determine_static` that did not cause circular imports).

I am happy to either move the `gb/add_classmethods` branch onto these new changes or to create a new PR, whichever is preferable.

The only piece of review feedback I have not addressed is "We could consider storing the wrapped function directly instead of unwrapping here" because I wanted some clarification there.

maximilianruesch:
@GBoydQ no worries!

> I am happy to either move the gb/add_classmethods branch onto these new changes or to create a new PR, whichever is preferable.

Surely you mean your current branch `gb/add_classfunctions`. But I would suggest you move your current branch to the new changes, if the current reviews would still apply. If they do not, and you would prefer a fresh review (that is unbiased by existing reviews) you may open a new PR. But it does not sound like that is the case here.

> The only piece of review feedback I have not addressed is "We could consider storing the wrapped function directly instead of unwrapping here" because I wanted some clarification there.

I will shortly comment with some clarifications. Edit: See https://github.com/Quantinuum/guppylang/pull/1699#discussion_r3658022838.

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: What we can probably do is let the `RawFunctionDef` take the wrapped function, and upon calling `parse`, the function is determined to be static or not (and this information will be available on the `ParsedFunctionDef`, perhaps via a new property, and unwrapped once (if appropriate) to be passed into `parse_py_func`.

@mark-koch is that roughly what you envisioned? I see no cleaner way to lift the `static` check without adding factory methods for `RawFunctionDef` and replacing all of its instantiations.

mark-koch: Hmm I think the best place to store whether the function was static or not would be inside `GuppyFunctionDefinition`. That is the object we have at hand when you try to register functions in the `DEF_STORE` which is the point at which you need to know if it's static or not

GBoydQ: ok fair enough, the reason I did not do this initially was that I didn't want method information to be stored in functions that are not methods, but in hindsight it would probably simplify things a bit.

maximilianruesch:
@GBoydQ I would not classify this as a breaking change. Adding a feature which was previously unsupported is not breaking, since previous user code did not compile anyways. Furthermore, users should not rely on features remaining to be unsupported.

GBoydQ:
> @GBoydQ I would not classify this as a breaking change. Adding a feature which was previously unsupported is not breaking, since previous user code did not compile anyways. Furthermore, users should not rely on features remaining to be unsupported.

I agree, however semver complained at me about this when I removed the previous breaking change by addressing your comments, I will look into it

maximilianruesch:
We recently adjusted the breaking changes check to not warn on breaks in the `guppylang_internals` interface anymore, since we consider every version bump of that package breaking. Thus, `!` now actually corresponds to behavioural breaks that a user of only `guppylang` can observe.

GBoydQ:
I have refactored to instead determine staticness during parsing and store it on parsed definitions as you've suggested, and also added support for comptime staticmethods.

The remaining failing test I had initially thought was a type inference error similar to ones that I had seen 'in the wild', but now think that it may be due to the fact that I am not calling `_instantiate_self` on protocol staticmethods.

I was experimenting with `Self` args a little (kind of necessary for protocol staticmethods to be useful), but I would prefer not to make this PR any larger + I do not understand `func_checker` and `protocol_checker` well enough to know the consequences of changes I would make there.

I would appreciate your thoughts, one possible option for now might be to not support staticmethods on protocols with type args.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -648,6 +648,35 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                    match typearg.ty:
+                        case BoundTypeVar() as ty:
+                            return self._check_bound_type_method(ty, node)
+                        case StructType() | EnumType() as ty:
+                            # case for when the type is known
+                            # what types actually should this be?
+                            ty_def = ty.defn
+                            if node.attr in DEF_STORE.type_members[ty_def.id] and (
+                                func := ENGINE.get_instance_func(ty_def, node.attr)
+                            ):
```
mark-koch: We should accept any type here, not just structs and enums.

`ENGINE.get_instance_func` should be able to handle that, not sure why there is the additional `DEF_STORE.type_members` check?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -648,6 +648,35 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                                return with_loc(
+                                    node, GlobalName(id=node.attr, def_id=func.id)
+                                ), func.ty
+
+            if node.value.id in self.ctx.globals:
+                defn = self.ctx.globals[node.value.id]
+                if (
+                    isinstance(defn, TypeDef)
+                    and node.attr in DEF_STORE.type_members[defn.id]
```
mark-koch: Same here, doesn't `get_instance_func` cover that?

guppylang-internals/src/guppylang_internals/definition/overloaded.py
```diff
@@ -120,6 +121,7 @@ def check_call(
     def synthesize_call(
         self, args: list[ast.expr], node: AstNode, ctx: "Context"
     ) -> tuple[ast.expr, Type]:
+        determine_static(self)
```
mark-koch: This looks like it has no effect. The side effect that this checks that all overloads have the same staticness is not really obvious at the call site here

guppylang-internals/src/guppylang_internals/definition/staticness.py
```diff
@@ -0,0 +1,85 @@
...
+            | RawFunctionDecl()
+            | RawTracedFunctionDef()
+        ):
+            if isinstance(defn.python_func, staticmethod):
+                return True, defn.python_func.__func__
+            else:
+                return False, None
+        case OverloadedFunctionDef():
+            # check all the methods in the overload are also static and error if not
+            # returns None regardless of staticness as there is nothing to unwrap
```
mark-koch: I think it would be clearer to just move this logic into `overloaded.py`. This moves the error to the place one would expect overloading errors to come from

And the remaining cases are fairly trivial, so maybe worth just inlining them, or just writing a small

```python
def determine_static(func: PyFunc) -> tuple[bool, PyFunc | None]
```

helper that does the unwrapping?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -379,6 +386,34 @@ def get_instance_func(self, ty: Type | TypeDef, name: str) -> CallableDef | None
...
+            and name in DEF_STORE.type_members[type_defn.id]
+        ):
+            return DEF_STORE.type_members[type_defn.id][name]
+        return None
+
+    def is_def_static(self, func_id: DefId) -> bool:
+        """Get staticness of parsed definition if it can be static."""
+        from guppylang_internals.definition.declaration import ParsedFunctionDecl
+        from guppylang_internals.definition.function import ParsedFunctionDef
+        from guppylang_internals.definition.traced import TracedFunctionDef
```
mark-koch: How about adding `is_static` as a required field to the `CallableDef` ABC and then simplifying this method to

```python
return isinstance(parsed, CallableDef) and parsed.is_static
```

This would make it more robust if more kinds of callable definitions are added in the future. Also this current version is missing overloaded or custom functions

guppylang-internals/src/guppylang_internals/definition/overloaded.py
mark-koch: Also annoying that we have to check this in `synthesize_call` but probably not worth refactoring that for this PR :/

tests/error/overload_errors/mixed_static.err
```diff
@@ -0,0 +1,11 @@
+Error: Invalid static overloads (at $FILE:15:4)
+   | 
+13 | 
+14 |     @guppy.overload(func1, func2)
+15 |     def overloaded() -> None: ...
+   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ Some overloads of method `overloaded` are static but others
+   |                                   are not
+   |                                           static: `func2`
+   |                                           non-static: `func1`
```
mark-koch: Very nice! Regarding the formatting, could you try turning this into a separate `Note` instead of adding it to the span?

You can do this by adding a new dataclass that implements the `Note` diagnostic ABC and then add it to the error using the `add_subdiagnostic` method

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -747,19 +779,68 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
         if func := ENGINE.get_instance_func(ty, node.attr):
             name = with_type(
                 func.ty, with_loc(node, GlobalName(id=func.name, def_id=func.id))
             )
-            # Make a closure by partially applying the `self` argument
             # TODO: Try to infer some type args based on `self`
-            result_ty = FunctionType(func.ty.inputs[1:], func.ty.output, func.ty.params)
-            return with_loc(node, PartialApply(func=name, args=[node.value])), result_ty
+            type_member = ENGINE.get_type_member(ty, node.attr)
+            if type_member and ENGINE.is_def_static(type_member):
```
mark-koch: Does `ENGINE.is_def_static(func.id)` suffice here?

mark-koch (commented):
Thanks for addressing the comments Greg and apologies for taking forever to get back to you. Really sorry for having this PR stuck in review limbo for so long 😅

Just a couple of small questions and suggestions, happy for @maximilianruesch to approve without me having another look

> I was experimenting with Self args a little (kind of necessary for protocol staticmethods to be useful), but I would prefer not to make this PR any larger + I do not understand func_checker and protocol_checker well enough to know the consequences of changes I would make there.
>
> I would appreciate your thoughts, one possible option for now might be to not support staticmethods on protocols with type args.

That sounds reasonable, I'd also be fine with just disallowing static methods on protocols and leaving that as a future feature. Whichever you prefer

guppylang-internals/src/guppylang_internals/definition/overloaded.py
GBoydQ: I have just inlined the simple staticness checks for everything other than overloads and added `is_overload_static` that determines if a overloaded method is static at decoration time

guppylang-internals/src/guppylang_internals/engine.py
GBoydQ: I have added `is_static` as a kwarg only field to `CallableDef`

tests/error/overload_errors/mixed_static.err
GBoydQ: Done

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
GBoydQ: I have changed this to be `case _` with the `BoundTypeVar` giving an error about staticmethods unsupported on protocols

GBoydQ:
Thanks both for your review.

@maximilianruesch I have added hints about calling instance methods on instances to the preexisting errors, and fixed the compilation stage error by adding `make_global_name`.
Also included a fix for staticmethod accesses on shadowed types

tests/error/enum_errors/use_enum_class5.py
maximilianruesch, line 1: Can you also delete the golden error file related to this class?

tests/error/struct_errors/instance_method_on_class_hint.err
maximilianruesch, line 1: This file / test probably deserves a new name

tests/error/struct_errors/field_access_on_class.err
```diff
@@ -6,6 +6,7 @@ Error: Instance field accessed on class (at $FILE:17:20)
    |                     ^^^^^ `field` is an instance field of `MyStruct`, not a static
    |                           field
 
-Help: Access it on an instance instead, e.g. `MyStruct(...).field`
+Help: `field` is an instance field. Maybe you mean to access it on an instance,
+e.g. `MyStruct(...).field`
```
maximilianruesch:
```suggestion
Help: `field` is an instance field. Maybe you meant to access it on an instance,
e.g. `MyStruct(...).field`?
```
Let us add some proper punctuation. Additionally, we usually use the past form to speak of past errors (similar to "did you mean to").

maximilianruesch (changes requested):

maximilianruesch (approved):
🚀 

Thank you for the incredible amount of work on this @GBoydQ, and thank you for the in-depth review @mark-koch!

(the semantic check fail can be ignored, since it only reports `staticmethod` being removed from unsupported Python std features. User code that was compilable before is still compilable now)

Issue #1596, opened 2026-03-25
[Feature]: Add hint when using `result` with `Measurement`


tatiana-s:
### Problem Statement

It would be nice to add a hint along the lines of "Call `read()` (blocking) to get a bool result from the measurement" when someone attempts to call `result` directly on measurements

### Proposed Solution

Would need to work out the best way to do this first:
- the current error is an overload error, which makes it hard to just add a hint to it (overload errors suppress previously thrown errors) 
- so adding another version of the overloaded result function is probably the way forward, however this means missing the overloaded information of all other available signatures
- on the last point, attempting to import anything from `qsystem/__init__.py` (`Measurement` in this case) leads to a weird indentation error during guppy parsing, not sure if this is a python import or guppy thing

### Component

Standard library

### Effort Estimate

Medium - Moderate complexity

PR #1707, opened 2026-04-30
feat!: Change the return type of measurement functions to a new `Measurement` type and remove opaque bools from the compiler

### LINKED ISSUE #1485: [Feature]: Change `measure()` to return a future bool
### Problem Statement

At runtime, measurements (and any quantum gates preceding them) are only actually performed when the result is needed. The real result of a measurement operation is a _future_ `bool`, which can later be read to instantiate the value. We want these semantics to be reflected in guppy.

### Proposed Solution

We should therefore make the return type of `measure()` a new type (e.g. `MeasureResult`, or just `Measurement`?) with a `read()` method. We can implement `__bool__()` on this type as syntactic sugar for `read()` (so that code like `if measure(q)` will continue to work as before).

We _may_ want to introduce more general `Future[T]` types when implementing this. (EDIT: This already exists.)

### Component

Guppy Compiler

### Effort Estimate

Hard - Significant effort required

### Additional Context

_No response_

tatiana-s:
Closes #1485

Requires a breaking tket release with the extension changes in [#1558 ](https://github.com/Quantinuum/tket2/pull/1558) before this can be merged 

BREAKING CHANGE: (guppylang.std.quantum): `measure` (and `measure_array` by extension) now return `Measurement` instead of `bool`
BREAKING CHANGE: (guppylang.std.qsystem/qsystem.functional): `measure` and `measure_reset` (and by extension the array versions) now return `Measurement` instead of `bool` (this matches the behaviour of `lazy_measure`/`lazy_measure_and_reset`)
BREAKING CHANGE: (guppylang_internals): `bool` is now lowered to `ht.Bool` again instead of `tket.bool`

The `deferred_measurement.ipynb` notebook has been adjusted to explain the motivation for the lazy behaviour instead of contrasting `measure` and `lazy_measure`


tests/error/modifier_errors/bad_custom_call.err
```diff
@@ -2,7 +2,10 @@ Error: Dagger constraint violation (at $FILE:9:4)
   | 
 7 | @guppy(dagger=True)
 8 | def test(x: qubit @owned) -> None:
-9 |     measure(x)
+9 |     measure(x).read()
   |     ^^^^^^^^^^ This function cannot be called in a dagger context
 
+Help: Consider adding the flag `(dagger=True)` to the decorator of the function
+`measure`
+
```
maximilianruesch: How was this not previously added? Is it because there was only one statement?

tests/integration/test_arithmetic.py
```diff
@@ -365,6 +365,9 @@ def main() -> nat:
 
 
 # TODO this should be a compile-time error
+# Fails with "pyo3_runtime.PanicException: Cannot convert negative integer -2 to
+# unsigned."
+@pytest.mark.xfail(reason="Rust PanicException cannot be caught from Python")
```
maximilianruesch, line 370: You can catch "Exception", and `PanicException` should be a subtype of that

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -568,39 +554,34 @@ def visit_StateResultExpr(self, node: StateResultExpr) -> Wire:
...
             qubit_arr_in = self.builder.add_op(
                 array_new(ht.Qubit, len(node.args) - 1), *qubits_in
             )
-            # Turn into standard array from borrow array.
-            qubit_arr_in = self.builder.add_op(
-                array_to_std_array(ht.Qubit, num_qubits_arg),
-                qubit_arr_in,
-            )
+        else:
+            qubit_arr_in = [self.visit(node.args[1])]
```
maximilianruesch: 1. Is the type of `qubit_arr_in` still current? In both cases it is at least a list I think, and
2. Does passing `Node` actually work? I presume that is returned by `self.visit`...

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -568,39 +554,34 @@ def visit_StateResultExpr(self, node: StateResultExpr) -> Wire:
...
-            qubit_arr_out = self.builder.add_op(
-                std_array_to_array(ht.Qubit, num_qubits_arg),
-                qubit_arr_out,
-            )
+        # Convert back from standard array to borrow array.
+        qubit_arr_out = self.builder.add_op(
+            std_array_to_array(ht.Qubit, num_qubits_arg),
+            qubit_arr_out,
+        )
+        qubits_out: Node | list[Wire]
```
maximilianruesch, line 577: Same question here

guppylang-internals/src/guppylang_internals/std/_internal/compiler/quantum.py
```diff
@@ -43,48 +43,36 @@ def from_halfturns_unchecked() -> ops.ExtOp:
 
 
 class InoutMeasureCompiler(CustomInoutCallCompiler):
-    """Compiler for the measure functions with an inout qubit
-    such as the `project_z` function - requiring conversion to tket.bool."""
+    """Compiler for the measure functions with an inout qubit"""
```
maximilianruesch:
```suggestion
    """Compiler for the measure functions with an inout qubit."""
```

guppylang-internals/src/guppylang_internals/std/_internal/compiler/quantum.py
```diff
@@ -43,48 +43,36 @@ def from_halfturns_unchecked() -> ops.ExtOp:
...
     ext: he.Extension
-
-    def __init__(self, opname: str | None = None, ext: he.Extension | None = None):
+    use_bool: bool
+
+    def __init__(
+        self,
+        opname: str | None = None,
+        ext: he.Extension | None = None,
+        use_bool: bool = False,
```
maximilianruesch: Bool arguments should most often be keyword only, so please insert a `*` as a preceding parameter.

guppylang-internals/src/guppylang_internals/std/_internal/compiler/quantum.py
```diff
@@ -43,48 +43,36 @@ def from_halfturns_unchecked() -> ops.ExtOp:
...
 class InoutMeasureCompiler(CustomInoutCallCompiler):
-    """Compiler for the measure functions with an inout qubit
-    such as the `project_z` function - requiring conversion to tket.bool."""
+    """Compiler for the measure functions with an inout qubit"""
 
     opname: str
     ext: he.Extension
-
-    def __init__(self, opname: str | None = None, ext: he.Extension | None = None):
+    use_bool: bool
```
maximilianruesch: Perhaps this should rather be "use_future" or even "use_future_over_bool", since a measurement is often returning a bool-like thing (in literature), so it is more useful to indicate what the alternative is. Either way, this should have a brief explanation in a comment that documents this argument (perhaps a docstring to `__init__` since that is the entrypoint).

guppylang/src/guppylang/std/bool.py
```diff
@@ -26,13 +26,13 @@ class bool:
     ``False`` using the standard truth testing procedure.
     """
 
-    @hugr_op(bool_logic_op("and"))
+    @hugr_op(logic_op("And"))
```
maximilianruesch, line 29: Incredibly sneaky case change, but ok

guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -197,43 +198,67 @@ def discard(self: "MaybeLeaked @ owned") -> None:
...
 
 N = guppy.nat_var("N")
 
 
 @guppy
 @no_type_check
-def measure_array(qubits: array[qubit, N] @ owned) -> array[bool, N]:
+def measure_array(qubits: array[qubit, N] @ owned) -> array[Measurement, N]:
     """Measure an array of qubits, returning an array of bools."""
     return array(measure(q) for q in qubits)
```
maximilianruesch, line 249: These two functions (this and the reset variant) should include the forward / backward references that the regular functions do in their docstrings. In particular, somewhere it should state that the measurement is lazy.

guppylang/src/guppylang/std/qsystem/functional.py
maximilianruesch, line 1: Perhaps the functions in this file should also state that they are lazy, or point to their wrapped symbols in the docstrings.

examples/postselect.ipynb
maximilianruesch, line 1: The metadata in this file (and other files) should have been cleaned by the pre-commit. Are you sure it ran / could you rerun it?

examples/canonical-qpe.ipynb
```diff
@@ -255,7 +255,7 @@
     "    discard_array(state)\n",
     "\n",
     "    # Create a result from the measured array\n",
-    "    result(\"c\", measure_array(measured))\n"
+    "    result(\"c\", collect_measurements((measure_array(measured))))\n"
```
maximilianruesch: Superfluous parantheses

tests/integration/test_pytket_circuits.py
```diff
@@ -38,6 +40,9 @@ def foo(q1: qubit, q2: qubit) -> None:
     validate(foo.compile_function())
 
 
+@pytest.mark.skip(
```
maximilianruesch: Instead of `pytest.mark.skip`, try to use `pytest.mark.xfail`. Instead of skipping the test, this will execute it and expect it to fail. This way we notice when tests suddenly work again (for example when the given issue is resolved), instead of having to try from time to time.

tests/integration/test_pytket_circuits.py
maximilianruesch, line 1: Just double checking that we are okay with breaking measurements in pytket circuits for now / until upgrading to a tket version that can decode them?

tests/integration/test_quantum.py
```diff
@@ -127,7 +127,7 @@ def test_measure_array(validate):
     """Build and measure array."""
 
     @guppy
-    def test() -> array[bool, 10]:
+    def test() -> array[Measurement, 10]:
```
maximilianruesch, line 130: If I understand correctly this is okay since we can coerce `Measurement` to `__bool__`? Do we have an explicit test for that? (May already exist, since `Measurement` existed before, but now it is very widely used since we lazify all measurements).

examples/deferred_measurement.ipynb
```diff
@@ -26,66 +29,57 @@
...
     "\n",
-    "In the example above, using `result` does mean forcing an immediate measurement of `q`, which isn't the best for performance. However, this isn't obvious from a user perspective. \n",
+    "Now `measure` (and adjacent functions such as `measure_array`) return a `Measurement` type, which is a handle to the measurement request. In order to block on the request and wait to receive a bool, the `read()` method should be used. This allows the runtime to defer measurements until they are actually needed, which can give more opportunities for parallelism and therefore better performance. Exposing this behaviour allows you to make more informed decisions about how to structure your program for better performance.\n",
     "\n",
-    "Now consider the same function using `lazy_measure`: "
+    "Consider the following program, where you might be tempted to use the output of `measure` directly in `result`:"
```
maximilianruesch: You will have to state which version is "previous", since this will become outdated / ambiguous fast when we make new releases. Something like "in Guppy v0" or similar. Remember that these examples are rendered on the docs page and have to be sufficiently self contained (i.e. not read like a pull request).

Also: Do we want to speak directly about what IS happening on the hardware, or more generally about the CAPABILITY of some hardwares to defer measurements (and thus require the concept of measurement requests), and that we have adjusted the model with v1 to capture this generality.

examples/deferred_measurement.ipynb
```diff
@@ -101,100 +95,101 @@
...
-   ]
+   "execution_count": 2
   },
   {
    "cell_type": "markdown",
    "id": "683a89c6",
    "metadata": {},
    "source": [
-    "Simply replacing the method results in an error because `lazy_measure` returns a value of type `Measurement`. In order to obtain a `bool`, we have to explicitaly use `read()` on the value. "
+    "Simply replacing the method results in an error because `result` expects a `bool`. In order to obtain it, we have to explicitly use `read()` on the value of type `Measurement`. This will block until the physical measurement is completed:"
```
maximilianruesch: We are not "replacing" the method anymore, since we are breakingly forcing an upgrade. This needs a reword, potentially to just "This results in an error because ...".

tests/integration/test_qsystem.py
```diff
@@ -36,17 +41,18 @@ def test_qsystem(validate):  # type: ignore[no-untyped-def]
...
         q1 = phased_x(q1, a1, a1)
         q1, q2 = zz_phase(q1, q2, a1)
         q1 = rz(q1, a1)
         q1, q2 = zz_max(q1, q2)
-        q1, b = measure_and_reset(q1)
+        q1, msmt1 = measure_and_reset(q1)
+        msmt1.read()
         q1 = reset(q1)
-        b = measure(q1)
+        msmt2 = measure(q1)
```
maximilianruesch, line 53: I hate that I can read this abbreviation...

examples/deferred_measurement.ipynb
```diff
@@ -101,100 +95,101 @@
...
+   "outputs": [],
+   "execution_count": 3
   },
   {
    "cell_type": "markdown",
    "id": "11043eea",
    "metadata": {},
    "source": [
-    "The program now type-checks and ends up compiling to the exact same operation order as `eager_read`. We can now see where exactly the `read()` happens, rather than it being done implicitly, so we can try to move it further down the program."
+    "The program now type-checks and we can see where exactly the `read()` happens, rather than it being done implicitly. Therefore we can try to move it further down the program, which allows for flexibility in when the quantum operations before the blocking read are performed:"
```
maximilianruesch:
```suggestion
    "The program now type-checks and we can see where exactly the `read()` happens, rather than it being done implicitly. Therefore we can try to move it further down the program, which allows for flexibility in when the quantum operations preceeding the blocking read are performed:"
```
In addition to this change, think about the wording in this file reading less like a changelog (e.g. avoiding the "rather than"). The vast majority of future users will (hopefully) only see v1. You may instead want to consider adding this to a new v0 -> v1 migration guide.

examples/deferred_measurement.ipynb
```diff
@@ -101,100 +95,101 @@
...
    "cell_type": "markdown",
    "id": "6e39f898",
    "metadata": {},
    "source": [
-    "Now that each measurement is only read at the end of the program, physical measurements can be deferred to a better point in execution.\n",
-    "\n",
-    "Of course in this case we could have also collected the `bool` returns of `measure` or used `measure_array` to achieve the same outcome, however there might be cases where other solutions are less obvious. It can therefore be useful to use `lazy_measure` in programs where you want more control over when measurements should happen.\n",
     "\n",
-    "For convenience, `__bool__` is implemented on the `Measurement` type as syntactic sugar for `read()`, so it is called automatically in conditionals for example:"
+    "For convenience, `__bool__` is implemented on the `Measurement` type as syntactic sugar for `read()`, so it is called automatically in conditionals, for example:"
```
maximilianruesch:
```suggestion
    "For convenience, the boolean coercion `__bool__` is implemented on the `Measurement` type as syntactic sugar for `read()`, so it is called automatically in conditionals, for example:"
```

examples/t_factory.ipynb
```diff
@@ -311,7 +399,12 @@
     "def main() -> None:\n",
     "    qs = array(some(qubit()) for _ in range(3))\n",
     "    mask = array(True, False, True)\n",
-    "    measure_mask(qs, mask)\n",
+    "    for opt_msmt in measure_mask(qs, mask):\n",
+    "        if opt_msmt.is_some():\n",
+    "            m = opt_msmt.unwrap()\n",
+    "            result(\"result\", m.read())\n",
+    "        else:\n",
+    "            opt_msmt.unwrap_nothing()\n",
```
maximilianruesch, line 408: Add a comment stating that we need to consume the array of measurement options since it is linear. Also, you added a new output that was not there previously. Is it mentioned somewhere how the values collate when you call `result` multiple times with the same string?

maximilianruesch (changes requested):
Big PR, lots of little changes. Great work! I massively trust the tests on this one.

A few comments here and there. Also I think the title (which will end up as the commit title) should mention that we drop the opaque guppy bools / that we *change* the return type of measurement functions, not just add some type.

tests/error/modifier_errors/bad_custom_call.err
tatiana-s: My guess is maybe because previously `measure` was defined through a `@hugr_op` and now it is just a guppy function?

tests/integration/test_arithmetic.py
tatiana-s: I did try that and it didn't work?

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
tatiana-s: Changed it to `Wire` since `self_visit` returns a `Wire` and `add_op` returns a `Node` which implements the `Wire` protocol to, probably got confused with the type below (though since `Node` is iterable this annotation also works, just means we don't need the `*` operator a few lines later now)

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
tatiana-s: This type is correct - `unpack_array` returns a `list[Wire]` and `add_op` returns a `Node` and they both can be passed to the `iter` later

I do agree it's a little bit confusing in the python interface because intuitively `add_op` returns the output wires of the op, but here we get the `Node` representing the op but then it can be used like the output wires, and it it's multiple wires it works because of 

```
    def __iter__(self) -> Iterator[OutPort]:
        return self.outputs()
```
implemented on it

guppylang/src/guppylang/std/bool.py
tatiana-s: Just the `tket.bool` extension being inconsistent with the built in hugr bool, but given it is being removed now doesn't matter anyway

examples/postselect.ipynb
tatiana-s: Definitely ran `just check` many times

tests/integration/test_pytket_circuits.py
tatiana-s: We actually can't merge this PR before a tket release anyway, and we can't do a tket release until I also implement some pytket conversion follow up PRs that should make this work again, so I will update this PR anyway at that point (so also will just leave as skip re: comment above as it's just for now)

tests/integration/test_quantum.py
tatiana-s: There is `test_lazy_measure_conditional` in `test_qsystem` from when we first introduced `Measurements`

examples/deferred_measurement.ipynb
tatiana-s: good question, probably worth a discussion

tatiana-s: will delete this notebook and then later make separate issues for migration guide / docs updates

tests/integration/test_arithmetic.py
maximilianruesch: What do you mean by "it did not work"? The panic went through? You can also catch `BaseException` even more generally...

examples/postselect.ipynb
maximilianruesch: I was talking about https://github.com/kynan/nbstripout, but indeed it does not seem to be part of our pre-commit. I recall having issues with this in Dec 25. Am I hallucinating???

tests/integration/test_pytket_circuits.py
maximilianruesch: Not needed since the skips will be removed before the PR is merged.

examples/t_factory.ipynb
```diff
@@ -311,7 +399,13 @@
     "def main() -> None:\n",
     "    qs = array(some(qubit()) for _ in range(3))\n",
     "    mask = array(True, False, True)\n",
-    "    measure_mask(qs, mask)\n",
+    "    # We need to consume the array of measurement options as they are linear.\n",
```
maximilianruesch, line 402: This comment and the next one have the same underlying reason (linear -> cannot be leaked), but both only display the truth partially.

maximilianruesch (changes requested):
LGTM for now. When merging / handling the latest tket release, please rerequest a review, and then I can approve.

guppylang/src/guppylang/std/quantum/__init__.py
```diff
@@ -392,6 +420,17 @@ def discard_array(qubits: array[qubit, N] @ owned) -> None:
     qubits.discard_all_taken()
 
 
+@guppy
+@no_type_check
+def collect_measurements(
```
acl-cqc, line 425: I guess it's been called `collect_measurements` for awhile but it feels like it ought to have `read` in the name at least, given the measurements are *already* in a collection (so no collecting is done) and there's nothing in the name to say that it's `read`ing!!

acl-cqc (commented):
Thanks @tatiana-s! A big PR already I admit but is this the time to add overrides to `result` that accept Measurement directly? (Or a followup?)

tatiana-s:
@acl-cqc at least we want https://github.com/Quantinuum/guppylang/issues/1596 but as a follow up (making results lazy themselves is quite a big follow up for later)

.github/workflows/pull-request.yaml
maximilianruesch, line 1: This file and the `semver-checks.yml` have changes unrelated to this PR. Remove em please?

pyproject.toml
maximilianruesch, line 1: Imo it would be cleaner to remove them, rather than add new commented lines. At least remove the changes from the PR please.

guppylang/src/guppylang/std/quantum/__init__.py
```diff
@@ -42,6 +43,28 @@ def discard(self: "qubit" @ owned) -> None:
...
+class Measurement:
+    """Represents the result of a lazy measurement which needs to be explicitly read
+    before being used."""
+
+    @hugr_op(quantum_op("Read", MEASUREMENT_EXTENSION))
+    @no_type_check
+    def read(self: "Measurement") -> bool:
+        """Read the measurement result, consuming it. Blocks until the result is
+        available if the measurement hasn't been performed yet since being requested.
+        """
```
maximilianruesch, line 60: This is not "consuming" anymore, is it?

guppylang/src/guppylang/std/quantum/__init__.py
```diff
@@ -42,6 +43,28 @@ def discard(self: "qubit" @ owned) -> None:
         discard(self)
 
 
+@custom_type(
+    ht.ExtType(MEASUREMENT_EXTENSION.get_type("Measurement")),
+    copyable=True,
+    droppable=True,
+)
+class Measurement:
```
maximilianruesch, line 51: Since we do not have threading, copying a future is fine (it can only be resolved in one place anyway). But this could be a footgun if I am not mistaken.

maximilianruesch (approved):
May have to upgrade to the latest `tket-py` for the tests to pass.

guppylang/src/guppylang/std/quantum/__init__.py
tatiana-s: The explicit handling of copying during replacement is the same as when we were replacing copyable bools with linear futures so it's been used for for a while

PR #1708, opened 2026-04-30
fix: Use of array subscript as argument of control modifiers

### LINKED ISSUE #1704: [Bug]: Use of arrays with controllers
### Description

The following program:
```python
@guppy
def main() -> None:
    array_qubits: array[qubit, 2] = array(qubit(), qubit())

    with control(array_qubits[0]):
        h(array_qubits[1])

    discard_array(array_qubits)
```
fail during compilation due to an AssertionError. The program should be valid

nicolaassolini-qntm:
closes #1704 

Fixes handling of array subscripts used as control modifier arguments.

- Call `check_place_assignable` on control arguments statement checking, so subscript controls get a synthesised `setitem_call`.
- Fix modifier HUGR generation so captured arguments are read after control arguments are compiled, avoiding stale array wires. 
- Return controller qubits borrowed from array subscripts back into their parent arrays after the modified indirect call. Use the synthesised setter during the HUGR generation to write returned control values back into their parent array slots.

- Updated integration test.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -158,3 +158,4 @@ devenv.local.nix
 **/.benchmarks/
 .vscode/settings.json
 guppy-exports/
+AGENTS.md
--- a/.gitignore
+++ b/.gitignore
@@ -158,4 +158,3 @@ devenv.local.nix
 **/.benchmarks/
 .vscode/settings.json
 guppy-exports/
-AGENTS.md
--- a/guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
@@ -15,6 +15,7 @@ from guppylang_internals.std._internal.compiler.arithmetic import convert_itousi
 from guppylang_internals.std._internal.compiler.array import (
     array_new,
     array_to_std_array,
+    barray_borrow,
     barray_return,
     standard_array_type,
     std_array_to_array,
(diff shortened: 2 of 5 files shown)


guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
```diff
@@ -171,9 +199,21 @@ def compile_modified_block(
...
             dfg[arg] = next(outports)
 
+    for subscript in control_subscripts:
+        # For nested places, use the synthesized setter; plain array subscripts can
+        # be restored directly with the borrow_array return op.
+        if subscript.setitem_call is not None:
+            dfg[subscript.setitem_call.value_var] = dfg[subscript]
+            expr_compiler.visit(subscript.setitem_call.call)
+        else:
+            _return_control_subscript(subscript, dfg, ctx)
```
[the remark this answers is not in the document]
qartik: @nicolaassolini-qntm what's the resolution here?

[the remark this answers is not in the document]
qartik: e.g. with
```py
def test_nested_element_control_subscript(validate):
    @guppy
    def main(qs: array[array[qubit, 2], 2], target: qubit) -> None:
        with control(qs[0][0]):
            h(target)

```
we get
```
selene_hugr_qis_compiler.HugrReadError:
Node(17) has a port Port(Outgoing, 0) of type borrow_array(2, qubit) with more than one connection.

```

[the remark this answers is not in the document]
nicolaassolini-qntm: You're right, I need to investigate a bit

nicolaassolini-qntm:
I have refactored the proposed solution, and it should be way better

.gitignore
```diff
@@ -158,3 +158,4 @@ devenv.local.nix
 **/.benchmarks/
 .vscode/settings.json
 guppy-exports/
+AGENTS.md
```
qartik:
```suggestion
```
I don't think we should be ignoring this file. If you have found your local agents file useful, perhaps consider sharing it as a PR for addition to the repo.

nicolaassolini-qntm: I'm still experimenting, I would like not to push it for now

qartik: I meant in a separate PR. Regardless, could you remove this gitignore line as it's irrelevant to this PR and may create surprises for people later?

nicolaassolini-qntm: Sure!

qartik (approved):

Issue #1597, opened 2026-03-25
[Feature]: Improve error rendering with modifiers


nicolaassolini-qntm:
### Problem Statement

An open issue, we need to go through the error tests of the modifier and improve the messages when they are not good enough

### Component

Guppy Compiler

### Effort Estimate

Easy - Small change or addition

nicolaassolini-qntm:
https://github.com/Quantinuum/guppylang/blob/e3ac40fcf017abec8814988edf270b35d4a4a654/tests/error/modifier_errors/variable_leak.py#L5-L6

nicolaassolini-qntm:
~~here:
https://github.com/Quantinuum/guppylang/blob/5b2dbc55cf9023d11a47cfb29c159da9e5a07906/tests/error/modifier_errors/bad_custom_call.err~~

~~Is it possible to succinctly explain the criteria for functions that can't be called in a dagger context?~~

nicolaassolini-qntm:
Improve error:
```python
@guppy(unitary=True)
def bar(q1: qubit) -> None:
    q2 = qubit()
    cx(q1, q2)
    with control(q2):
        h(q1)
    measure(q2)
```

nicolaassolini-qntm:
Also consider:
```python
@guppy
def foo[T: Copy](x: T) -> tuple[T]:
    with power(2):
        x = x
    return x
```

nicolaassolini-qntm:
All the issues addressed here have been solved in previous PRs

PR #1712, opened 2026-05-01
feat!: Implement error handling for modified variables outside modifiers

### LINKED ISSUE #1700: [Feature]: Raise an error when variables are used outside a modifier block
### Problem Statement

Variables assignment a modifier block are not reflected to the outside.
For instance:
```python
x = 3
with power(2):
    x += 1
y = x #here x = 3, not 5
```

### Proposed Solution

To avoid counterintuitive programs, we should raise an error when we try to use a variable that has been modified in the modifier block. (the mod. block should "consume" classical variable).

nicolaassolini-qntm:
closes #1700 

This PR add checks to raise an error when we use variables outside a modifier block when their value was assigned inside that modifier block. Moreover it improve the error rendering for undefined variables when the variables are only defined inside the modifier block.

- Track variables assigned inside modifier blocks during CFG analysis.
- Report `Variable assigned in modifier block` when a previously defined variable is assigned in a modifier and then used outside it.
- Add help notes for variables that are defined only inside a modifier block and later used outside it.
- Preserve valid reassignment behavior when a variable is reassigned after the modifier block before being used.
- Add regression coverage for nested modifiers, sequential modifiers, branches, captured classical variables, and variables defined only inside modifiers.

BREAKING CHANGE: Using a variable outside a modifier block after it was assigned inside that modifier block now raises a compilation error. 
 
BEGIN_COMMIT_OVERRIDE
feat!: Error handling for outside use of variables scoped to modifier blocks (#1712)
END_COMMIT_OVERRIDE


guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -887,13 +887,16 @@ class CheckedModifiedBlock(ast.With):
     ty: FunctionType
     #: Mapping from names to variables captured in the body.
     captured: Mapping[str, tuple["Variable", AstNode]]
+    #: Mapping from captured names to assignments inside the body.
+    modified_captured: Mapping[str, tuple["Variable", AstNode]]
```
mark-koch: I think it would be better to call these `shadowed` throughout the entire PR

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -734,6 +751,11 @@ def visit_CheckedModifiedBlock(self, node: CheckedModifiedBlock) -> None:
             if InputFlags.Inout in var.flags:
                 self._reassign_single_inout_arg(var, var.defined_at or use)
 
+        for var, assignment in node.modified_captured.values():
+            if var.ty.copyable:
+                for place in leaf_places(var):
+                    self.scope.use(place.id, assignment, UseKind.REDEFINED_IN_MODIFIER)
```
mark-koch: Mhmm tracking these as uses seems wrong since they're actually assignments. I assume that's also why you included the copyable check?

For example here:
```python
@guppy
def foo[T: Copy](x: T) -> tuple[T]:
    with power(2):
        x = x
    return x

@guppy
def main() -> int:
    return foo(0)[0]
```

we get an error about a leak, but that's confusing since x is used later. We should instead complain at the return point that x was assigned in the modifier

guppylang-internals/src/guppylang_internals/checker/errors/linearity.py
```diff
@@ -55,6 +55,26 @@ class MakeCopy(Help):
         )
 
 
+@dataclass(frozen=True)
+class ModifiedVariableUsedError(Error):
+    title: ClassVar[str] = "Variable modified in modifier block"
```
mark-koch:
```suggestion
    title: ClassVar[str] = "Variable assigned in modifier block"
```

Similar for the other uses of "modified" in the PR

mark-koch (commented):
I had a initial look through, but I'm not convinced that a new `UseKind` is the best way to implement this. Currently, you're marking variables as used that are actually not used in the block, which doesn't feel good. In particular, what we actually care about is assignments, so presumably the assignment should track whether it happened in modifier block, not the use?

Furthermore, I think it would be better to check this before linearity checking and maybe even before type checking. This would be more consistent with how we currently catch undefined variable errors (see `check_bb` in `cfg_checker.py`). Any way we can include this logic at that point?

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
mark-koch: Related,

```python
@guppy
def main() -> int:
    with power(2):
        x = 0
    return x
```

here we just get an undefined error, but it would be better to mention the modifier block

guppylang-internals/src/guppylang_internals/cfg/bb.py
```diff
@@ -232,6 +232,8 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> None:
         from guppylang_internals.cfg.analysis import LivenessAnalysis
 
         stats = {bb: bb.compute_variable_stats() for bb in node.cfg.bbs}
+        # NICOLA: Should use this stats: store here the vars assigned in the block,
+        # and save during the visitor if a used variable had been assigned in the block or not.
```
nicolaassolini-qntm: Refactor following this suggestion

nicolaassolini-qntm:
```python
@guppy
def foo[T: Copy](x: T) -> tuple[T]:
    with power(2):
        x = x
    return x
```
The error in this case should be improved, but maybe in a different PR? I added the example in #1597

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
```diff
@@ -262,11 +300,26 @@ def check_bb(
...
+        raise GuppyError(
+            _assigned_in_modifier_error(x, use, bb.vars.assigned_in_modifier_block[x])
+        )
+    # We also check that the variables used in the block was not assigned in a modifier
+    # block in a predecessor
+    for x, use in bb.vars.used.items():
+        if x in cfg.assigned_in_mod[bb] and x in cfg.live_before[bb]:
+            raise GuppyError(
+                _assigned_in_modifier_error(x, use, cfg.assigned_in_mod[bb][x])
+            )
```
mark-koch: The following would be better: for every successor block, check if one of the variables in `bb.vars.assigned_in_modifier_block` is live before that successor. Then you should be able to get rid of your custom dataflow analysis

guppylang-internals/src/guppylang_internals/cfg/bb.py
```diff
@@ -231,9 +247,27 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> None:
...
+            # each BB.
+            for var, ast_node in (
+                # We union the two dicts in this order to prioritize variables assigned
+                # in the outer modifier block
+                bb_stat.assigned_in_modifier_block | bb_stat.assigned
+            ).items():
+                # if already present we keep the first assignment
+                if var not in self.stats.assigned_in_modifier_block:
+                    self.stats.assigned_in_modifier_block[var] = ast_node
+            stats[bb] = bb_stat
```
mark-koch, line 269: Not sure I see what you're trying to do here? Can't we populate the fields during the visiting without having to do post processing?

tests/error/errors_on_usage/var_in_modifier1.err
```diff
@@ -0,0 +1,15 @@
...
+17 |     measure(q)
+18 |     return a
+   |            ^ `a` is not defined
+
+Help:
+   | 
+12 |     with power(3):
+13 |         a = 3
+   |         ----- a is defined inside a modifier block, but variables defined
+   |               in a modifier block are not available outside of it
```
mark-koch: How about:

```
`a` cannot be used...
```

and then a note:

```
... since it was defined in a modifier block
```

tests/error/modifier_errors/captured_classical_modified.err
```diff
@@ -0,0 +1,17 @@
...
+
+Note:
+   | 
+ 8 |     with power(2):
+ 9 |         x += 1
+   |         ------ `x` assigned here
+
+Help: Assignments inside modifier blocks are not reflected outside the block
+
+Guppy compilation failed due to 1 previous error
```
mark-koch, line 17: Looks like you have two different kinds of messages for the same error? Would be better to make them consistent

tests/error/errors_on_usage/var_in_modifier1.err
mark-koch: Also, backticks around the variable!

guppylang-internals/src/guppylang_internals/cfg/bb.py
nicolaassolini-qntm: The assigned_in_modifier_block variables are the one assigned in the block `bb_stat.assigned` plus that are `assigned_in_modifier_block` in bb_stats, i.e. variables that are assigned in a modifier block inside `bb`.
We are not going to visit `node.cfg.bbs` directly, what I can do is trigger the `compute_variable_stats` and the get the info that I need.

guppylang-internals/src/guppylang_internals/cfg/bb.py
```diff
@@ -151,6 +167,10 @@ def _handle_assign_target(self, lhs: ast.expr, node: ast.stmt) -> None:
         match lhs:
             case ast.Name(id=name):
                 self.stats.assigned[name] = node
+                # We remove the variable from `assigned_in_modifier_block` only if it
+                # has not been used yet
+                if name not in self.stats.badly_used_after_modifier_block:
+                    self.stats.assigned_in_modifier_block.pop(name, None)
```
mark-koch: This breaks the invariant you put above:

```python
# Variables whose last assignment in this BB happened inside a modifier block.
assigned_in_modifier_block: dict[VId, AstNode] = field(default_factory=dict)
```

As far as I can tell, you want to keep it inside `assigned_in_modifier_block` so you can look it up later? Why not store all the information you need directly in `badly_used_after_modifier_block`?

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
```diff
@@ -268,6 +295,15 @@ def check_bb(
...
                 raise GuppyError(VarNotDefinedError(use_bb.vars.used[x], x))
 
+            # We also check that the variables assigned in a modifier block are not
+            # in successor blocks
+            if x in bb.vars.assigned_in_modifier_block:
+                raise GuppyError(
+                    _assigned_in_modifier_error(
+                        x, use_bb.vars.used[x], bb.vars.assigned_in_modifier_block[x]
+                    )
+                )
```
mark-koch, line 305: I think it would make more sense to move this into the `if x in cfg.assigned_somewhere` branch since it only applies to locals.

In particular, `cfg.assigned_somewhere` should probably include variables assigned in modifier blocks?

guppylang-internals/src/guppylang_internals/cfg/bb.py
```diff
@@ -231,9 +251,27 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> None:
...
+            # We update the `assigned_in_modifier_block` field of the stats for
+            # each BB.
+            for var, ast_node in (
+                # We union the two dicts in this order to prioritize variables assigned
+                # in the outer modifier block
+                bb_stat.assigned_in_modifier_block | bb_stat.assigned
+            ).items():
+                # if already present we keep the first assignment
+                if var not in self.stats.assigned_in_modifier_block:
+                    self.stats.assigned_in_modifier_block[var] = ast_node
```
mark-koch: Why keep the first assignment? The new one shadows it so wouldn't it be better to point to that one?

guppylang-internals/src/guppylang_internals/cfg/analysis.py
```diff
@@ -144,7 +144,9 @@ def join(self, *ts: LivenessDomain[VId]) -> LivenessDomain[VId]:
     def apply_bb(self, live_after: LivenessDomain[VId], bb: BB) -> LivenessDomain[VId]:
         stats = self.stats[bb]
         return dict.fromkeys(stats.used, bb) | {
-            x: b for x, b in live_after.items() if x not in stats.assigned
+            x: b
+            for x, b in live_after.items()
+            if x not in stats.assigned and x not in stats.assigned_in_modifier_block
```
mark-koch, line 149: Isn't `assigned_in_modifier_block` a subset of `assigned`?

guppylang-internals/src/guppylang_internals/cfg/bb.py
```diff
@@ -126,8 +133,17 @@ def _update_used(self, node: ast.AST) -> None:
             # Should point to first use, so also check that the name is not already
             # contained
             x = name.id
-            if x not in self.stats.assigned and x not in self.stats.used:
+            if (
+                x not in self.stats.assigned
+                and x not in self.stats.used
+                and x not in self.stats.assigned_in_modifier_block
```
mark-koch, line 142: Same here, `assigned_in_modifier_block` is a subset of `assigned`, so no need to check it?

guppylang-internals/src/guppylang_internals/cfg/bb.py
```diff
@@ -231,9 +251,27 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> None:
...
+        stats: dict[BB, VariableStats[str]] = {}
+
+        for bb in node.cfg.bbs:
+            bb_stat = bb.compute_variable_stats()
+            # We update the `assigned_in_modifier_block` field of the stats for
+            # each BB.
+            for var, ast_node in (
+                # We union the two dicts in this order to prioritize variables assigned
+                # in the outer modifier block
+                bb_stat.assigned_in_modifier_block | bb_stat.assigned
```
mark-koch, line 266: Same here, `bb_stat.assigned` should already contain everything?

mark-koch (commented):
Nice! This looks a lot cleaner now 👍 

Just a couple of questions and maybe opportunities to simplify further

guppylang-internals/src/guppylang_internals/cfg/analysis.py
nicolaassolini-qntm: no it is not

guppylang-internals/src/guppylang_internals/cfg/bb.py
nicolaassolini-qntm: Before this pr, `assigned` didn't include variables assigned in the block; thus, I defined a new set of variables. In principle, I think we can use bb_stat.assigned for everything? Not sure if this can break how we detect that a variable defined in a modifier block inside a bb is badly used in another bb

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
nicolaassolini-qntm: We should change the Assigment analysis to have the variables assigned in the block registered in cfg.assigned_somewhere

guppylang-internals/src/guppylang_internals/cfg/bb.py
nicolaassolini-qntm: true

guppylang-internals/src/guppylang_internals/cfg/bb.py
mark-koch: Oh yes I see, sorry! In that case it's fine as is 👍

mark-koch (approved):

PR #1716, opened 2026-05-05
chore(ci): Benchmark main with CodSpeed

ferbetanzo:
Adds missing base benchmarking with CodSpeed, and changes the CodSpeed auth method to OIDC instead of the legacy token-based.

### DIFF
--- a/.github/workflows/benchmarks_base.yml
+++ b/.github/workflows/benchmarks_base.yml
@@ -32,7 +32,7 @@ jobs:
       - name: Set up Just
         uses: extractions/setup-just@v4
 
-      - name: Benchmark main with Bencher
+      - name: Benchmark with Bencher
         run: |
           just bench-upload \
                --token '${{ secrets.BENCHER_API_TOKEN }}' \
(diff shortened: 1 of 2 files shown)


.github/workflows/benchmarks_base.yml
```diff
@@ -28,10 +32,16 @@ jobs:
       - name: Set up Just
         uses: extractions/setup-just@v4
 
-      - name: Track base branch benchmarks with Bencher
+      - name: Benchmark main with Bencher
```
maximilianruesch: nit: "main" is already mentioned in the title of the entire workflow file, so we can just omit it here (and in the next step).

maximilianruesch (approved):

.github/workflows/benchmarks_base.yml
ferbetanzo: 👍🏼 Fixed in https://github.com/Quantinuum/guppylang/pull/1716/commits/5d8ea14238a5f23ac834868c01ff6b1f70d0b92a

PR #1717, opened 2026-05-05
chore(ci): Lock instead of freezing in PR CI during dependency sync

maximilianruesch:
This fails when the `uv.lock` is not up to date, which has been a problem in a number of instances (such that subsequent PRs find themselves needing to include seemingly random `uv.lock` changes). `--locked` vs `--frozen` strictly tightens the requirements for success.

Also happy to pursue a separate action instead of modifying this one if required.


aborgna-q (approved):
👍

PR #1719, opened 2026-05-06
chore: Switch main branch release series to a pre-release alpha

maximilianruesch:
Required changes to update the release-please PRs to follow the new alpha versioning strategy we will adopt for the break to v1.

This set of changes ensures the base version is still updated correctly, but the next release has the "-alpha.X" suffix. We have to manually update the string to be "beta" when we put the package into beta.

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -2,8 +2,6 @@
   "$schema": "https://raw.githubusercontent.com/googleapis/release-please/main/schemas/config.json",
   "separate-pull-requests": true,
   "pull-request-title-pattern": "chore: release${component} ${version}",
-  "bump-minor-pre-major": true,
-  "bump-patch-for-minor-pre-major": true,
   "include-component-in-tag": true,
   "include-v-in-tag": true,
   "extra-label": "X-release",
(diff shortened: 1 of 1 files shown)


release-please-config.json
```diff
@@ -28,7 +27,8 @@
       "component": "guppylang-internals",
       "package-name": "guppylang_internals",
       "draft": false,
-      "prerelease": false,
```
ss2165, line 31: do we actually need to change anything about internals?

maximilianruesch: We will need to at least make sure it is promoted to 1.X once. After this, we can go with bumping minors for breaking releases again. We have to do this to avoid clashes with the 0.X bumps from the `guppylang-0.21` branch.

Of course, that does not mean we have to put it into pre-release mode.

ss2165: bumping minors for breaking releases would not be semver adherent?

maximilianruesch: It would not, but imo semver does not really make sense for guppylang-internals anyway, since it breaks so often (almost every release). A simple build number would be more fitting, but since we do not want the version from the main and the 0.X branch to interfere with one another, we would have to prefix either. And then it can just be 1.x and 0.X.

This will also give us the flexibility to do a 2.X series (although that should never be a thing, but who knows).

ss2165 (approved):

PR #1722, opened 2026-05-06
refactor: Add some missing override annotations

acl-cqc:
Done for CustomCallChecker.{check,synthesize} and CallableDef.{check_call,synthesize_call}, and everything else in the same files


maximilianruesch (approved):
These are all fine, just wondering why you did not include all of the functions in the files you touched, e.g. in `CustomMonoFunctionDef` in `custom.py`.

acl-cqc:
Ah, sorry, my intention was to be consistent per subclass hierarchy, i.e. here we identify all implementations of the methods defined by CallableDef and CustomCallChecker.

Really CustomCallChecker's `check` might be better renamed differently to CheckableGenericDef's `check`, but I haven't done that here

PR #1723, opened 2026-05-07
feat!: Explicit effects system, prevent calls from annotated funcs to those with fewer effects

### LINKED ISSUE #1745: kwarg to @guppy for ordering requirements
e.g. `max_effects=[...]` or `max_resources=[...]`: a list being a subset of the available "tokens": initially, this could be the empty set, i.e. the only possible annotation would be max_effects=[]
* Identify correct name for annotation (and the things in the list) - are they effects, resources, tokens, state, globals, etc.; is the annotation only_, max_, or just effects, or something like uses or affects, etc. (If you include something in max_effects but then don't use it, for linking purposes the function still has to take the token, so it's only a "max" for what happens inside it)
* Semantics: such a function may call only functions that are also annotated with the same / a subset of effects/resources. Check in guppy compiler and raise error if not.
* All side-effecting functions - qalloc/qfree/result/panic/etc. - are unannotated, i.e. max_effects=[] functions cannot call them.

acl-cqc:
closes #1745.

However, note this is pure typechecking, and does not add any new order edges (ideally we would use the effect-system information to replace [`EXTENSION_OPS_WITH_SIDE_EFFECTS`](https://github.com/Quantinuum/guppylang/blob/82b9c770b6ac2f316fc982ce26a54e1e9811bce4/guppylang-internals/src/guppylang_internals/compiler/core.py#L416) but leaving that to the PR that solves #1746.

* The annotation takes a list of Effect, or a single Effect, or `None`. At present the only known effect is `ANY` but I expect to redefine `Effect: type = SomeEnum | ....` in #1752 if not before.
* Unannotated functions are assumed to have `[Effect.ANY]`
* Callable types can be given effects via an `@effects` modifier; without which, you get `[ANY]` by default.
   * Callables are invariant. We could perhaps be more flexible, while we are using Order edges not tokens, but that would be introducing subtyping; moreover this is consistent, and will be improved in #1760 (as this does cause some degradations).
* Nested functions are assumed to have the same effects as their containing function (for now; this could cause issues if you want to return one as a Callable return value)
* No changes made to stdlib except for int-add as I've used this is tests. The rest to follow.

BREAKING CHANGE: Functions taking Callable will be able to accept *either* only Callables with effects (the default) or (if annotated) those without, rather than both; those returning Callables will need an explicit annotation if the returnee is pure.


guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -182,6 +182,7 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
         arg_exprs: list[ast.expr] = [
             with_loc(state.node, with_type(var.ty, PlaceNode(var))) for var in arg_vars
         ]
+        # ALAN add max_effects to Tracing?
```
acl-cqc: Can this be a followup?

guppylang-internals/src/guppylang_internals/tys/__init__.py
```diff
@@ -0,0 +1,17 @@
+from collections.abc import Iterable
+from enum import Enum
+
+
+class Effect(Enum):
```
acl-cqc, line 5: At some point we'll need both an enum, and a union `Effect = EffectEnum | <variable>` (some subtype of typearg), but this is for now

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -508,6 +524,8 @@ def _to_hugr_function_type(self, ctx: ToHugrContext) -> ht.FunctionType:
         The resulting `FunctionType` can then be embedded into a Hugr `Type` or a Hugr
         `PolyFuncType`.
         """
+        # At some point we may want to represent the max_effects as input and
+        # perhaps output "token" types in Hugr, but for now we will use Order edges.
```
acl-cqc, line 591: Even Order edges not done in this PR so could end on the comma

guppylang/src/guppylang/decorator.py
```diff
@@ -88,6 +99,17 @@
 __all__ = ("GuppyKwargs", "custom_guppy_decorator", "guppy")
 
 
+class Effect(Enum):
```
acl-cqc: Not clear how much this needs to be separate from the `guppylang-internals` one. I didn't like exposing something from `guppylang_internals` as part of the public API but we do that elsewhere. Should I just `from guppylang_internals.... import Effect` and leave it at that? (I've included it in the `__all__` for this file which IIUC is like Rust `pub use` rather than `use`)

Could also move to `std/effects.py`

guppylang-internals/src/guppylang_internals/tys/printing.py
```diff
@@ -95,6 +96,11 @@ def _visit_FunctionType(self, ty: FunctionType, inside_row: bool) -> str:
         if len(ty.inputs) != 1:
             inputs = f"({inputs})"
         output = self._visit(ty.output, True)
+        arrow = (
+            "->"
+            if ty.declared_effects is None
+            else f"-{Effect.format_list(ty.declared_effects)}->"
```
acl-cqc, line 102: This suggests we may want to work harder to separate out cases where the effects are relevant from cases where they are not, and only render them when they are. (E.g. TooManyEffectsError and OverloadNoMatchError, where the effects are not even in a FunctionType; but then TypeMismatchError *does* need to render effects in detail in the mismatch is between Callables differing only in effects)...this does show up quite a bit in error messages ATM

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -470,6 +471,22 @@ def type_with_flags_from_ast(
                 flags |= InputFlags.Comptime
                 if not ty.copyable or not ty.droppable:
                     raise GuppyError(LinearComptimeError(node.right, ty))
+            case ast.Call(func=ast.Name(id="effects")) as fx:
+                if not isinstance(ty, FunctionType):
+                    raise GuppyError(InvalidFlagError(node.right))
```
acl-cqc: This InvalidFlagError is not right, it's probably a new kind of error specific to `@effects` (a construct of only very limited use i.e. in connection with Callable!)...I haven't put in tests of each case: LHS not a Callable type; multiple `@effects`; individual effects not valid.

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -470,6 +471,22 @@ def type_with_flags_from_ast(
...
+                if ty.declared_effects is not None:
+                    raise GuppyError(InvalidFlagError(node.right))
+                effects: list[Effect] = []
+                for e in fx.args:
+                    # We might want to support ast.Attribute with LHS "Effects"
+                    # and look at RHS
+                    if not isinstance(e, ast.Name):
+                        raise GuppyError(InvalidFlagError(node.right))
+                    try:
+                        effects.append(Effect.__from_str__(e.id))
```
acl-cqc: I mean, you need the relevant `import ANY` in scope to make python 3.12 happy, but 3.14 is happy even without that. Moreover - just like other type modifiers - the `effects(ANY)` doesn't even reflect *what* you've imported....

guppylang-internals/src/guppylang_internals/tys/printing.py
```diff
@@ -168,4 +176,5 @@ def signature_to_str(name: str, sig: FunctionType, has_var_args: bool = False) -
         for inp in sig.inputs
     )
     s += ", ..." if has_var_args else ""
+    # TODO Not clear how to display effects in a Python-like syntax? (skip for now)
```
acl-cqc, line 179: perhaps `def foo(x : int) -> int (with effects [])` ?

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -420,13 +421,25 @@ class FunctionType(ParametrizedTypeBase):
 
     unitary_flags: UnitaryFlags = field(default=UnitaryFlags.NoFlags, init=True)
 
+    """ Effects declared in source code (i.e. as third argument to Callable).
```
acl-cqc:
```suggestion
    """ Effects declared in source code, i.e. `Callable[...] @ effects(EFFECTS)`.
```

guppylang-internals/src/guppylang_internals/decorator.py
```diff
@@ -59,6 +59,8 @@
     from collections.abc import Callable, Sequence
     from types import FrameType
 
+    from guppylang import Effect
```
acl-cqc: This seems wrong - `guppylang_internals` is now referring to public `guppylang` - because here we have decorators used in stdlib etc. Maybe a good argument to just combine the enums in guppylang-internals and re-export?

guppylang/src/guppylang/std/qsystem/__init__.py
```diff
@@ -256,14 +256,15 @@ class Measurement:
     """Represents the result of a lazy measurement which needs to be explicitly read
     before being used."""
 
-    @custom_function(compiler=ReadFutureBoolCompiler())
+    # We do *not* model the pipeline as a side-effect
```
acl-cqc: is this a good comment? Should I just drop it?

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -182,6 +182,7 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
         arg_exprs: list[ast.expr] = [
             with_loc(state.node, with_type(var.ty, PlaceNode(var))) for var in arg_vars
         ]
+        # ALAN add effects to Tracing?
```
mark-koch: Is this resolved?

guppylang-internals/src/guppylang_internals/tys/parsing.py
mark-koch: Could you add tests for each before merging?

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -470,6 +471,22 @@ def type_with_flags_from_ast(
...
                     raise GuppyError(LinearComptimeError(node.right, ty))
+            case ast.Call(func=ast.Name(id="effects")) as fx:
+                if not isinstance(ty, FunctionType):
+                    raise GuppyError(InvalidFlagError(node.right))
+                if ty.declared_effects is not None:
+                    raise GuppyError(InvalidFlagError(node.right))
+                effects: list[Effect] = []
+                for e in fx.args:
+                    # We might want to support ast.Attribute with LHS "Effects"
+                    # and look at RHS
```
mark-koch: I think users shouldn't think of effects as an enum. In particular, libraries may declare their own effects so they can come from different namespaces

At the moment we only have ANY so this isn't an issue, but imo not worth spending a lot of time on this until we figure out how new effects should be declared

guppylang-internals/src/guppylang_internals/tys/printing.py
mark-koch: Mhmm I think this makes the error message so much worse that we shouldn't release it to users as is.

Maybe for now we could do an inversion of defaults: Regular arrow means ANY effects and have a special thing for empty effects?

guppylang-internals/src/guppylang_internals/tys/printing.py
mark-koch: Maybe `def foo(x : int) -> int @ effects(...)`?

But again, we should only print it if the effects are not the default

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -874,6 +910,12 @@ def unify(s: Type | Const, t: Type | Const, subst: "Subst | None") -> "Subst | N
         case FunctionType() as s, FunctionType() as t if s.params == t.params:
             if len(s.inputs) != len(t.inputs):
                 return None
+            if s.effects != t.effects:
```
mark-koch:
```suggestion
            if set(s.effects) != set(t.effects):
```

Also raises the question wether we should just store them as sets? On the other hand this could make error reporting non-deterministic...

guppylang/src/guppylang/decorator.py
mark-koch: I would prefer just exposing the `ANY` constant (see my comment above).

We could still have an `Effect` ABC superclass that all effects inherit from, but I wouldn't reveal that it's an enum. 

The implementation in `guppylang.std.effects` looks good to me 👍

guppylang/src/guppylang/decorator.py
```diff
@@ -99,6 +120,7 @@ class GuppyKwargs(TypedDict, total=False):
     power: bool
     max_qubits: int
     link_name: str
+    effects: list[Effect]
```
mark-koch: Can we add some validation to the user input? E.g. I could see users writing `@guppy(effects=ANY)` instead of `@guppy(effects=[ANY])`

Or maybe we should even allow this?

mark-koch: Similarly, `@guppy(effects=None)` could also seem natural?

guppylang/src/guppylang/std/angles.py
```diff
@@ -31,47 +31,47 @@ class angle:
 
     halfturns: float
 
-    @guppy
+    @guppy(effects=[])
```
mark-koch: I see you added effect annotations for angles, arithmetic and quantum. Was this needed for the tests and the rest of the standard library is follow up?

guppylang/src/guppylang/std/qsystem/__init__.py
mark-koch: Mhmm probably not too helpful...

tests/error/effects_errors/overload.err
```diff
@@ -0,0 +1,14 @@
+Error: Invalid call of overloaded function (at $FILE:24:11)
+   | 
+22 | @guppy(effects=[])
+23 | def bad_pure_func(x: float) -> float:
+24 |     return only_pure_for_int(x)
+   |            ^^^^^^^^^^^^^^^^^^^^ No variant of overloaded function `only_pure_for_int` takes
+   |                                 a `float` argument and returns `float` with effects no more
+   |                                 than []
```
mark-koch: The brackets read a bit weird.

Maybe "without effects"? Or at least we could add back ticks around the brackets?

tests/error/effects_errors/pure_calls_explicit_callable.err
```diff
@@ -0,0 +1,15 @@
+Error: Too many effects (at $FILE:8:10)
+  | 
+6 | @guppy(effects=[])
+7 | def main(impure_f: Callable[[int], int] @ effects(ANY)) -> int:
+8 |    return impure_f(5)
+  |           ^^^^^^^^ Callee of type `int -[ANY]-> int` has effects [ANY] that
+  |                    exceed those allowed
```
mark-koch: I don't think the signature is particularly helpful here and `ANY` appears twice.

How about:

```
Function `impure_f` has the effect `ANY` which is not allowed inside `main`
```

tests/error/effects_errors/pure_calls_explicit_callable.err
```diff
@@ -0,0 +1,15 @@
...
+
+Note:
+  | 
+5 | 
+6 | @guppy(effects=[])
+  |  ----------------- Allowed effects `[]` declared here
```
mark-koch: Similarly, the span already shows the effects, so maybe just:

```
Allowed effects of `main` declard here
```

mark-koch: Could even add a hint that suggests what they have to add?

tests/error/poly_errors/non_linear2.py
```diff
@@ -2,12 +2,15 @@
 
 from guppylang.decorator import guppy
 from guppylang.std.quantum.functional import h
+from guppylang.std.effects import effects
 
 T = guppy.type_var("T")
 
 
+# Pending https://github.com/Quantinuum/guppylang/issues/1752 we need to explicitly
+# declare the effects of `f`
```
mark-koch: I don't see the connection to #1752. Isn't this just because of invariance?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1497,6 +1517,9 @@ def to_bool(node: ast.expr, node_ty: Type, ctx: Context) -> tuple[ast.expr, Type
         return node, node_ty
     synth = ExprSynthesizer(ctx)
     exp_sig = FunctionType([FuncInput(node_ty, InputFlags.Inout)], bool_type())
+    # When we have effect variables, we should use an existential
+    # variable upper-bounded by those allowed in the context.
+    exp_sig = exp_sig.with_effects([])
```
mark-koch: Have you annotated all `__bool__` functions in the standard library?

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -159,7 +160,38 @@ def check_global_func_def(
...
+        max_effects_from=max_effects_from,
+    )
+
+
+def _find_guppy_decorator(decorators: list[ast.expr]) -> ast.expr | None:
+    for d in decorators:
+        if (
+            isinstance(d, ast.Call)
+            and isinstance(d.func, ast.Name)
+            and d.func.id == "guppy"
```
mark-koch: This sort of string matching will break for e.g. guppy-ft where they define their own decorators that internally call guppy

mark-koch: Don't know what the best solution is, but this seems too fragile. Especially if you're throwing an `InternalGuppyError` when the decorator is not found

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -168,7 +200,13 @@ def check_nested_func_def(
     ctx: Context,
 ) -> CheckedNestedFunctionDef:
     """Type checks a local (nested) function definition."""
-    func_ty = check_signature(func_def, ctx.globals)
+    # For now we assume the nested function has the same effects as that enclosing.
+    # We could do better by allowing a separate annotation (rather than a parameter
+    # to @guppy), but we will wait for callgraph analysis to compute precisely:
+    # nested functions are not part of any public API, so changes are not breaking.
```
mark-koch, line 189: Does this guarantee that programs that compile with this version will still compile once we do callgraph analysis?

guppylang-internals/src/guppylang_internals/decorator.py
mark-koch: Yeah this seems wrong. At the very least we should guard behind `if TYPE_CHECKING`

guppylang-internals/src/guppylang_internals/tracing/function.py
mark-koch: Looks like no. I think they need to be added to the `Context` in the line below. Otherwise effects won't be checked inside comptime functions.

Speaking of, would be good to add some tests for comptime functions

mark-koch (commented):
Nice! The implementation looks good to me, just some questions and suggestions for the interface and error messages

guppylang-internals/src/guppylang_internals/tys/printing.py
acl-cqc: Ok will have a think about hiding them (perhaps show them in TypeMismatchError?).

As to the latter, currently I have the printing distinguish between explicitly declared effects and implicit (so `-[]->` is if you've said there are none, `->` is if you didn't say anything, `-[ANY]->` is if you explicitly said ANY)....you think it's better to display the last as `->` ?

guppylang/src/guppylang/decorator.py
acl-cqc: Not sure which comment you are referring to, can you link?

`guppylang.std.effects` defines only the `@ effects` annotation, not the type of Effect...?

(in common with `owned`, `effects` is an alias for a `class Effects`, but perhaps the latter would be better as `class EffectsModifier` or `class EffectsAnnotation` or something to better distinguish `Effects` from `Effect`) ....?

guppylang/src/guppylang/decorator.py
acl-cqc: Yeah I don't see why we shouldn't allow.

Although, I am also wondering whether for uniformity we should allow `@guppy @ effects` (using the `__rmatmul__` to modify the thing on the left)...in which case should we allow the `effects=` kwargs as well (as alternative option, not both in same declaration) ?

guppylang-internals/src/guppylang_internals/checker/func_checker.py
acl-cqc: Not if they do things with Callable, e.g.
```
@guppy(effects=[ANY])
def outer() -> Callable[[int], int, ANY]:
   result(...) # Just to justify the ANY on the outer function, not otherwise relevant
   def inner(x: int) -> int:
      return x + 1
  return inner
```
callgraph analysis figures out that inner is pure, but Callable is invariant.

However with #1760 *and existentials for the effects of the returned Callable* then you could make it work if you declared `def outer() -> Callable[[int],int]`

guppylang-internals/src/guppylang_internals/checker/func_checker.py
acl-cqc: Ok yes good point about custom decorators. Tricky to see what to do though.
* It seems hard to identify a custom-decorator (that calls guppy) from the AST, there's so little to go on (we haven't even identified the custom-decorator definition/implementation that the `ast.Call` points to!).
* When the decorator executes, it calls `guppy`. So the latter could be an opportunity to put something into the e.g. `GuppyFunctionDefinition`. However at this point it's hard to identify which bit of AST lead to the call to guppy.
    * Best hack I can see might be for `guppy` to see the source file of the decorated function, look up the call stack (for callers to guppy) until it finds one in that source file (?!) and note that somehow. Doesn't work well as if the definition of the custom decorator is in the same source file as it's use...could we look up the call stack (calling `guppy`) until we find something that's listed as a decorator of the AST node for the function that `guppy` is running on? That seems hard, and possibly fragile, but I can have a go...[maybe in another PR?]
* Alternatively we make to make custom decorators do some work themselves (some will have to add an `effects=` parameter too but custom decorators might be pure-only or some such)...even this I'm not quite clear on though; we could add an extra `effects_declared_at` parameter to `@guppy` and get them to fill that in, but even finding the AST node to pass to that parameter could be quite hard for them.
* Would `@guppy @ effects(...)` help? (Would it even work? Perhaps if that was not using `__rmatmul__` but if `@effects` was its own decorator....then we could spot it independently of the `@guppy` / `@custom`, although it wouldn't work well if the custom-decorator specified the effects itself)

Least improvement is to fall back to the `def` line or the whole declaration if we can't find a decorator, rather than InternalError - I'll certainly do that.

acl-cqc: If we can't find the decorator, then the span of the FunctionDef includes the function body but not the decorator list, which isn't much good - so even just this fallback was more awkward than expected. But done. (We can make an issue for providing some way for custom decorators to override that and do better but for another PR?)

tests/error/poly_errors/non_linear2.py
acl-cqc: Sorry, I meant #1760, which follows on from that

guppylang-internals/src/guppylang_internals/tys/ty.py
acl-cqc: Yes excellent spot, given this is not realisable in a test at this point, but done. I think storing as lists is the way to go for determinism and closeness-to-source-code, but the comparison should be as `set`

tests/error/effects_errors/pure_calls_explicit_callable.err
acl-cqc: Yeah ok, so if we actually find a decorator, we can omit the explicit mention of what's allowed, nice idea....done.

tests/error/effects_errors/pure_calls_explicit_callable.err
acl-cqc: Yes. Got there in the end. Quite a few cases, as `int --> int` needs the "inferred" (atm, assumed ANY) effects adding explicitly.

guppylang/src/guppylang/decorator.py
acl-cqc: Added `@guppy(effects=None)` and `@guppy(effects=ANY)` for now. Let's have a think about using `@ effects` for decls too (not just Callable types) and see if that helps at all.

guppylang/src/guppylang/std/angles.py
acl-cqc: I think I'm gonna try to *undo* the library changes for now (except `__bool__` which will be required) and hope to reinstate in the followup which adds all the token types.

tests/error/effects_errors/overload.err
acl-cqc: Added backticks. Not sure special-casing "without effects" is worth it?

guppylang-internals/src/guppylang_internals/decorator.py
acl-cqc: This was actually in an `if TYPE_CHECKING` but I've changed it to using `guppylang_internals.tys.Effect` i.e. the original source of the re-export

guppylang-internals/src/guppylang_internals/tracing/function.py
acl-cqc: Yes done

guppylang-internals/src/guppylang_internals/tys/errors.py
```diff
@@ -172,6 +172,28 @@ class ComptimeArgShadowError(Error):
...
+    title: ClassVar[str] = "Invalid annotation"
+    span_label: ClassVar[str] = (
+        "Effects have already been applied to this Callable type"
+    )
+
+
+@dataclass(frozen=True)
+class InvalidEffectError(Error):
+    title: ClassVar[str] = "Invalid annotation"
+    span_label: ClassVar[str] = "Not a valid effect: {arg}"
```
mark-koch:
```suggestion
    span_label: ClassVar[str] = "Not a valid effect: `{arg}`"
```

tests/error/effects_errors/comptime_pure_calls_impure.err
```diff
@@ -1,15 +1,6 @@
...
-7 | @guppy(effects=[])
-  |  ----------------- Allowed effects declared here
-
-Guppy compilation failed due to 1 previous error
+Traceback (most recent call last):
+  File "$FILE", line 11, in <module>
+    main.compile()
+  File "$FILE", line 9, in main
+    return impure_func(5)
+guppylang_internals.error.GuppyComptimeError: Too many effects: Callee of type `int -> int` has effects `[ANY]` not allowed inside `main`
```
mark-koch, line 6: Any idea why the message does not reference `impure_func`? Is it because your looking for `ast.Name` to determine the callee?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1313,21 +1313,25 @@ def check_comptime_arg(
...
+        show_effects_allowed = mf.effects
         if isinstance(mf.decl, ast.expr):
             # We found the decorator that is the source of the effect constraint,
             # which will contain the allowed effects as an explicit argument
-            effects_allowed = None
+            show_effects_allowed = None
         # Otherwise, the error message points at all decorators, which may or may not
         # list the allowed effects, so list them explicitly
+
+        callee = loc_node.id if isinstance(loc_node, ast.Name) else func_ty
```
mark-koch, line 1347: Instead of determining the callee this way, maybe accept an optional `DefId` to `_check_effects`? Or the `CallableDef` directly? I guess it would need to be threaded through `synthesize_call` though :/

guppylang-internals/src/guppylang_internals/tys/errors.py
```diff
@@ -172,6 +172,28 @@ class ComptimeArgShadowError(Error):
     arg: str
 
 
+@dataclass(frozen=True)
+class EffectsNotApplicableError(Error):
+    title: ClassVar[str] = "Invalid annotation"
+    span_label: ClassVar[str] = "Effects may be applied only to a Callable type"
```
mark-koch:
```suggestion
    span_label: ClassVar[str] = "Effects may be applied only to a `Callable` type"
```

guppylang-internals/src/guppylang_internals/tys/errors.py
```diff
@@ -172,6 +172,28 @@ class ComptimeArgShadowError(Error):
...
+
+
+@dataclass(frozen=True)
+class EffectsRepeatedError(Error):
+    title: ClassVar[str] = "Invalid annotation"
+    span_label: ClassVar[str] = (
+        "Effects have already been applied to this Callable type"
```
mark-koch:
```suggestion
        "Effects have already been applied to this `Callable` type"
```

guppylang/src/guppylang/std/angles.py
mark-koch: Are you stil panning to undo this?

mark-koch (approved):
Nice, love the new error messages 👍

tests/error/effects_errors/comptime_pure_calls_impure.err
acl-cqc: Yes, as per your [other comment](https://github.com/Quantinuum/guppylang/pull/1723#discussion_r3379563074), here we have only the AST for the call itself, nothing more than that about the callee/target

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
acl-cqc: Yes, this is significantly better, but is quite a change to `check_call` / `synthesize_call`. I've raised #1846 for this...

PR #1734, opened 2026-05-12
test: Replace pytest-isolate with pytest-forked

### LINKED ISSUE #1640: [Maintenance]: Decoupling tests time out flake
### Description

The recently added decoupling tests were also recently modified to be isolated Python processes. Since the decorator required a timeout to be specified, these tests sometimes run into timeouts and thus fail on unrelated changes. E.g. on https://github.com/Quantinuum/guppylang/actions/runs/24074680707/job/70220008134.

We should investigate the timeouts and decide on further changes after.

### Motivation

_No response_

### Scope

_No response_

### Component

CI/Build System

### Priority

High - Important for project health

### Additional Context

_No response_

maximilianruesch:
Replace `pytest-isolate` with `pytest-forked`. Neither is in a good maintenance status, but `pytest-forked` is significantly simpler and works without supplying any options (such as timeouts). Thus, forked processes should now run without timeouts entirely. We need to observe how this changes the (flaky) #1640, but it is a worthwhile shot.

Closes #1640


aborgna-q (approved):
Thanks!

PR #1706, opened 2026-04-29
feat(guppy_internals)!: Allow indexes access and assignment in `dagger` contex

nicolaassolini-qntm:
closes #1703 
closes #1639

* Removed the explicit check and error for assignments within dagger contexts in `check_modified_block` and `check_invalid_under_dagger`, allowing assignments under dagger.
* Removed check for use of indexing inside dagger context.
* Removed tests and error files that checked for assignment errors under dagger, as these are no longer relevant.
* Updated existing error tests on valid dagger/unitary constraint violations.

BREAKING CHANGE: removed `visit_PlaceNode` in `guppylang_internals/checker/unitary_checker.py::BBUnitaryChecker` .

BEGIN_COMMIT_OVERRIDE
feat: Allow index access and assignment in dagger contexts (#1706)
END_COMMIT_OVERRIDE

PR #1738, opened 2026-05-14
feat(guppy_internals)!: Allow indexes access and assignment in dagger contex, fixing use of  array subscript as argument of control modifiers

nicolaassolini-qntm:
cherry pick pr #1706 and #1708 

Pr #1708: 
Fixes handling of array subscripts used as control modifier arguments.

- Call `check_place_assignable` on control arguments statement checking, so subscript controls get a synthesised `setitem_call`.
- Fix modifier HUGR generation so captured arguments are read after control arguments are compiled, avoiding stale array wires. 
- Return controller qubits borrowed from array subscripts back into their parent arrays after the modified indirect call. Use the synthesised setter during the HUGR generation to write returned control values back into their parent array slots.

- Updated integration test.

Pr #1706:
* Removed the explicit check and error for assignments within dagger contexts in `check_modified_block` and `check_invalid_under_dagger`, allowing assignments under dagger.
* Removed check for use of indexing inside dagger context.
* Removed tests and error files that checked for assignment errors under dagger, as these are no longer relevant.
* Updated existing error tests on valid dagger/unitary constraint violations.

BREAKING CHANGE: removed `visit_PlaceNode` in `guppylang_internals/checker/unitary_checker.py::BBUnitaryChecker` .


acl-cqc (approved):
Looks good thanks @nicolaassolini-qntm

Issue #1396, opened 2025-12-17
Enforce protocol implementations needing to be copyable/droppable


tatiana-s:
See https://github.com/Quantinuum/guppylang/pull/1275#discussion_r2603508425

PR #1754, opened 2026-05-19
feat!: Protocol parsing

### LINKED ISSUE #1675: Refactor protocol parsing to expect `guppylang.declare`
(In the `checking-fresh` branch, currently)

The protocols implementation expects the definitions in the protocol body to be unadorned, as in:
```python
@guppy.protocol
class Proto:
    def foo(s: S) -> T: ...
    def bar(t: T) -> None: ...
```
but we would like instead:
```python
@guppy.protocol
class Proto:
    @guppy.declare
    def foo(s: S) -> T: ...
    @guppy.declare
    def bar(t: T) -> None: ...
```

croyzor:
Closes #1675, #1396, #1653 

BREAKING CHANGE:
(guppy-internals)
* Add an extra arg to `NonGuppyMethodError` to say which annotation needs added
* Add extra arg to `try_parse_generic_base` to specify whether we're looking for a `Generic` super or a `Protocol`
* Add a parameter to type vars saying which protocols they must implement
* Add an argument to `type_check_args`

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/protocol.py
+++ b/guppylang-internals/src/guppylang_internals/definition/protocol.py
@@ -1,0 +1,189 @@
+import ast
+import sys
+from collections.abc import Mapping, Sequence
+from dataclasses import dataclass, field
+from typing import TYPE_CHECKING, ClassVar
+
+from guppylang_internals.ast_util import AstNode
+from guppylang_internals.checker.core import Globals
+from guppylang_internals.checker.errors.generic import (
+    UnexpectedError,
+    UnsupportedError,
+)
+from guppylang_internals.definition.common import (
+    CheckableDef,
+    CompiledDef,
+    DefId,
+    Definition,
+    ParsableDef,
+)
+from guppylang_internals.definition.declaration import ParsedFunctionDecl
+from guppylang_internals.definition.util import (
+    NonGuppyMethodError,
+    RedundantParamsError,
+    params_from_ast,
+    parse_py_class,
+    try_parse_generic_base,
+)
+from guppylang_internals.diagnostic import Help
+from guppylang_internals.engine import DEF_STORE, ENGINE
+from guppylang_internals.error import GuppyError
+from guppylang_internals.span import SourceMap, Span, to_span
+from guppylang_internals.tys.arg import Argument
+from guppylang_internals.tys.param import Parameter, check_all_args
+from guppylang_internals.tys.protocol import ProtocolInst
+from guppylang_internals.tys.ty import FunctionType
+
+if TYPE_CHECKING:
+    from guppylang_internals.diagnostic import Error
+
+if sys.version_info >= (3, 12):
+    from guppylang_internals.tys.parsing import parse_parameter
+
+
+@dataclass(frozen=True)
+class ProtocolDef(Definition):
+    """Abstract base class for protocol definitions."""
+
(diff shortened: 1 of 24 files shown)


guppylang/src/guppylang/decorator.py
```diff
@@ -311,6 +312,57 @@ def decorator(
...
+        for val in cls.__dict__.values():
+            if isinstance(val, GuppyDefinition):
+                DEF_STORE.register_type_member(defn.id, val.wrapped.name, val.id)
+        # We need the the `__firstlineno__` attribute to look up the source later.
+        cls = _set_firstlineno(cls, frame)
+        # We're pretending to return the class unchanged, but in fact we return
+        # a `GuppyDefinition` that handles the comptime logic
+        return GuppyDefinition(defn)  # type: ignore[return-value]
+
+    def require(
```
croyzor, line 340: This is just copy-pasted from the definition of `@guppy.declare`. There's currently no way to tell the difference!

tests/error/errors_on_usage/protocols.err
```diff
@@ -0,0 +1,8 @@
+Error: Unsupported (at $FILE:20:8)
+   | 
+18 | @guppy
+19 | def main() -> None:
+20 |     bar(FooStruct())
+   |         ^^^^^^^^^^^ Protocol checking is not supported
```
qartik, line 6: Is this a PR scope issue? Ie, checking protocols is future work?

guppylang-internals/src/guppylang_internals/definition/protocol.py
```diff
@@ -0,0 +1,189 @@
...
+                    py_func = getattr(self.python_class, name, None)
+
+                    if not isinstance(py_func, GuppyDefinition):
+                        raise GuppyError(
+                            NonGuppyMethodError(
+                                node, self.name, name, "protocol", "@guppy.require"
+                            )
+                        )
+                    assert isinstance(py_func, GuppyDefinition)
+                    func_defs[name] = py_func.id
```
qartik, line 131:
```suggestion
                    func_defs[name] = py_func.id
                    if not has_empty_body(node):
                        raise GuppyError(BodyNotEmptyError(node.body[0], name))
```

`@guppy.require` is still a function declaration, so its empty-body invariant can be enforced here.

qartik (commented):
One question about scope, one small suggestion. Otherwise, LGTM.

guppylang-internals/src/guppylang_internals/definition/protocol.py
qartik: also consider adding a test for this case.

tests/error/errors_on_usage/protocols.err
croyzor: Yes, exactly!

guppylang-internals/src/guppylang_internals/definition/protocol.py
croyzor: This invariant will already be enforced by the parser for `RawFunctionDecl`. There's a test for this in the incoming follow-up PR 🙂

qartik (approved):

guppylang/src/guppylang/decorator.py
croyzor: I've created #1764

Issue #1764, opened 2026-05-22
Differentiate between `@guppy.require` and `@guppy.declare`


croyzor:
As of #1754, these are the same thing which means we can't raise an error when we see
```
@guppy.protocol
class Foo
    @guppy.declare
    def foo ...
```
or
```
@guppy.struct
class Foo
    @guppy.require
    def foo ...
```

PR #1765, opened 2026-05-22
feat!: Protocol checking

### LINKED ISSUE #1210: Integrate initial protocol checking code into guppy

### LINKED ISSUE #1396: Enforce protocol implementations needing to be copyable/droppable

### LINKED ISSUE #1653: [Bug]: Misleading error for function type args

croyzor:
Closes #1396, closes #1210, closes #1653

---

BREAKING CHANGE: (guppylang-internals) Add `ProtocolCall` option to `AnyCall`


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -336,11 +337,7 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
             # protocol definition itself first.
             if isinstance(defn, ParsedProtocolDef):
                 assert isinstance(func_ty, FunctionType)
-                raise GuppyError(
-                    UnsupportedError(
-                        node.func, "Checking protocol method calls", singular=True
-                    )
-                )
+                raise RequiresMonomorphizationError
```
mark-koch: Why would `defn` ever be a `ParsedProtocolDef`?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -624,6 +621,28 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                proto_def = ENGINE.get_checked(proto.def_id, proto.type_args)
+                assert isinstance(proto_def, CheckedProtocolDef)
+                for member_name in proto_def.member_defs:
+                    member_ty = proto_def.member_sig(member_name)
+                    if node.attr == member_name:
+                        name_node = with_type(
+                            member_ty,
+                            with_loc(
+                                node,
+                                GlobalName(id=member_name, def_id=proto.def_id),
```
mark-koch: Returning a global name with the id of the full protocol feels a bit weird. Don't we want some sort of identifier for which method of the protocol is used?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
mark-koch: Ah I see because attribute accesses are turned into Global names with the protocol id.

I think it would be better to add a new node type that holds a protocol id and the name of the method that is used. Then we wouldn't need to cop out here

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -385,7 +384,7 @@ def parse_self_arg(arg: ast.arg, self_defn: TypeDef, ctx: TypeParsingCtx) -> Fun
         [param.to_existential()[0] for param in self_defn.params]
     )
     self_ty_placeholder = ExistentialTypeVar.fresh(
-        "Self", copyable=self_ty_head.copyable, droppable=self_ty_head.droppable
+        "Self", copyable=True, droppable=True
```
mark-koch: Why this change?

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -470,6 +474,26 @@ def parse_self_arg_proto(
         )
 
 
+def handle_implicit_proto_self_arg(
```
mark-koch, line 478: Could you add a docstring explaining what this function does

guppylang-internals/src/guppylang_internals/checker/protocol_checker.py
```diff
@@ -0,0 +1,180 @@
...
+    BoundTypeVar,
+    FunctionType,
+    Type,
+    unify,
+    unify_type_args,
+)
+
+
+@dataclass(frozen=True)
+class ImplProofBase(ABC):
```
mark-koch, line 33: Also docstrings for the things in this file

guppylang-internals/src/guppylang_internals/checker/protocol_checker.py
```diff
@@ -0,0 +1,180 @@
...
+    from guppylang_internals.engine import ENGINE
+
+    protocol = ENGINE.get_parsed(proto_inst.def_id)
+    raise GuppyError(FirstArgNotProtocol(None, protocol.name))
+
+
+def check_protocol(
+    ty: Type, protocol: ProtocolInst, loc: AstNode | None = None
+) -> tuple[ImplProof, Subst]:
+    """Check that `ty` implements `protocol`"""
```
mark-koch: Can you also mention that `protocol` may contain existential variables that will all (?) be solved by the returned substitution?

guppylang-internals/src/guppylang_internals/checker/protocol_checker.py
```diff
@@ -0,0 +1,180 @@
...
+    subst = {}
+    member_impls: dict[str, tuple[DefId, Inst]] = {}
+    for name in protocol_def.member_defs:
+        proto_sig = protocol_def.member_sig(name)
+        if len(proto_sig.inputs) > 0:
+            if isinstance(proto_sig.inputs[0].ty, BoundTypeVar):
+                proto_sig = _instantiate_self(proto_sig, protocol, ty)
+            else:
+                raise GuppyError(FirstArgNotProtocol(None, protocol_def.name))
+                proto_sig = _instantiate_self(proto_sig, protocol, ty)
```
mark-koch: Unreachable

guppylang-internals/src/guppylang_internals/tys/param.py
```diff
@@ -133,9 +135,16 @@ def check_arg(
...
                 if self.must_implement:
-                    raise GuppyError(
-                        UnsupportedError(loc, "Protocol checking", singular=True)
+                    from guppylang_internals.checker.protocol_checker import (
+                        check_protocol,
                     )
+
+                    for proto in self.must_implement:
+                        prfs = [prf for prf in arg.proto_proofs if proto == prf.proto]
+                        if len(prfs) == 0:
```
mark-koch:
```suggestion
                        if all(proto != prf.proto for prf in arg.proto_proofs)
```
?

mark-koch: Also a comment explaining when we would expect the `proto_proofs` table to be populated could be useful

guppylang-internals/src/guppylang_internals/tys/param.py
```diff
@@ -150,7 +159,7 @@ def to_existential(self) -> tuple[Argument, ExistentialVar]:
             self.name,
             self.must_be_copyable,
             self.must_be_droppable,
-            tuple(self.must_implement),
+            tuple(impl for impl in self.must_implement),
```
mark-koch: This doesn't change anything?

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -898,6 +898,13 @@ def _unify_type_var(var: ExistentialTypeVar, t: Type, subst: "Subst") -> "Subst
         return unify(var, subst[t], subst)
     if var in t.unsolved_vars:
         return None
+    # Check that `t` implements all protocols required by `var`.
+    if var.implements:
+        from guppylang_internals.checker.protocol_checker import check_protocol
+
+        for proto in var.implements:
+            _, proto_subst = check_protocol(t, proto)
```
mark-koch: Mhmm this would raise a user error if the protocol is not implemented. Usually, we return `None` from unification to indicate failure. Is this not a problem?

tests/error/protocol_errors/self_inferred_type.err
```diff
@@ -0,0 +1,9 @@
+Error: Unsupported (at $FILE:6:12)
+  | 
+4 | class MyProto:
+5 |     @guppy.require
+6 |     def foo(self) -> None: ...
+  |             ^^^^ Inference of type for `self` is not supported in protocol
+  |                  methods
```
mark-koch, line 7: Imo better to use the regular `MissingArgAnnotationError` error (maybe with a note that self inference on protcols is not supported)

tests/error/protocol_errors/self_no_self_arg.err
```diff
@@ -0,0 +1,8 @@
+Error: Invalid self annotation (at $FILE:7:15)
+  | 
+5 | class Foo:
+6 |     @guppy.require
+7 |     def foo(n: nat) -> None: ...
+  |                ^^^ `n` must be of type `self`
```
mark-koch: `Self` (captitalised) or `Foo`?

tests/error/protocol_errors_py312/doesnt_implement.err
```diff
@@ -0,0 +1,8 @@
+Error: Missing protocol implementation (at $FILE:25:22)
+   | 
+23 | @guppy
+24 | def baz(f: FooInt) -> None:
+25 |     return eat_foostr(f)
+   |                       ^ Type `FooInt` does not implement protocol `Foo`
```
mark-koch: Shouldnt it be "does not implement protocol `Foo[str]`"?

tests/error/protocol_errors_py312/no_impl.err
```diff
@@ -0,0 +1,8 @@
+Error: Missing protocol implementation (at $FILE:7:15)
+  | 
+5 | class Proto[T]:
+6 |     @guppy.require
+7 |     def foo[T](t: T) -> T: ...
+  |                ^^^^ Type `T` does not implement protocol `Proto`
```
mark-koch: A nice future improvement would be to add notes explaining why the protocol is not implemented

tests/integration/test_protocol_py312.py
```diff
@@ -0,0 +1,412 @@
...
+        # def bar(self: Self) -> "MyProto": ...
+
+        @guppy.require
+        def baz(self, y: int) -> int: ...
+
+    MyProto.check()
+
+
+## TODO: See if this can work in light of the monomorphisation changes assuming
+## check and compile are called on entrypoints
```
mark-koch: Are these TODOs resolved?

mark-koch (commented):
🔥 🔥 🔥

Sorry I didn't have the time to look through everything in detail yet, but here are some comments from a first read through. Will have a closer look next week

tests/error/protocol_errors_py312/no_impl.err
croyzor: True, created #1779

tests/error/protocol_errors/self_no_self_arg.err
croyzor: Fixed in https://github.com/Quantinuum/guppylang/pull/1765/commits/06cf7bc20b175ea2faa8f7cdf8cbc8440783a339

tests/error/protocol_errors_py312/doesnt_implement.err
croyzor: Fixed in https://github.com/Quantinuum/guppylang/pull/1765/commits/927b4c9b9d5cc1d98ea7cbb50b72e17787e3ca52

guppylang-internals/src/guppylang_internals/checker/func_checker.py
croyzor: Oops! reverted in https://github.com/Quantinuum/guppylang/pull/1765/commits/9b9baad9f3c6bbac3ca42cb43392deda0ac76137

guppylang-internals/src/guppylang_internals/tys/param.py
croyzor: looking at it now, it seems like the `proto_proofs` will never be populated...

croyzor: unless we have something like `implements Protocol[concrete], Protocol[T]` and `T := concrete`

tests/error/protocol_errors/self_inferred_type.err
croyzor: Sure, but I wont bother changing it since there's not #1770

guppylang-internals/src/guppylang_internals/tys/ty.py
croyzor: I think potentially yes, but it depends on how we utilise unification. If we only do it when we actually expect things to be equal, and raise an error on `None`, then this is just giving a better error message, compared to a generic `S /= T` that doesn't mention protocols

guppylang-internals/src/guppylang_internals/checker/protocol_checker.py
croyzor: done in https://github.com/Quantinuum/guppylang/pull/1765/commits/0c97c61d5871d8776d09ef8420629c292b2cad99

guppylang-internals/src/guppylang_internals/tys/param.py
croyzor: fair, undone

guppylang-internals/src/guppylang_internals/tys/ty.py
mark-koch: Ok I'm not sure about that, `unify` is used in a lot of places...

For example, we use `unify` to determine if an implementation matches a protocol signature. If it does not, we want to give a specific error, not the one from `unify`?

guppylang-internals/src/guppylang_internals/tys/param.py
mark-koch: Ok, so should we just remove `proto_proofs`?

guppylang-internals/src/guppylang_internals/tys/ty.py
croyzor: Good point, [I've wrapped it in a try-except block](https://github.com/Quantinuum/guppylang/pull/1765/commits/d132df80e7c17e5f82ad5ec89e0348168f7bd1d0), but it's not the most satisfying solution...

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
croyzor: Okay, brought back `ProtocolCall` here instead of raising a monomorphisation error

guppylang-internals/src/guppylang_internals/tys/arg.py
```diff
@@ -50,6 +50,9 @@ def unsolved_vars(self) -> set[ExistentialVar]:
     def bound_vars(self) -> set[BoundVar]:
         """The bound type variables contained in this argument."""
 
+    @abstractmethod
+    def __str__(self) -> str: ...
```
mark-koch, line 54: You can use the `TypePrinter` here similar to `TypeBase.__str__`

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -624,6 +631,28 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
                 # you loose access to all fields besides `a`).
                 expr = FieldAccessAndDrop(value=node.value, struct_ty=ty, field=field)
             return with_loc(node, expr), field.ty
+        elif isinstance(ty, BoundTypeVar):
+            from guppylang_internals.definition.protocol import CheckedProtocolDef
+
+            for proto in ty.implements:
+                proto_def = ENGINE.get_checked(proto.def_id, proto.type_args)
+                assert isinstance(proto_def, CheckedProtocolDef)
+                for member_name, member_id in proto_def.member_defs.items():
```
mark-koch, line 641: Do we need to check if there is a unique method that matches? I.e. what happens if a variable has two bounds that provide a method with the same name?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -624,6 +631,28 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                    member_ty = proto_def.member_sig(member_name)
+                    if node.attr == member_name:
+                        name_node = with_type(
+                            member_ty,
+                            with_loc(
+                                node,
+                                GlobalName(id=member_name, def_id=member_id),
```
mark-koch, line 648: Ohh I didn't know that the `@guppy.require` functions have their own id?

Looks like they are all mapped to the `ProtocolDef` that defined them? This feels a bit weird since now an id `x` can be mapped to a definition whose own id is not `x`?

I can think of two possible alternatives:
* Make a new unique definition object for each `@guppy.require`. They could be `CallableDef`s, in which case the special case for `ProtocolCall` above could be removed again
* Or, stop giving ids to `@guppy.require` functions and return something different than `GlobalName` here. Maybe something like `ProtocolMember(proto_id, member_name)` and then check for that in `visit_Call` above?

What do you think?

tests/error/protocol_errors_py312/no_impl.err
```diff
@@ -0,0 +1,8 @@
+Error: Missing protocol implementation (at $FILE:7:15)
+  | 
+5 | class Proto[T]:
+6 |     @guppy.require
+7 |     def foo[T](t: T) -> T: ...
+  |                ^^^^ Type `T` does not implement protocol `Proto[?T]`
```
mark-koch, line 6: What happens if I write `def foo[T: Proto[T]](t: T) -> T`? I guess `T` is undefined in the bound?

mark-koch (commented):
Thanks, looks great! Only remaining doubt is re using `GlobalName` to detect protocol calls

tests/error/protocol_errors_py312/no_impl.err
croyzor: Exactly right 🙂

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
croyzor: I'm not sure what you mean by
> an id x can be mapped to a definition whose own id is not x?

but your solutions sound neater anyway. It sounds like either of them would also help address #1764

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
croyzor: Good point! Currently we'll match the method from the protocol that comes first in the list of bounds on the type variable and raise a type mismatch error even if a later protocol bound has the correct method type!

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
mark-koch: Ahh I didn't realise you're building a `RawFunctionDecl` when doing `@guppy.require`.

Yeah I think it would be good to stop doing that and go with one of the two options

croyzor: You suggest
> Make a new unique definition object for each @guppy.require. 

The `RawFunctionDecl` in this case _is_ the definition object. It could be something else, sure...

> Or, stop giving ids to @guppy.require functions and return something different than GlobalName here.

Trying to do this reminded me why I didn't do it in this PR! It gets surprisingly messy if you try to check the protocol methods in `CheckableProtocol.check`. `check_signature` wants to get a `CheckedProtocol` but we're not finished yet!

mark-koch: > check_signature wants to get a CheckedProtocol but we're not finished yet!

Ah I see, yes that would be tricky

> The RawFunctionDecl in this case is the definition object. It could be something else, sure...

Yeah I didn't realise you were already doing that. I still thing it should be its own kind of definition, not a decl. But I see you already have #1764, so ok if you want to address this in a different PR

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
croyzor: I'm working on this, but I think it's a separable issue, so I've added #1856

guppylang-internals/src/guppylang_internals/tys/arg.py
```diff
@@ -86,7 +86,9 @@ def transform(self, transformer: Transformer) -> Argument:
         return transformer.transform(self) or TypeArg(self.ty.transform(transformer))
 
     def __str__(self) -> str:
-        return str(self.ty)
+        from guppylang_internals.tys.printing import TypePrinter
+
+        return TypePrinter().visit(self)
```
mark-koch, line 91: You can move this into `ArgumentBase` and remove the individual `__str__` implementations (`TypePrinter` can handle `Argument`s)

mark-koch (approved):

guppylang-internals/src/guppylang_internals/tys/arg.py
croyzor: Going to leave this because it confuses mypy that `TypePrinter.visit` accepts `Argument` and not `ArgBase`. My instinct is to delete the alias and rename `ArgBase` to `Argument`. What do you think?

PR #1768, opened 2026-05-27
feat!: Higher order functions with unitary modifiers

### LINKED ISSUE #1637: [Feature]: Allow unitary flag with higher order function
### Problem Statement

given:
```python
@guppy(dagger=True)
def test_ho(f: Callable[[qubit], None], q: qubit) -> None:
```
There is no way to specify flags for `f`

see: https://github.com/Quantinuum/guppylang/blob/e3ac40fcf017abec8814988edf270b35d4a4a654/tests/error/modifier_errors/higher_order.py

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

nicolaassolini-qntm:
closes #1637

This PR allows higher-order function parameters to specify which unitary modifier contexts they support.

Changes:
- Adds callable aliases for higher-order functions with unitary capabilities:
  - `Unitary[[...], R]`
  - `Daggerable[[...], R]`
  - `Powerable[[...], R]`
  - `Controllable[[...], R]`
  - `PowerControllable[[...], R]`
- Checks callable unitary flags when passing functions as higher-order arguments.
- Updates unitary diagnostic rendering and decorator hints.
- Adds integration and error tests for higher-order callable flag matching.

BREAKING CHANGE: `UnitaryCallError.render_flags` was removed.
BREAKING CHANGE: `CallableTypeDef.name` changed from `field(default='Callable', init=False)` to `field(default='Callable', kw_only=True)`.
BREAKING CHANGE: `CallableTypeDef.__init__(name)` changed from positional-or-keyword to keyword-only, `CallableTypeDef.__init__(defined_at)` positional parameter moved from position 3 to position 2, `CallableTypeDef.__init__(params)` positional parameter moved from position 4 to position 3.
BREAKING CHANGE: Added the field `missing_flags` to `UnitaryCallError.Hint` 

BEGIN_COMMIT_OVERRIDE
feat: Higher order functions with unitary modifiers (#1768)
END_COMMIT_OVERRIDE


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1066,6 +1070,12 @@ def check_type_against(
         if coerced := try_coerce_to(act, exp, node, ctx):
             return coerced, {}, ()
         raise GuppyTypeError(TypeMismatchError(node, exp, act, kind))
+
+    # If we have a function type, we also check that unitary flags match
+    if isinstance(act, FunctionType):
+        assert isinstance(exp, FunctionType)
```
doug-q, line 1076: Are you sure you want to leave these asserts in?

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -197,10 +204,24 @@ def _arg_from_instantiated_defn(
...
 
     match defn:
-        # Special case for the `Callable` type
-        case CallableTypeDef():
-            return TypeArg(_parse_callable_type(arg_nodes, node, ctx))
-            # Special case for the `Callable` type
+        # Special cases for the `Callable` type
+        case CallableTypeDef(name=name):
+            # arguments of the python Callable
+            if name == "Callable":
```
doug-q: This is gross, is there no better way?

doug-q (commented):
Thanks, looks mostly good. I am not convinced this works with callables in both positive and negative position. (https://cs.stackexchange.com/questions/42150/meaning-of-positive-position-and-negative-position-in-type-theory). I expect "is_weaker_than" to need to swap the order of it's arguments in different polarities.

Could you add tests for cases like
* Returning a callable stronger/weaker than necessary is Ok
* Taking a callable that takes a callable that is stronger/weaker than necessary

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
nicolaassolini-qntm: Yes, if you look at `unify`, if `act` is a `FunctionType` subst is not `None` only if also `exp` is a `FunctionType`. Also `try_coerce_to` fail if the two types are not numeric.

tests/error/experimental_errors/unitary_callable.py
```diff
@@ -0,0 +1,22 @@
...
+    f(q)
+    return unitary
+
+
+@guppy
+def test(q: qubit) -> None:
+    apply_control(h, q)
+
+
+test.compile_function()
```
doug-q: probably want a newline here

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -389,6 +390,106 @@ class UnitaryFlags(Flag):
...
+        "Unitary",
+        "Powerable",
+        "Daggerable",
+        "Controllable",
+        "PowerControllable",
+    ]:
+        """Returns the name of the corresponding Callable variant for this flag."""
+        match self:
+            case UnitaryFlags.NoFlags:
+                return "Callable"
```
doug-q, line 479: This is much better. You could instead access `callable_type_def.name` here right?

guppylang-internals/src/guppylang_internals/tys/builtin.py
```diff
@@ -209,6 +218,41 @@ def _option_to_hugr(args: Sequence[Argument], ctx: ToHugrContext) -> ht.Type:
 
 
 callable_type_def = CallableTypeDef(DefId.fresh(), None, None)
+unitary_type_def = CallableTypeDef(
```
doug-q, line 232: Is this the best way to do it, or would it be better to define Unitary/Daggerable/etc as sublcasses of CallableTypeDef?

doug-q (approved):
Thanks. Some questions about whether this is the best that can be, done, but looks great.

guppylang-internals/src/guppylang_internals/tys/builtin.py
nicolaassolini-qntm: I wuold like not to define a lot of different classes. I refactored a bit the class init, now it's more compact

guppylang-internals/src/guppylang_internals/tys/ty.py
nicolaassolini-qntm: I can't, e.g. the name would be "Control" not "Controllable"

PR #1770, opened 2026-05-27
feat!: Allow implicit self args in protocols

### LINKED ISSUE #1375: Enable the use of the `Self` type in protocol function definitions

croyzor:
Closes #1375

BREAKING CHANGE: Renamed `handle_implicit_proto_self_arg` to `handle_implicit_self_arg_proto` for symmetry with `parse_self_arg`, `parse_self_arg_proto`


guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -493,6 +486,13 @@ def handle_implicit_proto_self_arg(
     self_args = [param.to_bound() for param in self_defn.params]
     proto_inst = self_defn.check_instantiate(self_args, loc=arg)
     self_arg = BoundTypeVar("self", len(self_args), True, True, (proto_inst,))
+    ctx.param_var_mapping["self"] = TypeParam(
+        idx=0,
+        name="self",
+        must_be_copyable=True,
+        must_be_droppable=True,
```
hsemenenko, line 501: Is there an assumption that a protocol will always be non-linear? Would that preclude using `Self @ owned` as a type annotation in a protocol method?

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -158,24 +156,27 @@ def arg_from_ast(node: AstNode, ctx: TypeParsingCtx) -> Argument:
...
 
     from guppylang_internals.checker.cfg_checker import VarNotDefinedError
 
     match node:
         case ast.Name(id=x):
-            if x not in globals:
+            if x not in ctx.globals:
+                # if self_type := ctx.self_ty:
+                #    if isinstance(self_type, TypeDefn) and self_type.name == x:
+                #        return self_type
```
hsemenenko: Should this be removed?

hsemenenko (approved):

guppylang-internals/src/guppylang_internals/checker/func_checker.py
croyzor: Yes, this is the assumption for the MVP anyway, can be revisited later.
`Self @owned` should be fine, just redundant

guppylang-internals/src/guppylang_internals/tys/parsing.py
croyzor: Oops, yes!

guppylang-internals/src/guppylang_internals/checker/func_checker.py
hsemenenko: I think `Self @ owned` will cause a Guppy compiler error because nonlinear types cannot be owned.

PR #1783, opened 2026-05-29
chore: Do not run benchmarks on PR edit

maximilianruesch:
The inclusion of `edited` caused the benchmarks to run on every change of the PR title, which quickly blocks all runners available in the GH org.

### DIFF
--- a/.github/workflows/benchmarks_pr.yml
+++ b/.github/workflows/benchmarks_pr.yml
@@ -2,7 +2,7 @@ name: Benchmark PRs
 
 on:
   pull_request:
-    types: [opened, reopened, edited, synchronize]
+    types: [opened, reopened, synchronize]
 
 env:
   UV_VERSION: "0.9.5"


.github/workflows/benchmarks_pr.yml
```diff
@@ -2,7 +2,7 @@ name: Benchmark PRs
 
 on:
   pull_request:
-    types: [opened, reopened, edited, synchronize]
+    types: [opened, reopened, synchronize]
```
aborgna-q, line 5: This is the default set, so no need to specify it.

```suggestion
```

aborgna-q (approved):
Ouch -.-'

.github/workflows/benchmarks_pr.yml
maximilianruesch: Would like to keep it for being explicit, as it is always confusing which events are defaults.

Issue #1776, opened 2026-05-28
[Feature]: Make structs default to `frozen=False`


mark-koch:
### Problem Statement

Blocked by #1775

### Component

Guppy Compiler

### Effort Estimate

None

PR #1793, opened 2026-06-02
feat!: Mutable structs

### LINKED ISSUE #1775: [Feature]: Frozen annotation for structs
### Problem Statement

Add a new `frozen` kwarg to the `@guppy.struct` decorator. As a non-breaking change, this should default to `True` to match the current behaviour.

If `frozen=False`, the struct becomes affine and mutable

### Proposed Solution

_No response_

### Component

Guppy Compiler

### Effort Estimate

None

### Additional Context

_No response_

mark-koch:
Closes #1775

This PR adds the optional `frozen` kwarg to the `@guppy.struct` decorator. When set to `True`, the struct will support mutation of both classical and linear fields.

To make this change non-breaking, structs defaults to `frozen=True` and mutation of affine fields is allowed even if `frozen=True`. See #1776 for the issue for changing the default behaviour.

The main difficulty with this feature is linearity checking: If we use a projection of a mutable struct, we need to make sure that the struct value itself has not been moved. In particular, values can now be affine, even if all their projections are copyable individually. This clashes with the previous approach of linearity checking where struct values were tracked by just tracking all leaf projections.

Initially, I was planning to rewrite the entire linearity checker to stop decomposing places into their leaf projections. I still think this would be a good idea, however it felt too risky to get everything right in time for the 1.0 release. Instead, I found an alternative solution: Now, we track uses of values via all their projections, not only the leaf ones. This retains enough information to check if parents have been moved.

I recommend reviewing commit by commit.

Note that this PR is only breaking for `guppylang-internals`:

BREAKING CHANGE: `RawStructDef`, `ParsedStructDef`, and `CheckedStructDef` have a new required field `frozen`
BREAKING CHANGE: `linearity_checker.Scope.use` now requires a `Place` instead of a `PlaceId`

BEGIN_COMMIT_OVERRIDE
feat: Support mutable structs (#1793)
END_COMMIT_OVERRIDE

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
@@ -706,8 +706,8 @@ class BBLinearityChecker(ast.NodeVisitor):
         # On the other hand, we have to ensure that no linear places from the
         # outer scope have been used inside the comprehension (they would be used
         # multiple times since the comprehension body is executed repeatedly)
-        for x, use in inner_scope.used_parent.items():
-            place = inner_scope[x]
+        for use in inner_scope.used_parent.values():
+            place = use.origin_place
             # The only exception are values that are only borrowed from the outer
             # scope. These can be safely reassigned.
             if use.kind == UseKind.BORROW:
--- a/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
@@ -1008,7 +1008,6 @@ def check_cfg_linearity(
                 place = use_scope[x]
                 use = use_scope.used_parent[x]
                 if not place.ty.copyable and (prev_use := scope.used(x)):
-                    use = use_scope.used_parent[x]
                     # Special case if this is a use arising from the implicit returning
                     # of a borrowed argument
                     if isinstance(use.node, InoutReturnSentinel):
--- a/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
@@ -138,6 +138,11 @@ class Use(NamedTuple):
     #: The kind of use, i.e. is the value consumed, borrowed, returned, ...?
     kind: UseKind
 
+    #: Reference to the original place that invoked the use. For example, the use of a
+    #: struct is tracked as uses of all their leaf projections, but for each of them,
+    #: we remember the original root place that was actually used here.
+    origin_place: Place
+
 
 class Scope(Locals[PlaceId, Place]):
     """Scoped collection of assigned places indexed by their id.
@@ -171,7 +176,9 @@ class Scope(Locals[PlaceId, Place]):
             return self.parent_scope.used(x)
         return None
 
-    def use_leaf(self, place: Place, node: AstNode, kind: UseKind) -> None:
+    def use_leaf(
+        self, place: Place, node: AstNode, kind: UseKind, origin_place: Place
+    ) -> None:
         """Records a use of a leaf place.
 
         Works for places in the current scope as well as places in any parent scope.
@@ -179,12 +186,12 @@ class Scope(Locals[PlaceId, Place]):
         assert not immediate_child_places(place), "Must be a leaf"
         x = place.id
         if x in self.vars:
-            self.used_local[x] = Use(node, kind)
+            self.used_local[x] = Use(node, kind, place)
         else:
             assert self.parent_scope is not None
             assert x in self.parent_scope
-            self.used_parent[x] = Use(node, kind)
-            self.parent_scope.use_leaf(place, node, kind)
+            self.used_parent[x] = Use(node, kind, origin_place)
+            self.parent_scope.use_leaf(place, node, kind, origin_place)
 
     def use(self, place: Place, node: AstNode, kind: UseKind) -> None:
         """Records a use of a place.
@@ -192,9 +199,9 @@ class Scope(Locals[PlaceId, Place]):
         Works for places in the current scope as well as places in any parent scope.
         """
         for leaf in leaf_places(place):
-            self.use_leaf(leaf, node, kind)
+            self.use_leaf(leaf, node, kind, place)
         for proj in projections(place):
-            self.used_projections[proj.id] = Use(node, kind)
+            self.used_projections[proj.id] = Use(node, kind, place)
 
     def assign_leaf(self, place: Place) -> None:
         """Records an assignment of a leaf place."""
@@ -350,7 +357,7 @@ class BBLinearityChecker(ast.NodeVisitor):
                     raise GuppyError(err)
 
             # Also check if parents have been moved
-            use = Use(node, use_kind)
+            use = Use(node, use_kind, node.place)
             check_mutable_parent_already_used(node.place, use, self.scope)
 
             # Finally, mark the place as used
@@ -922,7 +929,7 @@ def check_mutable_parent_already_used(place: Place, use: Use, scope: Scope) -> N
             case _:
                 mutable = False
         if mutable and (prev_use := scope.used(parent.id)):
-            err = ParentAlreadyUsedError(use.node, place, use.kind)
+            err = ParentAlreadyUsedError(use.node, use.origin_place, use.kind)
             sub = ParentAlreadyUsedError.ParentUse(prev_use.node, parent, prev_use.kind)
             err.add_sub_diagnostic(sub)
             raise GuppyError(err)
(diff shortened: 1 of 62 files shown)


guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -778,9 +778,13 @@ def leaf_places(place: Place) -> Iterator[Place]:
...
-            stack += [
-                FieldAccess(place, field, place.defined_at) for field in place.ty.fields
-            ]
+            if place.ty.fields:
+                stack += [
+                    FieldAccess(place, field, place.defined_at)
+                    for field in place.ty.fields
+                ]
+            else:
+                yield place
```
mark-koch, line 799: Empty struct places should be treated as leafs, before we just omitted them

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -314,9 +314,8 @@ def visit_PlaceNode(
         # Places involving subscripts are handled differently since we ignore everything
         # after the subscript for the purposes of linearity checking.
         if subscript := contains_subscript(node.place):
-            # We have to check the item type to determine if we can move out of the
-            # subscript.
-            if not is_inout_arg and not subscript.ty.copyable:
+            # We are only allowed to move copyable data out of a subscript
+            if not is_inout_arg and not node.place.ty.copyable:
```
mark-koch, line 326: Driveby: Whether we can move out of a subscript depends on the type of the final projection, not the whole array element. For example:

```python
@guppy.struct(frozen=False)
class S:
    q: qubit
    x: int

@guppy
def foo(xs: array[S, 42]) -> None:
    a = xs[0].x  # Ok since the x projection is int, so we can copy it
    b = xs[0].q  # Error: Cannot move affine value out of array 
```

This came up while writing the array test, so I fixed it here

tests/error/linear_errors/mutable_affine_struct2.err
```diff
@@ -0,0 +1,15 @@
+Error: Copy violation (at $FILE:14:11)
+   | 
+12 |     if b:
+13 |         t = s
+14 |     return s
+   |            ^ Field `s.x` cannot be returned ...
```
nicolaassolini-qntm: should here say "Variable `s` cannot be returned ..." as in the mutable_affine_struct case?

tests/error/linear_errors/mutable_affine_struct7.err
```diff
@@ -0,0 +1,15 @@
+Error: Copy violation (at $FILE:13:12)
+   | 
+11 | def foo(s: MyStruct @owned, b: bool) -> None:
+12 |     while b:
+13 |         t = s
+   |             ^ Field `s.x` cannot be moved ...
```
nicolaassolini-qntm: same, why `s.x` and not `s`?

tests/error/struct_errors/mutate_classical.err
```diff
@@ -1,8 +1,8 @@
-Error: Unsupported (at $FILE:14:6)
+Error: Immutable (at $FILE:14:4)
    | 
 12 | def foo(s: MyStruct) -> tuple[MyStruct, bool]:
 13 |     t = s
 14 |     t.x += 1
-   |       ^ Mutation of classical fields is not supported
+   |     ^^^ Struct `MyStruct` is immutable
```
nicolaassolini-qntm, line 7: adding a Note that says "Set `frozen=False` to make the struct mutable"?

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -145,42 +146,72 @@ class Scope(Locals[PlaceId, Place]):
     """
 
     parent_scope: "Scope | None"
```
nicolaassolini-qntm, line 154: you commented all the fields a part of this, a small comment here?

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -145,42 +146,72 @@ class Scope(Locals[PlaceId, Place]):
...
         """Checks whether a place has already been used."""
-        if x in self.vars:
-            return self.used_local.get(x, None)
-        assert self.parent_scope is not None
-        return self.parent_scope.used(x)
+        if x in self.used_projections:
+            return self.used_projections[x]
+        if self.parent_scope:
+            return self.parent_scope.used(x)
+        return None
```
nicolaassolini-qntm, line 178: the new implementation of `used` only checks for place in `used_projections` or in the parent_scope, is it fine? before we was checking in `used_local`

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -834,6 +908,26 @@ def is_simple_literal_subscript(
     return (subscript, literal_idx)
 
 
+def check_mutable_parent_already_used(place: Place, use: Use, scope: Scope) -> None:
```
nicolaassolini-qntm, line 919:
```suggestion
def _check_mutable_parent_already_used(place: Place, use: Use, scope: Scope) -> None:
```
if it is meant to be just a local helper, (same for the other new helper functions)

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -834,6 +908,26 @@ def is_simple_literal_subscript(
...
+        match parent.ty:
+            case StructType(frozen=False):
+                mutable = True
+            case _:
+                mutable = False
+        if mutable and (prev_use := scope.used(parent.id)):
+            err = ParentAlreadyUsedError(use.node, place, use.kind)
+            sub = ParentAlreadyUsedError.ParentUse(prev_use.node, parent, prev_use.kind)
+            err.add_sub_diagnostic(sub)
+            raise GuppyError(err)
```
nicolaassolini-qntm:
```suggestion
    for parent in parent_places(place, include_self=True):
        match parent.ty:
            case StructType(frozen=False):
                if prev_use := scope.used(parent.id):
                    err = ParentAlreadyUsedError(use.node, place, use.kind)
                    sub = ParentAlreadyUsedError.ParentUse(prev_use.node, parent, prev_use.kind)
                    err.add_sub_diagnostic(sub)
                    raise GuppyError(err)

```

nicolaassolini-qntm (commented):
I think is good just little comments!

nicolaassolini-qntm:
I also requested Copilot, sometimes it is useful, at least it is better than me in spotting typos or "ugly" comments

tests/error/linear_errors/mutable_affine_struct2.err
mark-koch: Fixed!

tests/error/struct_errors/mutate_classical.err
mark-koch: Following #1795, they will be mutable by default, so the user must have explicitly written `frozen=True`

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
mark-koch: The leaf projections in `used_projections` are exactly the ones that are in `used_local | used_global` and `used_global` is exactly what is tracked in `parent_scope`, so they are equivalent when you're querying leafs.

The only difference is that `used_projections` also contains non-leafs, so now `used` now gives the correct answer for all projections, not just leafs

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
mark-koch: None of the other helpers in this file are underscored. Imo making a public/private distinction doesn't make much sense for `guppylang-internals` since everything is be definition internal

tests/error/struct_errors/mutate_classical.err
nicolaassolini-qntm: oh true!  In this case, the message is fine!

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -679,6 +707,9 @@ def _check_comprehension(
             elif not place.ty.copyable:
                 raise GuppyTypeError(ComprAlreadyUsedError(use.node, place, use.kind))
 
+            # Also check if parents have been moved
+            check_mutable_parent_already_used(place, use, self.scope)
```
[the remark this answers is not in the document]
mark-koch: No, but I added some tests for comprehensions

guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -906,6 +999,7 @@ def check_cfg_linearity(
             for x, use_bb in live.items():
                 use_scope = scopes[use_bb]
                 place = use_scope[x]
+                use = use_scope.used_parent[x]
                 if not place.ty.copyable and (prev_use := scope.used(x)):
                     use = use_scope.used_parent[x]
```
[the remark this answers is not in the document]
mark-koch: Fixed

nicolaassolini-qntm (approved):
Sounds good! Welcome to Mutable structs

PR #1801, opened 2026-06-03
docs: use scipy optimizer in QAOA example

### LINKED ISSUE #1545: [Docs]: Use a real-world classical optimizer in the QAOA MaxCut example
> [!NOTE]
> 
>
>

## Summary

The QAOA MaxCut notebook currently searches for good QAOA parameters by sampling cost and mixer angles uniformly at random. This keeps the example simple, but it makes the classical optimization part unrealistic and obscures the intended variational workflow.

Please refactor the notebook to use a real classical optimizer, such as `scipy.optimize.minimize`, while keeping the quantum/Guppy part of the example readable.

## Background

The notebook lives here:

- https://github.com/Quantinuum/guppylang/blob/main/examples/qaoa_maxcut_example.ipynb

The relevant functions are:

- `build_qaoa_instance`
- `energy_from_result`
- `eval_qaoa_energy`
- `solve_max_cut_instance`

The notebook currently says that the optimizer is deliberately naive and could be improved with SciPy. This issue is about doing that improvement.

This is separate from the deeper performance issue in https://github.com/Quantinuum/guppylang/issues/1546. The goal here is to improve the classical optimization workflow, not to solve recompilation or emulator setup costs.

## Minimal example of the current pattern

The current implementation repeatedly samples random angles and keeps the best result:

```python
for _ in range(iterations):
    guess_cost_angles = rng.uniform(0, 2 * np.pi, size=p_value)
    guess_mixer_angles = rng.uniform(0, 2 * np.pi, size=p_value)

    qaoa_energy, sys_result = eval_qaoa_energy(
        guess_mixer_angles,
        guess_cost_angles,
        graph,
        p_value,
        seed,
        shots,
    )
```

The requested workflow should instead present QAOA as a normal variational loop:

```python
from scipy.optimize import minimize

def objective(params: NDArray[np.float64]) -> float:
    cost_angles, mixer_angles = split_qaoa_params(params, p_value)
    energy = eval_qaoa_energy(
        mixer_angles,
        cost_angles,
        graph,
        p_value,
        seed,
        shots,
    )
    return -energy

result = minimize(objective, initial_params, method="COBYLA")
```

The exact optimizer and helper function names are up to the implementer, as long as the notebook remains clear and reproducible.

## Scope

Please update the QAOA notebook so that it:

- Uses a SciPy optimizer for the classical parameter search.
- Refactors `eval_qaoa_energy` so it returns only the energy needed by the optimizer.
- Runs one final QAOA forward pass using the best parameters found by the optimizer.
- Uses that final pass for the counts/bar-chart visualization.
- Keeps a fixed seed or otherwise makes the notebook output reasonably reproducible.
- Updates the explanatory markdown so readers understand which part is the classical optimizer and which part is the Guppy quantum program.

The notebook currently returns `QsysResult` from every energy evaluation. A good cleanup is to remove that from the optimization loop and only request/store the full result for the final best parameter set.

## Out of scope

This issue does not require solving the recompilation/performance problem tracked in #1546.

This issue does not require implementing parameterized Guppy entrypoints, caching compiled QAOA programs, or changing the emulator API.

This issue does not require finding globally optimal QAOA parameters for every graph. The target is a clearer and more realistic example.

## Environment and testing

Useful files:

- `examples/qaoa_maxcut_example.ipynb`
- `pyproject.toml`
- `tests/integration/test_notebooks.py`

Useful command:

```bash
just check
```

Note: the QAOA notebook is currently excluded from CI because it is long-running. Please still run the notebook manually, or document any runtime limitation in the PR description.

cmdr-chara:
This updates the QAOA MaxCut example to use `scipy.optimize.minimize` for the classical parameter search instead of sampling random angles.

The refactor keeps full `QsysResult` objects out of the optimizer loop. `eval_qaoa_energy` now returns only the scalar energy, and the notebook runs one final QAOA circuit with the best parameters for validation and plotting.

I also added SciPy to the examples dependency group because the notebook now imports it directly.

Testing:
- `python -m uv lock --check`
- `python -m uv run ruff format --check`
- `python -m uv run ruff check`
- full QAOA notebook executed manually from a temp copy; validation printed `Success ratio: 0.3845`

I also ran `python -m uv run pytest tests/integration/test_notebooks.py` locally on Windows/Python 3.13.7. It produced unrelated stochastic notebook snapshot diffs in `adaptive-qpe`, `repeat-until-success`, `state_results`, and `swap_test`. The changed QAOA notebook is excluded from that test by the project because it is long-running, so I executed it manually as above.

AI disclosure: I used OpenAI Codex with GPT-5.5 Extra High to help refactor the QAOA optimizer loop and review the PR text. I manually inspected the diff, removed unnecessary generated wording, and ran the notebook/checks locally.

Closes #1545


CalMacCQ:
Hi @cmdr-chara ,  Thanks for making a PR. I'll take a look at this and get back to you.

examples/qaoa_maxcut_example.ipynb
CalMacCQ, line 1: I've ran this locally and it looks promising. 

Scipy optimizer performance

```
optimizer status:  Return from COBYLA because the trust region radius reaches its lower bound.
best energy:  5.066499999999999
best mixer angles:  [0.861 0.389 0.197]
best cost angles:  [0.23  0.228 0.755]
CPU times: user 13.1 s, sys: 2.95 s, total: 16 s
Wall time: 1min 21s
```

Compare with the old naive random sampling approach

```
new highest energy found:  3.1499999999999986
new highest energy found:  3.299999999999997
new highest energy found:  4.306499999999993
new highest energy found:  4.876999999999997
new highest energy found:  4.937499999999999
highest energy:  4.937499999999999
best guess mixer angles:  [0.392 0.247 0.138]
best guess cost angles:  [0.592 0.738 0.608]
CPU times: user 31.7 s, sys: 3.88 s, total: 35.5 s
Wall time: 2min 6s
```

The scipy optimizer seems to reach a higher expectation value for the energy which is great. Also it seems the runtime performance has improved as well.

I'd stronly recommend commiting the fully executed notebook so that we can see all of the cell outputs.

What does this mean?
```
optimizer status:  Return from COBYLA because the trust region radius reaches its lower bound.
```

pyproject.toml
```diff
@@ -5,7 +5,7 @@ dev = [
   { include-group = "test" },
 ]
 docs = ["furo>=2024.8.6", "sphinx >=7.2.6,<9"]
-examples = ["matplotlib>=3.10.9", "networkx>=2.6,<4"]
+examples = ["matplotlib>=3.10.9", "networkx>=2.6,<4", "scipy>=1.15.3"]
```
CalMacCQ: Is it necessary to explicitly set scipy as a dependency here? The pytket package depends on scipy and pytket is a dependency of guppylang.

examples/qaoa_maxcut_example.ipynb
CalMacCQ, line 1: One thing I did observe when running this locally was the the "success ratio" (printed by the `validate_qaoa_experiment` function) is lower. This success ratio is the proportion of the shots which land in the two basis states representing the optimal colorings. From the plot this is lower with your implementation than with the random sampling based approach.

This is a little puzzling as the scipy optimizer achieves a higher expecation value than random sampling (expected).

Lets look at the two plots below... The green bars represent "success" shots. 

Your scipy based results: `success ratio=0.3385, expectation value = 5.0935`

Image

Old random sampling based results: `success ratio=0.4205, expectation value = 4.937499999999999`

Image


I suspect there is a bug somewhere as I would not expect the success ratio to be lower with a better optimizer. It could well have existed prior to your PR so it may be my fault.

Could you please investigate? Happy to discuss further if anything is unclear.

CalMacCQ (commented):
Thanks! This is a promising start. Your scipy optimizer improves both the expectation value and the runtime performance of QAOA which is great. The success ratio appears to be worse than the random sampling approach though which is mysterious. Maybe there was a bug in this example prior to your contribution that has only manifested now.

Also I would much prefer if you commited the fully executed notebook to github so we can review the cell outputs (the matplotlib plots and the expectation values).

examples/qaoa_maxcut_example.ipynb
cmdr-chara: Thanks, I committed the fully executed notebook outputs in `4aceac3e`.

The COBYLA message is a normal optimizer stopping condition rather than an error. It means the trust-region radius has reached its configured lower bound, so COBYLA considers that further local steps are no longer meaningful at that scale.

In other runs the status may instead be:

```text
Maximum number of function evaluations has been exceeded.

pyproject.toml
cmdr-chara: Good point, I agree this is redundant here.

Since `pytket` already depends on `scipy`, and `pytket` is already a dependency of `guppylang`, the examples group does not need to list `scipy` explicitly.

I removed the explicit `scipy>=1.15.3` entry and updated `uv.lock`.

I also checked locally with:

```bash
uv tree --package scipy --invert

examples/qaoa_maxcut_example.ipynb
cmdr-chara: I investigated this by checking both the plotted result bars and the sampled cut-value distribution.

The short version is that the two metrics are related but not identical. The scipy optimizer is maximizing **expected max-cut energy** over the sampled distribution. The `success ratio` reported by `validate_qaoa_experiment` only counts shots that land in the exact optimal colorings, which are the two green bars in the plot.

The plots show that the exact-optimal green bars can be lower, but that does not necessarily mean the energy is worse. In the executed notebook, much more probability mass moves into near-optimal `cut=5` states, even though the exact-optimal `cut=6` probability is lower. So the expected energy can improve while the exact-optimal success ratio decreases.

I added a distribution-by-cut-value diagnostic to make this visible numerically, because the bar chart alone shows basis-state counts but not the cut value represented by each non-optimal basis state.

cmdr-chara:
Updated this branch and re-executed the notebook locally.

I kept the SciPy optimizer flow, use the optimized angles directly in the final run, made the coloring check order-insensitive, and kept the cut-value distribution so the sampled result is easier to inspect.

The refreshed output gives optimized energy 5.1195 and optimal-coloring success ratio 0.3950.

CalMacCQ (approved):
Thank you for making the changes @cmdr-chara. Happy for this to go in now.

CalMacCQ:
Could you also comment on the https://github.com/Quantinuum/guppylang/issues/1545 so I can assign it to you?

nicolaassolini-qntm (approved):

Issue #1797, opened 2026-06-02
[Feature]: Allow emulation platform to be configurable


ss2165:
### Problem Statement

Soon qis-compiler will be configurable to choose the target platform (https://github.com/Quantinuum/tket2/pull/1381)

the emulator interface should support this in build configuration

This should then be tested, by parametrising some or all execution tests by platform. 

Specifically there should be unit tests that use platform ops from one platform and compile and run on the other to test cross-compilation.

### Effort Estimate

Medium - Moderate complexity

PR #1763, opened 2026-05-21
refactor: Split qsystem std lib into helios and sol

ss2165:
`guppylang.std.qsystem` is now split into two platform specific modules `helios` and `sol`. Functions can still be accessed from their old paths, but these are deprecated and aliases for the `helios` variants.

- Depends on: https://github.com/Quantinuum/tket2/pull/1567
- Depends on: https://github.com/Quantinuum/tket2/pull/1646

Closes #1676

PR #1807, opened 2026-06-03
feat: Frontend support for Helios platform configuration

cgh-qtnm:
This PR adds a new module `guppylang.std.qsystem.helios` and a function `set_platform_config` which sets Helios-specific platform options on a compiled HUGR. At the moment, this is only for development purposes - none of these options are supported by the backend yet and the interface should be considered unstable.

In addition, this module will also provide a definition of the metadata key. Let me know if that should go somewhere else.


tests/integration/test_qsystem.py
```diff
@@ -268,3 +271,68 @@ def test() -> int:
...
+
+    package = test.compile_function()
+    validate(package)
+    with pytest.raises(ValueError, match="must be >= 0"):
+        set_platform_config(
+            package,
+            squash_rxys=False,
+            enable_replay=True,
+            dd_threshold=-1,
+        )
```
nicolaassolini-qntm: We usually use the tests/error folder for test errors

nicolaassolini-qntm (changes requested):
Good for me, only a little change

cgh-qtnm:
Actually, I don't think this needs to merge yet. Sorry for the false alarm.

nicolaassolini-qntm:
> Actually, I don't think this needs to merge yet. Sorry for the false alarm.

No worries, notify me again when ready!

ss2165:
The module clashes with: https://github.com/Quantinuum/guppylang/pull/1763

and the scope seems to be related to https://github.com/Quantinuum/guppylang/issues/1797

cgh-qtnm:
@ss2165 I think this is ready to merge, although it won't actually have an effect until backend support is in place.

ss2165 (changes requested):
Thank you, some suggestions:


1. The config update function has been put in the guppy standard library, the standard library is a place for _guppy_ functions - this is a hugr transformation step. It probably belongs elsewhere.
2. guppy v1 has standardized on a decorator based approach for adding metadata to programs, see for example link_name: https://github.com/Quantinuum/guppylang/blob/cc0977af918569ef9747bb07ca17fdfcbe8376e9/guppylang/src/guppylang/library.py#L76 This seems the right approach here too
4. The feature is said to be experimental in the docstring - this should be enforced https://github.com/Quantinuum/guppylang/blob/cc0977af918569ef9747bb07ca17fdfcbe8376e9/guppylang-internals/src/guppylang_internals/experimental.py#L13

cgh-qtnm:
> It probably belongs elsewhere.

Sure, where would you suggest?

cgh-qtnm:
Closed in favor of https://github.com/Quantinuum/tket2/pull/1848

PR #1808, opened 2026-06-03
feat(std): add native PCG32 random number generator

### LINKED ISSUE #1578: [Feature]: Add native RNG implemented in Guppy
### Problem Statement

Guppy provides an RNG in the standard library that is implemented via a HUGR extension op that uses the platform RNG. As the platform RNG is global, this can lead to unexpected behaviour if multiple RNGs are defined in the same program.

Defining an RNG natively in Guppy should allow multiple RNGs to coexist in a single program.

### Proposed Solution

Implement PCG32 in guppy: https://github.com/wjakob/pcg32/blob/master/pcg32.h.

The native RNG should be supplementary to the existing RNG in `guppylang.std.qsystem.random`. 

### Component

Standard library

### Effort Estimate

Medium - Moderate complexity

### Additional Context

Implementation from @jake-arkinstall 

```python
@guppy.struct
class PCG32:
    state: nat
    inc: nat

    @guppy
    def nextUInt(self: "PCG32") -> tuple["PCG32", int]:
        """
        Return the advanced state and the 32 bit integer corresponding to
        the returned value.

        Unfortunately we can't mutate struct members so this pattern is necessary.
        Usage:

        ```python
        r = seeded_pcg32(1001)
        r, val = r.nextUInt()
        r, val = r.nextUInt()
        # and so on ...
        ```
        """
        PCG32_MULT: nat = 0x5851F42D4C957F2D
        new_state = nat((self.state * PCG32_MULT + self.inc))
        xorshifted = nat((((self.state >> 18) ^ self.state) >> 27) & 0xFFFFFFFF)
        rot = nat((self.state >> 59) & 0xFFFFFFFF)
        result = (xorshifted >> rot) | (xorshifted < ((~rot + 1) & 31)) & 0xFFFFFFFF
        if result >> 31:
            result = -(0xFFFFFFFF - result + 1)
        new_rng = PCG32(new_state, self.inc)
        return new_rng, result
```

luffy-orf:
## Summary

Adds a pure-Guppy PCG32 RNG at `guppylang.std.random` so random state lives in a local value instead of the platform-global qsystem RNG.

- Introduces `PCG32` and `seeded_pcg32(seed: int)`
- `next_int()` returns `(new_rng, signed_32_bit_int)` for immutable state threading
- Uses explicit 32-bit masking and signed conversion on top of `nat`
- Leaves `guppylang.std.qsystem.random.RNG` unchanged

Fixes the interference problem from #1578: with `seeded_pcg32(1)`, the second draw is always `-444364974`, even if another function uses `seeded_pcg32(2)`.

Closes #1578

## Test plan

- [x] `uv run pytest tests/integration/std/test_random.py`
- [x] Deterministic PCG32 sequence checks (seeds 1, 2, 54 vs reference vectors)
- [x] Issue motivating example: inner RNG does not affect outer stream
- [x] Quantum control-flow test: `rng0_1` is stable across branches
- [x] `uv run pytest tests/integration/test_qsystem.py` — no qsystem RNG regression
- [x] Full suite: `pytest -n auto` (1513 passed)
- [x] `ruff check`, `ruff format --check`, `mypy guppylang guppylang-internals`

BEGIN_COMMIT_OVERRIDE
feat: Add native PCG32 random number generator to std lib (#1808)
END_COMMIT_OVERRIDE

### DIFF
--- a/tests/integration/std/test_random.py
+++ b/tests/integration/std/test_random.py
@@ -1,0 +1,145 @@
+from guppylang import guppy
+from guppylang.std.builtins import owned, result
+from guppylang.std.quantum import h, measure, qubit, x
+from guppylang.std.random import seeded_pcg32
+
+
+def test_pcg32_compile(validate) -> None:
+    @guppy
+    def main() -> tuple[int, int]:
+        rng = seeded_pcg32(1)
+        rng, first = rng.next_int()
+        rng, second = rng.next_int()
+        return first, second
+
+    validate(main.compile_function())
+
+
+def test_pcg32_sequence_seed_1(run_int_fn) -> None:
+    @guppy
+    def main() -> int:
+        rng = seeded_pcg32(1)
+        rng, first = rng.next_int()
+        rng, second = rng.next_int()
+        return first + second
+
+    run_int_fn(main, 1307692281 + -444364974)
+
+
+def test_pcg32_sequence_seed_2(run_int_fn) -> None:
+    @guppy
+    def main() -> int:
+        rng = seeded_pcg32(2)
+        _, first = rng.next_int()
+        return first
+
+    run_int_fn(main, -8000311)
+
+
+def test_pcg32_deterministic_sequence(run_int_fn) -> None:
+    """First five outputs for initstate=42, initseq=54 match PCG32 reference."""
+
+    @guppy
+    def main() -> int:
+        rng = seeded_pcg32(54)
+        total = 0
+        for _ in range(5):
+            rng, value = rng.next_int()
+            total += value
+        return total
+
+    # Reference values from https://rosettacode.org/wiki/Pseudo-random_numbers/PCG32
+    expected = sum(
+        [
+            -1587805513,
+            2068313097,
+            -1172491472,
+            -2083327341,
+            -1079740341,
+        ]
+    )
+    run_int_fn(main, expected)
+
+
+def test_pcg32_motivating_example() -> None:
+    """Exact scenario from issue #1578: inner RNG must not affect outer."""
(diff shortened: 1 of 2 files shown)


tests/integration/std/test_random.py
```diff
@@ -0,0 +1,145 @@
...
+            -1172491472,
+            -2083327341,
+            -1079740341,
+        ]
+    )
+    run_int_fn(main, expected)
+
+
+def test_pcg32_motivating_example() -> None:
+    """Exact scenario from issue #1578: inner RNG must not affect outer."""
```
tatiana-s: If you add the full link to the issue it will be easier to look it up when looking at this code at a later date.

tatiana-s (commented):
Thank you for your contribution! This looks great in general, one thing I would maybe suggest is that in code with as many "magic" numbers as this, it would be good to have some more comments on where they came from or what they mean.

luffy-orf:
Thanks for the review! @tatiana-s  I’ve addressed both points:

- random.py - Added comments for the constants (32-bit masks, LCG multiplier, XSH-RR shift amounts, seeding steps) with references to the PCG spec and pcg32_random_r / pcg32_srandom_r.
- test_random.py - Updated the motivating-example (and quantum interference) docstrings to use the full issue link.

Let me know if you’d like any change is reuqired

tests/integration/std/test_random.py
luffy-orf: Done

tatiana-s:
Thanks!

tatiana-s (approved):

PR #1809, opened 2026-06-03
refactor!: Add shared recursive definition validation for structs and enums

### LINKED ISSUE #1732: [Refactor] Refactor recursive type-definition checks for structs and enums
> [!NOTE]
> 
>
>

## Summary

Guppy currently rejects recursive `@guppy.struct` and `@guppy.enum` definitions by calling `check_not_recursive`, which is the right user-facing behavior for now. However, the implementation of this check is fragile: the struct and enum implementations both detect recursion by temporarily replacing `check_instantiate` instance methods, then re-running type parsing over the fields.

The goal of this issue is to replace that hack with an explicit validation step that runs before type checking, ideally shared by structs and enums. The implementation should detect direct and mutual recursion across structs and enums, including through nested type expressions.

This is the follow-up to:

- Original issue: https://github.com/Quantinuum/guppylang/issues/1422
- Enum precursor issue: https://github.com/Quantinuum/guppylang/issues/1423
- Precursor PR adding enum recursion checks: https://github.com/Quantinuum/guppylang/pull/1450

## Background

The current checks live here:

- Struct recursion check: https://github.com/Quantinuum/guppylang/blob/e4324d5194ad5213719b1de7322159a11d124904/guppylang-internals/src/guppylang_internals/definition/struct.py#L248
- Enum recursion check: https://github.com/Quantinuum/guppylang/blob/e4324d5194ad5213719b1de7322159a11d124904/guppylang-internals/src/guppylang_internals/definition/enum.py#L364

Both implementations work by temporarily replacing the parsed definition's `check_instantiate` method with a dummy function that raises `UnsupportedError("Recursive definitions")`.

## Minimal examples

These examples should continue to fail with a clear Guppy user error. The implementation should change; the user-facing behavior should not become less strict.

Direct struct recursion:

```python
from guppylang import guppy


@guppy.struct
class Node:
    next: "Node"


Node.compile()
```

Mutual struct recursion:

```python
from guppylang import guppy


@guppy.struct
class A:
    b: "B"


@guppy.struct
class B:
    a: A


B.compile()
```

Mixed enum/struct recursion:

```python
from guppylang import guppy


@guppy.enum
class MaybeNode:
    Some = {"node": "Node"}


@guppy.struct
class Node:
    next: MaybeNode


Node.compile()
```

## Scope

Please implement a shared recursive-definition validation mechanism for parsed structs and enums.

The implementation should:

- Detect direct recursion, such as a struct field containing the same struct.
- Detect mutual recursion across multiple definitions.
- Detect recursion across structs and enums in either direction.
- Handle nested type expressions, such as `list[StructB]` and `tuple[MyStruct, int]`.
- Avoid temporarily mutating `check_instantiate`.
- Avoid duplicating the same recursion-checking logic separately in `struct.py` and `enum.py`.
- Preserve or improve the current user-facing error for unsupported recursive definitions.

A good direction is to build an explicit dependency graph between parsed struct/enum definitions and run cycle detection before checking the field types. Using `DefId` for graph nodes should make this more robust than comparing names.

It is also fine to address the TODO by changing `StructType` and `EnumType` so that they store a `DefId` rather than a checked definition, as long as the resulting implementation removes the current recursive checking hack and keeps existing behavior covered by tests.

## Out of scope

This issue does not ask for recursive structs or recursive enums to be supported. Recursive type definitions should still be rejected.

This issue also does not ask for changes to the public enum syntax, generated enum constructors, or unrelated struct/enum features.

## Environment and testing

There are already tests for several of these cases:

- `tests/error/struct_errors/recursive.py`
- `tests/error/struct_errors/mutual_recursive.py`
- `tests/error/struct_errors/struct_enum_recursive.py`
- `tests/error/enum_errors/recursive.py`
- `tests/error/enum_errors/mutual_recursive.py`
- `tests/error/enum_errors/enum_struct_recursive.py`

Useful commands:

```bash
uv run pytest tests/error/test_struct_errors.py tests/error/test_enum_errors.py
just check
```

Kitsunp:
Closes #1732

### Summary

This PR replaces the previous recursive type-definition checks for structs and enums with a shared explicit validation step.

Previously, `struct.py` and `enum.py` detected recursive definitions by temporarily replacing `check_instantiate` on parsed definitions, then re-running type parsing over the fields. This PR removes that monkeypatch-based approach and adds shared recursion validation for parsed struct and enum definitions before field type checking.

The checker walks struct fields and enum variant fields, resolves parsed struct/enum dependencies from type annotations, including nested annotations such as `list[T]` and `tuple[T, ...]`, and detects cycles using `DefId`.

### Changes

- Adds shared recursive-definition validation in `guppylang_internals.definition.util`.
- Reuses the shared validation from both `struct.py` and `enum.py`.
- Adds `annotation_nodes()` in `guppylang_internals.tys.parsing` for walking nested annotation AST nodes.
- Renames `_try_parse_defn()` to `try_parse_defn()` because it is now used outside `tys.parsing`.
- Preserves the existing user-facing error for unsupported recursive definitions.
- Adds nested cross-recursion error tests for struct/enum cycles.
- Adds a regression test for imported struct dependencies being resolved in their defining scope.

### Design Note

The issue mentions building an explicit dependency graph as a good direction. This implementation keeps the graph implicit: dependencies are extracted from parsed field/variant annotations, and cycle detection is performed with DFS over `DefId`s.

This keeps the change small while still covering the requested behavior: direct recursion, mutual recursion, mixed struct/enum cycles, nested type expressions, shared validation before field type checking, and no mutation of `check_instantiate`.

### Verification

```bash
uv run pre-commit run --all-files
UV_PROJECT_ENVIRONMENT=/tmp/guppy-1732-venv uv run pytest tests/integration/test_struct.py::test_imported_struct_dependency_uses_defining_scope tests/error/test_struct_errors.py tests/error/test_enum_errors.py -q
```

Results:

```text
pre-commit: passed
60 passed
```

### AI Transparency

I used OpenAI Codex to help inspect the repository patterns, compare the implementation against the issue requirements, draft tests, and run local verification. I manually reviewed the code, snapshots, design tradeoffs, and command outputs before preparing this PR.

BREAKING CHANGE: Internal parsing helpers were renamed because they are now used outside `tys.parsing`: `_try_parse_defn` is now `try_parse_defn`, and `_annotation_nodes` is now `annotation_nodes`. Internal callers should update imports and references to the new helper names.


tests/integration/test_struct.py
```diff
@@ -71,6 +72,18 @@ def main(a: StructA, b: StructB) -> None:
...
+def test_imported_struct_dependency_uses_defining_scope(validate):
+    @guppy.struct
+    class B:
+        a: struct_scope_defs.A
+
+    @guppy
+    def main() -> B:
+        return B(struct_scope_defs.A(struct_scope_defs.B(0)))
+
+    validate(main.compile_function())
```
nicolaassolini-qntm, line 84: why this test has been added?

nicolaassolini-qntm (changes requested):
Hi @Kitsunp, thanks for the contribution!

I think your solution is fine. I just have some minor comments.

Regarding the functions in `recursion.py`, I suggest moving them to a more appropriate location. In particular:

* `check_not_recursive`, `_check_not_recursive`, `_field_type_asts`, `_is_parsed_struct_or_enum`, and `_dependencies` should be moved to `guppylang-internals/src/guppylang_internals/definition/util.py`.
* All the other functions related to parsing the AST should be moved to `guppylang_internals.tys.parsing`.

tests/integration/test_struct.py
Kitsunp: I added this test based on the Codex Review comment about resolving dependencies in their defining scope.

The test covers a case where an imported struct definition refers to a type name that also exists in the caller module. In `struct_scope_defs.py`, `A` has a field `b: B`, which should resolve to `struct_scope_defs.B`. In `test_struct.py`, there is another local struct also named `B`.

This checks that when the recursion validator walks into the imported definition `struct_scope_defs.A`, it rebuilds the type parsing context from `A`’s defining scope instead of continuing to use the caller’s context. Otherwise, the checker could incorrectly resolve `A.b: B` to the local `B` and report a false recursive cycle.

Kitsunp:
Thanks for the review @nicolaassolini-qntm !

I updated the PR following your suggestion:

- Moved the definition-level recursion validation helpers to `guppylang_internals.definition.util`.
- Moved the AST annotation traversal helper to `guppylang_internals.tys.parsing`.
- Removed the separate `definition/recursion.py` module.

I kept the runtime imports of `ParsedStructDef` / `ParsedEnumDef` local inside the helper functions to avoid circular imports, since `struct.py` and `enum.py` already import from `definition.util`.

I also kept the imported-scope regression test. It covers the issue raised in the Codex review: when the recursion checker walks into a dependency defined in another module, that dependency's annotations need to be resolved in its defining frame rather than the caller's frame.

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -273,6 +273,16 @@ def _parse_delayed_annotation(ast_str: str, node: ast.Constant) -> ast.expr:
         return stmt.value
 
 
+def _annotation_nodes(node: ast.expr) -> Iterator[ast.expr]:
```
nicolaassolini-qntm:
```suggestion
def annotation_nodes(node: ast.expr) -> Iterator[ast.expr]:
```
this function is used outside this module

guppylang-internals/src/guppylang_internals/definition/util.py
```diff
@@ -2,32 +2,45 @@
...
 from guppylang_internals.diagnostic import Error, Help, Note
-from guppylang_internals.error import GuppyError
+from guppylang_internals.error import GuppyError, InternalGuppyError
 from guppylang_internals.ipython_inspect import is_running_ipython
 from guppylang_internals.span import SourceMap, Span, to_span
 from guppylang_internals.tys.param import Parameter
+from guppylang_internals.tys.parsing import (
+    TypeParsingCtx,
+    _annotation_nodes,
+    _try_parse_defn,
```
nicolaassolini-qntm: this function is now used also here, not only in `parsing`. could you rename it in `try_parse_defn`?

This will probrably make the pr breaking, thus you will have to update the title and the description

nicolaassolini-qntm (commented):
Good! Just two other small changes.

Also, could you rename the PR and make the title a bit more specific?

Kitsunp:
Updated, thanks. I renamed `_annotation_nodes` to `annotation_nodes` and `_try_parse_defn` to `try_parse_defn`, since both helpers are now used from outside `tys.parsing`.

Kitsunp:
@nicolaassolini-qntm would this be a specific enough title?

`refactor!: add shared recursive definition validation for structs and enums`

nicolaassolini-qntm:
> @nicolaassolini-qntm would this be a specific enough title?
> 
> `refactor!: add shared recursive definition validation for structs and enums`

yes!

Kitsunp:
@nicolaassolini-qntm I’ve updated the PR title and body.

nicolaassolini-qntm (approved):
Good job, thanks again!

nicolaassolini-qntm:
@Kitsunp can you post a comment on the linked issue? So I can assign you to the issue before closing it

PR #1813, opened 2026-06-04
fix: Upgrade tket and require pytket extra

maximilianruesch:
Upgrades tket to the next patch version to be able to require the `pytket` extra, which should enable package manager resolution to determine the best version to install given the `guppylang` and `tket` requirement for `pytket`.

This is to hopefully address problems in `lowest-direct` dependency release testing (see https://github.com/Quantinuum/guppylang/actions/runs/26953716772/job/79525233395?pr=1799) and is related to https://github.com/Quantinuum/tket/issues/2185.


guppylang-internals/pyproject.toml
```diff
@@ -35,7 +35,7 @@ dependencies = [
     "hugr ~= 0.16.0",
     "wasmtime>=38.0,<44.1",
     "pytket>=1.34",
-    "tket >= 0.13.0",
+    "tket[pytket]>=0.13.0",
```
aborgna-q, line 38:
```suggestion
    "tket[pytket]>=0.13.1",
```
?

maximilianruesch: The `pytket` extra is only defined in `0.13.1`. In `0.13.0`, `pytket` was a fixed dependency. I think this is what caused the weird behaviour in the first place, this resolution upgrade and suddenly there are less constraints. But I am unsure.

aborgna-q: Does uv not complain when using invalid extras‽

maximilianruesch: Wait that is odd, I thought I upgraded? Where has my change gone?

maximilianruesch: image

Indeed it does complain, but only as a warning. We could even leave this as `0.13.0` for now, since with the extra (and ignoring the warning) both versions require pytket... 💀

aborgna-q: 🐍 🤷

aborgna-q (approved):

Issue #1546, opened 2026-03-03
[Docs]: Speed up variational loop in QAOA example


CalMacCQ:
### Documentation Type

Other

### Documentation Location

QAOA example notebook added in #1331 

https://github.com/Quantinuum/guppylang/blob/main/examples/qaoa_maxcut_example.ipynb

### Issue Description

The example notebook runs very slowly. It takes up to 2 minutes to solve a maxcut instance on a graph with seven nodes. This is much slower than an equivalent program in pytket. The reason for this is that the main function has to be recompiled on every QAOA iteration. This may be resolvable if we have some way to compile the main function once and then re-evaluate for different parameters

This is the more difficult of two issues with the QAOA notebook. See also #1545 

### Suggested Improvement

(may have to wait for new features) figure out a way to avoid recompiling the main function on every iteration

CalMacCQ:
> Passing as comptime args to what? Are they not already comptime args to the qaoa instance? If you fetch them from the outside, you will have to make sure you refetch the guesses on every run (and thus still need to recompile). I struggle to see where the performance improvement will come from.

You're right I think. They're already comptime arguments effectively. I managed to confuse myself.

PR #1819, opened 2026-06-06
docs: use scipy optimizer in QAOA MaxCut example

### LINKED ISSUE #1545: [Docs]: Use a real-world classical optimizer in the QAOA MaxCut example

spital:
## Summary

- replace the random QAOA parameter sampling loop with a `scipy.optimize.minimize` variational loop using COBYLA
- split QAOA execution from the optimizer objective so `eval_qaoa_energy` returns only scalar energy
- run one final QAOA pass with the optimized parameters for validation and plotting
- document the classical optimizer vs. Guppy quantum-program boundary, including the angle coefficient convention
- add SciPy to the examples dependency group and refresh `uv.lock`

The performance/recompilation limitation remains documented in the notebook and is intentionally left to #1546.

## Testing

- `uv lock --check`
- `uv run ruff check`
- `uv run ruff format --check`
- `uv run pytest tests/integration/test_notebooks.py`
- `uv run jupyter nbconvert --to notebook --execute --ExecutePreprocessor.timeout=1200 --output-dir /tmp/tmp.P7p9GROppQ --output qaoa_executed.ipynb examples/qaoa_maxcut_example.ipynb`

Manual QAOA notebook execution reproduced the committed output:

```text
array([5.099, 1.736, 2.63 , 4.751, 4.831, 2.361, 2.2  ])
'Success ratio: 0.4015'
```

## AI disclosure

I used Codex, Claude, and Hermes as AI assistants to help prototype, review, compare related GitHub state, and draft this PR text. I manually reviewed the resulting diff, checked the GitHub issue/PR state, and locally ran the tests and notebook execution listed above. The implementation and PR are submitted with human oversight; any remaining mistakes are mine.

Closes #1545


spital:
Rebased onto latest `main` and force-pushed to clear the conflicts (only `pyproject.toml`/`uv.lock` from the ruff bump + 1.0.0-a2 release). Re-ran ruff, `uv lock --check`, and the notebook tests (15 passed) after the rebase.

pyproject.toml
```diff
@@ -5,7 +5,7 @@ dev = [
   { include-group = "test" },
 ]
 docs = ["furo>=2024.8.6", "sphinx >=7.2.6,<9"]
-examples = ["matplotlib>=3.10.9", "networkx>=2.6,<4"]
+examples = ["matplotlib>=3.10.9", "networkx>=2.6,<4", "scipy>=1.14,<2"]
```
CalMacCQ: I don't think it should be necessary to list scipy here. Scipy is a dependency of pytket which is in turn is a dependency of guppylang. Therefore I think it shouldn't be needed in pyproject.toml explicitly. Is the version upper bound relevant to your scipy optimizer? 

Scipy is listed as a dependency of pytket here -> https://github.com/Quantinuum/tket/blob/main/pytket/setup.py#L175

examples/qaoa_maxcut_example.ipynb
CalMacCQ, line 1: If you don't mind sharing... can I ask how you are running the notebook locally?

On my M4 macbook pro the notebook the cell running `solve_max_cut_instance` takes >90s to run on your branch locally (quite slow).

I noticed that in your commited change that its much faster.

Your output
```
CPU times: user 29.2 s, sys: 1.59 s, total: 30.8 s
Wall time: 36.3 s
```

Maybe there is something slowing my machine down that I'm now aware of. This seems like a good perfomance improvement in addition to the higher expectation value.

CalMacCQ (commented):
Thanks for your PR! Sorry for taking a few days to get to the review.

Your changes look good. The expectation value is higher and the cells seem to run quite a bit faster compared to my local execution.

Made one comment about the runtime and another about the scipy dependency.

pyproject.toml
spital: Good catch, you're right. scipy comes in transitively as a core dependency (`guppylang → pytket`, and pytket pins `scipy>=1.13.1`), so listing it in the `examples` group is redundant. I've dropped it from `pyproject.toml` and reverted the `uv.lock` lines, so the PR is now just the notebook change. The upper bound wasn't needed either, as the notebook only uses the stable `minimize(method="COBYLA")` API.

examples/qaoa_maxcut_example.ipynb
spital: I run it with `jupyter nbconvert --execute` or manually via `jupyter lab` (and reproduce the same numbers in a plain `uv run python`). On my AMD Ryzen 9 5900X the `solve_max_cut_instance` cell takes ~36 s, which is matching the committed output, so the timing isn't cherry-picked.

The run is deterministic (fixed seed + seeded shots), so COBYLA always does the same 77 objective evaluations regardless of machine; the wall-time difference is purely the per-evaluation cost. Profiling one forward pass here: ~336 ms is the Guppy→Selene **recompile** and only ~82 ms is the 7-qubit simulation. That recompile-per-evaluation is the bottleneck the notebook calls out in the markdown cell and that #1546 tracks.
It's single-threaded, so on your M4 each compile just costs more (~1.15 s/pass → ~90 s total).
Same amount of work, slower compile step. Nothing in the notebook forces extra iterations.

spital:
closing, issue closed with competing PR merged

PR #1824, opened 2026-06-08
fix: Comptime attribute access in struct definitions

### LINKED ISSUE #1816: [Bug]: Regression for using dataclasses inside struct comptime contexts
### Description

The following program compiles with `guppylang-v1.0.0-a0` but does not compile with `guppylang-v1.0.0-a1`:
```python
from guppylang import guppy, array, qubit, comptime
from dataclasses import dataclass

@dataclass
class Config:
    size: int

config = Config(size=5)

@guppy.struct
class Struct:
    qs: array[qubit, comptime(config.size)]

@guppy
def main() -> None:
    s = Struct(array(qubit() for _ in range(comptime(config.size))))
    for q in s.qs:
        discard(q)

main.compile()
```
For the latter, it instead gives
```
Error: Expected a module (at <In[1]>:13:30)
   | 
11 | @guppy.struct
12 | class Struct:
13 |     qs: array[qubit, comptime(config.size)]
   |                               ^^^^^^ Expected a module

```

### Steps to Reproduce

_No response_

### Expected Behaviour

There should be no regression.

### Actual Behaviour

There is a regression.

### Component

_No response_

### Environment

```markdown

```

### Additional Context

Following a git bisect, the commit 96dfb01edc3df0b3b9c70499f4b0dc23da80d022 has been identified as the first bad commit, introduced by PR #1809. This is likely to be the culprit, since it modified struct validation, and there may have been accidental changes. A regression test case should be added.

maximilianruesch:
The recursiveness checker introduced in #1809 also recursed into comptime calls that were nested inside type annotations in structs. In these, arbitrary Python expressions are allowed, which are not necessarily types.

Currently, I do not think that there is a way to have a Guppy type result from a comptime expression, so for now I just forbid the recursive type annotation search from recursing into comptime expressions.

Fixes #1816

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/parsing.py
+++ b/guppylang-internals/src/guppylang_internals/tys/parsing.py
@@ -274,6 +274,12 @@ def _parse_delayed_annotation(ast_str: str, node: ast.Constant) -> ast.expr:
 
 
 def annotation_nodes(node: ast.expr) -> Iterator[ast.expr]:
+    """Iterates over all the Guppy type annotations (recursively) contained in the given
+    expression. Parses delayed annotations, and does not recurse into comptime
+    expressions."""
+    if is_comptime_expression(node):
+        return
+
     if isinstance(node, ast.Constant) and isinstance(node.value, str):
         node = _parse_delayed_annotation(node.value, node)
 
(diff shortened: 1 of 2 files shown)


mark-koch (approved):
Thanks! 

> Currently, I do not think that there is a way to have a Guppy type result from a comptime expression, so for now I just forbid the recursive type annotation search from recursing into comptime expressions.

Exactly! Once we have comptime versions of types, we'll need a separate recursiveness check anyways, so this fine

PR #1828, opened 2026-06-08
feat!: Use mutable state for PCG32 and encapsulate internals

### LINKED ISSUE #1821: [Feature]: Use mutable struct for `PCG32` and make internals private
### Problem Statement

* Turn `PCG32` into a mutable struct
* Update the interface to change the state in place instead of returning a new object
* Make `PCG32.{state, inc}` private by renaming to `_state` and `_inc`

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

mark-koch:
Closes #1821

BREAKING CHANGE: `PCG32.next_int` now mutates the rng in place instead of returning a new state
BREAKING CHANGE: Members of `PCG32` are now private

### DIFF
--- a/guppylang/src/guppylang/std/random.py
+++ b/guppylang/src/guppylang/std/random.py
@@ -34,29 +34,26 @@ def _uint32_to_signed(value: nat) -> int:
     return int(masked)
 
 
-@guppy.struct
+@guppy.struct(frozen=False)
 class PCG32:
     """A deterministic 32-bit random number generator using the PCG32 algorithm.
 
     Unlike :py:class:`guppylang.std.qsystem.random.RNG`, PCG32 keeps its state in a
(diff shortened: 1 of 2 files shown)


guppylang/src/guppylang/std/random.py
```diff
@@ -34,29 +34,26 @@ def _uint32_to_signed(value: nat) -> int:
     return int(masked)
 
 
-@guppy.struct
+@guppy.struct(frozen=False)
```
maximilianruesch: By now structs are mutable by default, so this can be removed again, yes?

mark-koch: Yes, indeed 👍

nicolaassolini-qntm (approved):
👍

Issue #1746, opened 2026-05-18
Output Order edges for each effect


acl-cqc:
When inserting order edges (see below), build one chain per effect/resource; include each op/call only in the chain for effects/resources the op/called-function uses, i.e. calls to functions with `max_effects=[]` are not included in the chain, but any unannotated function is included in all chains (a conservative overapproximation).

Initially there are no effects/resources, so we'll have one chain only for functions that have no max-effects annotation (i.e. effects are unknown / the universe), but in principle it should be a map from Effect to last-node that had that effect.

This will require some refactoring: currently order edges are added in the [`track_hugr_side_effects` contextmanager](https://github.com/Quantinuum/guppylang/blob/857e799a0df90f00d0ab97b2d056725ea2b67277/guppylang-internals/src/guppylang_internals/compiler/core.py#L455) using an actual ["monkey-patch" method assignment](https://github.com/Quantinuum/guppylang/blob/857e799a0df90f00d0ab97b2d056725ea2b67277/guppylang-internals/src/guppylang_internals/compiler/core.py#L509-L510), so only gets to see the *Hugr node*, whereas we'll need to inspect the *guppy* FunctionType/callee/something.

PR #1829, opened 2026-06-08
refactor!: Replace track_hugr_side_effects monkeypatch with hierarchy of builder classes

acl-cqc:
* Unparametrize `DFBuilder` (the parameter wasn't really used), and make into a subclass hierarchy, allowing subclasses to define a couple of type-specific methods (`set_loop_outputs` and `set_block_outputs`)
* Subclasses also know about parent hierarchy and how to propagate edges upwards
* Add a few more methods, thus allowing to hide `raw_builder`
* Use DFBuilder more, much more, including tracing, indeed hopefully everywhere that calls add_op

There are a couple of optional-extra commits that I'm willing to retract, designed to remove the possibility of accidentally forgetting to use the new Builder classes...thoughts welcome on benefits/removal of

* commit dcd1c788588885148881d08cc83a40e1c061127e changes `Context.declare_global_func` to return an `hf.Function | FunctionBuilder` rather than `hf.Function, bool`; a lot of `isinstance` :(

* commit abe78a480e2e661bd8fa80d9d5680a3aeb49339e adds a CondBuilder for conditionals (*not* a DFBuilder); duplicating yet more from the underlying hugr hierarchy, and needing contextmanager methods (hugr Conditional is a contextmanager which checks that all cases have been constructed - hugr Case is also a contextmanager but I've stopped us from using that as it doesn't do anything. I *could* add ctx-mgr methods to CaseBuilder too...).

This is a big step towards https://github.com/Quantinuum/guppylang/issues/1746 but doesn't intend to change any part of the Hugr. (Rather it opens the way to replacing `DFBuilder.add_op` with something that takes a list of effects.)

**Question** should I move these classes (`*Builder` - DFBuilder, CaseBuilder, CondBuilder, ...) into `compiler/builder.py`?

BREAKING CHANGE: (guppylang-internals) DFBuilder no longer parametrized, now an ABC - construct (Case/TailLoop/Function/Block)Builders with relevant (grand/)parents. This *must* be used for all add_ops to preserve (currently-total) ordering of side effects.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/compiler/core.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/core.py
@@ -340,7 +340,7 @@ class DFBuilder(ABC, ToNode):
                 self.raw_builder.hugr[self._last_stateful_op].op, ops.Output
             )
         node = op_node.to_node()
-        if self._last_stateful_op != node: # avoid self-loops when propagating
+        if self._last_stateful_op != node:  # avoid self-loops when propagating
             self.raw_builder.add_state_order(self._last_stateful_op, node)
             self._last_stateful_op = node
 
@@ -384,12 +384,7 @@ class DFBuilder(ABC, ToNode):
     def add_tail_loop(
         self, just_inputs: Sequence[Wire], rest: Sequence[Wire]
     ) -> "TailLoopBuilder":
-        # Can't state this in the return type without adding a second type parameter
-        # to DFBuilder, which is horrendous (at least until we can provide
-        # default values for type parameters, which arrives in PEP 0696 / python 3.13).
-        # So just get a static check here
-        raw: TailLoop = self.raw_builder.add_tail_loop(just_inputs, rest)
-        return TailLoopBuilder(raw, self)
+        return TailLoopBuilder(self.raw_builder.add_tail_loop(just_inputs, rest), self)
 
     def load(
         self, const: ToNode | val.Value, const_parent: ToNode | None = None
@@ -471,7 +466,6 @@ class BlockBuilder(DFBuilder):
 
     def set_block_outputs(self, branching: Wire, *other_outputs: Wire) -> None:
         self.raw_builder.set_outputs(branching, *other_outputs)
-        self._outputs_set = True
         if self._last_stateful_op is not None:
             self._handle_side_effects(self.raw_builder.output_node)
 
--- a/guppylang-internals/src/guppylang_internals/compiler/core.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/core.py
@@ -271,15 +271,13 @@ class DFBuilder(ABC, ToNode):
 
     Manages attaching debug information, which requires keeping track of the most
     relevant AST node for each operation being compiled with `current_ast_node`.
-
-    The underlying builder can still be accessed through `raw_builder`.
     """
 
     current_ast_node: AstNode | None = field(default=None, kw_only=True)
     _last_stateful_op: Node | None = field(default=None, init=False)
 
     @abstractproperty
-    def raw_builder(self) -> hf.Function | Case | TailLoop | Block:
+    def _raw(self) -> hf.Function | Case | TailLoop | Block:
         """The underlying Hugr dataflow graph builder."""
 
     @contextmanager
@@ -299,27 +297,27 @@ class DFBuilder(ABC, ToNode):
     def define_function(
         self, name: str, input_types: ht.TypeRow, output_types: ht.TypeRow
     ) -> hf.Function:
-        return self.raw_builder.module_root_builder().define_function(
+        return self._raw.module_root_builder().define_function(
             name,
             input_types,
             output_types,
         )
 
     def to_node(self) -> Node:
-        return self.raw_builder.to_node()
+        return self._raw.to_node()
 
     @property
     def input_node(self) -> Node:
-        return self.raw_builder.input_node
+        return self._raw.input_node
 
     def inputs(self) -> Sequence[Wire]:
-        return self.raw_builder.inputs()
+        return self._raw.inputs()
 
     def set_outputs(self, *outputs: Wire) -> hf.Function | Case | TailLoop | Block:
-        self.raw_builder.set_outputs(*outputs)
+        self._raw.set_outputs(*outputs)
         if self._last_stateful_op is not None:
-            self._handle_side_effects(self.raw_builder.output_node)
-        return self.raw_builder
+            self._handle_side_effects(self._raw.output_node)
+        return self._raw
 
     def add_op(
         self,
@@ -331,7 +329,7 @@ class DFBuilder(ABC, ToNode):
         """Adds an op to the dataflow graph builder. Set `set_debug_info=False` to
         avoid automatic debug information attachment.
         """
-        op_node = self.raw_builder.add_op(op, *args)
+        op_node = self._raw.add_op(op, *args)
         if may_have_side_effect(op):
             self._handle_side_effects(op_node)
 
@@ -347,12 +345,10 @@ class DFBuilder(ABC, ToNode):
             self._propagate_side_effects()
             self._last_stateful_op = self.input_node
         else:
-            assert not isinstance(
-                self.raw_builder.hugr[self._last_stateful_op].op, ops.Output
-            )
+            assert not isinstance(self._raw.hugr[self._last_stateful_op].op, ops.Output)
         node = op_node.to_node()
         if self._last_stateful_op != node:  # avoid self-loops when propagating
-            self.raw_builder.add_state_order(self._last_stateful_op, node)
+            self._raw.add_state_order(self._last_stateful_op, node)
             self._last_stateful_op = node
 
     @abstractmethod
@@ -370,7 +366,7 @@ class DFBuilder(ABC, ToNode):
         """Calls a static function in the graph. Set `set_debug_info=False` to
         avoid automatic debug information attachment.
         """
-        call = self.raw_builder.call(
+        call = self._raw.call(
             func, *args, instantiation=instantiation, type_args=type_args
         )
         self._handle_side_effects(call)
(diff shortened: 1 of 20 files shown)


guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -275,11 +260,11 @@ def __contains__(self, place: Place) -> bool:
     def __copy__(self) -> "DFContainer":
         # Make a copy of the var map so that mutating the copy doesn't
         # mutate our variable mapping
-        return DFContainer(self.builder.raw_builder, self.ctx, self.locals.copy())
+        return DFContainer(self.builder, self.ctx, self.locals.copy())
```
acl-cqc, line 257: Share the DFBuilder as this contains `last_stateful_op`

guppylang-internals/src/guppylang_internals/compiler/core.py
acl-cqc: Need to add some doc here

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
acl-cqc, line 17: This is only used once - tempting to inline

guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -316,14 +331,32 @@ def add_op(
...
+        else:
+            assert not isinstance(self._raw.hugr[self._last_stateful_op].op, ops.Output)
+        node = op_node.to_node()
+        if self._last_stateful_op != node:  # avoid self-loops when propagating
+            self._raw.add_state_order(self._last_stateful_op, node)
+            self._last_stateful_op = node
+
+    @abstractmethod
+    def _propagate_side_effects(self) -> None:
+        """Subclasses must implement"""
```
mark-koch: Could you make this more descriptive?

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -819,9 +818,9 @@ def array_read_bool(ctx: CompilerContext) -> hf.Function:
         params=[],
         body=ht.FunctionType([OpaqueBool], [ht.Bool]),
     )
-    func, already_defined = ctx.declare_global_func(ARRAY_READ_BOOL, sig)
-    if not already_defined:
-        func.set_outputs(func.add_op(read_bool(), func.inputs()[0]))
+    func = ctx.declare_global_func(ARRAY_READ_BOOL, sig)
+    if isinstance(func, FunctionBuilder):
+        func = func.set_outputs(func.add_op(read_bool(), func.inputs()[0]))
```
mark-koch, line 822: I think the previous version was a lot more readable. What's the motivation for changing this interface?

guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
```diff
@@ -55,14 +59,14 @@ def compile_modified_block(
...
-    func_builder = dfg.builder.raw_builder.module_root_builder().define_function(
+    func_builder = dfg.builder.define_function(
         str(modified_block), hugr_ty.input, hugr_ty.output
     )
     add_metadata(
         func_builder,
         additional_metadata={"unitary": modified_block.ty.unitary_flags.value},
     )
-
+    func_builder = FunctionBuilder(func_builder)
```
mark-koch: Should `define_function` already return a `FunctionBuilder`? (and then store those inside e.g. `CompiledFunctionDef`?)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
```diff
@@ -125,13 +124,18 @@ def build_unwrap_either(
...
-        error = build_static_error(case, error_signal, error_msg)
+
+    case = conditional.add_case(ok_case_num)
+    case.set_outputs(*case.inputs())
+
+    case = conditional.add_case(panic_case_num)
+    error = build_static_error(case, error_signal, error_msg)
+    with ExitStack() as es:
+        if builder.current_ast_node is not None:
+            es.enter_context(case.set_ast_context(builder.current_ast_node))
```
mark-koch: Why is this needed?

guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -316,14 +331,32 @@ def add_op(
...
         return op_node
 
+    def _handle_side_effects(self, op_node: ToNode) -> None:
+        if self._last_stateful_op is None:
+            self._propagate_side_effects()
+            self._last_stateful_op = self.input_node
+        else:
+            assert not isinstance(self._raw.hugr[self._last_stateful_op].op, ops.Output)
+        node = op_node.to_node()
+        if self._last_stateful_op != node:  # avoid self-loops when propagating
```
mark-koch: `_last_stateful_op` could only ever be `node` if it's the input, right?

mark-koch (approved):
Nice! Not as awful to review as you made it sound :)

> commit https://github.com/Quantinuum/guppylang/commit/dcd1c788588885148881d08cc83a40e1c061127e changes Context.declare_global_func to return an hf.Function | FunctionBuilder rather than hf.Function, bool; a lot of isinstance :(

Not a big fan of this, I feel the old version looked clearer imo

> commit https://github.com/Quantinuum/guppylang/commit/abe78a480e2e661bd8fa80d9d5680a3aeb49339e adds a CondBuilder for conditionals (not a DFBuilder);

I think that's good, otherwise we wouldn't get the automatic effect propagation, right?

> Question should I move these classes (*Builder - DFBuilder, CaseBuilder, CondBuilder, ...) into compiler/builder.py?

Yes I think it would be good to move them into their own module 👍

guppylang-internals/src/guppylang_internals/compiler/core.py
acl-cqc: If you're in a Case and you find a first stateful node, you propagate upwards and mark the Conditional as stateful.
Then if you're in the other Case and find a first stateful node in that arm...you propagate upwards again and would mark the Conditional again.

guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
acl-cqc: Debug information. ExitStack was necessary because `set_ast_context` only accepts a Node (and rejects None), whereas the old explicit `DFGBuilder(...)` (or now `CaseBuilder(...)`) accepts an option

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
acl-cqc: The idea was that you should only build it using a FunctionBuilder...it would have been better to have changed `hf.Function` for the already-built case to `ops.FuncDefn`.

I've reverted this now...

acl-cqc: (reverted)

guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
acl-cqc: Yeah, it was the latter I was holding off on. Doesn't feel like we really want a FunctionBuilder floating around - it really should only be used once. But technically an `hf.Function` was just as bad...however, I've hidden the builder.

This does mean that providing the builder as an explicit kwarg now looks bad (name begins with underscore), so the desired semantics could be provided via:
```
    func_bldr: InitVar[FunctionBuilder] = field(default=None)
    _func_bldr: FunctionBuilder = field(init=False)

    def __post_init__(self, func_bldr: FunctionBuilder) -> None:
        object.__setattr__(self, "_func_bldr", func_bldr) # bypass frozen=True
```

indeed one could even turn `_func_builder` into a `FunctionBuilder | Node` and replace the builder with just the node when building is done (see d33aeceb23906f439aa7b000b4a79af3c2c88631 but having to use `object.__setattr__` again, bah - I've reverted that. Sadly one cannot have the new field mutable when inheriting from a frozen superclass....oh for composition rather than inheritance!)

guppylang-internals/src/guppylang_internals/std/_internal/compiler/prelude.py
acl-cqc: I've now updated `set_ast_context` to take an option, avoiding the need for the ExitStack

acl-cqc:
Ok so I have verified equivalence vs before this PR for all the integration tests (`just export-integration-tests`). Binary comparison of hugrs produces many false differences, but comparing op class (e.g. `FuncDefn`) confirms no changes to nodes (there are changes to op instances - between runs of the same codebase, both before and after this PR), and no changes to the edges in all bar two testcases, https://github.com/Quantinuum/guppylang/blob/97c69624a1181804e8b2c1685941a394bae84d19/tests/integration/test_pytket_circuits.py#L201 and https://github.com/Quantinuum/guppylang/blob/97c69624a1181804e8b2c1685941a394bae84d19/tests/integration/test_pytket_circuits.py#L211, which start to get new/extra Order edges with this PR.

This is because `track_hugr_side_effects` is only applied in `CompilerContext.iterate_worklist`, which calls  `compile_inner` of items on the worklist. For `ParsedPytketDef` the Hugr is built in `compile_outer`, called via `CompilerContext.build_compiled_def`. So if a non-pytket function is on the worklist, but that calls a pytket one, the callee will be processed inside `compile_inner` for the caller, during the scope of `track_hugr_side_effects`. However, if the pytket function is compiled *directly* (e.g. via `compile_function`), as in these tests, then `CompilerContext.build_compiled_def` called `ParsedPytketDef.compile_outer` *outside* `CompilerContext.iterate_worklist` and thus without `track_hugr_side_effects` being applied.

Whilst adding extra Order edges to pytket circuits seems bad, the old scheme does not seem the right criteria for removing them (in only a few cases!), we should have a general solution regardless of how we get to `ParsedPytketDef.compile_outer` - this should happen as part of the effect system work. I've added a note in `definition/pytket_circuits.py`,

PR #1846, opened 2026-06-11
feat(effects)!: Improve error messages by passing callee into check_call/synthesize_call

acl-cqc:
This significantly improves a lot of the error messages for effects violations, and perhaps might be applied to other errors too?

Note that tensor'd functions are gonna have terrible messages both before and after, I haven't got a test of these yet as they are marked experimental, note comment - we'd need significant refactoring to sort that.

BREAKING CHANGE: (guppylang-internals) top-level `check_call` and `synthesize_call` methods take a tuple of `str | DefId | None` to identify the callee

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -1445,7 +1445,7 @@ def _check_effects(
 
 
 def synthesize_call(
-    target: FunctionType,
+    func_ty: FunctionType,
     args: list[ast.expr],
     node: AstNode,
     ctx: Context,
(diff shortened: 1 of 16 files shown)


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1328,14 +1345,32 @@ def check_comptime_arg(
...
+def _identify_callee(node: ast.expr) -> str | None:
+    match node:
+        case PlaceNode(place=Variable(name=n)):
+            return n
+        case _:
+            return None
+
+
+def _check_effects(
+    target: tuple[str | DefId | None, FunctionType], ctx: Context, node: AstNode
```
tatiana-s: In my opinion it would be better to have a class instead of tuples for readability - especially given the call graph will just store this information instead of using it here directly so this would also simplify the type of the graph

acl-cqc: No problem with making this a class, but is it the same as the kind of "node" in the callgraph? Might you want all callgraph nodes to have concrete DefIds, as only such could have outgoing edges?

acl-cqc: Turned into extra parameter!

PR #1850, opened 2026-06-12
feat: Add platform configuration to emulator 

### LINKED ISSUE #1797: [Feature]: Allow emulation platform to be configurable

ss2165:
Closes #1797

Tests passing locally with selene-sim 0.3.0 alpha, ~requires a release to merge~ 0.3.0a1 released.

Integration execution tests deterministically but randomly choose a platform to get coverage across both platforms.


guppylang/src/guppylang/emulator/builder.py
```diff
@@ -95,6 +117,11 @@ def with_build_arg(self, key: str, value: Any) -> Self:
         which saves LLVM IR into the build directory rather than
         saving bitcode.
         """
+        if key == "platform":
+            raise ValueError(
+                "Use with_platform() to set the target platform instead of "
```
croyzor: This is good for redirecting people to the right thing, but if the `custom_args` are customisable, one of them could have key `platform`?

guppylang/src/guppylang/emulator/builder.py
```diff
@@ -122,7 +143,8 @@ def build(self, package: Package, n_qubits: int) -> EmulatorInstance:
             progress_bar=self._progress_bar,
             strict=self._strict,
             save_planner=self._save_planner,
-            **self._custom_args or {},
+            platform=self._platform,
+            **custom_args,
```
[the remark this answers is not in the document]
croyzor: this explains the above

guppylang/src/guppylang/emulator/builder.py
```diff
@@ -20,6 +20,12 @@
     from selene_core import BuildPlanner, QuantumInterface, Utility
 
 
+#: The quantum platform to target when building an emulator instance.
+#: ``"helios"`` is the default and targets the Helios QIS.
+#: ``"sol"`` targets the Sol QIS.
+Platform = Literal["helios", "sol"]
```
croyzor, line 26: Seems like this should be an enum, but my python-foo isn't good enough to say for sure

tests/integration/conftest.py
```diff
@@ -85,7 +104,9 @@ def f(
         expected: Any,
         num_qubits: int | None = None,
         args: list[Any] | None = None,
+        platform: Platform | None = None,
     ):
+        resolved_platform = platform if platform is not None else default_platform
```
croyzor:
```suggestion
        resolved_platform = platform or default_platform
```

guppylang/src/guppylang/emulator/builder.py
ss2165: we use this literal pattern for platform throughout the stack now + with literals you get the same type safety as with enums (but saves an import)

tests/integration/test_qsystem.py
```diff
@@ -8,30 +8,44 @@
...
-        pytest.param(helios_mod, id="helios"),
-        pytest.param(sol_mod, id="sol"),
+        pytest.param(QsysMod(helios_mod, "helios"), id="helios"),
+        pytest.param(QsysMod(sol_mod, "sol"), id="sol"),
     ]
 )
-def qsys_mod(request):  # type: ignore[no-untyped-def]
-    """Fixture providing either the helios or sol qsystem module."""
+def qsys_mod(request: pytest.FixtureRequest) -> QsysMod:
+    """Fixture providing the helios or sol qsystem module with its platform name."""
```
croyzor:
```suggestion
    """Fixture providing the helios or sol qsystem module, with its platform name."""
```

tests/integration/conftest.py
```diff
@@ -71,11 +78,23 @@ def validate_impl(package: Package | PackagePointer | Hugr, name=None):
     return validate_impl
 
 
+@pytest.fixture
+def random_platform(request) -> Platform:
```
croyzor: It seems bad to introduce randomness into the testing. This is deterministic across runs, but what do we do if I add a new test and old ones start failing because the test tree has changed?

I'm not sure what's a better way of getting test coverage of both platforms without running every test twice, maybe just running the qsystem integration tests twice?

ss2165: do you have an alternative in mind that could get us equal coverage between the platforms on the integration tests?

guppylang/src/guppylang/emulator/builder.py
ss2165: yes I think this is overkill, reverting

tests/integration/conftest.py
croyzor: Maybe we could have a config option (argument to pytest or environment variable) that specifies the platform, and use it to run both variations in CI? Similar to how I never test against other python versions locally

ss2165: thanks, will try

ss2165: Done, added a step to python 3.12 check to run only the execution tests again with sol platform

croyzor (approved):

PR #1853, opened 2026-06-12
refactor!: Split internal decorators into modules and unify calling frame detection

maximilianruesch:
Does several things (closely tied so one PR):
- Splits the module `guppylang_internals.decorator` into several submodules according to their function and moves the external module decorator factory factory (yes) into `guppylang_internals.std._internals.decorator`. This helps separate concerns for upcoming GPU functionality in `guppylang_internals` (see #1854), and is more consistent with existing compilers, definitions and so on in the package.
- Moves `guppylang_internals.wasm_util` into `guppylang_internals.std._internals.wasm`, again for consistency reasons (wasm was the only functionality module with a top level util file, and instead we store all common utils for the std lib in `guppylang.std._internal`).
- Cleans up all places in `guppylang_internals` where calling frame detecting was used (three different kinds), since having three different ones made it pretty hard to distinguish between them and keep them properly maintained.

For some backward compatibility / convenience I still expose the decorators on the `guppylang_internals.decorators` path, but I can also change that if need be.

BREAKING CHANGE: Splits the module `guppylang_internals.decorator` into submodules, removing some internal functionality.
BREAKING CHANGE: Moves `guppylang_internals.wasm_util` into `guppylang_internals.std._internals.wasm`.
BREAKING CHANGE: Refactors calling frame detection and related methods from `guppylang_internals.tracing.util` and several other places to `guppylang_internals.frame_util`.

BEGIN_COMMIT_OVERRIDE
refactor: Split internal decorators into modules and unify calling frame detection (#1853)
END_COMMIT_OVERRIDE


guppylang-internals/src/guppylang_internals/frame_util.py
```diff
@@ -0,0 +1,49 @@
...
+from types import FrameType, ModuleType, TracebackType
+
+
+def get_calling_frame(*, skip_main_lang: bool = True) -> FrameType:
+    """Finds the first frame that called this function that either has an unknown
+    module, or the module is outside the compiler packages.
+
+    :param skip_main_lang: Whether to skip frames that belong to modules in the main
+        guppylang language package. Set to False if you want to resolve calling frames
+        in e.g. the sta lib.
```
nicolaassolini-qntm:
```suggestion
        in e.g. the std lib.
```

guppylang/src/guppylang/decorator.py
```diff
@@ -197,15 +197,15 @@ def decorator(
...
         # Set `return_class=True` to match the old behaviour until this deprecated
         # method is removed
-        import guppylang_internals.decorator
+        import guppylang_internals.decorator.ty
 
-        return guppylang_internals.decorator.extend_type(defn, return_class=True)
+        return guppylang_internals.decorator.ty.extend_type(defn, return_class=True)
 
-    @deprecated("Use @guppylang_internal.decorator.custom_type instead.")
+    @deprecated("Use @guppylang_internal.decorator.custom.custom_type instead.")
```
nicolaassolini-qntm, line 208: the imports at the top of the module are from `guppylang_internals.decorator`, I think we should have the suggestion and the imports in this file coherent

guppylang-internals/src/guppylang_internals/decorator/__init__.py
```diff
@@ -0,0 +1,12 @@
...
+from .wasm import wasm, wasm_module
+
+__all__ = [
+    "custom_function",
+    "custom_type",
+    "extend_type",
+    "hugr_op",
+    "wasm",
+    "wasm_module",
+]
```
nicolaassolini-qntm, line 12: I guess that you added this to avoid breanking imports in the `std`. Do we really need this?

nicolaassolini-qntm (commented):
Just a few questions

guppylang-internals/src/guppylang_internals/decorator/__init__.py
maximilianruesch: More of a convenience. Perhaps not really needed, but useful?

nicolaassolini-qntm (approved):

PR #1854, opened 2026-06-12
feat: Support for GPU modules

maximilianruesch:
Supports external GPU modules (effectively integrating https://github.com/Quantinuum/guppy-gpu into this repository).

For this, a further refactor (after #1853 already did some lifting) of the external module decorator factory factory was made, to generalise it beyond just wasm and remove some of the complexity. Protocol were added to clarify meaning in callbacks (especially for multiple arguments of the same type).

I pose the open question of whether we should go the step of removing one factory layer, and just let `create_type_def` fetch the filename and config name from the outer scope, further reducing complexity.

Based on #1853


guppylang-internals/src/guppylang_internals/std/_internal/gpu.py
```diff
@@ -0,0 +1,27 @@
+from dataclasses import dataclass
+from typing import ClassVar
+
+from guppylang_internals.diagnostic import Error
+from guppylang_internals.tys.ty import Type
+
+
+@dataclass(frozen=True)
+class GpuError(Error):
```
nicolaassolini-qntm, line 9: can we have a test for the new errors?

nicolaassolini-qntm: Also this files contains only errors, and ther is already a guppylang_internals/std/_internal/compiler/gpu.py file, I would consider calling this file `gpu_errors.py`

guppylang-internals/src/guppylang_internals/decorator/gpu.py
```diff
@@ -0,0 +1,93 @@
...
+        GpuModuleDiscardCompiler(),
+        init_arg=False,
+    )
+    return create_decorator(filename, config_filename)
+
+
+@overload
+def gpu(arg: Callable[P, T]) -> GuppyFunctionDefinition[P, T]: ...
+@overload
+def gpu(arg: int) -> Callable[[Callable[P, T]], GuppyFunctionDefinition[P, T]]: ...
```
nicolaassolini-qntm:
```suggestion
def gpu(fn_id: int) -> Callable[[Callable[P, T]], GuppyFunctionDefinition[P, T]]: ...
```
This make the linting of @gpu(42) nicer

guppylang-internals/src/guppylang_internals/decorator/gpu.py
```diff
@@ -0,0 +1,93 @@
...
+        type_def_wrapper,
+        GpuModuleInitCompiler(),
+        GpuModuleDiscardCompiler(),
+        init_arg=False,
+    )
+    return create_decorator(filename, config_filename)
+
+
+@overload
+def gpu(arg: Callable[P, T]) -> GuppyFunctionDefinition[P, T]: ...
```
nicolaassolini-qntm: add a comment saying that this is for `@gpu`

guppylang-internals/src/guppylang_internals/definition/gpu.py
```diff
@@ -0,0 +1,123 @@
...
+
+if TYPE_CHECKING:
+    from guppylang_internals.checker.core import Globals
+
+QSYSTEM_GPU_EXTENSION = gpu()
+
+
+class RawGpuFunctionDef(RawCustomFunctionDef):
+    def sanitise_type(self, loc: AstNode | None, fun_ty: FunctionType) -> None:
+        # Place to highlight in error messages
```
nicolaassolini-qntm: I dont understand the comment

guppylang-internals/src/guppylang_internals/definition/gpu.py
```diff
@@ -0,0 +1,123 @@
...
+                raise GuppyError(FirstArgNotModule(loc, ty))
+        for inp in fun_ty.inputs[1:]:
+            if not self.is_valid_gpu_type(inp.ty):
+                raise GuppyError(UnconvertibleType(loc, inp.ty))
+        if not self.is_valid_gpu_type(fun_ty.output):
+            match fun_ty.output:
+                case NoneType():
+                    pass
+                case _:
+                    raise GuppyError(UnconvertibleType(loc, fun_ty.output))
```
nicolaassolini-qntm: add a small comment/doc string to declare what we're checking here

nicolaassolini-qntm (commented):
Overall good, just some small comments

nicolaassolini-qntm:
since it is big, I'm asking for a review of the 🤖 too

guppylang-internals/src/guppylang_internals/definition/gpu.py
maximilianruesch: Leftover from refactor, I added some new comments and shuffled the functions around a bit to make this clearer

guppylang-internals/src/guppylang_internals/std/_internal/gpu.py
maximilianruesch: Similar to how there is `wasm.py`, I would leave the naming as is (but arguably, the wasm file does not contain just errors).

nicolaassolini-qntm (approved):
LGTM. :shipit: 

Just a final suggestion, in the `GPU signature error`, when a bad type is found, I would add a `Note` saying something like: "Only Numeric types are allowed in gpu functions."

Issue #1019, opened 2025-06-10
Remove deprecated standard library reexports


mark-koch:
* `std/quantum_functional` (now in `std/quantum/functional`)
* Remove deprecated `prelude` reexports (see comments in `prelude.py`)

PR #1855, opened 2026-06-15
refactor!: Remove deprecated decorators, functions, and sync dummy interface

maximilianruesch:
Removes all the deprecated decorators and `@guppy` members, as well as syncing the dummy interface to the current class, and finally removing a deprecated function from the `guppylang_internals` engine. These members raised Python deprecation warnings, so there has been _some_ effort to communicate the deprecation.

Deprecated reexports are generally handled in #1019, and I found no other references to the word "deprecated" (or similar) in the main language or internals.

Closes #1643 

BREAKING CHANGE: Removes deprecated members of `guppy` (such as `@guppy.hugr_op`).
BREAKING CHANGE: Removes the `register_extension` function from the compilation engine.


ss2165 (approved):
thanks

Issue #1858, opened 2026-06-15
[Feature]: Find a better way to check array indexes


nicolaassolini-qntm:
### Problem Statement

To solve #1669, we have to turn off the checking on indexes. We should find a better way to perform this check, considering a static analysis 

### Proposed Solution

I suggest the good old interval analysis

PR #1859, opened 2026-06-15
fix: Disable imprecise array indexing check

### LINKED ISSUE #1669: [Bug]: Guarded array indexing fails bounds check after monomorphization
### Description

Reported by @GBoydQ

```python
@guppy
def array_bounds(a_reg: array[qubit, n]) -> None:
    if n <= 5:
        x(a_reg[4])

@guppy
def main() -> None:
    a_reg = array(qubit() for _ in range(3))
    array_bounds(a_reg)
    discard_array(a_reg)
```

This used to compile, but now fails after the monomorphisation change from #1441.

We should turn off indexing checks until we have better static analysis to determine index validity within if-guards

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
close #1669

To prevent an unsound result, we disabled the check.

An issue about the improvement of the array indexing check has been opened: #1858


tests/error/array_errors/array_index_take.py
maximilianruesch, line 1: unrelated change in this file

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -189,6 +189,8 @@ class ArrayIndexChecker(CustomCallChecker):
     at compile time if it's not.
     """
 
+    expr_index: int
+
```
maximilianruesch, line 193: Unrelated change?

maximilianruesch (changes requested):
Please add a test that checks the guarded case (as shown in the referenced issue) and verifies it works.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
nicolaassolini-qntm: I added it for clarity because it was missing, do you suggest a separate PR?

maximilianruesch: If it was missing, how did things work previously? Was it just dynamically declared? Mypy should have picked up on this.

tests/error/array_errors/array_index_take.py
nicolaassolini-qntm: ruff formatting, I guess

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
nicolaassolini-qntm: The class is defined without `@dataclass` and in the `__init__ `we define `self.expr_index`. I think it is good enough for Mypy.

tests/error/array_errors/array_index_take.py
maximilianruesch: Highly unlikely, as ruff is checked in CI. Try again.

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
maximilianruesch: Ah yes. That should also mean that now you can remove the `: int` type hint from the constructor assignment.

nicolaassolini-qntm:
@maximilianruesch done it

maximilianruesch (approved):

PR #1871, opened 2026-06-16
refactor!: rename result and state_result to output and state_output

### LINKED ISSUE #1656: [Maintenance]: Rename `result` to `output`
### Description

Rename the `result` stdlib function to `output`

The existing functions and stdlib module should be deprecated aliases which may optionally be removed when guppy reaches v1. Blocked by #1654 so deprecation warnings can be issued.

### Motivation

"result" doesn't accurately describe the behaviour of the function, which is much more like "print". It is also overloaded, "Results" are also what is returned by execution in the python environment. 

### Scope

guppylang.std.result

### Component

_No response_

### Priority

Medium - Should be done soon

### Additional Context

_No response_

ss2165:
Closes #1656 

Retains deprecated aliases so shouldn't break v0 user code

Updated internal symbols to match, updated integration tests and examples.
Added unit tests to check deprecated aliases continue to validate.

Downside: HUGR extension/op names now don't match the user facing function. Not worth the extension break/migration IMO.

BREAKING CHANGE: Result compiler/checker symbols renamed to Output equivalents


guppylang-internals/src/guppylang_internals/nodes.py
```diff
@@ -532,14 +532,14 @@ def __init__(self, args: list[ast.expr], func_ty: FunctionType) -> None:
...
+    """A `state_output(tag, *args)` expression."""
 
     tag_value: Const
     tag_expr: ast.expr
     args: list[ast.expr]
     func_ty: FunctionType
-    #: Array length in case this is an array result, otherwise `None`
+    #: Array length in case this is an array output, otherwise `None`
     array_len: Const | None
     _fields = ("tag_value", "tag_expr", "args", "func_ty", "has_array_input")
```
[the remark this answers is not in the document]
ss2165: unrelated

guppylang/src/guppylang/std/debug.py
```diff
@@ -3,18 +3,22 @@
...
 
     Args:
-        tag: A string literal representing the tag of the result.
+        tag: A string literal representing the tag of the state output.
         args: The qubits whose quantum state is to be reported. The order they are given
         in corresponds to the order in which the state will be reported.
     """
+
+
+# Deprecated alias for `state_output`.
```
aborgna-q: State the guppy version that deprecated it.

I assume we don't have deprecation warnings?

aborgna-q (approved):
👍

guppylang/src/guppylang/std/debug.py
ss2165: very unlikely to land in time, warnings in general are a big compiler change

PR #1877, opened 2026-06-16
feat(std): add bounded random int to PCG32

### LINKED ISSUE #1874: [Feature]: Add bounded random int to `std.random`
### Problem Statement

Recently, we got a contribution for Guppy-based RNG -> https://github.com/Quantinuum/guppylang/pull/1808

This RNG produces 32-bit ints. Usually, people would use bounded random ints, though: coin flips, dice rolls... The naive way of doing so is `random_int % bound`, but this may introduce biases.

The reference implementation of PCG32 provides a safe way to implement bounded random int, and it's fairly simple: see `pcg32_boundedrand_r` from `pcg_basic.c`, which you can download [here](https://www.pcg-random.org/download.html). It's literally six lines of C code:

```
//     Generate a uniformly distributed number, r, where 0 <= r < bound
uint32_t pcg32_boundedrand_r(pcg32_random_t* rng, uint32_t bound)
{
    uint32_t threshold = -bound % bound;

    for (;;) {
        uint32_t r = pcg32_random_r(rng);
        if (r >= threshold)
            return r % bound;
    }
}
```

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

luffy-orf:
## Summary
- Add `PCG32.next_int_bounded(bound)` using PCG's `pcg32_boundedrand_r` rejection sampling (uniform `[0, bound)` without modulo bias)
- Extract `_next_uint32()` shared by `next_int()` and bounded draws
- Harden PCG32: 64-bit state masking and safe rotate amounts (fixes hugr const-fold shift overflows in emulation)
Fixes #1874
## Test plan
- [x] `uv run pytest tests/integration/std/test_random.py`
- [x] `uv run pre-commit run --all-files`
- [x] `uv run pytest -n auto` (1614 passed)

### DIFF
--- a/guppylang/src/guppylang/std/random.py
+++ b/guppylang/src/guppylang/std/random.py
@@ -23,6 +23,14 @@ def _mask32(value: nat) -> nat:
 
 @guppy
 @no_type_check
+def _mask64(value: nat) -> nat:
+    # 2**64 - 1: PCG32 keeps its state in a 64-bit word (see pcg32_random_r).
+    uint64_mask: nat = 18446744073709551615
+    return value & uint64_mask
+
+
+@guppy
+@no_type_check
 def _uint32_to_signed(value: nat) -> int:
     """Convert a 32-bit unsigned value stored in a nat to a signed 32-bit int."""
     # 2**31: high bit of a 32-bit word; values at or above this map to negative ints.
@@ -45,11 +53,30 @@ class PCG32:
 
         rng = seeded_pcg32(1)
         value = rng.next_int()
+        roll = rng.next_int_bounded(6)
         another = rng.next_int()
     """
 
     _state: nat
     _inc: nat
+
+    @guppy
+    @no_type_check
+    def _next_uint32(self: PCG32) -> nat:
+        """Advance the generator and return the next 32-bit output word."""
+        # LCG multiplier N from the PCG paper / pcg32_random_r (64-bit state, 32-bit
+        # output).
+        pcg32_mult: nat = 6364136223846793005
+        old_state = self._state
+        self._state = _mask64(nat(old_state * pcg32_mult + self._inc))
+        # XSH-RR output permutation: xor-shift then random rotate (see pcg32_random_r).
+        xorshifted = _mask32(
+            ((_mask64(old_state) >> nat(18)) ^ _mask64(old_state)) >> nat(27)
+        )
+        # oldstate >> 59 is at most 31 for 64-bit state; mask for safe hugr shifts.
+        rot = _mask32(_mask64(old_state) >> nat(59)) & nat(31)
+        rot_inv = _mask32(nat(32) - rot) & nat(31)
+        return _mask32((xorshifted >> rot) | (xorshifted << rot_inv))
 
     @guppy
     @no_type_check
@@ -59,17 +86,30 @@ class PCG32:
         Returns a signed 32-bit integer, matching the shape of
         :py:meth:`guppylang.std.qsystem.random.RNG.random_int`.
         """
-        # LCG multiplier N from the PCG paper / pcg32_random_r (64-bit state, 32-bit
-        # output).
-        pcg32_mult: nat = 6364136223846793005
-        old_state = self._state
-        self._state = nat(old_state * pcg32_mult + self._inc)
-        # XSH-RR output permutation: xor-shift then random rotate (see pcg32_random_r).
-        xorshifted = _mask32(((old_state >> nat(18)) ^ old_state) >> nat(27))
-        rot = _mask32(old_state >> nat(59))
-        rot_inv = _mask32((~rot + nat(1)) & nat(31))
-        output = _mask32((xorshifted >> rot) | (xorshifted << rot_inv))
-        return _uint32_to_signed(output)
+        return _uint32_to_signed(self._next_uint32())
+
+    @guppy
+    @no_type_check
+    def next_int_bounded(self: PCG32, bound: int) -> int:
+        """Generate a uniformly random integer in ``[0, bound)``.
+
+        Uses rejection sampling (``pcg32_boundedrand_r`` from the PCG reference) to
+        avoid the bias introduced by ``next_int() % bound``.
+
+        Args:
+            bound: Upper bound (exclusive); must be positive and less than ``2**31``.
+
+        Returns:
+            A value in ``[0, bound)``, matching the shape of
+            :py:meth:`guppylang.std.qsystem.random.RNG.random_int_bounded`.
+        """
+        # uint32_t threshold = -bound % bound in pcg32_boundedrand_r.
+        two_to_32: nat = 4294967296
+        threshold = _mask32(two_to_32 - _mask32(nat(bound))) % nat(bound)
+        while True:
+            r = self._next_uint32()
+            if r >= threshold:
+                return int(r % nat(bound))
 
 
 @guppy
@@ -92,6 +132,6 @@ def seeded_pcg32(seed: int) -> PCG32:
     # pcg32_srandom_r: advance twice after mixing initstate into state.
     rng = PCG32(nat(0), inc)
     rng.next_int()
-    rng._state += initstate
+    rng._state = _mask64(rng._state + initstate)
     rng.next_int()
     return rng
--- a/guppylang/src/guppylang/std/random.py
+++ b/guppylang/src/guppylang/std/random.py
@@ -10,6 +10,7 @@ from __future__ import annotations
 from typing import no_type_check
 
 from guppylang import guppy
+from guppylang.std.builtins import panic
 from guppylang.std.num import nat
 
 
@@ -25,8 +26,10 @@ def _mask32(value: nat) -> nat:
 @no_type_check
 def _mask64(value: nat) -> nat:
     # 2**64 - 1: PCG32 keeps its state in a 64-bit word (see pcg32_random_r).
-    uint64_mask: nat = 18446744073709551615
-    return value & uint64_mask
+    # Combine two 32-bit masks; hugr cannot const-fold ``(1 << 64) - 1``.
+    low = _mask32(value)
+    high = _mask32(value >> nat(32))
+    return low | (high << nat(32))
 
 
 @guppy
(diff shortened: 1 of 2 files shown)


tests/integration/std/test_random.py
PabloAndresCQ, line 1: I've locally tested against `RNG` from `std.qsystem.random` and I am getting the same values as this Guppy implementation in `std.random`, which is _great_.

```python
from guppylang import guppy
from guppylang.std.builtins import owned, output
from guppylang.std.quantum import h, measure, qubit, x
from guppylang.std.random import seeded_pcg32
from guppylang.std.qsystem.random import RNG

@guppy
def main() -> None:
    rng = seeded_pcg32(55555)
    output("a", rng.next_int_bounded(2))
    output("b", rng.next_int_bounded(6))
    output("c", rng.next_int_bounded(100))

    orig_rng = RNG(55555)
    output("A", orig_rng.random_int_bounded(2))
    output("B", orig_rng.random_int_bounded(6))
    output("C", orig_rng.random_int_bounded(100))
    orig_rng.discard()

results = main.emulator(2).coinflip_sim().with_seed(42).run().results[0].entries
print(results)
```

Results are:
```
[('a', 1), ('b', 5), ('c', 32), ('A', 1), ('B', 5), ('C', 32)]
```

I have tried with different seeds and it works. I was not certain `.qsystem`'s also used PCG32, but it seems to be the case. If this is intentional, it'd be great to add tests comparing results. What do you think @hsemenenko?

PS: Works too for the non-bounded versions! (would be good to add tests too)

PabloAndresCQ, line 71: I don't think the tests as-is is actually showing the intended behaviour (that `inner_rng` does not affect `outer_rng`).

Would you mind capturing this in the test? I'd suggest something like:

```python
    def main() -> None:
        outer = seeded_pcg32(1)
        first = outer.next_int()

        _ = uses_inner_rng()

        second = outer.next_int()
        output("first", first)
        output("second", second)

        # Repeat with no inner RNG
        outer = seeded_pcg32(1)
        first = outer.next_int()
        second = outer.next_int()
        output("other_first", first)
        output("other_second", second)
```

Then compare `first` with `other_first` and `second` with `other_second`.

PabloAndresCQ: Is this test adding anything not already captured by `test_pcg32_motivating_example`?

PabloAndresCQ (approved):
Thanks! This looks great. I've double checked against the reference implementation in C. I've also checked locally that the random_bound_int from the reference implementation and the Guppy implementation agree (and that they agree with the values in the test file).

I've made a couple of comments that are not from this PR, but from the original one that @luffy-orf also contributed with.

PS: I'm approving since, from the user perspective, this is sufficient. However, @hsemenenko, please have a look at the interface + my comments on tests, and mark as "Request changes" if you deem it necessary.

tests/integration/std/test_random.py
luffy-orf: Thanks for checking this locally really helpful confirmation that both paths use the same PCG32 stream.

I've added test_pcg32_matches_qsystem_random, which runs both seeded_pcg32 and RNG with the same seed (55555) in a single emulator program and asserts pairwise equality for:

next_int() / random_int()
next_int_bounded(2/6/100) / random_int_bounded(2/6/100)
The test also pins the observed sequence (636174845, 1, 0, 27) as a regression check. I verified in CI that PCG32 and qsystem values always match each other; the exact bounded values may differ from your local run depending on environment, but the cross-check is the important part.

Happy to extend this to additional seeds if you'd find that useful.

luffy-orf: Good catch you're right that the old test only checked expected values, not that the inner call was actually harmless.

I've updated test_pcg32_motivating_example following your suggestion: after the scenario with uses_inner_rng(), it re-runs the outer sequence without any inner call and asserts:

first == other_first
second == other_second
That directly demonstrates non-interference rather than relying on hard-coded outputs alone.

luffy-orf: Agreed with the strengthened motivating example, test_pcg32_independent_streams was redundant (it covered the same scenario, just returning first + second instead of proving isolation explicitly).

I've removed it. Compile coverage for bounded/unbounded draws is still covered by the other tests in the file (test_pcg32_compile, test_pcg32_bounded_compile, sequence tests, etc.).

guppylang/src/guppylang/std/random.py
```diff
@@ -21,6 +21,14 @@ def _mask32(value: nat) -> nat:
     return value & uint32_mask
 
 
+@guppy
+@no_type_check
+def _mask64(value: nat) -> nat:
+    # 2**64 - 1: PCG32 keeps its state in a 64-bit word (see pcg32_random_r).
+    uint64_mask: nat = 18446744073709551615
```
hsemenenko: Could we use bitshift here to help with readability?
```suggestion
    uint64_mask: nat = (1 << 64) - 1
```

guppylang/src/guppylang/std/random.py
```diff
@@ -59,17 +86,30 @@ def next_int(self: PCG32) -> int:
...
+
+        Args:
+            bound: Upper bound (exclusive); must be positive and less than ``2**31``.
+
+        Returns:
+            A value in ``[0, bound)``, matching the shape of
+            :py:meth:`guppylang.std.qsystem.random.RNG.random_int_bounded`.
+        """
+        # uint32_t threshold = -bound % bound in pcg32_boundedrand_r.
+        two_to_32: nat = 4294967296
```
hsemenenko: Similarly as above, consider using bitshift here.
```suggestion
        two_to_32: nat = 1 << 32
```

guppylang/src/guppylang/std/random.py
hsemenenko: The seed is `int` but is converted into `nat` below. If a user provides a negative int, the program fails at runtime. 

Does the seed need to be `nat`? If so, the input should be `nat` here.

guppylang/src/guppylang/std/random.py
```diff
@@ -59,17 +86,30 @@ def next_int(self: PCG32) -> int:
...
-        xorshifted = _mask32(((old_state >> nat(18)) ^ old_state) >> nat(27))
-        rot = _mask32(old_state >> nat(59))
-        rot_inv = _mask32((~rot + nat(1)) & nat(31))
-        output = _mask32((xorshifted >> rot) | (xorshifted << rot_inv))
-        return _uint32_to_signed(output)
+        return _uint32_to_signed(self._next_uint32())
+
+    @guppy
+    @no_type_check
+    def next_int_bounded(self: PCG32, bound: int) -> int:
```
hsemenenko: `bound` should be `nat` and also cannot be 0. The function should include a test and `panic` if 0 is provided.

Could you also add a negative test in this case.

tests/integration/std/test_random.py
hsemenenko: Yep, the qsystem RNG uses PCG32 so it makes sense to have some tests to compare between them. Such tests will never be comprehensive, so I don't think adding more seeds is necessary.

tests/integration/std/test_random.py
```diff
@@ -84,31 +84,60 @@ def main() -> None:
...
+    assert results["pcg_int"] == results["qsys_int"]
+    assert results["pcg_bnd2"] == results["qsys_bnd2"]
+    assert results["pcg_bnd6"] == results["qsys_bnd6"]
+    assert results["pcg_bnd100"] == results["qsys_bnd100"]
+    assert (
+        results["pcg_int"],
+        results["pcg_bnd2"],
+        results["pcg_bnd6"],
+        results["pcg_bnd100"],
+    ) == (636174845, 1, 0, 27)
```
hsemenenko: Consider using `.collated_shots()` here to simplify. This can also be applied to the other tests.

```suggestion
    results = main.emulator(0).coinflip_sim().with_seed(42).run().collated_shots()[0]
    assert results["pcg_int"] == results["qsys_int"] == [636174845]
    assert results["pcg_bnd2"] == results["qsys_bnd2"] == [1]
    assert results["pcg_bnd6"] == results["qsys_bnd6"] == [0]
    assert results["pcg_bnd100"] == results["qsys_bnd100"] == [27]
```

tests/integration/std/test_random.py
hsemenenko: This simulation uses `coinflip_sim` only for a single shot. With a seed of `42`, the measurement is `0` so that branch is never executed. We should consider using `ClassicalReplay` to run both branches and test the results.

hsemenenko (changes requested):
Thanks for adding this, It looks great! I've added a few comments of things to change. None of the comments should require major changes but I think are useful additions.

tests/integration/std/test_random.py
PabloAndresCQ: Good catch

luffy-orf:
Thanks @hsemenenko all your review comments are addressed in the latest push:

_mask64 readability: two_to_32 now uses 1 << 32. For the 64-bit mask, (1 << 64) - 1 triggers a hugr const-fold shift overflow during emulation, so _mask64 combines two 32-bit masks instead (with an inline comment).
seeded_pcg32: seed parameter is now nat, since initseq is unsigned.
next_int_bounded: bound is now nat, panics on bound == 0, and test_pcg32_bounded_bound_zero_panics covers that path.
Tests: qsystem parity and motivating-example tests use .collated_shots(); the quantum interference test uses ClassicalReplay with four measurement patterns to exercise both branches.

Ready for another look when you have a moment

guppylang/src/guppylang/std/random.py
```diff
@@ -103,18 +106,20 @@ def next_int_bounded(self: PCG32, bound: int) -> int:
             A value in ``[0, bound)``, matching the shape of
             :py:meth:`guppylang.std.qsystem.random.RNG.random_int_bounded`.
         """
+        if bound == nat(0):
+            panic("PCG32.next_int_bounded: bound must be positive")
```
hsemenenko: After a discussion with @PabloAndresCQ, could this be changed to use `exit` rather than `panic`? The reason is that `exit` will only exit the current shot and then continue to the next shot, while `panic` will exit and not run any subsequent shots.
```suggestion
            exit("PCG32.next_int_bounded: bound must be positive")
```

hsemenenko (changes requested):
Looking great, thank you. Apologies for not realising that `1<<64` would overflow, but great job coming up with a work around. 

Just one minor change with the bound panic and I can approve.

luffy-orf:
@hsemenenko Done - switched to exit for the zero-bound case and updated the test to verify shot 0 exits while shot 1 continues (test_pcg32_bounded_bound_zero_exits). Thanks for catching that distinction with @PabloAndresCQ!

hsemenenko (approved):

PR #1880, opened 2026-06-17
test: Benchmark checking and compilation for tracing workloads

### LINKED ISSUE #1839: [Maintenance]: Comptime / Tracing benchmarks
### Description

We are currently only testing main compilation. We should make sure to also test the tracing part of the compiler, e.g. for functions annotated with `@guppy.comptime`. 

### Motivation

Functions annotated with `@guppy.comptime` call into separate code paths than `@guppy`, and thus warrant additional benchmarks.

This would be especially useful to do before the v1 release, as we can then be more sure that we are not introducing performance regressions.

### Scope

_No response_

### Component

Testing Infrastructure

### Priority

Medium - Should be done soon

### Additional Context

After these benchmarks are introduced, we should take a look at https://github.com/Quantinuum/guppylang/tree/mr/perf/calling-frame-detection, which aims to improve performance here.

maximilianruesch:
Prepares improvements to comptime / tracing by adding two new benchmarks for checking and compiling `@guppy.comptime`. Note that this bumps the benchmark time by 1 minute (up to 18 from 17). This is due to the mode being simulation for stability / hardware agnostic behaviour, rather than walltime, slowing benchmarking significantly.

As a driveby, limits the number of concurrent benchmark runs on PRs to 1, cancelling those in progress, since they were hogging runners without real reason.

Closes #1839

### DIFF
--- a/tests/benchmarks/test_tracing.py
+++ b/tests/benchmarks/test_tracing.py
@@ -82,7 +82,7 @@ def test_circuit_comptime_check(benchmark, circuit_seed):
     def circuit_comptime_check():
         build_guppy_circuit_comptime(*circuit_seed).check()
 
-    benchmark.pedantic(circuit_comptime_check(), rounds=25)
+    benchmark.pedantic(circuit_comptime_check, rounds=25)
 
 
 def test_circuit_comptime_compile(benchmark, circuit_seed):
--- a/tests/benchmarks/test_tracing.py
+++ b/tests/benchmarks/test_tracing.py
@@ -78,10 +78,15 @@ def circuit_seed():
     return circ, seed
 
 
+def test_circuit_comptime_check(benchmark, circuit_seed):
+    def circuit_comptime_check():
+        build_guppy_circuit_comptime(*circuit_seed).check()
+
+    benchmark.pedantic(circuit_comptime_check(), rounds=25)
+
+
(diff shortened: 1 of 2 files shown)


tests/benchmarks/test_tracing.py
```diff
@@ -0,0 +1,90 @@
...
+        build_guppy_circuit_comptime(*circuit_seed).check()
+
+    benchmark(circuit_comptime_check)
+
+
+def test_circuit_comptime_compile(benchmark, circuit_seed):
+    def circuit_comptime_compile():
+        build_guppy_circuit_comptime(*circuit_seed).compile()
+
+    benchmark.pedantic(circuit_comptime_compile, rounds=25)
```
ferbetanzo: Why do we need multiple rounds if we are doing simulation benchmarking?

ferbetanzo (commented):
Just a small question.

tests/benchmarks/test_tracing.py
maximilianruesch: That was for local runs of benchmark for reducing the stddev. It is ignored in simulation mode. WIll remove.

ferbetanzo (approved):
Looks great. Thank you!

nicolaassolini-qntm (approved):
:shipit:

PR #1882, opened 2026-06-17
fix: Improve precision on span of missing return type error to function header

### LINKED ISSUE #1868: [Feature]: Improve error message for missing return type
### Problem Statement

As pointed out here: https://github.com/Quantinuum/guppylang/blob/9634dd2c43e65c46b40c1f32b94ae29984a3c907/guppylang-internals/src/guppylang_internals/checker/func_checker.py#L310 
the span covers the entire definition. It would be nicer if it only covered the header up to the `:`.

E.g. given
```python
from guppylang import guppy


@guppy
def foo():
    return

foo.compile()
```
the error is:
```
Error: Missing type annotation (at <In[1]>:5:0)
  | 
3 | 
4 | @guppy
5 | def foo():
  | ^^^^^^^^^^
6 |     return
  | ^^^^^^^^^^ Return type must be annotated
```

### Proposed Solution

We would like to have:
```
  | 
3 | 
4 | @guppy
5 | def foo():
  | ^^^^^^^^^^ Return type must be annotated
```

### Effort Estimate

luffy-orf:
## Summary
- Fixes #1868 by highlighting only the function signature (up to the trailing `:`) for `MissingReturnAnnotationError`, instead of the entire function definition
- Adds `function_header_span()` in `span.py` to compute the correct diagnostic span
- Updates golden error snapshots for missing return type cases

## Test plan
- [x] `uv run pytest tests/error/test_misc_errors.py -k "return_not_annotated" -v`
- [x] `uv run pytest tests/error/test_misc_errors.py -v`
- [x] `uv run ruff check guppylang-internals/src/guppylang_internals/span.py guppylang-internals/src/guppylang_internals/checker/func_checker.py`


guppylang-internals/src/guppylang_internals/span.py
```diff
@@ -126,6 +126,34 @@ def to_span(x: ToSpan) -> Span:
...
+def function_header_span(func_def: ast.FunctionDef) -> Span:
+    """Returns a span covering only the function header up to and including `:`."""
+    start = to_span(func_def).start
+    source = get_source(func_def)
+    file = get_file(func_def)
+    line_offset = get_line_offset(func_def)
+    if source is None or file is None or line_offset is None:
+        if func_def.body:
+            return Span(start, to_span(func_def.body[0]).start)
+        return Span(start, to_span(func_def).end)
```
maximilianruesch: If this is a possibility for the user, you may want to add test cases covering this (or assert that they are present already), and if it is not, `assert` that these cases are not present and move on?

maximilianruesch:
```suggestion
    if source is None or file is None or line_offset is None:
        if func_def.body:
            return Span(start, to_span(func_def.body[0]).start)
        return to_span(func_def)
```
The work is done anyway, and if I understand correctly these should be equivalent. Imo the intention of "giving up" is more understandable this way.

guppylang-internals/src/guppylang_internals/span.py
```diff
@@ -126,6 +126,34 @@ def to_span(x: ToSpan) -> Span:
     return Span(start, end)
 
 
+def function_header_span(func_def: ast.FunctionDef) -> Span:
+    """Returns a span covering only the function header up to and including `:`."""
```
maximilianruesch, line 130: Also describe cases where this is not possible, i.e. what the span will cover when one cannot identify the header span. So perhaps something like (with suitable line breaks for max line length):
```suggestion
    """Returns a span covering only the function header up to and including `:` on a best-effort basis, falling back to the full function definition."""
```
However, if you can prove that your below cases can all make the span exact (via a short argument), you do not need to add this.

guppylang-internals/src/guppylang_internals/span.py
```diff
@@ -126,6 +126,34 @@ def to_span(x: ToSpan) -> Span:
...
+            return Span(start, to_span(func_def.body[0]).start)
+        return Span(start, to_span(func_def).end)
+
+    lines = source.splitlines()
+    line_idx = func_def.lineno - 1
+    paren_depth = 0
+    for i in range(line_idx, len(lines)):
+        col_begin = func_def.col_offset if i == line_idx else 0
+        for col in range(col_begin, len(lines[i])):
+            char = lines[i][col]
```
maximilianruesch:
```suggestion
        for char in lines[i][col_begin:]:
```
A bit of Python splicing magic. Also handles out of bounds equally as gracefullly. Probably improves performance a touch too, but I am not worried about that here (we are already exclusively in a failure case).

You can do a very similar thing to the outer loop.

guppylang-internals/src/guppylang_internals/span.py
```diff
@@ -126,6 +126,34 @@ def to_span(x: ToSpan) -> Span:
...
 
 
+def function_header_span(func_def: ast.FunctionDef) -> Span:
+    """Returns a span covering only the function header up to and including `:`."""
+    start = to_span(func_def).start
+    source = get_source(func_def)
+    file = get_file(func_def)
+    line_offset = get_line_offset(func_def)
+    if source is None or file is None or line_offset is None:
+        if func_def.body:
```
maximilianruesch: Can you be more explicit here, i.e.

```suggestion
        if len(func_def.body) > 0:
```

maximilianruesch (changes requested):
Hey, thank you for the PR! The error message is definitely improved!

I am a bit careful about the slightly hacky approach (with the manual parsing), but I see that the AST module does not expose enough information to handle things in another way really. Nonetheless, I have some suggestions to improve the code as well as some suggestions!

luffy-orf:
@maximilianruesch  Thanks for the review! Addressed all suggestions:

- Removed the fallback branch and asserted that source metadata is always present (all `check_signature` call paths go through `annotate_location`).
- Simplified the loops using slicing/`enumerate`.
- Added `tests/test_function_header_span.py` covering single-line, annotated, multiline, and spaced-colon signatures.

acl-cqc:
As an alternative approach https://github.com/Quantinuum/guppylang/pull/1846 tackles a similar problem using subcomponents of the AST node. In that PR I specifically wanted to include the decorators, which you might not want here, but exclude the function body (the main problem with falling back to the definition!). Does that work here?

maximilianruesch:
@acl-cqc I cannot immediately see how that changeset is relevant here. Specifically you improve error messages at call sites, whereas here we improve error messages at function definition sites. I also could not find measures taken in that PR that improve on function definition sites. Could you elaborate on this please?

tests/test_function_header_span.py
```diff
@@ -0,0 +1,42 @@
...
+    return line[span.start.column : span.end.column]
+
+
+@pytest.mark.parametrize(
+    ("source", "expected_header"),
+    [
+        ("def foo():\n    return", "def foo():"),
+        ("def foo(x: bool):\n    return x", "def foo(x: bool):"),
+        ("def foo(\n    x: bool,\n):\n    return x", "def foo(\n    x: bool,\n):"),
+        ("def foo() :\n    return", "def foo() :"),
```
maximilianruesch: These test cases should be actual `.err` test cases following those in `misc_errors`. We usually do not add unit tests, all of our tests should be "integration" only.

guppylang-internals/src/guppylang_internals/span.py
maximilianruesch: Ah, I did not see you used `i` and `col` outside of the array accesses as well. It was cleaner before, my bad 😅

acl-cqc:
> @acl-cqc I cannot immediately see how that changeset is relevant here. Specifically you improve error messages at call sites, whereas here we improve error messages at function definition sites. I also could not find measures taken in that PR that improve on function definition sites. Could you elaborate on this please?

So sorry, both for slow response, and especially, for linking to the wrong thing! :( The relevant bit was actually in #1723, which has gone onto a feature branch `effects`, I quote from `EffectLimitDecl.for_def`:
```
            # Could not identify decorator, so include all in context; union with
            # returns will include name etc. inbetween but avoid the function body.
            elems = func_def.decorator_list
            if func_def.returns is not None:
                elems += [func_def.returns]

            def union(s1: Span, s2: Span) -> Span:
                r = s1 | s2
                assert r is not None  # Function def should not cross file boundary
                return r

            decl = reduce(union, (to_span(e) for e in elems))
```

maximilianruesch:
That is a neat trick, however that only includes the function name when there is an actual return. In our case, there is no return (because that is exactly what the error message is for), so I think unfortunately this will not work here. I propose to stick with the original.

acl-cqc:
> That is a neat trick, however that only includes the function name when there is an actual return. In our case, there is no return (because that is exactly what the error message is for), so I think unfortunately this will not work here. I propose to stick with the original.

Well you could just go to the end of the arguments? I haven't tested to see whether that would include the closing `)` and it definitely wouldn't the `:`, so if that's particularly important?

If the effects stuff goes in, then sharing the union+`reduce` (and using a different subset of the FunctionDef ast children) would mean very little new code. Not clear that it is going in tho...but I wonder if an AST `minus` rather than union would do the trick (FunctionDef minus body).

Not as good as I thought tho, sorry, so your call :)

maximilianruesch:
I would rather a complete solution that is well tested (and this one is not particularly complicated) rather than one that just barely (and not really) works for our use case and is less understandable. We can consider a refactor (for both) once we get effects in. Or even, effects can consider using this function, and simply adding the decorators on top, for its fallback. Deduplicate the other way.

maximilianruesch (changes requested):
Sorry for the delay and hickup in review, hoping to get this improvement merged soon 🚀

luffy-orf:
@maximilianruesch Replaced the unit test with integration `.err` cases in `misc_errors`:
- `return_not_annotated_multiline` - multiline signature
- `return_not_annotated_spaced_colon` - space before `:` (uses `# fmt: off` to preserve formatting)

Single-line cases were already covered by `return_not_annotated` and `return_not_annotated_none1`.

maximilianruesch (approved):

maximilianruesch:
I removed the fmt comments, since they seem to be specific to your setup. You may want to check your IDE setup and keep it from editing golden files without your permission.

Issue #1747, opened 2026-05-18
Define all known effects, update guppy funcs/@hugr_op's


acl-cqc:
* First, finalise full list of tokens - see [design doc](https://cqc.atlassian.net/wiki/spaces/~221321769/pages/4339269633/List+of+token+types+v2) and also https://github.com/Quantinuum/guppy-docs/pull/211 https://github.com/Quantinuum/guppy-docs/pull/149
* Define the system set of tokens, i.e. legal values to specify to the annotation.
* Code to wire them up. `DFBuilder._handle_side_effects` (https://github.com/Quantinuum/guppylang/pull/2024) will need extending as we are thinking that panics are in parallel but totally ordered by exits.
* Annotate all ops (result, panic, etc.) appropriately
* Update tket2 BorrowSquashPass, as the extra order edges will probably stop it from working as it stands

Note that in future we'll want to be able to specify python-like global variables as "effects"/whatever too, although such globals may not matter yet (e.g. for this issue).

PR #1883, opened 2026-06-17
feat(effects): insert Order edges according to effects

### LINKED ISSUE #1746: Output Order edges for each effect

acl-cqc:
NOTE: This is into the effects branch, i.e. based off #1723; needs reworking to go onto `main`.

Closes #1746 

* Change builder add_op to take a `OpWithEffects = Pure | (Op, Sequence[Effect])`, update uses everywhere
* `_last_side_effect` now a map from Effect, propagate_side_effects takes list of Effect too
* also add some tests demonstrating current state of arrays: indexing (`arr[3]` i.e. `array.__getitem__`) is not permitted in pure functions - this may change as part of #1747.
* Current op-level wiring does totally order array panics, consistent with previous, this causes many new Order edges to be added into the benchmarks, hence the increase in hugr size.

Need to add some tests that actually show things (in particular function calls) being wired up with parallel Order edges rather than sequential, both for pure functions, and for functions with different (nonzero) effects - the latter will probably require "cheating" and passing a non-Effect in (with typechecking turned off)...

Current typedef of `OpWithEffects` as a union is introduced in c4e01247e0e8470a4289fb43babc7a7f99fe02ca, before that 'twas a class - felt a bit awkward, but suggestions welcome for better ways. (I wanted to make every add_op an error unless updated; given that's now the case, could make add_op take a hugr ExtOp *or* a tuple, but not sure that's good, an easy mistake to make)


acl-cqc:
Superceded by #2024 which targets `main`

PR #1884, opened 2026-06-18
perf: More efficient calling frame detection

maximilianruesch:
Using `inspect.getmodule` is overkill when all you are trying to retrieve is the name of the module (you can access that through the file-global `__name__` as well). At the same time, `inspect.getmodule` accesses the file cache of the Python interpreter, potentially doing a lot of IO. During tracing workloads, calling frame detection is run for most expressions for purposes of nice error messages, and thus the `inspect.getmodule` is called in a rather hot loop. Any improvements here are vital for tracing performance.


nicolaassolini-qntm (approved):
17% faster, nice :shipit:

PR #1885, opened 2026-06-18
feat: Allow setting optimization level

### LINKED ISSUE #1843: [Feature]: Optimization level option in compilation call
### Problem Statement

When calling `.compile` or `.emulator` on a definition, there is currently no way to specify optimizations to apply (or avoid applying) to the program.

We want the normal compilation workflow to do some simple optimizations and structural simplifications by default, but adding this would be a breaking change unless.

We should include an API for specifying compilation levels / passes when compiling a guppy program, and specify in the v1.0 migration guide how to disable them.
This issue does not require actually running optimizations, just defining the breaking API update.

### Proposed Solution

Include some new definitions:

```py
class OptimizationLevel(Enum):
    Default = "default"
    """
    Optimization set used by default for all guppy program compilations.
    """

    Minimal = "minimal"
    """
    Only apply structural rewrites required for execution.
    """

    Classical = classical
    """
    Only apply classical optimizations.

    Some gate rebasing/dead quantum code elimination may be applied as needed.
    """

class OptimizerInstance:
    """Builder struct used while defining the optimizations to apply
    to a guppy program during compilation.
    """
    ...
```

`GuppyFunctionDefinition` gains a new method for setting this level:
```py
class GuppyFunctionDefinition:
    def with_opt_level(self, level: OptimizationLevel) -> OptimizerInstance: ...
```

The optimizer instance should then provide the same `compile_*` and `emulator` methods, so it can be used with method chaining. To simplify this definition, we define a common protocol implemented both by `OptimizerInstance` and `GuppyFunctionDefinition`:

```py
class GuppyCompilableProgram(Protocol):
    def emulator(self, ...) -> EmulatorInstance: ...
    def compile(self) -> Package: ...
    def compile_entrypoint(self) -> Package: ...
    def compile_function(self) -> Package: ...
```

Finally, advanced users may want to run specific `hugr.ComposablePass`es before emulating their program.
`OptimizerInstance` should allow a chainable call to add passes to the list that will get called;

```py
class OptimizerInstance:
    def with_optimization(self, pass: ComposablePass) -> OptimizerInstance:
        """Add an additional optimization pass to be run while compiling a Guppy program"""
```

### Component

Guppy Compiler

### Effort Estimate

Medium - Moderate complexity

### Additional Context

_No response_

aborgna-q:
Closes #1843

Note that using the default compilation level requires having `tket` available.

Some GuppyNormalization sub-passes have been disabled in the default optimization level, pending some bug investigation.

To emulate the old behaviour, disable optimizations by calling `with_minimal_opt` before compilation/emulation.
```py
some_fn.with_minimal_opt().compile()
```


nicolaassolini-qntm:
~~I think this PR is also closing: #1895~~

guppylang/src/guppylang/defs.py
```diff
@@ -117,8 +133,35 @@ def __getattr__(self, name: str) -> Any:
...
+    """A guppy definition for a program that can be compiled or emulated."""
+
+    def emulator(
+        self,
+        n_qubits: int | None = None,
+        builder: EmulatorBuilder | None = None,
+        libs: list[Package] | None = None,
+        platform: Platform = "helios",
+    ) -> EmulatorInstance:
+        """Compile the program and build an emulator instance for it."""
```
ss2165: are these docstrings supposed to be comprehensive or placeholders to be overriden?

guppylang/src/guppylang/defs.py
```diff
@@ -117,8 +133,35 @@ def __getattr__(self, name: str) -> Any:
...
+        ...
+
+    def compile(self) -> Package:
```
ss2165, line 184: this is ostensibly just an alias for compile_entrypoint - consider implementing it as a default to enforce that?

guppylang/src/guppylang/optimizer.py
```diff
@@ -0,0 +1,124 @@
...
+                ]
+            case OptimizationLevel.Minimal:
+                return []
+
+
+def _apply_passes(package: Package, passes: Sequence[ComposablePass]) -> Package:
+    if not passes:
+        return package
+
+    for pass_ in passes:
```
ss2165: im assuming this mean we can't compose the composable passes from here and run in one go?

guppylang/src/guppylang/optimizer.py
```diff
@@ -0,0 +1,124 @@
...
+        for module in package.modules:
+            pass_.run(module, inplace=True)
+    return package
+
+
+@dataclass
```
ss2165: frozen?

tests/integration/test_array.py
```diff
@@ -47,7 +48,9 @@ def main() -> None:
         foo(array(True))
         foo(array(True, False))
 
-    package = main.compile_function()
+    # Skip optimizations during compilation to preserve the UnsignedIntVal
+    # instantiation.
+    package = main.with_opt_level(OptimizationLevel.Minimal).compile_function()
```
ss2165: we might want a utility shorthand for "with minimal opt"

ss2165 (commented):
first pass, haven't looked at tests yet

guppylang/src/guppylang/optimizer.py
aborgna-q: It's equivalent, although some combinations may be able to do inter-pass optimization (e.g. avoiding redundant serialization roundtrips).

I updated the code

guppylang/src/guppylang/defs.py
aborgna-q: Yes, I'm adding it to the protocol.

Note that we cannot import `GuppyCompilableProgram` here due to cyclic imports, so I'm leaving the method implementation here too.

guppylang/src/guppylang/defs.py
aborgna-q: Copied the return/raises part from the implementations. It's a bit verbose, but better to repeat the docs than to have them missing in some sphinx page.

guppylang/src/guppylang/optimizer.py
```diff
@@ -59,17 +59,29 @@ def passes(self) -> list[ComposablePass]:
...
                 return [
                     passes.NormalizeGuppy(
+                        remove_tuple_untuple=True,
+                        remove_dead_funcs=True,
+                        remove_redundant_order_edges=True,
+                        squash_borrows=True,
+                        # Multiple tests assume no function inlining.
+                        # They must be updated before this can be enabled.
+                        inline_funcs=False,
+                        # Tket reports invalid constructed HUGRs
```
ss2165, line 68: is it being called under minimal?

guppylang/src/guppylang/defs.py
```diff
@@ -129,8 +145,56 @@ def __getattr__(self, name: str) -> Any:
...
+        """Compile an execution entrypoint to a HUGR package.
+
+        Alias for :py:meth:`compile_entrypoint`.
+
+
+        Returns:
```
ss2165, line 190: emulator not updated?

ss2165 (changes requested):

guppylang/src/guppylang/optimizer.py
aborgna-q: Updated a few tests with `with_minimal_opt` and enabled `inline_funcs` in the default optimization level.

ss2165 (approved):

PR #1886, opened 2026-06-18
refactor!: Remove deprecated reexports and leave custom import errors

### LINKED ISSUE #1019: Remove deprecated standard library reexports

maximilianruesch:
Removes deprecated reexports, leaving custom import errors as a convenience for migration. Also adds a test for those. We can remove this custom infrastructure once this change has been released for a couple of iterations of Guppy v1.

BREAKING CHANGE: Various symbols are not importable from reexports in `guppylang.std.quantum_functional` or `guppylang.std.builtins` anymore and can be imported from their appropriate varying paths.

Closes #1019


ss2165 (commented):
Nice! I think we should remove the moved symbols from __all__ now

maximilianruesch:
> Nice! I think we should remove the moved symbols from all now

Fair point, that gives another mode of warning through type checkers.

maximilianruesch:
Also I forgot to nuke `PriorityQueue` from `guppylang.std.collections.stack`. Will do both.

ss2165 (approved):
:shipit:

PR #1887, opened 2026-06-19
chore: Overhaul release workflow using home-cooked solution

### LINKED ISSUE #1866: [Maintenance]: Alternative versioning schemes for compiler internals
### Description

Following reasons in #1864, we need to move the versioning scheme for the compiler internals to a more sustainable scheme. As almost every release is breaking anyway, we can just mark every release as breaking by moving to exact string based versioning (such as build numbers).

A proposal is to move to `<mainlang-major>-<build-number>`, to provide:
1. An indication which main lang major track this internals version is for
2. A monotonically increasing build number to sort releases by recency
3. (via the `-`) a clear non-compatibility with semver frameworks
   This may not be feasible if we want to adhere to https://packaging.python.org/en/latest/discussions/versioning/#valid-version-numbers (which we should), but we could employ the suggested two-part versioning as `X.X` (with a `.`).

The feasibility of this and the interaction with existing tools such as uv resolution or release-please still needs to be explored, e.g. via a test project hooked up to `test.pypi.org`.

### Motivation

_No response_

### Scope

_No response_

### Component

_No response_

### Priority

None

### Additional Context

_No response_

### LINKED ISSUE #1867: [Maintenance]: Combine release workflows
### Description

Following #1864 and in particular #1866, we need to increment versions manually and inspect the changelog manually. In the current setup, the most frequent action is to release internals, wait until main lang is updated, release main lang.

Instead, we could combine the release workflows, manually decide what is breaking and what is not, and release both packages at once. This has the added benefit of reducing CI time, simplifying workflow setups for release checks, and a faster turnaround for releases (as well as one fewer open PRs ;)).

### Motivation

_No response_

### Scope

_No response_

### Component

_No response_

### Priority

None

### Additional Context

_No response_

maximilianruesch:
Uses a homecooked release PR solution that suits our needs.
 
Disclosure: The base of this PR was generated by Copilot, but was later refactored. The initial pitch included tooling tests, but I have removed those (in an extra commit). We could restore them, but they need some updating in that case.

Closes #1866 
Closes #1867


ss2165:
Please include some test CI runs of the relevant workflows

maximilianruesch:
@ss2165 Most of the PR based workflows have been tested in https://github.com/Quantinuum/guppylang/pull/1889, I am careful with testing tag + release based stuff in the repository itself; it is too easy to fetch (and subsequently repush) unwanted tags.

ss2165 (commented):
I have reviewed the DEVELOPMENT.md high level workflow. Couple of queries:

1. Major releases are exceedingly rare, I think we should also have a minor release guard
2. Pushing to registry can fail, this would leave a tag without a matching release needing manual tag cleanup. Would it make sense for tag creation to be the final stage rather than the trigger?
3. has the patch release workflow been tested? 
For that matter have all ways of hitting test pypi been exercised?

maximilianruesch:
> Major releases are exceedingly rare, I think we should also have a minor release guard

The advantage would be that we could force releasers to make a conscious choice before starting a new minor train. The disadvantage is that minor releases are common, and as such it may get in the way. Was that your original intent?

> Pushing to registry can fail, this would leave a tag without a matching release needing manual tag cleanup. Would it make sense for tag creation to be the final stage rather than the trigger?

While true, the current tag based workflow allows me to bundle the entrypoint to "please make a release" to just the tag. Imo a registry push fail does not mean that the actual code is wrong. Doing another release is fine, and people can still pull the intermediate version from GitHub if they want to.

> has the patch release workflow been tested? For that matter have all ways of hitting test pypi been exercised?

Nope, forgot about test pypi, doing that now.

aborgna-q:
> Pushing to registry can fail, this would leave a tag without a matching release needing manual tag cleanup. Would it make sense for tag creation to be the final stage rather than the trigger?

This is not a problem imho. If a release fails even after all checks ran correctly on the draft release, and it cannot be fixed by re-running the workflow, we can just remove the release & tag, commit any fix, and try again.

That will change if we enable immutable releases, but that's another can of works.

aborgna-q (approved):
Since the wheel building workflow is mostly unchanged, we should be able to test the tag -> draft release part locally once this is merged. We should be able to do without the test pypi run.

PR #1890, opened 2026-06-19
feat: runtime entrypoint arguments for the emulator

### LINKED ISSUE #1546: [Docs]: Speed up variational loop in QAOA example

### LINKED ISSUE #1892: [Feature]: emulation with entrypoint parameters
### Problem Statement

For variational-style programs the same program is run many times just changing parameters. This is very slow because changing parameters recompiles the whole program. 

### Proposed Solution

By compiling programs that take entrypoint parameters, i.e. `def main(foo: int) -> None` etc. we can re-run the same compiled program with different inputs. 

### Component

Guppy Compiler

### Effort Estimate

Medium - Moderate complexity

### Additional Context

_No response_

### LINKED ISSUE #1943: [Bug]: QAOA MaxCut example notebook fails to execute on main (raw measure_array output), and is not covered by notebook CI
### Description

I spent some time on the QAOA MaxCut example during UnitaryHack 2026 (issue #1545); my own attempt was #1819, and the change that landed was #1801 ; While re-testing `examples/qaoa_maxcut_example.ipynb` against current `main` afterwards, I noticed two small, related things that seem worth tracking, so I'm reporting them here rather than touching any code.

### Steps to Reproduce

```bash
# 1) the notebook is excluded from the executed set:
rg -n "qaoa_maxcut_example|Remove long running QAOA|1546" tests/integration/test_notebooks.py

# 2) the notebook suite is green even though the notebook above fails to execute:
```

### Expected Behaviour

If the example notebooks are meant to stay executable, the QAOA notebook is either included in notebook execution tests, or covered by a smaller targeted smoke test for the API-sensitive part.

### Actual Behaviour

The suite passes while the notebook is broken. The exclusion is:

```python
# Remove long running QAOA notebook from C.I. tests
# Hopefully we can add it back in when we can speed it up.
# https://github.com/Quantinuum/guppylang/issues/1546
example_notebooks.remove(
    Path(__file__).parent.parent.parent / "examples" / "qaoa_maxcut_example.ipynb"
)
```

and the suite reports (on this revision):

```text
14 passed in 53.42s
```

so `tests/integration/test_notebooks.py` is green at the same time the `nbconvert --execute` command in Problem 1 fails.

I'm not suggesting re-enabling the heavy notebook. A lighter guard might be enough to catch this class of drift, for example a tiny compile/emulator smoke test over the `output(... measure_array ...)` pattern (the standalone reproducer above runs in well under a second).

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

ss2165:
Closes #1892 

## Runtime entrypoint arguments for the emulator

Adds the Guppy-side support for passing **runtime arguments to a program's entrypoint**, so a program can be compiled once and run many times with different parameter values. 

The motivating use case is variational workloads (e.g. QAOA): today each parameter update forces a full recompile; with this, the parameters become entrypoint arguments and only execution repeats.

### Scope
Emulator-only (selene). `compile()` / `compile_entrypoint()` are unchanged and still reject entrypoint args.

### API
```python
inst = main.emulator(n_qubits=N).run(theta=1.5, k=3)          # constant args across shots
inst = main.emulator(n_qubits=N).run_per_shot([{...}, {...}])  # per-shot args
```
Supported arg types: `bool`, signed `int`, `float`, and 1-D arrays of those. `nat` and nested arrays are rejected with a clear error.

### What's here
- `emulator/_args.py` — arg classification, validation, and entrypoint surgery.
- `defs.py` — `emulator()` wiring, arg-spec extraction, unsupported-arg error.
- `emulator/builder.py`, `emulator/instance.py` — builder plumbing + `run()` / `run_per_shot()`.
- Unit + integration tests; the QAOA maxcut example rewritten to use runtime args.
- Re-enables the QAOA notebook in CI (Closes #1546). Stochastic platform-dependent cells are ignored in the regression testing.


examples/qaoa_maxcut_example.ipynb
ss2165, line 1: small output changes due to different rng, cut=0: case is missing in one of the cell outputs because the single shot that caused it no longer occurs

guppylang/src/guppylang/defs.py
```diff
@@ -156,13 +167,22 @@ def emulator(
         Returns:
             An `EmulatorInstance` that can be used to run the function in an emulator.
         """
-        mod = self.compile()
+        mod = self.compile_function()
 
         if libs is not None:
             mod = mod.link(*libs)
 
+        arg_specs = self._entrypoint_arg_specs()
```
maximilianruesch: Imo this should be moved to just above the if statement where it is used (except if you want it to fail early, but I doubt that instantiating the emulator builder is not negligible overhead; we should fix a LOT of stuff before we do such micro-optimisations).

Additionally, you can use the `if ... := ...` syntax to shorten this even more.

guppylang/src/guppylang/defs.py
```diff
@@ -218,15 +279,9 @@ def compile_entrypoint(self) -> Package:
...
-        monomorphized_id = (self.id, ())
-        compiled_def = ENGINE.compiled.get(monomorphized_id)
-        if (
-            isinstance(compiled_def, CompiledCallableDef)
-            and len(compiled_def.ty.inputs) > 0
-        ):
-            # Check if the entrypoint has arguments
-            defined_at = cast("ast.FunctionDef", compiled_def.defined_at)
+        result = self._compiled_entrypoint_with_inputs()
+        if result is not None:
```
maximilianruesch: Use `... := ...` here? I am a bit torn about that syntax, but with parantheses it offers sufficient readability to  allow the line reduction.

guppylang/src/guppylang/emulator/instance.py
```diff
@@ -211,11 +241,80 @@ def stabilizer_sim(self) -> Self:
...
+            )
+        _validate_per_shot_args(self._arg_specs, args)
+        if self.shot_offset != 0:
+            raise EntrypointArgValueError(
+                "`run_per_shot` is not compatible with a non-zero shot offset "
+                f"(got {self.shot_offset}); per-shot arguments are indexed from 0."
+            )
+        set_shots = self._options._shots
+        if set_shots is not None and set_shots != len(args):
+            raise EntrypointArgValueError(
```
maximilianruesch: Perhaps it does not matter tooooo much, but this error class is (according to its docstring) misused in the second and third check here.

guppylang/src/guppylang/defs.py
```diff
@@ -189,7 +209,48 @@ def emulator(
...
+            compiled_def.ty.inputs,
+            defined_at.args.args,
+            strict=True,
+        ):
+            if (reason := unsupported_arg_reason(inp.ty)) is not None:
+                raise GuppyError(
+                    UnsupportedEntrypointArgError(span=to_span(ast_arg), reason=reason)
+                )
+            specs.append(EntrypointArgSpec(name=name, ty=inp.ty))
+        return tuple(specs)
```
maximilianruesch: I do not see a reason this needs to cast / constructed to a tuple here, you could just return a `Sequence` (which is what `.build` accepts).

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
...
+    from hugr.package import Package
+
+_BORROW_ARRAY_DEF = BORROW_ARRAY_EXT.types["borrow_array"]
+
+#: Name of the no-input wrapper function generated around the original entrypoint
+#: when runtime arguments are injected via ``read_arg``.
+_ARGS_WRAPPER_ENTRYPOINT_NAME = "__entrypoint_with_args"
+
+#: Human-readable description of the supported entrypoint argument types.
+SUPPORTED_ARG_TYPES_MSG = "`bool`, `int`, `float`, and arrays of those"
```
maximilianruesch: This could (and imo should) just be a local variable in the single function that uses it.

If you want to keep it here to synchronise with the `ArgValue` type union, please
1. Make this private (prefix with `_`)
2. Remove the doc comment, as it is entirely subsumed by the variable name and does not add value
3. Add a regular comment that this should be kept in sync with `ArgValue`
4. Add a slight wording adjustment that nested arrays are not supported (or rather what is supported are "non-nested" or "one-dimensional" or "shallow" arrays, wording at your disposal).

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
...
+        )
+    for spec in specs:
+        reason = _value_error(spec.ty, record[spec.name])
+        if reason is not None:
+            raise EntrypointArgValueError(
+                f"Invalid value{where} for entrypoint argument `{spec.name}`: {reason}"
+            )
+
+
+def _validate_per_shot_args(
```
maximilianruesch: Why are these functions (this and the above one) marked as private if they are used outside of this module (and the module itself is already private)?

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
+"""Support for runtime arguments to emulator entrypoint functions.
```
maximilianruesch: Imo just the first sentence would suffice, since this is a private module. Instead, all of these words should be on a docs page somewhere.

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
...
+    name: str
+    ty: Type
+
+
+class EntrypointArgValueError(ValueError):
+    """Raised when runtime argument values don't match the entrypoint signature."""
+
+
+def _is_supported_scalar(ty: Type) -> bool:
+    """Whether ``ty`` is a supported scalar entrypoint argument type."""
```
maximilianruesch:
```suggestion
```
Does imo not add much

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
...
+    if is_bool_type(ty):
+        return True
+    if isinstance(ty, NumericType):
+        return ty.kind in (NumericType.Kind.Int, NumericType.Kind.Float)
+    return False
+
+
+def _is_supported_arg_type(ty: Type) -> bool:
+    """Whether ``ty`` may be used as an entrypoint runtime argument."""
```
maximilianruesch:
```suggestion
```
Same here

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
...
+                return None if isinstance(value, (int, float)) else "expected a `float`"
+    if is_array_type(ty):
+        elem = get_element_type(ty)
+        n = _array_length(ty)
+        # str and bytes are Sequence subtypes in Python's ABC but are not valid
+        # array arguments; exclude them so they fall through to the error below.
+        if not isinstance(value, Sequence) or isinstance(value, (str, bytes)):
+            return f"expected an array of length {n}"
+        # TODO: selene's ArgProvider currently rejects empty lists; this restriction
+        # is expected to be lifted in https://github.com/Quantinuum/selene/pull/187.
```
maximilianruesch, line 132: What is the current error message? Is it somewhat friendly? Perhaps add a test for it in the meantime?

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
...
+                return None if isinstance(value, int) else "expected an `int`"
+            case NumericType.Kind.Float:
+                if isinstance(value, bool):
+                    return "expected a `float`, got a `bool`"
+                return None if isinstance(value, (int, float)) else "expected a `float`"
+    if is_array_type(ty):
+        elem = get_element_type(ty)
+        n = _array_length(ty)
+        # str and bytes are Sequence subtypes in Python's ABC but are not valid
+        # array arguments; exclude them so they fall through to the error below.
```
maximilianruesch, line 128: Ewwww

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
...
+            return wrapper.add_op(std_array_to_array(elem_ty, length_arg), std_wire)[0]
+        return wrapper.add_op(tket_exts.argument.read_arg(name, ty))[0]
+
+    def _has_inputs(module: Hugr[Any]) -> bool:
+        op = module[module.entrypoint].op
+        return isinstance(op, ops.FuncDefn) and len(op.inputs) > 0
+
+    module = next((m for m in package.modules if _has_inputs(m)), None)
+    if module is None:
+        raise ValueError("No entrypoint with input parameters found in package.")
```
maximilianruesch, line 225: Can we not just return `package` and have this function be a noop?

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,275 @@
...
+    input_types = list(op.inputs)
+    output_types = list(op.signature.body.output)
+    if len(arg_names) != len(input_types):
+        raise ValueError(
+            "Mismatch between entrypoint parameter names "
+            f"({len(arg_names)}) and HUGR inputs ({len(input_types)})."
+        )
+
+    wrapper = Function.new_nested(
+        ops.FuncDefn(_ARGS_WRAPPER_ENTRYPOINT_NAME, [], visibility="Public"),
```
maximilianruesch: Make it very explicit (e.g. via a comment) that the choice to now have two public functions is concious.

tests/error/misc_errors/entrypoint_unsupported_tuple_arg.err
```diff
@@ -0,0 +1,10 @@
+Error: Unsupported entrypoint argument type (at $FILE:5:8)
+  | 
+3 | 
+4 | @guppy
+5 | def foo(x: tuple[int, float]) -> None:
+  |         ^^^^^^^^^^^^^^^^^^^^ Type `(int, float)` is not supported as an entrypoint
+  |                              argument. Supported types are `bool`, `int`, `float`, and
+  |                              arrays of those.
```
maximilianruesch: Perhaps a bit late to raise this, but why are tuples not supported? They are fixed length arrays as well, just with more types to accept... Type checking then just recurses into the tuple and inspects each argument separately.

Perhaps a follow up issue?

tests/error/misc_errors/entrypoint_unsupported_array_elem_arg.err
```diff
@@ -0,0 +1,10 @@
+Error: Unsupported entrypoint argument type (at $FILE:7:8)
+  | 
+5 | 
+6 | @guppy
+7 | def foo(x: array[nat, 3]) -> None:
+  |         ^^^^^^^^^^^^^^^^ Arrays of `nat` are not supported as entrypoint arguments.
+  |                          Supported element types are `bool`, `int`, `float`, and
+  |                          arrays of those.
```
maximilianruesch: `nat` has a special error message when used on its own, imo it should have a similarly special message when used with an `array`.

tests/integration/test_emulator_args.py
```diff
@@ -0,0 +1,91 @@
...
+    def main(xs: array[float, 3]) -> None:
+        result("first", xs[0])
+        result("third", xs[2])
+
+    py_xs = [1.0, 2.0, 3.0]
+    res = main.emulator(n_qubits=1).run(xs=py_xs)
+    assert res == EmulatorResult([[("first", py_xs[0]), ("third", py_xs[2])]])
+
+
+def test_bool_arg() -> None:
```
maximilianruesch: What about all the other supported types? Imo they deserve a positive test.

tests/emulator/test_args.py
maximilianruesch, line 1: Should most of these tests not be integration tests that test these features with real Guppy programs, instead of tests of mostly private and implementation detail heavy stuff?

tests/emulator/test_instance.py
```diff
@@ -435,3 +442,74 @@ def test_emulator_instance_full_configuration_workflow():
         assert isinstance(call_kwargs["simulator"], Quest)
 
         mock_result.assert_called_once_with(mock_result_stream)
+
+
+def _arg_instance(
```
maximilianruesch:
```suggestion
def _required_theta_arg_instance(
```

tests/emulator/test_instance.py
```diff
@@ -435,3 +442,74 @@ def test_emulator_instance_full_configuration_workflow():
...
+    run_shots_side_effect: Callable[..., Any] | None = None,
```
maximilianruesch: Am I blind? Is this argument ever used?

tests/emulator/test_instance.py
```diff
@@ -435,3 +442,74 @@ def test_emulator_instance_full_configuration_workflow():
...
+def test_run_requires_args_when_expected() -> None:
+    instance, _ = _arg_instance()
+    with pytest.raises(EntrypointArgValueError, match=r"Missing.*`theta`"):
+        instance.run()
+
+
+def test_run_per_shot_runs_len_shots() -> None:
+    instance, mock_selene = _arg_instance()
+    instance.run_per_shot([{"theta": 1.0}, {"theta": 2.0}, {"theta": 3.0}])
+    # n_shots should be overridden to the number of per-shot records
```
maximilianruesch:
```suggestion
    # n_shots should be set to the number of per-shot records
```
It is not an override since if you provide a value there is an explicit conflict.

tests/emulator/test_instance.py
```diff
@@ -435,3 +442,74 @@ def test_emulator_instance_full_configuration_workflow():
...
+    instance = instance.with_shots(5)
+    with pytest.raises(EntrypointArgValueError, match="with_shots"):
+        instance.run_per_shot([{"theta": 1.0}, {"theta": 2.0}])
+
+
+def test_run_per_shot_rejects_explicit_with_shots_one() -> None:
+    # An explicit `with_shots(1)` is distinguishable from the unset default and
+    # still conflicts with a multi-record call.
+    instance, _ = _arg_instance()
+    instance = instance.with_shots(1)
```
maximilianruesch, line 505: Imo at the same time you should assert that the `shots` property returns `1` when no prior value was provided, catching test drift if the default ever changes.

maximilianruesch (changes requested):
💥 🚀 🧑‍🍳

guppylang/src/guppylang/defs.py
ss2165: yes we can just return a list

guppylang/src/guppylang/emulator/_args.py
ss2165: yes the important things should be on the top level docstrings or compiler guide, for the scope of this PR that means the .emulator and .run* APIs

guppylang/src/guppylang/emulator/_args.py
ss2165: Explicit over implicit - this function takes arg names as a parameter so is explicitly for known entrypoints with args, I prefer to enforce that contract

tests/error/misc_errors/entrypoint_unsupported_tuple_arg.err
ss2165: heterogeneous tuples are not like arrays, we'd need to inject constructor logic. Big follow up but I would say not big impact - I like the symmetry with `output`

tests/emulator/test_args.py
ss2165: There was an excess of unit tests due to copilot TDD and insufficient positive integration test coverage as you point out above, I've attempted to re-address this balance while leaving some unit tests useful for regression behind.

guppylang/src/guppylang/defs.py
```diff
@@ -192,7 +210,48 @@ def emulator(
                 )
             )
 
-        return builder.build(mod, n_qubits=qubits)
+        return builder.build(mod, n_qubits=qubits, arg_specs=arg_specs)
+
+    @pretty_errors
+    def _entrypoint_arg_specs(self) -> Sequence[EntrypointArgSpec]:
```
maximilianruesch:
```suggestion
    def _entrypoint_arg_specs(self) -> list[EntrypointArgSpec]:
```
Return types should be as narrow as possible (at least in private functions).

guppylang/src/guppylang/emulator/_args.py
```diff
@@ -0,0 +1,255 @@
...
+            "Mismatch between entrypoint parameter names "
+            f"({len(arg_names)}) and HUGR inputs ({len(input_types)})."
+        )
+
+    # The wrapper is added as a second public function alongside the original
+    # entrypoint (whose visibility is left unchanged). Having two public
+    # functions in the module is a deliberate choice to allow the original
+    # entrypoint to be called as a library function.
+    # Name of the no-input wrapper generated around the original entrypoint.
+    wrapper_name = "__entrypoint_with_args"
```
maximilianruesch: Should the wrapper name be generated from the former entrypoint name?

tests/integration/test_emulator_args.py
```diff
@@ -0,0 +1,161 @@
...
+    ("make_main", "value"),
+    [
+        pytest.param(_bool_echo, True, id="bool"),
+        pytest.param(_int_echo, 7, id="int"),
+        pytest.param(_float_echo, 1.5, id="float"),
+    ],
+)
+def test_scalar_arg_roundtrip(
+    make_main: Callable[[], GuppyFunctionDefinition[..., None]],
+    value: float,
```
maximilianruesch:
```suggestion
    expected: float,
```
or
```suggestion
    expected_value: float,
```
And similar for the test immediately below

tests/integration/test_notebooks.py
```diff
@@ -7,22 +7,28 @@
...
-    Path(__file__).parent.parent.parent / "examples" / "qaoa_maxcut_example.ipynb"
-)
 # Turn paths into strings, otherwise pytest doesn't display the names
 example_notebooks = [str(f) for f in example_notebooks]
 
+# Cells in qaoa_maxcut_example.ipynb whose outputs depend on the optimizer
+# convergence path, which varies across scipy/numpy versions and platforms.
+# The notebook is still executed in full — only the exact numerical outputs
+# of these cells are excluded from the regression comparison.
+_QAOA_STOCHASTIC_CELLS = (
```
maximilianruesch: Perhaps we should have a simple cell that tells us whether convergence was achieved or not (if it can even diverge). But that is a pretty minor point and I am happy to look past it. Or even the succeeding cells will provide vastly different values for diverging runs, so this may already be covered.

tests/integration/test_notebooks.py
```diff
@@ -7,22 +7,28 @@
...
+
 
 @pytest.mark.parametrize("notebook", example_notebooks)
 def test_example_notebooks(nb_regression, notebook: Path):
     nb_regression.diff_ignore += (
         "/metadata/language_info/version",
         "/cells/*/outputs/*/data/image/png",
     )
+    if "qaoa_maxcut_example" in str(notebook):
+        nb_regression.diff_ignore += _QAOA_STOCHASTIC_CELLS
```
maximilianruesch: Perhaps a `DIFF_IGNORE_CELLS` dict would be appropriate, where keys are checked as here, and values are added to nb_regression if their keys are found in the notebook string.

Just in an effort to keep this function general and all the "configuration" above and outside it. Also means you do not have to name the ignored diff twice, it could literally be
```
DIFF_IGNORE_CELLS = {
  # Reasoning for exclusion
  "qaoa_maxcut_example": (
    "cell1", # name the cell
    "cell2, # name the other cell
  )
}
```
Would imo be cleaner.

tests/emulator/test_args.py
maximilianruesch: I am way happier about this since it does not use / test `_` prefixed functions anymore, and thus it is way less "implementation detail testing" and way more "module testing" 👍

maximilianruesch (approved):

tests/integration/test_notebooks.py
ss2165: @CalMacCQ is intending to do a determinism pass on this notebook as a follow-up anyway, left as out of scope

PR #1897, opened 2026-06-22
feat(guppylang-internals)!: Infer unitary flags when loading PyTket circuits

### LINKED ISSUE #1325: Infer unitary flags for pytket circuits

nicolaassolini-qntm:
closes #1325 

Adds unitary flag inference when loading PyTKET circuits

BREAKING CHANGE: (pytket_circuits.py) added the field `unitary_flags_value: int` to `ParsedPytketDef`


tests/error/modifier_errors/pytket_w_measurement.err
CalMacCQ, line 1: Should there be a space here in this file for `b =guppy_circ(q)`?

Maybe its suppoed to be `b = guppy_circ(q)`

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
```diff
@@ -446,5 +463,22 @@ def _signature_from_circuit(
...
+    """Helper function for inferring unitary flags from a pytket circuit."""
+
+    from pytket.circuit import Circuit, OpType  # Decoupled import
+    from pytket.utils.stats import gate_counts  # Decoupled import
+
+    assert isinstance(input_circuit, Circuit)
+
+    counter = gate_counts(input_circuit)
+
+    if counter[OpType.Measure] > 0 or counter[OpType.Reset] > 0:
```
CalMacCQ: This seems like a fine check for now. Unfortuantely pytket doesn't have a `Op.is_unitary` property (maybe it should). Perhaps @cqc-alec knows of a more robust way to blacklist non-unitary ops.

What is the expected behaviour here for `OpType`s which act on classical bits? I.e. [Conditional](https://docs.quantinuum.com/tket/api-docs/optype.html#pytket.circuit.OpType.Conditional) or [SetBits](https://docs.quantinuum.com/tket/api-docs/optype.html#pytket.circuit.OpType.SetBits). Will there be an error upon `guppy.load_pytket` in these cases?

guppylang-internals/src/guppylang_internals/tys/errors.py
```diff
@@ -243,3 +243,12 @@ class HigherOrderHint(Help):
         )
         callable_name: str
         function_description: str
+
+    @dataclass(frozen=True)
+    class PyTketHint(Help):
+        func_name: str
+        message: ClassVar[str] = (
+            "The function `{func_name}` corresponds to a PyTKET circuit containing a "
```
CalMacCQ: I would suggest using `pytket` rather than `PyTKET` here as this is consistent with the name of the pypi package.

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
```diff
@@ -162,13 +171,16 @@ class ParsedPytketDef(CallableDef, CompilableDef):
...
         use_arrays: Whether the circuit function should use arrays as input types.
+        unitary_flags_value: The integer value of the unitary flags for this circuit.
     """
 
     input_circuit: Any
     use_arrays: bool
 
     source_span: Span | None  # Only set for load_pytket for debug purposes.
 
+    unitary_flags_value: int
```
CalMacCQ, line 182: Maybe I am missing something but why is `unitary_flags_value` an `int` here?

I looked at the definition of `UnitaryFlags`

```python
class UnitaryFlags(Flag):
    """Flags that can be set on functions to indicate their unitary properties.

    The flags indicate under which conditions a function can be used
    in a unitary context.
    """

    NoFlags = 0
    Control = auto()
    Dagger = auto()
```

Do "non-zero" values relate to `Control` and `Dagger` somehow?

guppylang-internals/src/guppylang_internals/tys/errors.py
```diff
@@ -243,3 +243,12 @@ class HigherOrderHint(Help):
         )
         callable_name: str
         function_description: str
+
+    @dataclass(frozen=True)
+    class PyTketHint(Help):
```
CalMacCQ: I would suggest `PytketHint` rather than `PyTketHint` to be consistent with other pytket related classes in guppy internals e.g. `RawPytketDef`

CalMacCQ (commented):
Looks good! Thanks for doing this. Just a few questions and minor suggestions.

guppylang-internals/src/guppylang_internals/definition/pytket_circuits.py
nicolaassolini-qntm: I only need `unitary_flags.value`, i.e. the value corresponding to the flag to be stored in the metadata not the flag it self

nicolaassolini-qntm:
Addressed your suggestions. I will ask Alec tomorrow if there is a better way to check for unitary gates

CalMacCQ (approved):
Looks good to me, thanks @nicolaassolini-qntm !

tatiana-s (approved):

Issue #1914, opened 2026-06-24
[Feature]: Treat types implementing `__call__` as `Callable`


mark-koch:
### Problem Statement

This is non-breaking, so optional for the 1.0 release

### Effort Estimate

None

PR #1860, opened 2026-06-15
feat!: Rename `Callable` type

mark-koch:
Closes  #1912

BREAKING CHANGE: `guppylang_internals.engine.callable_type_def` renamed to `function_type_def`
BREAKING CHANGE: `guppylang_internals.tys.builtin.CallableTypeDed` renamed to `FunctionTypeDef`
BREAKING CHANGE: `guppylang_internals.tys.errors.InvalidCallableTypeError` renamed to `InvalidFunctionTypeError`
BREAKING CHANGE: `guppylang_internals.tys.errors.CallableComptimeError` renamed to `FunctionTypeComptimeError`

PR #1929, opened 2026-06-25
feat!: Add `Callable` protocol

### LINKED ISSUE #1913: [Feature]: Add `Callable` protocol, only satisfied by `Function` types
### Problem Statement

Make protocol definitions customisable and add a `Callable` protocol that only accepts `Function` types with matching signatures

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

mark-koch:
Closes #1913

BREAKING CHANGE: `guppylang_internals.tys.errors.InvalidFunctionTypeError` now requires a new `kind` argument

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -131,6 +131,7 @@ from guppylang_internals.nodes import (
 from guppylang_internals.span import Span, to_span
 from guppylang_internals.tys.arg import Argument, ConstArg, TypeArg
 from guppylang_internals.tys.builtin import (
+    CallableProtocolInst,
     bool_type,
     float_type,
     frozenarray_type,
@@ -357,12 +358,23 @@ class ExprChecker(AstVisitor[tuple[ast.expr, Subst]]):
             return self.visit_Call(node, ty)
 
         # Otherwise, it must be a function as a higher-order value - something
-        # whose type is either a FunctionType or a Tuple of FunctionTypes
+        # whose type is either a FunctionType, a genric parameter with a `Callable`
+        # bound, or a Tuple of FunctionTypes
         if isinstance(func_ty, FunctionType):
             args, subst, inst = check_call(func_ty, node.args, ty, node, self.ctx)
             check_inst(func_ty, inst, node)
             node.func = instantiate_poly(node.func, func_ty, inst)
             return with_loc(node, LocalCall(func=node.func, args=args)), subst
+
+        if isinstance(func_ty, BoundTypeVar):
+            for protocol in func_ty.implements:
+                if isinstance(protocol, CallableProtocolInst):
+                    args, subst, inst = check_call(
+                        protocol.sig, node.args, ty, node, self.ctx
+                    )
+                    assert inst == (), "Callables are not generic"
+                    node.func = instantiate_poly(node.func, protocol.sig, inst)
+                    return with_loc(node, LocalCall(func=node.func, args=args)), subst
 
         if isinstance(func_ty, TupleType) and (
             function_elements := parse_function_tensor(func_ty)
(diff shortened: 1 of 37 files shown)


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -265,7 +266,13 @@ def check(
         # When checking against a variable, we have to synthesize
         if isinstance(ty, ExistentialTypeVar):
             expr, syn_ty = self._synthesize(expr, allow_free_vars=False)
-            return with_type(syn_ty, expr), {ty: syn_ty}
+            expr, subst, inst = check_type_against(
+                syn_ty, ty, expr, self.ctx, self._kind
+            )
```
mark-koch, line 271: Before, we just greedily returned the substiution `{ty: syn_ty}` but this breaks down if the existential has a bound that itself contains existentials. In that case, we want a subsitution that also solves the existentials in the bound.

By calling `check_type_against` we hook into the existing machinary to do that stuff

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1092,6 +1120,7 @@ def check_type_against(
         # Finally, check that the instantiation respects the linearity requirements and
         # if the unitary flags match
         check_inst(act, inst, node)
+        exp = exp.substitute(subst)
```
mark-koch, line 1123: Now `exp` could be an existential variable, so we should apply the substitution to get the actual function type

guppylang-internals/src/guppylang_internals/checker/protocol_checker.py
```diff
@@ -37,9 +37,6 @@ class ImplProofBase(ABC):
     proto: ProtocolInst
     ty: Type
 
-    def __post_init__(self) -> None:
-        assert all(not arg.unsolved_vars for arg in self.proto.type_args)
-
```
mark-koch, line 42: A bit annoying, but we have to get rid of this check since we're now actually constructing proofs that still contain unsolved vars.

Concretely this happens when unifying with things that have bounds with existentials

guppylang-internals/src/guppylang_internals/definition/protocol.py
```diff
@@ -174,7 +175,7 @@ def check_instantiate(
 class CheckedProtocolDef(ProtocolDef, CompiledDef):
     """A fully checked protocol definition."""
 
-    defined_at: ast.ClassDef
+    defined_at: ast.ClassDef | None
```
mark-koch, line 178: We need to allow `None` here for compiler internal defined protocols (such as callable)

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -203,8 +205,23 @@ def _arg_from_instantiated_defn(
...
+            )
+        # Special cases for the `Callable` protocol
+        case CallableProtocolDef():
+            sig = _parse_function_type(arg_nodes, node, ctx, "Callable")
+            proto_inst = CallableProtocolInst(sig)
+            param = TypeParam(
+                len(ctx.param_var_mapping),
+                name=str(proto_inst),
+                must_be_copyable=True,
+                must_be_droppable=True,
```
mark-koch, line 223: For now, all callable things are copyable and droppable. We can relax this once we consider more things callable

guppylang-internals/src/guppylang_internals/tys/builtin.py
```diff
@@ -184,6 +195,70 @@ def check_instantiate(
...
+
+    def check_implemented_by(
+        self, ty: "Type", loc: ToSpan | None
+    ) -> "tuple[ImplProof, Subst]":
+        from guppylang_internals.checker.protocol_checker import (
+            AssumptionImplProof,
+            ConcreteImplProof,
+        )
+
+        if isinstance(ty, FunctionType):
```
mark-koch, line 240: For now, we only consider `Function` types to be callable.

In the future, we should add a check for a `__call__` method here (see #1914)

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -970,15 +971,17 @@ def _unify_type_var(var: ExistentialTypeVar, t: Type, subst: "Subst") -> "Subst
...
     if var.implements:
-        from guppylang_internals.checker.protocol_checker import check_protocol
-
-        try:
-            for proto in var.implements:
-                _, proto_subst = check_protocol(t, proto)
+        for proto in var.implements:
+            try:
+                loc = DUMMY_SPAN  # We catch the error later so the span doesn't matter
+                _, proto_subst = proto.check_implemented_by(t, loc)
```
mark-koch, line 977: We need to pass a location here since the protocol checker sometimes tries to construct errors with span labels

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -970,15 +971,17 @@ def _unify_type_var(var: ExistentialTypeVar, t: Type, subst: "Subst") -> "Subst
...
                 subst |= proto_subst
-        except GuppyError:
-            return None
-    return {var: t, **subst}
+            except GuppyError:  # noqa: PERF203
+                # At this point, we only use protocol checking to infer types. If the
+                # protocol is not satisfied, we still keep going. The error will be
+                # raised later when we check the inferred instantiation.
+                pass
```
mark-koch, line 983: This is because protocol checking can fail for lots of different reasons.atm. For example, if the protocol *definition* is invalid, then the error will only be thrown once we check an *instance*. If we consider this as a unification failure, we'd get lots of confusing error messages (especially now that we apply more aggressive unification due to the additional call to `check_type_against`)

tests/error/comptime_arg_errors/callable.err
mark-koch, line 1: I forgot to rename these files in #1860, so I change them to `Callable` here and add new `Function` versions below

croyzor (approved):

PR #1930, opened 2026-06-25
refactor!: Move libraries into their own module

### LINKED ISSUE #1923: [Feature]: Move library creation into its own module
### Problem Statement

Currently, libraries are created as `guppy.library(...)`. However, the `@guppy` decorator (or class, more broadly), should only include functions that create Guppy symbols.

We will want to move creation of libraries (and more broadly, all library stuff), into a module `guppylang.libraries`. As a follow up, a `@link_name` decorator will follow, replacing the current `link_name=...` kwarg on `@guppy`. See #1925.

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

maximilianruesch:
Creates a new module `guppylang.library` where library creation and definitions now reside.

Closes #1923

BREAKING CHANGE: Libraries are now created using `guppylang.library.GuppyLibrary.from_members` instead of `guppylang.guppy.library`.
BREAKING CHANGE: `guppylang.defs.GuppyLibrary` has moved to `guppylang.library.GuppyLibrary`.


guppylang/src/guppylang/library.py
```diff
@@ -0,0 +1,66 @@
...
+
+        @guppy
+        def foo() -> int:
+            return 42
+        @guppy
+        def bar() -> int:
+            return 7
+
+        # Compilable collection containing `foo` and `bar`.
+        lib = GuppyLibrary.library(foo, bar)
```
nicolaassolini-qntm:
```suggestion
        lib = GuppyLibrary.from_members(foo, bar)
```
?

nicolaassolini-qntm (changes requested):
Just one comment

guppylang/src/guppylang/library.py
maximilianruesch: Yeah oops

nicolaassolini-qntm (approved):

PR #1933, opened 2026-06-25
chore: Update tket dependency from 0.14.1 to 0.14.2 and fix modifier tests

nicolaassolini-qntm:
Since tket 1.14.2 adds `ModifierResolverPass` to `NormalizeGuppy`, this PR refactors the modifier integration tests to ensure that CI validation does not fail. As part of this refactor, it adds a `validate` flag to the `@pytest.fixture`, allowing specific HUGRs to be excluded from export during validation.

### DIFF
--- a/tests/integration/test_modifier.py
+++ b/tests/integration/test_modifier.py
@@ -218,8 +218,9 @@ def test_nested_modifiers(validate):
 def test_free_linear_variable_in_modifier(validate):
     T = guppy.type_var("T", copyable=False, droppable=False)
 
-    @guppy.declare(controllable=True)
-    def use(a: T) -> None: ...
+    @guppy(controllable=True)
+    def use(a: T) -> None:
+        pass
 
     @guppy.declare
     def discard(a: T @ owned) -> None: ...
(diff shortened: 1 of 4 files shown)


tests/integration/test_modifier.py
```diff
@@ -348,6 +307,11 @@ def main(q: qubit) -> None:
     validate(main.compile_function())
 
 
+# Skipped because tket2 still contains some bugs with higher-order functions.
+# Waiting for:
+# - https://github.com/Quantinuum/guppylang/issues/1917 and
+# - https://github.com/Quantinuum/tket2/issues/1710
+@pytest.mark.skip("CI fails, waiting for fix on tket2")
```
maximilianruesch:
```suggestion
@pytest.mark.xfail("Waiting for fix on tket2")
```
Also amend this message for the other use here, and try and use xfail instead of skip throughout this file.

maximilianruesch (changes requested):

tests/integration/test_modifier.py
```diff
@@ -348,6 +307,11 @@ def main(q: qubit) -> None:
     validate(main.compile_function())
 
 
+# Skipped because tket2 still contains some bugs with higher-order functions.
```
maximilianruesch:
```suggestion
# Tket2 still contains some bugs with higher-order functions.
```

tests/integration/test_modifier.py
```diff
@@ -363,15 +327,19 @@ def main(ctrl: qubit, q: qubit) -> None:
...
```
maximilianruesch:
```suggestion
# Tket2 still contains some bugs with higher-order functions.
```

maximilianruesch (approved):

Issue #1926, opened 2026-06-24
[Feature]: Turn `max_qubits` decorator kwarg into its own decorator


mark-koch:
### Problem Statement

Should be implemented via the @metatdata decorator from #1924, and renamed to `expected_qubits`.

### Effort Estimate

None

Issue #1925, opened 2026-06-24
[Feature]: Turn `link_name` decorator kwarg into its own decorator


mark-koch:
### Problem Statement

Should be implemented via the `@metatdata` decorator from #1924

### Effort Estimate

None

PR #1937, opened 2026-06-25
feat!: Add `@metadata` decorator

### LINKED ISSUE #1924: [Feature]: Add metadata decorator
### Problem Statement

Add a new `@metadata` decorator (notice not starting with `@guppy.`!) to add metadata to guppy definitions.

The decorator should be placed *below* the `@guppy` decorator that annotates the definition:

```python
@guppy
@metadata("key1", "value1")
@metadata("key2", "value2")
def foo() -> None: ...
```

A Python error should be thrown at decoration-time if a user tries to put `@metadata` above `@guppy`.

### Proposed Solution

Can be implemented by making the `@metadata` decorator populate a special attribute of the decorated Python function via `setattr`  

### Component

Guppy Compiler

### Effort Estimate

None

### Additional Context

_No response_

nicolaassolini-qntm:
closes #1924 

This PR adds a new decorator (`@metadata`) that can be used to set arbitrary metadata to guppy definitions.

BREAKING CHANGE: Internals: changed `ReservedMetadataKeysError.message`.


guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -60,6 +60,11 @@ def set_max_qubits(self, max_qubits: int) -> None:
     def set_unitary_flags(self, value: int) -> None:
         self._node_metadata[MetadataUnitaryFlags] = value
 
+    def set_generic_metadata(self, key: str, value: JsonType) -> None:
+        if key in self._RESERVED_KEYS:
```
maximilianruesch:
```suggestion
        if key in FunctionMetadata.reserved_keys():
```

guppylang/src/guppylang/decorator.py
```diff
@@ -649,6 +657,35 @@ def foo(default_reg: array[qubit, 2],
         return GuppyFunctionDefinition(defn)
 
 
+class _MetadataDecorator:
```
maximilianruesch: This can be just a simple decorator function (or rather a function producing a decorator). And then the signature can specify `key: str, value: Any`!

guppylang/src/guppylang/decorator.py
```diff
@@ -649,6 +657,35 @@ def foo(default_reg: array[qubit, 2],
...
+                }
+                return f
+
+            return decorator
+
+        got = 0 if len(args) == 1 and _is_bare_decorated_target(args[0]) else len(args)
+        raise TypeError(f"@metadata requires exactly 2 arguments, got {got}")
+
+
+def _is_bare_decorated_target(arg: Any) -> bool:
```
maximilianruesch: This should not be needed anymore

guppylang-internals/src/guppylang_internals/dummy_decorator.py
```diff
@@ -55,6 +55,18 @@ def load_pytket(self, *args: Any, **kwargs: Any) -> Any:
         return None
 
 
+class _DummyMetadata:
```
maximilianruesch: This can go when metadata becomes a regular function

tests/metadata/test_metadata_decorator.py
maximilianruesch, line 1: Love this test. Can you add one that shows that one can wrap metadata to define their own custom fixed-key metadata decorator (like we plan in #1925 or #1926)?

maximilianruesch (changes requested):

tests/error/test_decorator_errors.py
```diff
@@ -0,0 +1,36 @@
+import pytest
+
+from guppylang.decorator import guppy, metadata
+
+
+
+def test_metadata_decorator_position():
```
maximilianruesch, line 6: An awful lot of whitespace here, innit?

tests/error/test_decorator_errors.py
```diff
@@ -0,0 +1,36 @@
...
+        match=r"metadata\(\) missing 1 required positional argument: 'value'",
+    ):
+        @guppy
+        @metadata("key1")
+        def foo() -> None:
+            pass
+
+    with pytest.raises(
+        TypeError,
+        match=r"metadata\(\) missing 1 required positional argument: 'value'",
```
maximilianruesch, line 29: The error here is not great, but I guess it is the best we can do without more effort?

tests/metadata/test_metadata.py
```diff
@@ -91,3 +93,18 @@ def test_add_metadata_property_max_qubits():
     add_metadata(mock_hugr_node, guppy_metadata)
 
     assert mock_hugr_node.metadata.as_dict() == {"tket.hint.max_qubits": 5}
+
+
+def test_metadata_reserved_keys_using_decorator():
```
maximilianruesch: Imo this test should be in the metadata decorator test, because it concerns the decorator.

tests/metadata/test_metadata_decorator.py
```diff
@@ -0,0 +1,40 @@
...
+        data
+        for _, data in hugr.nodes()
+        if isinstance(data.op, FuncDefn) and data.op.f_name.endswith(".foo")
+    ]
+
+    assert len(funcs) == 1
+    assert funcs[0].metadata["test-key"] == {"nested": ["value"]}
+
+
+def test_wrapping_metadata_decorator():
```
maximilianruesch, line 25: Love it

guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -30,11 +30,22 @@ class MetadataAlreadySetError(Fatal):
...
-        "The following metadata keys are reserved by Guppy but also provided in "
-        "additional metadata: `{keys}`"
-    )
+    message: ClassVar[str] = "{rendered_message}"
     keys: set[str]
+    singular: bool
+
+    @property
+    def rendered_message(self) -> str:
+        if self.singular:
```
maximilianruesch: What if this is instead determined in here via the length check on `keys` (that you do anyway somewhere else?). Imo "singular" or "plural" should be an implementation detail here, it does not need to be part of the API surface of this Error.

maximilianruesch (changes requested):

tests/error/test_decorator_errors.py
nicolaassolini-qntm: "A python error for a python programmer"
This is the default python error, which is as bad as all python errors. `@metadata` is quite simple, you can know how to use it from the docstring, I think the default error is good enough.

maximilianruesch (approved):
:shipit:

PR #1941, opened 2026-06-26
feat!: Add  `@link_name` decorator

### LINKED ISSUE #1925: [Feature]: Turn `link_name` decorator kwarg into its own decorator

nicolaassolini-qntm:
closes #1925.

Linked names are no longer specified using the `link_name` keyword argument in the `@guppy` decorator. Instead, they should be specified using the dedicated `@link_name` decorator.

BREAKING CHANGE: Using the `link_name` keyword argument in `@guppy` now raises an error.
BREAKING CHANGE: `GuppyKwargs.link_name` has been removed.


guppylang/src/guppylang/decorator.py
```diff
@@ -693,6 +694,25 @@ def decorator(f: Any) -> Any:
     return decorator
 
 
+def link_name(name: str) -> Any:
```
maximilianruesch: This decorator should live in `guppylang.library`, not here.

guppylang/src/guppylang/decorator.py
```diff
@@ -78,7 +78,9 @@
     OverloadedFunctionDef,
 )
 
-__all__ = ("GuppyKwargs", "custom_guppy_decorator", "guppy")
+__all__ = ("GuppyKwargs", "custom_guppy_decorator", "guppy", "link_name", "metadata")
+
+LINK_NAME_KEY = "link_name"
```
maximilianruesch: This should be a private attribute, prefix with `_`.

guppylang/src/guppylang/decorator.py
```diff
@@ -855,17 +878,26 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
     return ParsedGuppyKwargs(
         flags=flags,
         metadata=metadata,
-        link_name=link_name,
     )
 
 
+def _get_link_name(f: Callable[..., Any]) -> str | None:
+    """Returns the link name for a Guppy function, if it has one."""
```
```suggestion
```
This doc comment does not really add anything.

Apart from removing it, you can say more about how the function does this, e.g. something like "extracts the link name from annotations made with the `@metadata` decorator or its wrappers" but most of that is clear from the implementation. I do not think we need it.

guppylang/src/guppylang/decorator.py
```diff
@@ -846,7 +865,11 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
     if "max_qubits" in kwargs:
         metadata.set_max_qubits(kwargs.pop("max_qubits"))
 
-    link_name = kwargs.pop("link_name", None)
+    if LINK_NAME_KEY in kwargs:
+        raise TypeError(
+            "`link_name` keyword argument is not allowed in @guppy decorator, "
+            "use @link_name decorator instead"
+        )
```
maximilianruesch, line 857:
```suggestion
        raise TypeError(
            "`link_name` keyword argument has been removed from the `@guppy` decorator, "
            "use the `@link_name` in `guppylang.library` decorator instead."
        )
```
This makes it clearer 1. where the new decorator resides, 2. improves formatting, and 3. adjusts wording to signal that the keyword _has been_ allowed at some point, but is no longer. Eventually, we will remove this more specific error message (same as we will remove the more specific import errors for outdated reexports), probably after v0 support has truly ended.

guppylang/src/guppylang/decorator.py
```diff
@@ -855,17 +878,26 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
...
+@hide_trace
 def _add_generic_metadata(f: Callable[..., Any], metadata: FunctionMetadata) -> None:
     """Adds the given metadata to the function's `__guppy_metadata__` attribute, which
     is used by the compiler to store metadata for Guppy functions.
     """
     custom_metadata = getattr(f, "__guppy_metadata__", {})
     assert isinstance(custom_metadata, dict)
     for key, value in custom_metadata.items():
+        if key == LINK_NAME_KEY:
+            continue  # link_name is handled separately
```
maximilianruesch: We should not have special handling for specific keys here, that violates the purpose of this abstraction to "metadata". Imo there is no harm in populating the metadata, although it may be a bit confusing later on for users, but we do not really care (it may even be useful, e.g. finding the link name that something was annotated with, rather than the final calculated link name that is used as the function name). Thus:
```suggestion
```

guppylang/src/guppylang/decorator.py
```suggestion
LINK_NAME_KEY = "guppylang.library.link_name"
```
or
```suggestion
LINK_NAME_KEY = "guppylang.link_name"
```
since we now allow this to exist in metadata and it is not just a placeholder anymore.

guppylang/src/guppylang/decorator.py
```diff
@@ -78,7 +78,9 @@
     OverloadedFunctionDef,
 )
 
-__all__ = ("GuppyKwargs", "custom_guppy_decorator", "guppy")
+__all__ = ("GuppyKwargs", "custom_guppy_decorator", "guppy", "link_name", "metadata")
```
maximilianruesch: I prefer these to be line-broken (even if not necessary), it makes the diff easier to see.

guppylang/src/guppylang/decorator.py
```diff
@@ -855,17 +878,26 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
     return ParsedGuppyKwargs(
         flags=flags,
         metadata=metadata,
-        link_name=link_name,
     )
 
 
+def _get_link_name(f: Callable[..., Any]) -> str | None:
```
maximilianruesch: You can either also just move this to the new module, or put the private things into the internals package and import both into `guppylang.library` and here.

tests/integration/test_link_name.py
```diff
@@ -277,3 +288,17 @@ def main() -> None:
         "tests.integration.test_link_name.FileLevelEnum.superb_name_defn",
         "tests.integration.test_link_name.FileLevelEnum.superb_name_decl",
     }
+
+
+def test_error_when_using_old_kwarg():
+    """Asserts that using the old `link_name` kwarg raises an error."""
```
maximilianruesch:
```suggestion
    """Asserts that using the old `link_name` kwarg raises a helpful error."""
```
Because that is the entire point of this test, generic errors we have for other kwargs anyway, but not helpful ones.

maximilianruesch (changes requested):

guppylang/src/guppylang/decorator.py
```diff
@@ -846,7 +849,11 @@ def _parse_kwargs(kwargs: GuppyKwargs) -> ParsedGuppyKwargs:
     if "max_qubits" in kwargs:
         metadata.set_max_qubits(kwargs.pop("max_qubits"))
 
-    link_name = kwargs.pop("link_name", None)
+    if "link_name" in kwargs:
+        raise TypeError(
+            "`link_name` keyword argument has been removed from the `@guppy` decorator,"
+            " use the `@link_name` in `guppylang.library` decorator instead."
```
```suggestion
            " use the `@link_name` decorator from `guppylang.library` instead."
```
Oop, the wording was bad on my end apparently

guppylang/src/guppylang/library.py
```diff
@@ -1,14 +1,21 @@
...
 
 from guppylang_internals.definition.common import DefId
 from guppylang_internals.engine import DEF_STORE, ENGINE
 from hugr.package import Package
 from typing_extensions import Self
 
+from guppylang.decorator import metadata
 from guppylang.defs import GuppyDefinition, _update_generator_metadata
 
+LINK_NAME_METADATA_KEY = "guppylang.library.annotated_link_name"
```
maximilianruesch: @mark-koch Any thoughts on whether this should be public (since the behaviour is public and thus we want warnings when we change the key) or not?

nicolaassolini-qntm:
Did the rewording

guppylang/src/guppylang/library.py
mark-koch: Would lean towards private since it feels more like a compiler implementation detail. In particular, we don't view changing the Hugr output as a breaking language change, right?

nicolaassolini-qntm: Changing the Hugr output may be a breaking change if it alters some compilation passes. It is not breaking by itself, but it may break something in relation to hugr or tket

maximilianruesch: In this case it should be fine, as in it does not tangent other places. So let us privatise this (and move `_get_link_name` accordingly).

maximilianruesch (approved):

PR #1942, opened 2026-06-26
feat!: Add `@expected_qubits` decorator and remove `max_qubits` kwarg

### LINKED ISSUE #1926: [Feature]: Turn `max_qubits` decorator kwarg into its own decorator

nicolaassolini-qntm:
closes #1926

This PR turns the key  kwarg`max_qubits` of `@guppy` decorator into `@expected_qubits` decorator.

BREAKING CHANGE: guppylang: using the kwarg `max_qubits` of `@guppy` raises a Type Error now, using `@expected_qubits` is required

BREAKING CHANGE: guppylang-internals: renamed `guppylang_internals/metadata/max_qubits.py` to `guppylang_internals/metadata/expected_qubits.py`
BREAKING CHANGE: guppylang-internals: renamed `MetadataMaxQubits` to `MetadataExpectedQubitsHint`
BREAKING CHANGE:  guppylang-internals: renamed `FunctionMetadata.set_max_qubits` and `FunctionMetadata.get_max_qubits` in `FunctionMetadata.set_expected_qubits` and `FunctionMetadata.get_expected_qubits`


guppylang-internals/src/guppylang_internals/metadata/expected_qubits.py
```diff
@@ -1,12 +1,14 @@
...
 from hugr.metadata import Metadata
 from hugr.utils import JsonType
 
 
 @dataclass(frozen=True)
-class MetadataMaxQubits(Metadata[int]):
-    KEY = "tket.hint.max_qubits"
+class MetadataExpectedQubits(Metadata[int]):
+    KEY = "tket.hint.expected_qubits"
+    ALIASES: ClassVar[list[str]] = ["tket.hint.max_qubits"]
```
maximilianruesch: Aliases will not work in this direction. Rather, we will need to rename the key in tket (where it is hopefully defined) and add an alias there to where it can be read.

We do not have any compatibility guarantees in the Guppy stage itself, so we do not need an alias (which would be used for reading only anyway) here.

guppylang/src/guppylang/decorator.py
```diff
@@ -874,7 +892,10 @@ def _add_generic_metadata(f: Callable[..., Any], metadata: FunctionMetadata) ->
     custom_metadata = getattr(f, "__guppy_metadata__", {})
     assert isinstance(custom_metadata, dict)
     for key, value in custom_metadata.items():
-        metadata.set_generic_metadata(key, value)
+        if key == MetadataExpectedQubits.KEY:
+            metadata.set_expected_qubits(value)
+        else:
+            metadata.set_generic_metadata(key, value)
```
maximilianruesch: Oof, this is again violating the abstraction principle... But yeah, this time there is no easy way around it, perhaps except for allowing overrides to reserved metadata, which would imo be fine. But that we can also decide later.

maximilianruesch (changes requested):

guppylang-internals/src/guppylang_internals/metadata/expected_qubits.py
nicolaassolini-qntm: addressed https://github.com/Quantinuum/tket2/pull/1765

nicolaassolini-qntm:
I renamed MetadataExpectedQubits in MetadataExpectedQubitsHint to make it more similar to tket2 equivalent

maximilianruesch (approved):

PR #1827, opened 2026-06-08
feat!: Implicit loading of Python values

mark-koch:
Closes #1158 

BREAKING CHANGE: `guppylang_internals.checker.errors.comptime_errors.IllegalComptimeExpressionError` renamed to `UnsupportedPythonValueError`
BREAKING CHANGE: `guppylang_internals.tys.errors.IllegalComptimeTypeArgError` renamed to `IllegalPythonTypeArgError`

PR #1944, opened 2026-06-26
feat!: Implicit loading of Python values

### LINKED ISSUE #1158: Allow captures from Python scope without explicit comptime?
We should decide if we want this for 1.0

ss2165:
Based on #1827 from @mark-koch 

Adds support for coercion of python `int` to guppy `nat` on top to support e.g. `range(x)` where x is a python value in scope.

Also updates examples and integration tests where appropriate to use more concise syntax

BREAKING CHANGE: guppylang_internals.checker.errors.comptime_errors.IllegalComptimeExpressionError renamed to UnsupportedPythonValueError
BREAKING CHANGE: guppylang_internals.tys.errors.IllegalComptimeTypeArgError renamed to IllegalPythonTypeArgError

Closes #1158


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -295,6 +295,20 @@ def visit_Constant(self, node: ast.Constant, ty: Type) -> tuple[ast.expr, Subst]
...
 
+    def visit_Name(self, node: ast.Name, ty: Type) -> tuple[ast.expr, Subst]:
+        """Check a name against an expected type, propagating the hint for Python
+        values so that e.g. a non-negative Python int variable is accepted where a
+        ``nat @comptime`` is expected (mirroring what ``visit_Constant`` does for
+        literals)."""
+        if node.id in self.ctx.globals:
+            match self.ctx.globals[node.id]:
+                case PythonObject(obj=val):
+                    if python_value_to_guppy_type(val, node, ty) is not None:
```
mark-koch: Success of `python_value_to_guppy_type` doesn't necessarily mean that the final type matches `ty`. It is only used as a hint, not for emitting errors.

See `ExprChecker.visit_ComptimeExpr` for how to handle this. Could even factor it out into helper?

examples/ghz_and_graph.ipynb
```diff
@@ -75,8 +75,8 @@
     "    @guppy\n",
     "    def main() -> None:\n",
     "        # allocate number of qubits specified from outer\n",
-    "        # python function using a `comptime` expression.\n",
-    "        q = array(qubit() for _ in range(comptime(n_qb)))\n",
+    "        # Python variable referenced implicitly as a compile-time value.\n",
```
nicolaassolini-qntm: maybe
```
# Allocate the number of qubits specified by an outer Python
# variable, implicitly referenced as a compile-time value.

```

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -411,15 +442,14 @@ def visit_ComptimeExpr(
...
-        if act := python_value_to_guppy_type(python_val, node.value, ty):
-            subst = unify(ty, act, {})
-            if subst is None:
-                self._fail(ty, act, node)
-            act = act.substitute(subst)
-            subst = {x: s for x, s in subst.items() if x in ty.unsolved_vars}
-            return with_type(act, with_loc(node, ast.Constant(value=python_val))), subst
-
-        raise GuppyError(IllegalComptimeExpressionError(node.value, type(python_val)))
+        if (result := self._check_python_value(python_val, node.value, ty)) is not None:
```
nicolaassolini-qntm: I find more logic first checking `if python_value_to_guppy_type` then `_check_python_value`. 
I would also consider moving `python_value_to_guppy_type` out of `_check_python_value`, thus in `visit_Name` we can do:
```
act = python_value_to_guppy_type(value, node, ty)
if act is None:
    return None
```
first, and then call `_check_python_value` with act

tests/integration/test_comptime_expr.py
nicolaassolini-qntm, line 75: is this `comptime` redundant?

nicolaassolini-qntm, line 138: do we need this comptime?

nicolaassolini-qntm, line 185: all the tests below use comptime for single variables, do we want to keep them?

nicolaassolini-qntm (changes requested):
Good, just some small comments

tests/integration/test_comptime_expr.py
ss2165: since this file tests comptime_expr explicitly better to keep I think

ss2165: since this file tests comptime_expr explicitly better to keep I think

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
ss2165: done

nicolaassolini-qntm (approved):
good, thanks!

PR #1945, opened 2026-06-26
feat: Choose a valid protocol when multiple protocols give a method

### LINKED ISSUE #1856: [Feature]: Improve resolution for protocols with the same method
### Problem Statement

Following #1688, there is work to be done on resolving between protocols with the same methods. For example, in the following code:
```python
@guppy.protocol
class Foo[T]:
    @guppy.require
    def foo(self, t: T) -> T: ...

class bar[A, B, T: Foo[A], Foo[B]](t: T, b: B) -> B:
    return t.foo(b)
```
checking `bar` will raise an error, because it checks the protocols that `T` should implement in the order they're written. Once the protocol checker sees that `T` contains a method `foo` (from `Foo[A]`) it is happy to proceed, but then the call fails because `A` doesn't match the argument type `B`.

The protocol should be able to resolve the `foo` call to the `Foo[B]` protocol instance

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

croyzor:
Closes #1856.

Catches the case where a method `foo` is called on a bound type variable which is required to implement multiple protocols with a `foo` method. The `t.foo` ast node is turned into an `UnresolvedProtocolMethod` node, which gives the list of matching protocol implementations. When this is applied to arguments, we try to check each of the possible methods against the arguments and use the first one that works.

Previous behaviour was to use the first protocol that has a `foo` method.


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -340,7 +344,10 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
             if isinstance(defn, CallableDef):
                 return defn.check_call(node.args, ty, node, self.ctx)
 
-            from guppylang_internals.definition.protocol import ParsedProtocolDef
+            from guppylang_internals.definition.protocol import (
+                CheckedProtocolDef,
+                ParsedProtocolDef,
+            )
 
             # Protocol methods don't have their own definition, we have to look up the
```
tatiana-s, line 388: nit: maybe mention something in comment that makes it more clear why there are two cases handling protocol methods, here and the unresolved ones below?

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -266,7 +266,7 @@ def _arg_from_proto(
 ) -> Argument:
     """Parses a protocol definition with type args into an argument."""
     proto_args = [arg_from_ast(arg_node, ctx) for arg_node in arg_nodes]
-    inst = proto_defn.check_instantiate(proto_args, node)
+    inst = proto_defn.check(ctx.globals).check_instantiate(proto_args, node)
```
tatiana-s: Why is this change necessary now?

tests/error/protocol_errors_py312/no_protocols_match.err
```diff
@@ -0,0 +1,13 @@
+Error: No given protocols apply (at $FILE:22:11)
+   | 
+20 | @guppy
+21 | def bar_flt[T: (FooNat, FooInt)](t: T, n: float) -> nat:
+22 |     return t.foo(n)
+   |            ^^^^^^^^ Couldn't find a protocol implementation providing `foo` with
+   |                     a valid signature.
+
+Note: Protocol `FooNat` provides method `foo` with signature:  `(T, nat) -> nat`
```
tatiana-s: I like the hints :)

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -653,25 +717,46 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                    return with_loc(
+                        node, PartialApply(func=name_node, args=[node.value])
+                    ), ty_without_self
+                case _:
+                    return UnresolvedProtocolMethod(
+                        node.value,
+                        ty,
+                        valid_proto_impls,
+                        node.attr,
+                    ), ty  # Dummy type!!
```
tatiana-s: Is there any big risk of the dummy type accidentally being used and causing issues? But I guess trying to change the signature of this method would cause more issues in itself

tatiana-s (approved):
A couple of comments but looks good to me in general 👍

acl-cqc:
Not sure whether this is part of v1.0 milestone or not, and haven't looked at the code in detail but one thought:

Would it be possible to combine this with OverloadedFunctionDef? That seems basically the same idea.

That said, what is "the first one"? In the order in which the protocols were listed in the type of the method receiver? I +1 a feature to allow specifying which protocol explicitly, but that can be a followup/separate PR.

croyzor:
> Would it be possible to combine this with OverloadedFunctionDef? That seems basically the same idea.

Good point, I'll add that as a future improvement

> That said, what is "the first one"? In the order in which the protocols were listed in the type of the method receiver?

Yes, exactly

> I +1 a feature to allow specifying which protocol explicitly, but that can be a followup/separate PR.

Absolutely!

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -382,6 +389,63 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
                     node.func = instantiate_poly(node.func, protocol.sig, inst)
                     return with_loc(node, LocalCall(func=node.func, args=args)), subst
 
+        if isinstance(node.func, UnresolvedProtocolMethod):
```
mark-koch: The same logic also needs to go into `ExprSynthesizer.visit_Call`

guppylang-internals/src/guppylang_internals/tys/parsing.py
croyzor: No idea, reverted

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
croyzor: Yes! I was quite worried about this, so after talking with Mark I've decided to simplify things drastically by just calling `RequiresMonomorphizationError` instead of doing the actual type checking. Waiting until after monomorphisation disambiguates things nicely!

The downside to this is that we don't get the nice error message showing the signatures from each protocol anymore :-(

PR #1946, opened 2026-06-29
fix: allow 'frozen' kwarg in _DummyGuppy mock class

CalMacCQ:
This is a fix to the `_DummyGuppy` class which mocks the guppy decorator to be more compatible with Sphinx. ~Still need to test if this works by pinning to a commit (hence draft)~.

Now tested with [4561d49](https://github.com/Quantinuum/guppylang/pull/1946/commits/4561d49698309ed044a9ffd830c2a24b97401910) and it has resolved all of the docs issues! I think I just had to bring the signature/return type of the mock method into alignment with `@guppy.struct`.


tatiana-s (approved):
nice

PR #1949, opened 2026-06-29
feat: Add in-place array reverse

GBoydQ:
Small addition of in-place reverse method for arrays.

### DIFF
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -260,13 +260,13 @@ class array(builtins.list[_T], Generic[_T, _n]):
 
     @guppy
     @no_type_check
-    def reverse(self: array[L, n]) -> None:
+    def reverse_in_place(self: array[L, n]) -> None:
         """Reverse array elements in-place.
 
         .. code-block:: python
 
             qs = array(qubit() for _ in range(2))
-            qs.reverse()
+            qs.reverse_in_place()
         """
         for i in range(n // 2):
             mem_swap(self[i], self[n - i - 1])
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -261,6 +261,13 @@ class array(builtins.list[_T], Generic[_T, _n]):
     @guppy
     @no_type_check
     def reverse(self: array[L, n]) -> None:
+        """Reverse array elements in-place.
+
+        .. code-block:: python
+
+            qs = array(qubit() for _ in range(2))
+            qs.reverse()
+        """
         for i in range(n // 2):
             mem_swap(self[i], self[n - i - 1])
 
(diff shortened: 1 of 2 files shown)


guppylang/src/guppylang/std/array.py
```diff
@@ -257,6 +258,19 @@ def __new__(cls, *args: _T) -> builtins.list[_T]:  # type: ignore[no-redef]
             return list(args[0])
         return [*args]
 
+    @guppy
+    @no_type_check
+    def reverse(self: array[L, n]) -> None:
```
maximilianruesch: Not a full review, but this should be called `reverse_in_place` to make the semantics clear, because usual array handling treats them as immutable (except for take and put I guess).

tests/integration/test_array.py
```diff
@@ -861,3 +861,21 @@ def main() -> int:
...
     run_int_fn(main, expected=1)
+
+
+def test_array_reverse(validate, run_int_fn)->None:
+    @guppy
+    def main() -> int:
+        a = array(1,2,3)
+        a.reverse()
+        return a[0]
+    run_int_fn(main, expected=3)
```
nicolaassolini-qntm, line 883:
```suggestion
    validate(main.compile())
    run_int_fn(main, expected=3)
```

guppylang/src/guppylang/std/array.py
nicolaassolini-qntm:
```suggestion
    def reverse_in_place(self: array[L, n]) -> None:
```

nicolaassolini-qntm (dismissed):
I agree with the renaming proposed by @maximilianruesch, I also left another small comment.
Moreover, the CI test is failing due to a ruff error. You should be able to fix it by running `just check`.

tests/integration/test_array.py
```diff
@@ -861,3 +861,21 @@ def main() -> int:
         return int(bs[0])
 
     run_int_fn(main, expected=1)
+
+
+def test_array_reverse(validate, run_int_fn)->None:
```
maximilianruesch: Also add a test that reverses an array with an even number of members

tests/integration/test_array.py
```diff
@@ -861,3 +861,21 @@ def main() -> int:
...
+    @guppy
+    def main() -> int:
+        a = array(1,2,3)
+        a.reverse()
+        return a[0]
+    run_int_fn(main, expected=3)
+
+
+def test_array_reverse_linear(validate)->None:
```
maximilianruesch: This test would also pass if the reverse method did nothing for linear arrays. Please add some mutation to one of the qubits (e.g. a bit flip) and measure that qubit once the array is reversed to ensure it works here as well.

maximilianruesch (changes requested):

maximilianruesch (approved):

maximilianruesch:
I was a bit confused about the `x -> el` change, removed it, realised it is because it shadows `x` from the outside which ruff does not like (fair enough), force pushed to clean the history.

PR #1950, opened 2026-06-29
feat!: Add function item type

### LINKED ISSUE #1934: [Feature]: Add type for function items
### Problem Statement

Besides the `Function` type denoting opaque function values, we also need a type for function items that uniquely identify a single definitions.

This is required to to force the modifiable function protocols to only be implemented for statically known functions.

Equivalent concept in Rust: https://doc.rust-lang.org/reference/types/function-item.html

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

mark-koch:
Closes #1934

BREAKING CHANGE: Guppy functions now have a function item type by default. It implicitly coerces into the opaque `Function` type

### DIFF
--- a/tests/integration/test_higher_order.py
+++ b/tests/integration/test_higher_order.py
@@ -57,9 +57,9 @@ def test_conditional(validate):
     @guppy
     def main(b: bool) -> int:
         if b:
-            baz = foo
+            baz: Function[[], int] = foo
         else:
-            baz = bar
+            baz: Function[[], int] = bar
         return baz()
 
     validate(main.compile_function())
--- a/tests/integration/test_side_effect_ordering.py
+++ b/tests/integration/test_side_effect_ordering.py
@@ -137,10 +137,7 @@ def test_call(validate):
     # Check that we have the expected order edges between the allocations and calls
     hugr = compiled.modules[0]
     [a1, a2] = find_ext_nodes(hugr, QUANTUM_EXTENSION.get_op("QAlloc").qualified_name())
-    [d1] = [node for node, data in hugr.nodes() if isinstance(data.op, ops.Call)]
-    [d2] = [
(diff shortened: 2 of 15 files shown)


tests/integration/test_higher_order.py
```diff
@@ -57,9 +57,9 @@ def bar() -> int: ...
     @guppy
     def main(b: bool) -> int:
         if b:
-            baz = foo
+            baz: Function[[], int] = foo
         else:
-            baz = bar
+            baz: Function[[], int] = bar
```
mark-koch, line 62: We need an explicit `Function` annotation here to coerce both branches into an opaque function value. Otherwise we get the error from the test above

tests/integration/test_side_effect_ordering.py
```diff
@@ -137,10 +137,7 @@ def test() -> None:
     # Check that we have the expected order edges between the allocations and calls
     hugr = compiled.modules[0]
     [a1, a2] = find_ext_nodes(hugr, QUANTUM_EXTENSION.get_op("QAlloc").qualified_name())
-    [d1] = [node for node, data in hugr.nodes() if isinstance(data.op, ops.Call)]
-    [d2] = [
-        node for node, data in hugr.nodes() if isinstance(data.op, ops.CallIndirect)
-    ]
+    [d1, d2] = [node for node, data in hugr.nodes() if isinstance(data.op, ops.Call)]
```
mark-koch, line 140: All calls are direct now 🎉

aborgna-q (approved):
The "function item" name doesn't seem too intuitive to me ("concrete function" maybe?), but in general LGTM.

mark-koch:
> The "function item" name doesn't seem too intuitive to me

Hmm yeah fair. I just copied the terminology from Rust: https://doc.rust-lang.org/reference/types/function-item.html

Maybe `FunctionDefType`?

PR #1951, opened 2026-06-29
feat!: Turn modifiers into protocols

### LINKED ISSUE #1917: [Feature]: Turn `Unitary`, `Controllable` and `Daggerable` into a protocol
### Problem Statement

While doing this, we should also think about:

```
@guppy
def foo(f: Unitary[[qubit], None], q: qubit) -> None:
    with dagger:
        f(q)


@guppy
def main(b: bool) -> None:
    q = qubit()
    foo(x if b else h, q)
    discard(q)
```
and possibly discarding runtime selection such as:
```
foo(x if b else h, q)
```

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

_No response_

### Additional Context

_No response_

mark-koch:
Closes #1917

`Unitary`, `Controllable` and `Daggerable` are now protocols and no longer experimental

BREAKING CHANGE: `guppylang_internals.experimental.check_unitary_callable_enabled` removed
BREAKING CHANGE: `guppylang_internals.tys.builtin.FunctionTypeDef` no longer stores `flags`
BREAKING CHANGE: `guppylang_internals.tys.builtin.{unitary_type_def, daggerable_type_def, controllable_type_def}` renamed to `*_protocol_def` and turned into `ModifiableFunctionProtocolDef`s


tests/error/modifier_errors/higher_order_bad_return.py
```diff
@@ -1,22 +0,0 @@
...
-from guppylang.std.quantum import h
-
-
-@guppy(daggerable=True)
-def dagger_only(q: qubit) -> None:
-    pass
-
-
-@guppy
-def apply_control(f: Controllable[[qubit], None], q: qubit) -> Unitary[[qubit], None]:
```
mark-koch, line 12: Returning protocols is not supported yet, so this test needs to be deleted

tests/integration/test_modifier.py
```diff
@@ -392,22 +395,8 @@ def apply_plain(f: Function[[qubit], None], q: qubit) -> None:
...
-
-    @guppy
-    def take_daggerable_consumer(
-        consumer: Function[[Daggerable[[qubit], None], qubit], None], q: qubit
-    ) -> None:
-        consumer(unitary_fun, q)
-
     @guppy
     def main(q: qubit) -> None:
-        take_plain_consumer(apply_dagger, q)
```
mark-koch, line 412: This line was never correct: `take_plain_consumer` expects a function that takes any other function as input, but `apply_dagger` only takes daggerable things

Not sure why this worked before, seems like a covariance-contravariance mix up

tests/integration/test_modifier.py
```diff
@@ -392,22 +395,8 @@ def apply_plain(f: Function[[qubit], None], q: qubit) -> None:
...
-        take_daggerable_consumer(apply_plain, q)
```
mark-koch, line 413: This would be ok if the `Daggerable` bound was universally quantified inside `Function[[Daggerable[[qubit], None], qubit], None]`. However, the bound is scoped to the outer `take_daggerable_consumer`, so we're actually invariant here.

Added both cases as error tests

tests/integration/test_modifier.py
```diff
@@ -337,14 +333,21 @@ def apply_unitary(f: Unitary[[qubit], None], ctrl: qubit, q: qubit) -> None:
             with control(ctrl):
                 f(q)
 
-    # Tket2 still contains some bugs with higher-order functions.
-    # Thus validating exported hugr files will fail on CI.
-    # Waiting for:
-    # - https://github.com/Quantinuum/guppylang/issues/1917 and
-    # - https://github.com/Quantinuum/tket2/issues/1710
-    validate(apply_unitary.compile_function(), export=False)
```
mark-koch, line 348: For some reason, the pass is still failing on this Hugr. Will investigate

mark-koch: Ok this is still broke due to https://github.com/Quantinuum/tket2/issues/1761

Will disable the export again

tests/error/type_errors/fun_item_ty_mismatch.err
croyzor, line 6: This seems like it shouldn't be an error. I understand it's not if we annotate `baz: Function[[], int]`, but that coercion should probably happen automatically?

At least, suggest that change in the error message

mark-koch: Very odd: this was changed in #1950, not sure why it shows up in this diff??

Doing the coercion automatically would be nice but not trivial since I think you'd need to re-check basic blocks until you hit a fixed point? Agree that a hint would be nice though

mark-koch: I'll push a merge, then the diff hopefuly gets simpler?

guppylang-internals/src/guppylang_internals/tys/parsing.py
```diff
@@ -223,12 +224,8 @@ def _arg_from_instantiated_defn(
...
     match defn:
         # Special cases for the `Function` type
-        case FunctionTypeDef(flags=flags, name=name):
-            if flags != UnitaryFlags.NoFlags:
-                check_unitary_callable_enabled(flags.callable_name(), node)
-            return TypeArg(
-                _parse_function_type(arg_nodes, node, ctx, name, flags=flags)
-            )
+        case FunctionTypeDef(name=name):
+            return TypeArg(_parse_function_type(arg_nodes, node, ctx, name))
```
croyzor, line 228: why are the flags dropped?

mark-koch: Because `Function[[*args], ret]` denotes a function type without any flags. Before `Unitary` etc were special cases of `FunctionTypeDef`, but now they're protocols so they need their own kind of definition

croyzor: Make perfect sense, thanks :+1:

guppylang-internals/src/guppylang_internals/tys/builtin.py
```diff
@@ -261,6 +251,95 @@ def __str__(self) -> str:
...
+class ModifiableFunctionProtocolInst(ProtocolInst):
+    """Protocol instance associated with the builtin `Unitary`, `Controllable`, and
+    `Daggaerable` protocols.
+    """
+
+    sig: FunctionType = field(hash=False)
+
+    def __init__(self, sig: FunctionType):
+        assert not sig.parametrized
+        type_args = (*(TypeArg(inp.ty) for inp in sig.inputs), TypeArg(sig.output))
```
croyzor:
```suggestion
        type_args = (*(inp.ty.to_arg() for inp in sig.inputs), sig.output.to_arg())
```
?

guppylang-internals/src/guppylang_internals/tys/builtin.py
```diff
@@ -261,6 +251,95 @@ def __str__(self) -> str:
...
+        super().__init__(type_args, callable_protocol_def.id)
+        object.__setattr__(self, "sig", sig)
```
croyzor, line 296: why not `self.sig = sig`?

croyzor (approved):
LGTM!

guppylang-internals/src/guppylang_internals/tys/builtin.py
mark-koch: Because it's a frozen dataclass, so you can't assign in the constructor :/

PR #1952, opened 2026-06-29
feat!: Better error for Protocols as types in struct fields

### LINKED ISSUE #1915: [Bug]: Implicit protocol bounds in structs don't work
### Description

```python
@guppy.protocol
class Proto:
    @guppy.require
    def foo(self: "Proto") -> None:
        ...


@guppy.struct
class Ty:
    @guppy
    def foo(self) -> None:
        pass

@guppy.struct(frozen=True)
class S:
    p: Proto

@guppy
def foo(s: S) -> None:
    pass

@guppy
def main() -> None:
    foo(S(Ty()))

main.compile()
```

Fails with

```
Error: Type mismatch (at /file.py:120:10)
    | 
118 | @guppy
119 | def main() -> None:
120 |     foo(S(Ty()))
    |           ^^^^ Expected expression of type `Proto`, got `Ty`
```

Also, worryingly

```python
foo.compile_function()
```

crashes the compiler with

```
guppylang_internals.error.InternalGuppyError: Tried to convert generic variable to Hugr. This should have been monomorphized away.
```

Maybe we should disallow implicit protocol bounds in structs (and enums) for now?

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

croyzor:
Closes #1915.
As discussed this morning, we don't want to fix this bug by allowing the syntax, because it's much clearer to write out such cases with an explicit struct type argument that takes a protocol bound

---
BREAKING CHANGE: (guppy-internals) modify `try_parse_defn` to take `TypeParsingContext` instead of `globals`


tests/error/protocol_errors/struct_field_sugar.py
```diff
@@ -0,0 +1,29 @@
...
+
+@guppy.struct
+class Ty:
+    @guppy
+    def foo(self) -> None:
+        pass
+
+@guppy.struct(frozen=True)
+class S:
+    p: Proto # This should fail with a nice error
```
tatiana-s: Probably redundant to leave this comment in for an error test?

tatiana-s (approved):
LGTM 👍 (besides merge conflicts)

PR #1955, opened 2026-06-29
feat: Add hint to error when passing `Measurement` values to `output`

### LINKED ISSUE #1596: [Feature]: Add hint when using `result` with `Measurement`

tatiana-s:
Closes #1596


tests/error/output_errors/measurement_hint.err
```diff
@@ -0,0 +1,22 @@
...
+Note: Available overloads are:
+  def output(tag: str @comptime, value: int) -> None
+  def output(tag: str @comptime, value: nat) -> None
+  def output(tag: str @comptime, value: bool) -> None
+  def output(tag: str @comptime, value: float) -> None
+  def output(tag: str @comptime, value: array[int, n]) -> None
+  def output(tag: str @comptime, value: array[nat, n]) -> None
+  def output(tag: str @comptime, value: array[bool, n]) -> None
+  def output(tag: str @comptime, value: array[float, n]) -> None
+
```
nicolaassolini-qntm, line 21: not sure if it is still needed with the other `Note:`

in any case if the other error is renamed from `Note` to `Help` I think we can keep this, may be usefull.

guppylang-internals/src/guppylang_internals/std/_internal/compiler/platform.py
```diff
@@ -119,3 +129,62 @@ def tag_to_hugr(tag_arg: Argument, ctx: CompilerContext, loc: AstNode) -> tys.Ty
...
+
+    @dataclass(frozen=True)
+    class MeasurementOutputError(Error):
+        title: ClassVar[str] = "Unsupported output value type"
+        span_label: ClassVar[str] = (
+            "Values of type `Measurement` cannot be passed to `output` directly"
+        )
+
+        @dataclass(frozen=True)
+        class ReadHint(Note):
```
nicolaassolini-qntm: maybe?
```suggestion
        class ReadHint(Help):
```

guppylang-internals/src/guppylang_internals/std/_internal/compiler/platform.py
```diff
@@ -119,3 +129,62 @@ def tag_to_hugr(tag_arg: Argument, ctx: CompilerContext, loc: AstNode) -> tys.Ty
...
+            message: ClassVar[str] = (
+                "Use `.read()` on the value to get a bool (this will block execution "
+                "until the measurement result is available)"
+            )
+
+        @dataclass(frozen=True)
+        class ReadArrayHint(Note):
```
nicolaassolini-qntm:
```suggestion
        class ReadArrayHint(Help):
```

guppylang/src/guppylang/std/builtins.py
```diff
@@ -33,7 +33,7 @@
 )
 from guppylang.std.option import Option, nothing, some
 from guppylang.std.platform import exit, output, panic, result
-from guppylang.std.quantum import qubit
+from guppylang.std.quantum import Measurement, qubit
```
nicolaassolini-qntm, line 36: Since we are re-exporting measurement from `builtins` I think we should update the `from guppylang.std.quantum import Measurement` in the tets, to avoid keep using the old imports

guppylang-internals/src/guppylang_internals/std/_internal/compiler/platform.py
```diff
@@ -119,3 +129,62 @@ def tag_to_hugr(tag_arg: Argument, ctx: CompilerContext, loc: AstNode) -> tys.Ty
...
+        expected_ty = self.func.ty.inputs[1].ty
+        is_array = is_array_type(arg_ty)
+
+        if is_array:
+            if (not is_array_type(expected_ty)) or (
+                get_element_type(arg_ty) != get_element_type(expected_ty)
+            ):
+                raise GuppyTypeError(TypeMismatchError(self.node, expected_ty, arg_ty))
+        elif arg_ty != expected_ty:
+            raise GuppyTypeError(TypeMismatchError(self.node, expected_ty, arg_ty))
```
nicolaassolini-qntm, line 177: these checks are needed to ensure that the overloading is failing wiht the wrong function definition right?

nicolaassolini-qntm (changes requested):
just a couple of questions

tests/error/output_errors/measurement_hint.err
tatiana-s: I think it is useful to keep in order to make it more consistent with the error you get when you put some other wrong type into an overloaded function

guppylang/src/guppylang/std/builtins.py
tatiana-s: Hm qubit is usually still imported from quantum even though it is in builtins - maybe part of the bigger conversation about import conventions?

guppylang-internals/src/guppylang_internals/std/_internal/compiler/platform.py
tatiana-s: Yes exactly, with the way that overloaded function checking constructs the hint about available overloads, if you don't first type check then you would also throw this error for any other types that are not measurements

guppylang/src/guppylang/std/builtins.py
nicolaassolini-qntm: Agree, could mention this in the big import issue (I think there is an issue to summarise "the import conventions")?

nicolaassolini-qntm (approved):
all good! thanks

PR #1965, opened 2026-06-30
feat: Point users of native Python lists experimental feature to array via helptext

aidanCQ:
When I first tried using lists, it wasn't immediately clear why I couldn't or what I should use instead. This should hopefully point users in the right direction.

The text still needs work.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/experimental.py
+++ b/guppylang-internals/src/guppylang_internals/experimental.py
@@ -59,8 +59,7 @@ class ExperimentalFeatureError(Error):
 @dataclass(frozen=True)
 class NativePythonListsSuggestion(Help):
     message: ClassVar[str] = (
-        "Native Python lists are not supported here. "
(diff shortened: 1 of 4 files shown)


guppylang-internals/src/guppylang_internals/experimental.py
```diff
@@ -56,6 +56,14 @@ def __post_init__(self) -> None:
         self.add_sub_diagnostic(ExperimentalFeatureError.Suggestion(None))
 
 
+@dataclass(frozen=True)
+class NativePythonListsSuggestion(Help):
+    message: ClassVar[str] = (
+        "Native Python lists are not supported here. "
+        "Use `array` from `guppylang.std.quantum` instead."
```
maximilianruesch:
```suggestion
        "Use `array` from `guppylang.std.array` instead."
```

aborgna-q (approved):
Nice!

PR #1967, opened 2026-06-30
feat: Disallow protocols as return types

### LINKED ISSUE #1948: [Bug]: Returning protocol impls should not be allowed for now
### Description

```python
@guppy.protocol
class Proto:
    @guppy.require
    def foo(self: "Proto") -> None:
        pass

@guppy.struct
class Foo:
    @guppy
    def foo(self) -> None:
        pass

@guppy
def main() -> Proto:
    return Foo()

main.check()
```

gives the confusing error message

```
Error: Type mismatch (at file.py:134:11)
    | 
132 | @guppy
133 | def main() -> Proto:
134 |     return Foo()
    |            ^^^^^ Expected expression of type `Proto`, got `Foo`
```

This is because returning a `Proto` should be treated as an existential, but the current implementation treats it like universal quantification:

```python
def main[P: Proto]() -> P: ...
```

We should reject returning protcol impls for now until we have actusally implemented existential returns

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

croyzor:
Closes #1948

### DIFF
--- a/tests/error/protocol_errors/proto_return.err
+++ b/tests/error/protocol_errors/proto_return.err
@@ -3,6 +3,6 @@ Error: Protocols are not allowed as return types (at $FILE:15:13)
 13 | 
 14 | @guppy
 15 | def bar() -> Proto:
-   |              ^^^^^ Protocol `Proto` as return type
+   |              ^^^^^ `Proto` is a protocol, which is not allowed as a return type
 
 Guppy compilation failed due to 1 previous error
--- a/tests/error/protocol_errors/proto_return.err
+++ b/tests/error/protocol_errors/proto_return.err
@@ -1,0 +1,8 @@
+Error: Protocols are not allowed as return types (at $FILE:15:13)
+   | 
+13 | 
+14 | @guppy
+15 | def bar() -> Proto:
+   |              ^^^^^ Protocol `Proto` as return type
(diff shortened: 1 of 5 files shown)


tests/error/protocol_errors/proto_return.err
```diff
@@ -0,0 +1,8 @@
+Error: Protocols are not allowed as return types (at $FILE:15:13)
+   | 
+13 | 
+14 | @guppy
+15 | def bar() -> Proto:
+   |              ^^^^^ Protocol `Proto` as return type
```
nicolaassolini-qntm: I would rephrase this in
"The type `Proto` represents a protocol; protocols are not allowed as return types"

croyzor: I want to avoid calling it a type. Have updated to 
> `Proto` is a protocol, which is not allowed as a return type

nicolaassolini-qntm: Even better

nicolaassolini-qntm (approved):
Good for me, you just have to address Ruff’s complaints

PR #1971, opened 2026-06-30
feat!: Drop support for Python 3.10 and 3.11

### LINKED ISSUE #1970: [Feature]: Drop support for Python 3.10 and 3.11
### Problem Statement

With Python 3.10 soon to be end-of-life (October this year) and 3.11 following "soon" after (October next year), we have decided to drop support for those versions with Guppy v1 to benefit from the performance improvements and additional language features allowing us to get rid of tech debt.

Following an internal poll, there seem to be no remaining users of 3.10. Given that 3.11 has been a release that was not published in any major Linux distribution either, we feel confident to raise the minimum to 3.12.

We may revert this decision in the following weeks (the RC period) leading up to the full Guppy v1 release. Until then, we may implement it.

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

maximilianruesch:
Drops support for two old Python versions (reasoning explained in #1970). Also raises the minimum versions tested in CI (now the matrix points to 3.12, 3.13, 3.14).

Closes #1970

BREAKING CHANGE: Python 3.10 and 3.11 are no longer considered supported by this project, even though using them may still continue to work.


croyzor (approved):

.github/workflows/pull-request.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.10", "3.12"]
+        python-version: ["3.12", "3.13"]
```
ss2165: we should consider skipping 3.13 imo, save some CI hours. 3.12 was introduced because it was a frequent source of errors

guppylang/pyproject.toml
ss2165:
```suggestion
```

ss2165 (changes requested):

.github/workflows/pull-request.yaml
maximilianruesch: I left it as a matrix to facilitate future introduction of 3.15 (this october)

ss2165 (approved):

PR #1976, opened 2026-07-01
feat: add @inline decorator

acl-cqc:
Heavily based on #1942, using `InlineAnnotation` and `InlineAnnotationValue` from tket.metadata, importing carefully to pass "decoupled" tests.

Includes a test that tket optimizations respect the annotations as we expect.


guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -67,6 +69,9 @@ def set_debug_info(self, debug_info: DebugRecord) -> None:
     def set_expected_qubits(self, expected_qubits: int) -> None:
         self._node_metadata[MetadataExpectedQubitsHint] = expected_qubits
 
+    def set_inline(self, inline: Literal["best_effort", "never"]) -> None:
```
cgh-qtnm: It would be good to check the value of `inline` at runtime, to surface errors earlier for users who don't run type checkers.

guppylang-internals/src/guppylang_internals/metadata/inline.py
```diff
@@ -0,0 +1,25 @@
+from dataclasses import dataclass
+from typing import Literal
+
+from hugr.metadata import Metadata
+from hugr.utils import JsonType
+
+
+@dataclass(frozen=True)
+class MetadataInline(Metadata[Literal["best_effort", "never"]]):
```
cgh-qtnm: `Literal["best_effort", "never"]` is repeated several times, probably worth extracting it into a common def.

cgh-qtnm (approved):
Mostly looks good, a couple nits. You may also want to get a review from someone more familiar with Guppy compilation.

guppylang-internals/src/guppylang_internals/metadata/inline.py
```diff
@@ -0,0 +1,29 @@
...
+from hugr.utils import JsonType
+
+InlineOptions = Literal["best_effort", "never"]
+
+INLINE_OPTIONS: Sequence[InlineOptions] = ["best_effort", "never"]
+
+
+@dataclass(frozen=True)
+class MetadataInline(Metadata[InlineOptions]):
+    KEY = "tket.inline"
```
aborgna-q: This is already defined in `tket.metadata.InlineAnnotation`

There's also `InlineAnnotationValue` for the INLINE_OPTIONS

acl-cqc: Hah! Yes, this is getting too drawn out...need to merge soon and close the thread....

InlineAnnotationValue is like my InlineOptions. I've used `typing.get_args` to get the possible strings, but I can understand if you think that's too much of a hack. (IMHO it'll do until tket exposes something better?)

aborgna-q (approved):

PR #1977, opened 2026-07-01
refactor!: Remove `max_qubits` from `GuppyKwargs` TypedDict

nicolaassolini-qntm:
BREAKING CHANGE: Remove `max_qubits` from `GuppyKwargs`


maximilianruesch (approved):
Make sure (via grep) that there are no docs mentioning it here

PR #1981, opened 2026-07-01
fix: Fix inout updates for overloaded functions in a comptime context

### LINKED ISSUE #1920: [Bug]: Comptime inout updates don't work with overloaded functions
### Description

As `trace_call` requires `func.ty.inputs` to not be 0 in order to call `update_packed_value` but all overloaded function definitions have the signature `() -> None` at that point, comptime inout updates don't work for them.

For example:

```python
@guppy.declare
def op1(q: qubit) -> None: ...

@guppy.declare
def op2(q: qubit, x: int) -> None: ...

@guppy.overload(op1, op2)
def op() -> None: ...

@guppy.comptime
def main() -> None:
    qs = array(qubit() for _ in range(2))

    for i in range(2):
        op(qs[i])

    for q in qs:
        discard(q)
```

fails as `op` is regarded as having consumed the qubits so they can no longer be used for discard.

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

tatiana-s:
Closes https://github.com/Quantinuum/guppylang/issues/1920


croyzor (approved):
nice!

PR #1983, opened 2026-07-01
chore: Add guidance on backporting and clean up patch release guidance

maximilianruesch:


DEVELOPMENT.md
```diff
@@ -276,35 +276,22 @@ After a release is made, the main branch needs to be updated to reflect the newe
...
-   ```sh
-   git tag guppylang-v1.0.1
-   git tag guppylang-internals-v1.0.1
-   git push origin guppylang-v1.0.1 guppylang-internals-v1.0.1
-   ```
-
-   `release-publish.yml` picks up each tag, creates the draft GitHub release, and —
-   once you publish it — `python-wheels.yml` builds and uploads the wheels.
+A new commit should always land on the main branch (there may be exceptions to this if a bug can be proven to not be present on the main branch but in a released version to be patched).
+The corresponding PR may then be labelled with `B-backport-nominated`, and if accepted by a Guppy team member, it will also be labelled with `B-backport-accepted`.
```
aborgna-q: The users with permission to add the `B-backport-nominated` label are the same ones that can set `B-backport-accepted`.

Do we need some guidance on the process to go from one to the other?
We could say "the team has discussed and approved the backport internally", or even "if accepted by _another_ Guppy team member".

aborgna-q (approved):

DEVELOPMENT.md
maximilianruesch: We can add the "another", and know that in some trivial cases we can ignore that word.

PR #1985, opened 2026-07-02
feat!: Drop support for Python 3.10 and 3.11

maximilianruesch:
(cherry picked from commit ac2209cb6675de74deb357d965fc22c473a7732e)

BREAKING CHANGE: Python 3.10 and 3.11 are no longer considered supported by this project, even though using them may still continue to work.


.github/workflows/pull-request.yaml
```diff
@@ -48,7 +48,7 @@ jobs:
 
     strategy:
       matrix:
-        python-version: ["3.10", "3.12"]
+        python-version: ["3.12"]
```
aborgna-q, line 51: Shouldn't we be checking 3.13?

aborgna-q (approved):

.github/workflows/pull-request.yaml
maximilianruesch: Nope, see https://github.com/Quantinuum/guppylang/pull/1971#discussion_r3505635350

cqc-alec (approved):

PR #1990, opened 2026-07-02
test: Merge python <3.12 and python 3.12 tests

croyzor:


maximilianruesch (commented):
I have yet to look at most of the changes, but as a general thought: How will we tell apart the tests that use 3.10 syntax (which we still support) and those that use 3.12?

croyzor:
> I have yet to look at most of the changes, but as a general thought: How will we tell apart the tests that use 3.10 syntax (which we still support) and those that use 3.12?

If this is necessary then we should probably close the PR, but I'm not convince that it is

PR #1991, opened 2026-07-02
refactor: Update codebase to python 3.12

### LINKED ISSUE #2119: [Maintenance]: `ruff` still checking against python 3.10
### Description

Since we are no longer supporting python 3.10, we should probably move to python 3.12 *syntax* - see [ruff.toml](https://github.com/Quantinuum/guppylang/blob/2bca2da08625dd0b92ea293fcf058454c49bec9c/ruff.toml#L2).

At the moment this throws up about 90 lints in the codebase, many of which can't be fixed automatically by ruff.

### Motivation

python 3.12 allows significant new syntax, see e.g. #2118, and we should probably practice what we preach about minimum-supported-version

### Scope

Update ruff.toml, fix the lints; also rename the tests/**/*_py312.py files by removing the suffix and remove [that rule](https://github.com/Quantinuum/guppylang/blob/2bca2da08625dd0b92ea293fcf058454c49bec9c/ruff.toml#L87) from ruff.toml too.

### Component

_No response_

### Priority

None

### Additional Context

_No response_

croyzor:
Closes #2119

Note on breakages:
1. I do not view the removal of `typing.Generic` as a superclass to be a breaking change. It just _looks_ like a breaking change. The `Generic` superclass is just a hack to achieve what we can now do with type arg syntax and not a real class. (Reviewers smash that :+1: button if you agree)
2. I can't delete the guppy type variables from the `std.collections` libs because that would be breaking 😔

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/ty.py
+++ b/guppylang-internals/src/guppylang_internals/tys/ty.py
@@ -3,7 +3,7 @@ from collections.abc import Mapping, Sequence
 from dataclasses import dataclass, field, replace
 from enum import Enum, Flag, auto
 from functools import cached_property, total_ordering
-from typing import TYPE_CHECKING, ClassVar, Literal, TypeAlias, assert_never, cast
+from typing import TYPE_CHECKING, ClassVar, Literal, assert_never, cast
 
 import hugr.std.float
 import hugr.std.int
@@ -948,9 +948,7 @@ class EnumType(ParametrizedTypeBase):
 
 
 #: The type of parametrized Guppy types.
-ParametrizedType: TypeAlias = (
-    FunctionType | TupleType | OpaqueType | StructType | EnumType
-)
+type ParametrizedType = FunctionType | TupleType | OpaqueType | StructType | EnumType
 
 
 #: The type of Guppy types.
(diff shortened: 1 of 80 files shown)


guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -960,8 +961,34 @@ def transform(self, transformer: Transformer) -> "Type":
     | ParametrizedType
 )
 
+
+def is_type(t: Any) -> Type | None:
+    """Typechecker nonsense"""
```
cqc-alec: Ugh.

cqc-alec (approved):

guppylang-internals/src/guppylang_internals/tys/ty.py
croyzor: Oops, I meant to take that comment out! But sadly it does seem necessary...

croyzor: I've now seen the light of making this a `TypeGuard` function, which is a bit nicer 🤷

guppylang-internals/src/guppylang_internals/tys/ty.py
```diff
@@ -962,29 +962,21 @@ def transform(self, transformer: Transformer) -> "Type":
...
-        return t
-    if isinstance(t, OpaqueType):
-        return t
-    if isinstance(t, StructType):
-        return t
-    if isinstance(t, EnumType):
-        return t
-    return None
+def is_type(t: Any) -> TypeGuard[Type]:
+    return any(
```
cqc-alec: Can also write:
```python
    return isinstance(
        t,
        (
            BoundTypeVar,
            ExistentialTypeVar,
            NumericType,
            NoneType,
            FunctionDefType,
            FunctionType,
            TupleType,
            OpaqueType,
            StructType,
            EnumType,
        ),
    )
```
Matter of taste I guess.

croyzor: I didn't know you could use `isinstance` like that! 😮

PR #1997, opened 2026-07-02
fix: Make auto type params for protocol bounds unique

### LINKED ISSUE #1989: [Bug]: Protocols in type signature are expected to be the same
### Description
Writing a function with two `Callable` arguments, causes typechecking to fail

### Steps to reproduce
Failing example:
```python
from guppylang import guppy
from guppylang.std.builtins import qubit
from typing import Callable

@guppy
def take_two(a: Callable[[qubit], None], b: Callable[[qubit], None], q: qubit) -> None:
    a(q)
    b(q)

@guppy
def fn(q: qubit) -> None:
    return

@guppy
def call_two() -> None:
    q = qubit()
    return take_two(fn, fn, q)

call_two.compile()
```

### Expected Behaviour
This program should type check

### Actual Behaviour
```
Error: Type mismatch (at /Users/craig.roy/work/guppy/callable.py:17:24)
   | 
15 | def call_two() -> None:
16 |     q = qubit()
17 |     return take_two(fn, fn, q)
   |                         ^^ Expected argument of type `Callable[[qubit], None]`, got
   |                            `def fn(q: qubit) -> None`

```

croyzor:
Closes #1989

The `param_var_mapping` in our `TypeParsingContext` is usually populated by the explicit type args to a function (or type vars inherited from another scope), i.e., in
```python
def foo[A, B: Proto, C: nat](...)
```
we would parse the `(...)` with a context containing 
```python
param_var_mapping = {'A': BoundVar(...), 'B': BoundVar(...implements=[Proto]), 'C': BoundConst(ty=nat...)`
```
which is all well and good. When we see any of `A`, `B`, `C` in the function signature, we want to substitute the var for it's value in the `param_var_mapping`.

But we also allow users to write protocols in place of types, as in
```python
def foo(b: Proto)
```
which is desugared to 
```python
def foo[B: Proto](b: B)
```
the key difference here that went unnoticed before is that if we see _another_ `Proto` type
```python
def foo(b: Proto, d: Proto)
```
we DON'T want it to desugar to
```python
def foo[B: Proto](b: B, d: B)
```
but rather
```python
def foo[B: Proto, D: Proto](b: B, d: D)
```
Unfortunately, the case we don't want is what comes from naïvely adding `{'B': BoundVar(impl=[Proto])}` to the `param_var_mapping`. We actually want the parameter to be unique to the specific type that introduced it, and not be visible to be looked up to the rest.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/parsing.py
+++ b/guppylang-internals/src/guppylang_internals/tys/parsing.py
@@ -246,7 +246,12 @@ def _arg_from_instantiated_defn(
                 must_be_droppable=True,
                 must_implement=[proto_inst],
             )
-            ctx.param_var_mapping[param.name] = param
+            # Create a fresh parameter to take this `Callable` protocol bound.
+            # If we see another callable in the signature, we *don't* want it to resolve
+            # to this one.
+            # Hence, the key here is assumed to be unique, which is assumed because we
+            # don't otherwise have numerals as param vars.
+            ctx.param_var_mapping[str(len(ctx.param_var_mapping))] = param
             return param.to_bound()
         # Special case for the `Unitary`, `Controllable`, and `Daggerable` protocols
         case ModifiableFunctionProtocolDef(flags=flags):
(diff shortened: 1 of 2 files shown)


tatiana-s (approved):
Makes sense 👍

PR #2003, opened 2026-07-06
fix!: Dagger flags cancelation inside `daggarable` functions

### LINKED ISSUE #1984: [Bug]: Dagger flags are wrongly inferred with nested functions
### Description

Even dagger modifiers nullify themselves, thus if we have:
```python
with dagger:
  with dagger:
    body
```
The metadata of the inner dagger block does not mark the function block as daggerable.

When we have:
```python
@guupy(unitary=True)
def foo():
  with dagger:
    body
```
 the metadata of the inner dagger is marked only as controllable, beacuse the dagger block is nullifing the dagger component of the unitary function.

### Steps to Reproduce

_No response_

### Expected Behaviour

In 
```python
@guupy(unitary=True)
def foo():
  with dagger:
    body

@guupy(daggerable=True)
def foo():
  with dagger:
    body

```
 Since we don't know the context where the functions may be called, the metadata of the modifier block should still be considered daggarable.

This bug also involve type checking

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

nicolaassolini-qntm:
Closes #1984.

This PR fixes an issue where `dagger` blocks incorrectly cancelled the `daggerable` property of the enclosing function:  a `daggerable` function should always enforce the `daggerable` constraint within its body. Previously, a `dagger` block inside a `daggerable` function was treated as non-`daggerable` because the block's `dagger` modifier cleared the `daggerable` flag inherited from the function decorator. However, since the context in which a function is called is not known when checking its body, the unitary constraints declared in the function signature must always be enforced.

The bug was generating a bad `cfg.unitary_flags` in the modifier blocks, affecting both the metadata and the type checking.

BREAKING CHANGE: guppylang-internals: remaned parameter `unitary_flags` `CFGBuilder.build` to `function_unitary_flags`

### DIFF
--- a/guppylang-internals/src/guppylang_internals/cfg/builder.py
+++ b/guppylang-internals/src/guppylang_internals/cfg/builder.py
@@ -78,6 +78,9 @@ class CFGBuilder(AstVisitor[BB | None]):
 
     cfg: CFG
     globals: Globals
+    # The accumulated unitary flags correspond to the flags accumulated inside a
+    # function via the modifier `block`.
+    accumulated_flags: UnitaryFlags
 
     def build(
         self,
@@ -85,6 +88,7 @@ class CFGBuilder(AstVisitor[BB | None]):
         returns_none: bool,
         globals: Globals,
         unitary_flags: UnitaryFlags = UnitaryFlags.NoFlags,
+        accumulated_flags: UnitaryFlags = UnitaryFlags.NoFlags,
     ) -> CFG:
         """Builds a CFG from a list of ast nodes.
 
@@ -95,6 +99,7 @@ class CFGBuilder(AstVisitor[BB | None]):
         self.cfg = CFG()
         self.cfg.unitary_flags = unitary_flags
         self.globals = globals
+        self.accumulated_flags = accumulated_flags
 
         final_bb = self.visit_stmts(
             nodes, self.cfg.entry_bb, Jumps(self.cfg.exit_bb, None, None)
@@ -356,7 +361,17 @@ class CFGBuilder(AstVisitor[BB | None]):
         for item in node.items:
             item.context_expr, bb = ExprBuilder.build(item.context_expr, self.cfg, bb)
             modifiers.push(self._handle_withitem(item))
-        accumulated_flags = self.cfg.unitary_flags.accumulate(modifiers.flags())
+
+        # accumulate_flags accumulates flags to handle nested dagger blocks (even dagger
+        # nullify themself).
+        accumulated_flags = self.accumulated_flags.accumulate(modifiers.flags())
+        # we overwrite the accumulate_flags with self.cfg.unitary_flags to ensure that
+        # the flags declared via the `@guppy` decorator are always preserved. This is
+        # done to ensure that a dagger block inside a @guppy(daggarable=true) function
+        # has the dagger flag (and thus the type checker can correctly handle it)
+        outer_unitary_flags = self.cfg.unitary_flags
+        accumulated_flags |= outer_unitary_flags
+
         if UnitaryFlags.Dagger in accumulated_flags:
             # `original_ast_body` is only needed for checking invalid constructs under
             # dagger, so we only need to save it if dagger is present.
@@ -364,7 +379,13 @@ class CFGBuilder(AstVisitor[BB | None]):
         else:
             original_ast_body = None
 
-        cfg = CFGBuilder().build(node.body, True, self.globals, accumulated_flags)
+        cfg = CFGBuilder().build(
+            node.body,
+            True,
+            self.globals,
+            outer_unitary_flags,
+            accumulated_flags,
+        )
         new_node = ModifiedBlock(
             cfg,
             modifiers,
--- a/guppylang-internals/src/guppylang_internals/cfg/builder.py
+++ b/guppylang-internals/src/guppylang_internals/cfg/builder.py
@@ -78,10 +78,10 @@ class CFGBuilder(AstVisitor[BB | None]):
 
     cfg: CFG
     globals: Globals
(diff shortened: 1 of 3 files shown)


tests/error/modifier_errors/decorator_dagger_is_preserved.py
mark-koch, line 1: This test has nothing to do with the changes in this PR? Don't you want to check metadata?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -356,15 +375,30 @@ def visit_With(self, node: ast.With, bb: BB, jumps: Jumps) -> BB | None:
...
+
         if UnitaryFlags.Dagger in accumulated_flags:
             # `original_ast_body` is only needed for checking invalid constructs under
             # dagger, so we only need to save it if dagger is present.
             original_ast_body = copy.deepcopy(node.body)
         else:
             original_ast_body = None
 
-        cfg = CFGBuilder().build(node.body, True, self.globals, accumulated_flags)
+        cfg = CFGBuilder().build(
```
mark-koch, line 395: Wouldn't the easier fix be to just pass `self.unitary_flags` here instead of `accumulated_flags`?

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -78,13 +78,22 @@ class CFGBuilder(AstVisitor[BB | None]):
...
     cfg: CFG
     globals: Globals
+    # The unitary flags correspond to flags declared via the `@guppy` decorator on the
+    # function. When the builder is constructing a `with` block, this field contains the
+    # flags corresponding to the outer function. This allows us to prevent accumulated
+    # flags from nullifying the outer-function flags (e.g. an even number of daggers).
+    function_unitary_flags: UnitaryFlags
+    # The accumulated unitary flags correspond to the flags accumulated inside a
+    # function via the modifier `block`.
+    accumulated_flags: UnitaryFlags
```
mark-koch, line 88: `accumulated_flags` is never used? (apart from the recursive call to build)

tests/error/modifier_errors/decorator_dagger_is_preserved.py
nicolaassolini-qntm: The bug is for both metadata and flags used during type checking. I can add a new test for metadata too (maybe update the pr depscrition?)

mark-koch: But you didn't make any changes to type checking?

nicolaassolini-qntm: The type checking uses the cfg.unitary_flags, thus, changing the cfg.unitary_flags setting affects the type checking. The added example before was not discarded

mark-koch: Oh sorry yes I see. I think the test is fine then

guppylang-internals/src/guppylang_internals/cfg/builder.py
nicolaassolini-qntm: I use it at line 381 to accumulated the block flags

guppylang-internals/src/guppylang_internals/cfg/builder.py
nicolaassolini-qntm: do you mean self.function_unitary_flags?

guppylang-internals/src/guppylang_internals/cfg/builder.py
mark-koch: Yeah, but the only thing you do with the resulting `accumulated_flags` is pass it back into `build`. It is never read during checking or metadata insertion, right?

mark-koch: Ah nevermind, you assign it to `self.cfg.unitary_flags` I see

nicolaassolini-qntm: yes, I need one variable to store the accumulation with nested modifier, and one to keep the outer function flags

guppylang-internals/src/guppylang_internals/cfg/builder.py
mark-koch: I mean keep the single `unitary_flags` field on this class as before and set `accumulated_flags = self.unitary_flags | modifiers.flags()` here?

nicolaassolini-qntm: When there are more nested modifier blocks, I need one variable to store the outer function flag and another one to store the "dynamic" flag of the modifiers.
If I have only one variable, how do I know if self.unitary_flags is related to the function (thus we need to do `self.unitary_flags | modifiers.flags()`) or to the previous modifier (we need to use the `.accumulate` method that deletes even daggers)

mark-koch: Why do we need to delete even daggers? If I add `with dagger`, then the body ought to be daggerable imo, even if it's an even dagger

nicolaassolini-qntm:
```
with dagger:
  with control:
    with dagger:
       Does the dagger constraint apply here?

mark-koch: Imo it should

mark-koch: Nevermind, I changed my mind. I think this is ok as is 👍

mark-koch (approved):

PR #2011, opened 2026-07-07
fix: correct docstring indentation in EmulatorBuilder.build

CalMacCQ:
Before this change
Screenshot 2026-07-07 at 11 37 07

Gave the corresponding Sphinx warning
```
/Users/callum.macpherson/Desktop/guppy-docs/.venv/lib/python3.14/site-packages/guppylang/emulator/builder.py:docstring of guppylang.emulator.builder.EmulatorBuilder.build:7: ERROR: Unexpected indentation. [docutils]
```

After this change (tested in local docs build with the commit [7592833](https://github.com/Quantinuum/guppylang/pull/2011/commits/7592833fea1a804cbd2dd3c4e6ed4e63770a1778))

Screenshot 2026-07-07 at 11 40 18

We run the docs build as part of the Guppylang release process now. Hopefully once the strict build (fail on warnings) is reenabled then we can catch this sort of thing prior to release.

### DIFF
--- a/guppylang/src/guppylang/defs.py
+++ b/guppylang/src/guppylang/defs.py
@@ -238,17 +238,24 @@ class GuppyFunctionDefinition(GuppyDefinition, GuppyCompilableProgram, Generic[P
         See :py:mod:`guppylang.emulator` for more details on the emulator.
 
         Args:
-            n_qubits: The number of qubits to allocate for the function. If it is not
-            provided, the function has to declare the expected number of qubits it needs
-            with the decorator `@expected_qubits`.
-            builder: An optional `EmulatorBuilder` to use for building the emulator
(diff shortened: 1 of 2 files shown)


cqc-alec (approved):
Looks like there is still a problem with lack of line breaks between argument descriptions...

CalMacCQ:
> Looks like there is still a problem with lack of line breaks between argument descriptions...

Yes good spot. managed to solve this in sphinx configuation without changes to the source files.

CalMacCQ:
@cqc-alec I managed to clean this up by using more explicit indentation and also using the `sphinx.ext.napoleon` extension to parse the google stype docstrings.

Screenshot 2026-07-07 at 12 21 11

PR #2021, opened 2026-07-07
chore: Omit internals from semver break CI

maximilianruesch:
Since we now treat every version bump on internals as a breaking change, we can simply omit internals from the semver break CI and treat a lot fewer PRs as "actually breaking the user", indicated via the `!`.


aborgna-q (approved):
Should we update DEVELOPMENT.md too?

Under "Contributing to Guppy" it says

> Where the scope is optional, and the `!` is only included if this is a semver breaking change that requires a major version bump.

We could clarify what kind of breaking change requires that.

qartik (approved):

Issue #1748, opened 2026-05-18
Build callgraph of guppy functions, use to calculate effects of unannotated funcs


acl-cqc:
This fits inbetween checking (+monomorphization) and compilation, but we'll need to modify checking (specifically `check_call` and `synthesize_call`) to build the callgraph.

The intervening analysis phase uses a graph library to identify cycles (connected components); all functions in a CC will have the same effects. Effects of unannotated functions can then be calculated by a tree traversal (union of effects in the body of a function/cycle plus those in any callee outside of the cycle).

Whether we should error on finding a call from a `max_effects=` function to one that's unannotated is debateable - we could add a `strict_effects=<bool>` flag but we can probably allow this and instead enforce that every `@guppy.declare` is explicitly annotated along with every library export (otherwise it's very easy to introduce breaking changes).

acl-cqc:
Looking at #1723, `class EffectLimitDecl` currently stores the effect of an explicit `effects=` annotation; `Context` stores an `EffectLimitDecl | None` (note in how few places Context is constructed- tracing and `check_global_func_def` - we'll also want to start making a new Context in `check_nested_func_def`.)

So, replace the EffectLimitDecl in Context with "source node in callgraph", and where `_check_effects` checks legality of calls, instead just record all the calls. (We could raise errors for calls where both caller and callee are annotated but this is probably just going to repeat code.) That is, this builds the callgraph.

Callgraph nodes can then have an associated `EffectLimitDecl` to record any explicit annotation. After the callgraph is built and then analysed (i.e. SCCs), we'll then need to check calls from annotated functions: certainly to unannotated functions, i.e. if the analysis has figured out the callee has too many effects; probably to all functions (i.e. also annotated ones), as that's more uniform than dealing with these first/earlier.

PR #2024, opened 2026-07-08
refactor: remove EXTENSION_OPS_WITH_SIDE_EFFECTS, pass effects alongside DataflowOp

### LINKED ISSUE #1746: Output Order edges for each effect

acl-cqc:
closes #1746

* Removes the EXTENSION_OPS_WITH_SIDE_EFFECTS blacklist
* Instead `add_op` takes an `OpWithEffects` i.e. a `(DataflowOp, Iterable[Effect])`. Using a tuple so that factory methods returning both are straightforward.
* Zero change to hugrs, verified just prior to daf68f609c320eb6d86d9d8c7c840f1fccff9a6f i.e. with the assert that the old blacklist was the same as the value provided with the op.
* Adds a `@abstractproperty def call_effects` to `CallableDef` allowing to override for `@hugr_op` / `@custom_function`, but this PR maintains the same effects as before including that every user-defined `@guppy` function  has all effects". (We can improve upon both later, for user-defined functions [using a callgraph](#1748)).
* Also add a couple of tests documenting current behaviour of arrays

About 40 tests require the fallback to `Effect.ANY` if we cannot resolve target of a LocalCall :-(. There are ways we can try and do better but they would add significant complexity, beyond #1748, so no plans yet.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
@@ -11,7 +11,7 @@ from hugr.build import Block, Case, Cfg, Conditional, TailLoop
 from hugr.build import function as hf
 from hugr.hugr.node_port import ToNode
 from hugr.metadata import HugrDebugInfo
-from hugr.ops import CallIndirect, DataflowOp, Output
+from hugr.ops import DataflowOp, Output
 from typing_extensions import Self, override
 
 from guppylang_internals.ast_util import AstNode
@@ -98,13 +98,7 @@ class DFBuilder(ABC, ToNode):
         """Adds an op to the dataflow graph builder. Set `set_debug_info=False` to
         avoid automatic debug information attachment.
         """
-        from guppylang_internals.compiler.core import may_have_side_effect
-
         op, effects = op
-
-        assert frozenset(effects) == frozenset(
-            [Effect.ANY] if may_have_side_effect(op) else []
-        )
         op_node = self._raw.add_op(op, *args)
         self._handle_side_effects(op_node, effects)
 
--- a/guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
@@ -21,13 +21,11 @@ from guppylang_internals.metadata.debug_info_util import (
 )
 from guppylang_internals.tys import Effect
 
-
-@dataclass(frozen=True)
-class Pure:
-    op: DataflowOp
+OpWithEffects: TypeAlias = tuple[DataflowOp, Iterable[Effect]]
 
 
-OpWithEffects: TypeAlias = Pure | tuple[DataflowOp, Iterable[Effect]]
+def pure(op: DataflowOp) -> OpWithEffects:
+    return (op, [])
 
 
 @dataclass
@@ -100,7 +98,7 @@ class DFBuilder(ABC, ToNode):
         """Adds an op to the dataflow graph builder. Set `set_debug_info=False` to
         avoid automatic debug information attachment.
         """
-        op, effects = (op.op, []) if isinstance(op, Pure) else op
+        op, effects = op
         op_node = self._raw.add_op(op, *args)
         self._handle_side_effects(op_node, effects)
 
--- a/guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
@@ -115,7 +115,11 @@ class DFBuilder(ABC, ToNode):
         node = op_node.to_node()
         to_propagate = set()  # Effects newly added to our container
 
-        def get_last_node(e: Effect) -> Node:
+        def get_prev_node(e: Effect) -> Node:
+            """Gets the previous node that had the given effect,
+            or Input (marking this container as having that effect) if none.
+            Then records the current `node` as the last node, for any later call."""
+
             last = self._last_side_effect.get(e)
             if last is None:
                 to_propagate.add(e)
@@ -125,9 +129,9 @@ class DFBuilder(ABC, ToNode):
             self._last_side_effect[e] = node
             return last
 
-        prev_nodes = {get_last_node(e) for e in effects}
-        # Avoid cycles and duplicate edges:
-        prev_nodes.discard(node)
+        prev_nodes = {get_prev_node(e) for e in effects}
+        prev_nodes.discard(node)  # Avoid cycles
+        # Avoid duplicate Order edges when two nodes share multiple effects:
         for prev in self._raw.hugr.incoming_order_links(node):
             prev_nodes.discard(prev)
 
@@ -140,7 +144,7 @@ class DFBuilder(ABC, ToNode):
     @abstractmethod
     def _propagate_side_effects(self, effects: Iterable[Effect]) -> None:
         """Subclasses must implement to mark the container node
-        as side-effecting within any parent/ancestor builder"""
+        as having the given Effects within any parent/ancestor builder."""
 
     def call(
         self,
--- a/guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
@@ -11,7 +11,7 @@ from hugr.build import Block, Case, Cfg, Conditional, TailLoop
 from hugr.build import function as hf
 from hugr.hugr.node_port import ToNode
 from hugr.metadata import HugrDebugInfo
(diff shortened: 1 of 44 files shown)


guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
```diff
@@ -1,24 +1,33 @@
...
 from guppylang_internals.ast_util import AstNode
-from guppylang_internals.compiler.core import may_have_side_effect
 from guppylang_internals.metadata.debug_info_util import (
     debug_conditions_fulfilled,
     make_location_record,
 )
+from guppylang_internals.tys import Effect
+
+
+@dataclass(frozen=True)
```
tatiana-s: Feels a bit odd to me that there is a specific dataclass for `Pure` that could just be the `DataflowOp` but not for the tuple where it might be nice to access things by name rather than index?

guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
```diff
@@ -101,26 +111,44 @@ def add_op(
...
+    def _handle_side_effects(self, op_node: ToNode, effects: Iterable[Effect]) -> None:
+        """Updates Hugr to reflect `op_node` having effects `effects`.
+        Does nothing if effects is empty (or the node already has those effects)."""
         node = op_node.to_node()
-        if self._last_side_effect != node:  # avoid self-loops when propagating
-            self._raw.add_state_order(self._last_side_effect, node)
-            self._last_side_effect = node
+        to_propagate = set()  # Effects newly added to our container
+
+        def get_last_node(e: Effect) -> Node:
```
tatiana-s: Just for clarity, is this `last` node here referring the last in the sense of hierarchy as in the outer most container? I find this method a little hard to understand so may be good to have a few more comments especially if anyone tries to understand it later without much context

guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
```diff
@@ -13,7 +13,7 @@
     Signature,
 )
 from guppylang_internals.checker.core import Place, Variable
-from guppylang_internals.compiler.builder import BlockBuilder, DFBuilder
+from guppylang_internals.compiler.builder import BlockBuilder, DFBuilder, ops
```
tatiana-s, line 16: Is the idea that `ops` only ever refers to the module in guppy as opposed to hugr now? Otherwise it may be confusing between both

guppylang-internals/src/guppylang_internals/decorator/custom.py
```diff
@@ -41,6 +42,7 @@ def custom_function(
     name: str = "",
     signature: FunctionType | None = None,
     unitary_flags: UnitaryFlags = UnitaryFlags.NoFlags,
+    effects: Iterable[Effect] = (),  # ALAN TODO remove default?
```
tatiana-s: Worth removing any specific `ALAN` todos before merging?

guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
tatiana-s: And related a lot of the diff seems to be remembering to wrap ops in `Pure`, what's the main argument for this?

tatiana-s (commented):
Nice! Generally looks good to me in terms of the approach, have left a few more specific comments and also as you mention yourself there are the failing tests. In terms of splitting PRs, I can't see the diff to see the asserts you mention, I guess the question is whether any of them could lead to user facing assertion errors that should be proper errors? If not, I don't see the harm of leaving them in and splitting into a refactor and any changes to ops, however I also think this PR is not so big that splitting is absolutely necessary so I'd also be happy to keep it all in one.

guppylang-internals/src/guppylang_internals/decorator/custom.py
acl-cqc: Definitely. I'm not 100% sure about these default-removal "TODO"s - the original take of the effects system I had them all explicit, and it did lead me to find a few that I would otherwise have missed. But let's say I've done that, I could still leave the default in.... any thoughts here?

guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
acl-cqc: Yes...so files that use both I've done `use hugr.ops as hops`, for consistency I could change to that even in files not using the new `ops`...

guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
acl-cqc: Explicitness, avoids the possibility of passing in a dataflow op without thinking about whether it's pure or not ;).

I admit I *could* remove this, and argue that I've found them all now, but I think it's good to distinguish 'coz I'm sure we'll be adding more in the future.

A slightly different way would be to avoid the union type and define `OpWithEffects = tuple[DataflowOp, Iterable[Effect]]` (or even a dataclass rather than tuple), and then have `fn pure(d: DataflowOp) -> OpWithEffects: return (d, ())`...I'm not sure why I didn't like this originally but could go that way. Any thoughts/preferences?

acl-cqc: I've gone for a simple tuple with a factory `fn pure` (as `(op, [])` isn't very explicit about what the empty list is), but avoided `@dataclass OpWithEffects` as that'd have required writing an even-noisier `OpWithEffects(op, ...)` everywhere and when you see `(op, [Effect.ANY])` it's reasonably obvious that the second is a collection of Effects...

guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
acl-cqc: I've renamed "last" to "prev" in a way that I hope is more consistent/clear, and added a comment about "get_prev_node" (perhaps too much comment?).

Everything here (except gathering `to_propagate` and then calling `_propagate_side_effects`) is at this builder object's level of the hierarchy (builders have parents/grandparents, see fields); I haven't added any more comment about *that*, if you think that'd help can you suggest something?

guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
acl-cqc: Done, tho in a bunch of cases I just imported one thing from hugr.ops, I think there is no confusion about e.g. `ExtOp` or `DataflowOp`.

guppylang-internals/src/guppylang_internals/decorator/custom.py
acl-cqc: left default here, removed ALANs tho have converted some to EFFECTS for the next PR

acl-cqc:
@tatiana-s I've pulled the other changes out into #2089 as that PR will do more yet and we can avoid changing effect semantics twice (this doesn't go the whole way to what we'll want for #1747 now). Also I think leaving this as pure refactor means, no reason not to put it in ;). Can you have another look please?

guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
tatiana-s: Comment looks good to me

guppylang-internals/src/guppylang_internals/compiler/builder/__init__.py
tatiana-s: I like this better 👍

tatiana-s (approved):
Thanks, looks good to me now and seems like a good idea to get this in as a refactor to start with!

Slight nit, for purposes of a readable change log change "rm" in the title to the full "remove"?

PR #2025, opened 2026-07-08
fix: Don't emit coercion hint for temporary variables

mark-koch:
Fixing a bug that was introduced in #2013

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
@@ -246,6 +246,12 @@ class BranchTypeError(Error):
         var: str
         ty: Type
 
+        @property
+        def rendered_span_label(self) -> str:
+            if is_tmp_var(self.var):
+                return f"Consider coercing this value to `{self.ty}`"
+            return f"Consider adding a type annotation: `{self.var}: {self.ty} = ...`"
+
     @dataclass(frozen=True)
     class CoerceBothHint(Help):
         message: ClassVar[str] = (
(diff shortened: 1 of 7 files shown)


nicolaassolini-qntm:
Do you think it is possible to add a hint saying: "Function must be statically determined"?

nicolaassolini-qntm (commented):
If we can improve the error with an hint that would be nice, otherwise is fine

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
```diff
@@ -254,6 +261,15 @@ class CoerceBothHint(Help):
         ty: str
         extra: str = ""
 
+        @property
+        def rendered_message(self) -> str:
+            if is_tmp_var(self.var):
+                return f"Consider coercing both values to `{self.ty}`"
+            return (
+                f"Consider adding type annotations{self.extra}: "
```
nicolaassolini-qntm, line 263: Is there a test for this case?

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
```diff
@@ -245,6 +246,12 @@ class CoerceOneHint(Help):
         var: str
         ty: Type
 
+        @property
```
nicolaassolini-qntm, line 246: these class has both `rendered_span_label` and `snan_label` I'm quite confused

nicolaassolini-qntm (commented):
LGTM, just some clarifications

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
mark-koch: Yes, the original test added in #2013

guppylang-internals/src/guppylang_internals/checker/cfg_checker.py
mark-koch: Oops should have removed `snan_label`

nicolaassolini-qntm (approved):
:shipit:

PR #2034, opened 2026-07-10
refactor: Remove ProtocolCall

acl-cqc:
There were a couple of cases in `expr_checker.py` that asserts demonstrated never happened - roughly equivalent to:
```
@guppy.protocol
class Proto:
    @guppy.require
    def foo() -> None: ...

Proto.foo()
```
Removing these then leaves the ProtocolCall ast node as no longer used, so remove that, too.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -122,7 +122,6 @@ from guppylang_internals.nodes import (
     MakeIter,
     PartialApply,
     PlaceNode,
-    ProtocolCall,
     SubscriptAccessAndDrop,
     TensorCall,
     TupleAccessAndDrop,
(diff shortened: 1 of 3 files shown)


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -381,24 +380,13 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
-                        proto_id=node.func.def_id,
-                        args=args,
-                        type_args=inst,
-                    ),
-                ), subst
+            # This would correspond to something like:
+            #   Callable(x,y)
+            # ...but even picking a protocol which it might make sense to call,
+            # that still doesn't.
+            assert not isinstance(defn, ParsedProtocolDef)
```
croyzor: Looking at the code, this should correspond to calling a static method from a specific protocol, like
```python
@guppy.protocol
class Proto:
    @guppy.require
    def foo() -> None: ...

Proto.foo()
```
which we can't do! So im happy to remove

croyzor: I'd rather just delete the code than add this assertion though

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1008,20 +996,11 @@ def visit_Call(self, node: ast.Call) -> tuple[ast.expr, Type]:
...
-                        proto_id=node.func.def_id,
-                        args=args,
-                        type_args=inst,
-                    ),
-                ), return_ty
+            # This would correspond to some source code like
+            #   Callable(x)
+            # but even picking a protocol which supports some notion of calling,
+            # this still makes no sense.
+            assert not isinstance(defn, ParsedProtocolDef)
```
croyzor: Same comments as above

acl-cqc:
Thanks @croyzor! I think the important thing about the asserts is that now at least two of us have seen the code passes tests with them, i.e. the removed code never executed. So yes, removed now...

croyzor (approved):

PR #2035, opened 2026-07-10
feat: Analyse call graph and use it to infer effects to reduce order edges

### LINKED ISSUE #1748: Build callgraph of guppy functions, use to calculate effects of unannotated funcs

acl-cqc:
closes #1748.

`check_call` and `synthesize_call` take an extra parameter that identifies the callee (and the caller is in the `Context`), this is then added to the call data stored inside the ENGINE. This records both edges, and effects from calls to things that are not nodes (e.g. CustomFuncDefs).

A single call to `compute_effects` inbetween checking and compilation, turns the data into a graph, analyses this using SCCs, and computes a set for each function. When GlobalCalls to functions are compiled to Hugr, this only generates order edges to the Call node if required by the effects of the called function.

There are a few other places where we have to fall back to using `[Effect.ANY]` too; there are things we can do to improve these but it's not clear how urgent these will be.

NOTE: Unlike the [last effects PR](#2024) which was pure refactor, this changes many Hugrs by removing Order edges to some function calls. However there are still more opportunities for removal (I have not, for example, removed the Effect.ANY from struct or enum constructors, left in by #2024 to preserve previous behaviour) - some of these are long-term goals, others I could do here but it seems a bit gratuitous and I think that's better left for the next PR addressing #1747.


guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
```diff
@@ -24,6 +25,16 @@ def compile_global_func_def(
     """Compiles a top-level function definition to Hugr."""
     cfg = compile_cfg(func.cfg, builder, builder.inputs(), ctx)
     builder.set_outputs(*cfg)
+    if not ctx.effects[(func.id, func.mono_args)].issuperset(
```
acl-cqc: This is a critical check of compiler correctness; if it fires it means that compilation (typically of a custom function) is generating Hugr ops with some effect that was not predeclared, typically meaning that we need to add the effect to the `@custom_function` decorator

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
```diff
@@ -24,6 +25,16 @@ def compile_global_func_def(
...
+        builder._last_side_effect.keys()
+    ):
+        raise InternalGuppyError(
+            f"Function {func.name} compiled to have side effects not expected"
+            " during checking; callgraph analysis will be incomplete."
+        )
+    # Inequality (actual effects < expected) does not lead to wrong behaviour,
```
acl-cqc, line 34: There are a good number of cases of this (where checking expects effects that compilation does not generate), some simple (e.g. struct/enum constructors) and some much more complex (accuracy/precision). Reducing these is a good goal but I'm not sure whether we can ever actually eliminate.

tests/integration/test_callgraph.py
```diff
@@ -0,0 +1,323 @@
...
+from hugr.std import PRELUDE
+
+from guppylang import guppy
+from guppylang_internals.engine import ENGINE
+from guppylang.std.builtins import owned, panic
+from guppylang.std.quantum import qubit
+from guppylang.std.quantum.functional import cx, h
+
+
+def test_simple():
```
acl-cqc: I could move some of these (that only do checking not compilation) into a separate `tests/test_callgraph.py` i.e. outside tests/integration, if that's worth doing?

guppylang/src/guppylang/std/option.py
```diff
@@ -34,7 +34,7 @@ def is_nothing(self: "Option[L]") -> bool:
     def is_some(self: "Option[L]") -> bool:
         """Returns `True` if the option is a `some` value."""
 
-    @custom_function(OptionUnwrapCompiler())
+    @custom_function(OptionUnwrapCompiler(), effects=[Effect.ANY])
```
maximilianruesch, line 37: Can we make this the default of `custom_function`? I do think ANY effect is sound, even though `[]` may be more precise in a number of cases. But imo defaults should be sound.

Then you can revert all of the little changes to `custom_function` calls, hopefully reducing noise a bit (although you might add it again in a few other places).

guppylang-internals/src/guppylang_internals/definition/value.py
```diff
@@ -35,11 +35,22 @@ def load(self, dfg: "DFContainer", ctx: "CompilerContext", node: AstNode) -> Wir
...
+        """The maximum set of effects that may occur when calling the function."""
+
+
 @dataclass(frozen=True)
 class CallableDef(ValueDef):
     """Abstract base class for definitions that represent functions and to which
     calls can be type-checked."""
 
+    # Does not inherit CallableEffects as some functions do not know
+    # their effects until after checking.
```
maximilianruesch, line 53:
```suggestion
    # Does not inherit CallableEffects as some functions do not know
    # their effects until after checking.
    
```
nit: Add a space so we can ensure future devs do not mistake it for having anything to do with `ty: FunctionType`

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -285,25 +295,26 @@ class CompiledFunctionDef(CheckedFunctionDef, CompiledCallableDef, CompiledHugrN
...
         docstring: The docstring of the function.
         link_name: The external name for this function (applied to the Hugr node, and
             other representations, regardless of whether the function is actually
             visible for linking)
+        mono_args: Partial monomorphization of the generic type parameters.
         cfg: The type- and linearity-checked CFG for the function body.
         _func_bldr: used to build the function body in `compile_inner`; clients
                    should use `hugr_node`
+        effects: effects of calling the function, computed after checking
+                but before compilation begins
```
maximilianruesch, line 378:
```suggestion
        effects: Effects of calling the function, computed between checking and compilation
```

guppylang-internals/pyproject.toml
```diff
@@ -34,6 +34,8 @@ dependencies = [
     "wasmtime>=38.0,<48.1",
     "pytket>=2.18.1",
     "tket[pytket]~=0.15.6",
+    "networkx>=3.4.2",
+    "types-networkx>=3.6.1.20260612",
```
maximilianruesch: I do not think you ever use the type hints as of now?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -370,6 +403,13 @@ def register_generic_use(self, defn: CheckableGenericDef, type_args: Inst) -> No
         if not finder.bound_vars:
             self.to_check_worklist[defn.id, type_args] = defn
 
+    def register_call_graph_node(self, mono_id: MonoDefId) -> None:
```
maximilianruesch, line 460: I do not think explicit registration is necessary, especially because it is not required for adding edges to it. I rather think `register_call` should just ensure that an appropriate data node is present in the callgraph.

guppylang/src/guppylang/std/list.py
```diff
@@ -35,10 +36,11 @@
 class list[T]:
     """Mutable sequence items with homogeneous types."""
 
-    @custom_function(ListGetitemCompiler())
+    @custom_function(ListGetitemCompiler(), effects=[Effect.ANY])
     def __getitem__(self: list[L], idx: int) -> L: ...
 
-    @custom_function(ListSetitemCompiler())
+    @custom_function(ListSetitemCompiler(), effects=[Effect.ANY])
+    # ALAN couldn't write a test that triggered this, setting not supported??
```
maximilianruesch: Can you elaborate on this? Do you mean you did not manage to trigger the `__setitem__` function? Did you encounter any special error messages?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -533,6 +573,9 @@ def _compile(
     def _compile_impl(
         self, def_ids: list[DefId]
     ) -> tuple[ModulePointer, list[CompiledDef]]:
+        # Run effects checking based on call graph analysis.
+        effects = compute_effects(self.call_graph)
```
maximilianruesch: I think no harm is done just putting this directly above the construction of the `CompilerContext`, since nothing much happens in between.

Also I think the comment does not add much, and can be removed.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -266,6 +278,27 @@ def reset(self) -> None:
         self.to_check_worklist = {}
         self.generic_to_check_worklist = {}
         self.types_to_check_worklist = {}
+        self.call_graph = {}
+
+    def register_call(
```
maximilianruesch, line 308: I am not a fan of conflating the callgraph and the effects storage. I thus propose to pull them apart like this:
- Have two separate files from the `effects_checker.py` file, one named e.g. `callgraph.py` (housing the data node definition and basic dict -> networkx translation), and one named e.g. `effects.py` (containing the computation of actual effects for each function node, bottom up)
- Split the callgraph datastructure in the engine to 1. a dict mapping functions to sets / iterables of their called functions (the callgraph), and 2. a dict mapping functions to some set of effects they have *as a minimum* (analysis can add more later based on callees)
- Split this function into `register_call` (which just takes a `Context`, `CallableDef` and `Inst`) and `register_effects` (which takes a `Context` and an `Iterable[Effect]`). I know there is a slight reduction in performance if a function tries to make its effects known regardless of its callees, BUT I do feel like the callgraph being incomplete in such cases is way more of an issue, especially after calling a function called `register_call`.
- Let `compute_effects` take the (finished / converted) callgraph and the existing effects storage that is based on calls to `register_effects`.

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -226,6 +249,12 @@ def check_nested_func_def(
         if InputFlags.Comptime not in inp.flags
     ]
     def_id = DefId.fresh()
+    mono_args: Inst = ()
+
+    # Store nested functions in the call graph under their own DefIDs,
+    # although calls to them will be via local variables until we solve
+    # https://github.com/Quantinuum/guppylang/issues/2038
```
maximilianruesch: That issue is closed, can you update this comment?

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -159,9 +163,11 @@ def check_global_func_def(
         # Comptime inputs are turned into generic args, so are not included here
         if InputFlags.Comptime not in inp.flags
     ]
-    generic_args = {
-        param.name: arg for param, arg in zip(generic_ty.params, type_args, strict=True)
-    }
+    generic_args = unique_named_generic_args(generic_ty.params, type_args)
```
maximilianruesch: Which test cases fail if you use the old implementation?

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -170,9 +176,26 @@ def check_global_func_def(
...
+    result = {}
+    for param, arg in zip(params, type_args, strict=True):
+        name = param.name
+        if name in result:
+            for i in itertools.count():
+                new_name = f"{name}_{i}"
+                if new_name not in result:
+                    name = new_name
+                    break
+        result[name] = arg
```
maximilianruesch:
```suggestion
        name = param.name
        i = 0
        while name in result:
            name = f"{name}_{i}"
            i += 1
        result[name] = arg
```
I suppose this would also work the same way and be a bit less confusing? Also, should it start at 1, so we do not use zero indexing when the base variant is already present?

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -170,9 +176,26 @@ def check_global_func_def(
         func_def.name,
         globals,
         modified_block_name_base=link_name,
+        current_caller=current_caller,
     )
 
 
+def unique_named_generic_args(
```
maximilianruesch:
```suggestion
def _unique_named_generic_args(
```
Private tool, private function? Although not really mandatory, since we are in guppylang-internals anyway, but more an encapsulation signal for future devs here.

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
```diff
@@ -18,7 +19,10 @@
 
 
 def check_modified_block(
-    modified_block: ModifiedBlock, bb: BB, ctx: Context
+    modified_block: ModifiedBlock,
+    bb: BB,
+    ctx: Context,
+    current_caller: MonoDefId,
```
maximilianruesch: There should be no need for this argument, since the value is already present in `ctx` (which is supported by the only callsite of this function).

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -417,9 +421,12 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> ast.stmt:
...
         if not self.bb:
             raise InternalGuppyError("BB required to check with block!")
-
+        if not self.ctx.current_caller:
+            raise InternalGuppyError("current_caller required to check with block!")
         # check the body of the modified block
-        checked_modified_block = check_modified_block(node, self.bb, self.ctx)
+        checked_modified_block = check_modified_block(
+            node, self.bb, self.ctx, current_caller=self.ctx.current_caller
+        )
```
maximilianruesch: You can revert this change after https://github.com/Quantinuum/guppylang/pull/2035/changes#r3880999840

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -417,9 +421,12 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> ast.stmt:
 
         if not self.bb:
             raise InternalGuppyError("BB required to check with block!")
-
+        if not self.ctx.current_caller:
+            raise InternalGuppyError("current_caller required to check with block!")
```
maximilianruesch: Imo this check should be inside the `check_modified_block`, where the value is actually required.

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
```diff
@@ -24,6 +25,16 @@ def compile_global_func_def(
     """Compiles a top-level function definition to Hugr."""
     cfg = compile_cfg(func.cfg, builder, builder.inputs(), ctx)
     builder.set_outputs(*cfg)
+    if not ctx.effects[(func.id, func.mono_args)].issuperset(
+        builder._last_side_effect.keys()
+    ):
+        raise InternalGuppyError(
+            f"Function {func.name} compiled to have side effects not expected"
+            " during checking; callgraph analysis will be incomplete."
+        )
```
maximilianruesch, line 33: Can you add the surplus effects to the error message?

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
maximilianruesch: We could perhaps add `debug` (or even `info` or `warn`) messages to the compiler once we have the framework to do it?

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
```diff
@@ -24,6 +25,16 @@ def compile_global_func_def(
     """Compiles a top-level function definition to Hugr."""
     cfg = compile_cfg(func.cfg, builder, builder.inputs(), ctx)
     builder.set_outputs(*cfg)
+    if not ctx.effects[(func.id, func.mono_args)].issuperset(
+        builder._last_side_effect.keys()
```
maximilianruesch: You should not access private properties of builders, regardless of being in a compiler-internal package.

I instead propose you add a read-only `effects` property-getter to the builder, that forwards `self._last_side_effect.keys()`. That also resolves the confusion around "what is last" here!

guppylang-internals/src/guppylang_internals/compiler/modifier_compiler.py
```diff
@@ -69,9 +68,7 @@ def compile_modified_block(
     # compile body
     cfg = compile_cfg(modified_block.cfg, func_builder, func_builder.inputs(), ctx)
     func_builder.set_outputs(*cfg)
-    # EFFECTS we should pull out the actual effects from func_builder,
-    # e.g. func_builder._last_side_effect.keys(), but preserving previous behaviour
-    effects = [Effect.ANY]
+    effects = func_builder._last_side_effect.keys()
```
maximilianruesch: See https://github.com/Quantinuum/guppylang/pull/2035/changes#r3881091759

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1534,8 +1554,20 @@ def check_comptime_arg(
...
+            return None
+
+
 def synthesize_call(
-    func_ty: FunctionType, args: list[ast.expr], node: AstNode, ctx: Context
+    func_ty: FunctionType,
+    args: list[ast.expr],
+    node: AstNode,
+    ctx: Context,
+    callee: CallableDef | Iterable[Effect],
```
maximilianruesch: I am not a fan of this potentially being a list of effects. If that ever needs to be the case, it should be done separately, via a separate `register_effects`.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1534,8 +1554,20 @@ def check_comptime_arg(
     return subst
 
 
+def _identify_callee(node: ast.expr) -> str | None:
+    match node:
+        case PlaceNode(place=Variable(name=n)):
+            return n
+        case _:
+            return None
```
maximilianruesch: Is this function used anywhere?

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -539,14 +540,20 @@ class DefaultCallChecker(CustomCallChecker):
 
     @override
     def check(self, args: list[ast.expr], ty: Type) -> tuple[ast.expr, Subst]:
-        # Use default implementation from the expression checker
-        args, subst, inst = check_call(self.func.ty, args, ty, self.node, self.ctx)
+        # Use default implementation from the expression checker,
+        # but pass known effects (not DefId - we do not wish to compute them)
+        args, subst, inst = check_call(
+            self.func.ty, args, ty, self.node, self.ctx, self.func
+        )
```
maximilianruesch: You are passing the whole function, which may be something that knows of effects. Shouldn't you pass the effects explicitly here? Should `check_call` and `synthesize_call` not take just optional effects to register, and otherwise just register the callee in the callgraph anyway?

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
maximilianruesch, line 1: In this file as well, I am wondering how you decide whether to pass a list of effects or the whole function to the `check_call` / `synthesize_call`, and whether we can narrow the argument types they consume accordingly.

tests/integration/test_callgraph.py
maximilianruesch: Those that have anything to do with effects should move to a file `test_effects_via_callgraph`. Then the remaining ones can move to a file as you suggested.

tests/integration/test_callgraph.py
```diff
@@ -0,0 +1,323 @@
...
+        def inner() -> int:
+            return 42
+
+        return inner()
+
+    outer.check()
+
+    data = ENGINE.call_graph.get((outer.id, ()))
+    assert data is not None
+    # Check the outer function call exactly one function (the nested function).
```
maximilianruesch:
```suggestion
    # Check the outer function calls exactly one function (the nested function).
```

guppylang-internals/src/guppylang_internals/checker/effects_checker.py
```diff
@@ -0,0 +1,78 @@
...
+    # First construct a networkx DiGraph based on the call graph info for analysis.
+    call_graph: nx.DiGraph[MonoDefId] = nx.DiGraph()
+    for mono_def_id in func_data:
+        finder = BoundVarFinder()
+        (_, args) = mono_def_id
+        # Include only real/concrete instantiations, the others will not be compiled.
+        for arg in args:
+            arg.visit(finder)
+        if not finder.bound_vars:
+            call_graph.add_node(mono_def_id)
```
maximilianruesch: I suggest you pull out
```python
def _is_concrete(inst: Inst) -> bool:
    finder = BoundVarFinder()
    for arg in args:
        arg.visit(finder)
    return not finder.bound_vars
```
to clarify this a bit. The comment should stay above the if statement, that can then just be
```python
# Include only real/concrete instantiations, the others will not be compiled.
if _is_concrete(args):
    call_graph.add_node(mono_def_id)
```

guppylang-internals/src/guppylang_internals/checker/effects_checker.py
```diff
@@ -0,0 +1,78 @@
...
+        if mono_def_id not in call_graph.nodes:
+            continue
+        effects = set(data.other_callee_effects)
+        for tgt in data.callee_defs:
+            assert tgt in call_graph.nodes
+            call_graph.add_edge(mono_def_id, tgt)
+        call_graph.nodes[mono_def_id]["effects"] = effects
+
+    # Then compute strongly components to find cycles in the call graph. Every node
+    # in a component must have the same effects.
```
maximilianruesch:
```suggestion
    # Then compute strongly components to find cycles in the call graph. Every node
    # in a component will receive the same effects as a least upper bound.
```

guppylang-internals/src/guppylang_internals/checker/effects_checker.py
```diff
@@ -0,0 +1,78 @@
...
+            call_graph.add_node(mono_def_id)
+
+    for mono_def_id, data in func_data.items():
+        if mono_def_id not in call_graph.nodes:
+            continue
+        effects = set(data.other_callee_effects)
+        for tgt in data.callee_defs:
+            assert tgt in call_graph.nodes
+            call_graph.add_edge(mono_def_id, tgt)
+        call_graph.nodes[mono_def_id]["effects"] = effects
```
maximilianruesch: You should be able to do this during the first for-loop as well, then you do not need the continue. The effects should leave the callgraph "compilation" entirely as proposed in https://github.com/Quantinuum/guppylang/pull/2035/changes#r3880935337

maximilianruesch (changes requested):
Big PR, a lot of noisy changes, and thus also a lot of comments. I am generally happy (it is shaping up to be a rather clean call graph construction compared to languages like Java) save for the intertwining of effects and the callgraph construction itself, where I see some potential for improvement. Otherwise a lot of smaller comments.

Hoping to get this across the line soon 🚀

guppylang-internals/src/guppylang_internals/definition/function.py
acl-cqc: Unsure what the benefit is here. I slightly wonder whether this is *part of* compilation (even if abstract from whether we are compiling to Hugr or, y'know, anything else) - really it's the start of a whole new "analysis" phase - but your suggested wording is no different. perhaps "between checking and the rest of compilation"?

Related, is `checker/effects_checker.py` in the right place? It could be `guppylang_internals/effects.py` or perhaps we should have `guppylang_internals/analysis/` containing both callgraph and effects?

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: I've clarified a bit: the idea is that we only want nodes in the callgraph that *might* have outgoing edges. (I mean, we *could* add a node for *every* call target that has a MonoDefId, i.e. also Declarations and others...at the moment this is kinda handy as its "the things for which we want the analysis to compute some effects").

Sadly it turns out that many "call" targets don't have MonoDefId as they are not really defs - they look like calls at first but turn into something different in checking (e.g. ArrayUnpack and others) - so we still have to store these effects (under the caller).

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Done!! There were a couple of issues (making sure we compute effects for user-defined functions that have no calls; not all callees having MonoDefId's) but it's mostly as you suggest, and indeed I think this is much better :-), thanks!

guppylang/src/guppylang/std/list.py
acl-cqc: https://github.com/Quantinuum/guppylang/blob/b96b18fea772671f1d89747369e0ed25e9cd2920/guppylang-internals/src/guppylang_internals/checker/stmt_checker.py#L187-L195

guppylang-internals/src/guppylang_internals/checker/func_checker.py
acl-cqc: Yeah, we don't have any testcase that's affected by this one, I did it to parallel the change in `definition/traced.py` which was necessary for `test_comptime` in `test_protocol_py312.py` (where there are two synthetic Type-arguments called Animal)...

The alternative...ok, [I have rolled back](37d68e576162e400ce772c1d04f1f291b74e7294) `unique_named_generic_args`...is just to pass `type_args: Inst` into `trace_function` *as well as `generic_args`*, I mean these two ought to be redundant but nvm....

guppylang-internals/src/guppylang_internals/checker/func_checker.py
acl-cqc: Firstly, I'm trying to avoid adding a suffix (even `_0`) if none is necessary, because this affects the error message in testcase

(Secondly, your code produces `_0_1_2_3` etc. because you are updating name each time around the loop, but never mind that)

Thirdly, `unique_named_generic_args` now removed, so nvm!

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
acl-cqc: Yeah, ok, fair. There might be a better name than `effects` but it'll do if neither of us thinks of one soonish ;)

guppylang-internals/src/guppylang_internals/checker/effects_checker.py
acl-cqc: Trying to preserve that only nodes specifically provided to `register_call_graph_node` are nodes in the graph...but this has changed a lot

guppylang-internals/src/guppylang_internals/checker/effects_checker.py
acl-cqc: I've written something a bit like that; it's not just "a least upper bound" but a very specific one

guppylang-internals/src/guppylang_internals/checker/effects_checker.py
acl-cqc: Done

guppylang-internals/src/guppylang_internals/definition/custom.py
acl-cqc: Passing the whole function seems correct: if the callee knows its effects (it's a CallableEffects), we'll register its effects under its own MonoDefId (not the caller)

guppylang/src/guppylang/std/option.py
acl-cqc: In https://github.com/Quantinuum/guppylang/pull/2269

guppylang-internals/src/guppylang_internals/definition/value.py
acl-cqc: Done

guppylang-internals/pyproject.toml
acl-cqc: Am now as they are in `class CallGraph`

acl-cqc:
Thanks @maximilianruesch, very thorough review. Think I've addressed everything, noting
* See #2269 separately
* Tests split into two files (many added to existing `test_side_effect_ordering.py`)
* Should we move callgraph.py and effects_checker.py? The latter is perhaps misnamed, too
* Haven't looked at #2038 so whether to reopen that or there's something stupid I'm doing here

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: I like `guppylang_internals/analysis`, so we can put a `callgraph.py`, an `effects.py` and any other files related to that in there.

> perhaps "between checking and the rest of compilation"?

Sounds good to me

guppylang/src/guppylang/std/list.py
maximilianruesch: I think you can remove your comment here then.

guppylang-internals/src/guppylang_internals/checker/func_checker.py
maximilianruesch: Addresing the "Firstly", my code also does not add a suffix if not neccesary. For "Secondly", you would simply adjust `i = 1` as the initialiser. But I concede to "Thirdly" :)

guppylang-internals/src/guppylang_internals/checker/stmt_checker.py
```diff
@@ -417,7 +422,6 @@ def visit_ModifiedBlock(self, node: ModifiedBlock) -> ast.stmt:
 
         if not self.bb:
             raise InternalGuppyError("BB required to check with block!")
-
```
maximilianruesch, line 420: Spurious line change 😅

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
maximilianruesch: Did you see this comment? (link for convenience: https://github.com/Quantinuum/guppylang/pull/2035#discussion_r3881056881)

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
```diff
@@ -63,7 +72,7 @@ def compile_local_func_def(
 
     # If we have captured variables and the body contains a recursive occurrence of
     # the function itself, then we provide the partially applied function as a local
-    # variable
+    # variable.
```
maximilianruesch: Now seems like a rather spurious change, and it is 30+ lines away from any other changed lines 😅

tests/test_callgraph.py
```diff
@@ -0,0 +1,81 @@
...
+    def caller2() -> int:
+        return leaf()
+
+    @guppy
+    def root() -> int:
+        return caller1() + caller2()
+
+    root.check()
+
+    # After checking we should have call graph node for root, caller1, caller2 (but not
```
maximilianruesch:
```suggestion
    # After checking we should have call graph nodes for root, caller1, caller2 (but not
```

guppylang-internals/src/guppylang_internals/analysis/callgraph.py
```diff
@@ -0,0 +1,84 @@
...
+
+        for mono_def_id, callees in calls.items():
+            if mono_def_id not in self._graph.nodes:
+                continue
+            for tgt in callees:
+                if tgt in self._graph.nodes:
+                    self._graph.add_edge(mono_def_id, tgt)
+
+
+def is_concrete(mono_def_id: MonoDefId) -> bool:
```
maximilianruesch: If you are not using this function outside of this file, I suggest you privatise it via `_`-prefix.

guppylang-internals/src/guppylang_internals/analysis/callgraph.py
```diff
@@ -0,0 +1,84 @@
...
+
+@dataclass(frozen=True)
+class CallGraphComponent:
+    """Represents a strongly connected component in the call graph."""
+
+    members: frozenset[MonoDefId]
+
+    """ Calls from functions in this component to functions outside it.
+    Each tuple is (source, target component, target), with the target component being
+    None if the target is not in the call graph."""
```
maximilianruesch:
```suggestion
    """Calls from functions in this component to functions outside it.
    Each tuple is (source, target component, target), with the target component being
    None if the target is not in the call graph."""
```
Also I wonder whether "target is not in the call graph" will ever occur, and what value the target mono def id will take. Is this the "call graph may be incomplete" case we have talked about?

guppylang-internals/src/guppylang_internals/checker/effects_checker.py
maximilianruesch: I believe the new wording of "tightest-possible least-upper-bound" to be a bit superfluous, since what is the least upper bound if not the tightest possible one? Otherwise it would not be least 😅 But I am fine with leaving as is, it is clear what is meant.

guppylang-internals/src/guppylang_internals/analysis/effects.py
```diff
@@ -0,0 +1,38 @@
...
+from guppylang_internals.checker.callgraph import CallGraph
+from guppylang_internals.tys import Effect
+
+if TYPE_CHECKING:
+    from guppylang_internals.engine import MonoDefId
+
+
+def compute_effects(
+    call_graph: CallGraph,
+    other_callee_effects: Mapping["MonoDefId", set[Effect]],
```
maximilianruesch:
```suggestion
    func_effects: Mapping["MonoDefId", set[Effect]],
```
There is no `other` <insert meme>

guppylang-internals/src/guppylang_internals/analysis/callgraph.py
```diff
@@ -0,0 +1,84 @@
...
+    @cached_property
+    def condensed(self) -> Sequence[CallGraphComponent]:
+        """Returns a list of SCCs of the call graph, in topological sort order
+        (caller before callee)."""
+        components = list(nx.strongly_connected_components(self._graph))
+        condensed = nx.condensation(self._graph, scc=components)
+
+        rev_result = []
+        comps: dict[MonoDefId, CallGraphComponent] = {}
+        # Process callees first
```
maximilianruesch:
```suggestion
        # Process callees first so we can point to them from callers
```

guppylang-internals/pyproject.toml
maximilianruesch: Still, since these type stubs are only used by type checkers, you should be able to get away with only registering this dependency in the global `pyproject.toml` as a dev-dependency.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -291,6 +302,27 @@ def reset(self) -> None:
...
+        # current_caller is not set for e.g. comptime but should be here:
+        assert ctx.current_caller is not None
+        assert ctx.current_caller in self.call_graph
+        self.call_graph[ctx.current_caller].append((callee.id, inst))
+        if isinstance(callee, CallableEffects):
+            # ALAN this'll happen a lot, we should switch to a set
+            self.register_effects((callee.id, inst), callee.call_effects)
+
+    def register_effects(self, caller: MonoDefId, effects: "Iterable[Effect]") -> None:
+        """Registers known effects for a caller."""
```
maximilianruesch: We should expand on this a bit, perhaps with
```
"""Registers additional known effects for the caller. Use this function when effects must be known without having concrete callees to name as a cause."""
```
or something similar?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -291,6 +302,27 @@ def reset(self) -> None:
...
+        callee: "CallableDef",
+        inst: Inst,
+    ) -> None:
+        """Registers a function call in the call graph."""
+        # current_caller is not set for e.g. comptime but should be here:
+        assert ctx.current_caller is not None
+        assert ctx.current_caller in self.call_graph
+        self.call_graph[ctx.current_caller].append((callee.id, inst))
+        if isinstance(callee, CallableEffects):
+            # ALAN this'll happen a lot, we should switch to a set
```
maximilianruesch: What should switch to a set?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -291,6 +302,27 @@ def reset(self) -> None:
...
+        self.call_graph = {}
+        self.other_callee_effects = {}
+
+    def register_call(
+        self,
+        ctx: "Context",
+        callee: "CallableDef",
+        inst: Inst,
+    ) -> None:
+        """Registers a function call in the call graph."""
```
maximilianruesch: This function does more than that, it also registers the effects.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -395,19 +396,29 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
         # Otherwise, it must be a function as a higher-order value - something
         # whose type is either a FunctionType, a generic parameter with a `Callable`
-        # bound, or a Tuple of FunctionTypes
+        # bound, or a Tuple of FunctionTypes. Try each in turn...
+        effects: Iterable[Effect]
         if isinstance(func_ty, FunctionType):
-            args, subst, inst = check_call(func_ty, node.args, ty, node, self.ctx)
+            effects = [Effect.ANY]  # worst-case/conservative assumption
+            args, subst, inst = check_call(func_ty, node.args, ty, node, self.ctx, None)
+            register_effects(self.ctx, effects)
```
```suggestion
            args, subst, inst = check_call(func_ty, node.args, ty, node, self.ctx, None)
            register_effects(self.ctx, [Effect.ANY]) # worst-case/conservative assumption
```
Can we also change this in the other applicable places in the file?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -395,19 +396,29 @@ def visit_Call(self, node: ast.Call, ty: Type) -> tuple[ast.expr, Subst]:
...
+                    assert self.ctx.current_caller is not None
+                    assert all(
+                        isinstance(a, TypeArg) and isinstance(a.ty, BoundTypeVar)
+                        for a in self.ctx.current_caller[1]
+                    )
                     args, subst, inst = check_call(
-                        protocol.sig, node.args, ty, node, self.ctx
+                        protocol.sig, node.args, ty, node, self.ctx, None
                     )
+                    # register_effects(self.ctx, effects=[]) # Pointless
```
```suggestion
```
If it is pointless, why not remove it?

guppylang-internals/src/guppylang_internals/analysis/callgraph.py
```diff
@@ -0,0 +1,84 @@
...
+                    for tgt in self._all_calls[src]
+                    if tgt not in members
+                ),
+            )
+            comps.update(dict.fromkeys(members, cgc))
+            rev_result.append(cgc)
+        return list(reversed(rev_result))
+
+    def __init__(self, calls: Mapping[MonoDefId, Iterable[MonoDefId]]):
+        self._all_calls = calls  # deepcopy?
```
maximilianruesch: I do not think a deepcopy is necessary, but you could add one out of extra caution.

maximilianruesch (approved):
Only some smaller changes left. Thank you for all the updates 🚀 

I still believe there should be some further separation / caution between the callgraph and effects, but we should have a look at this another time, especially also concerning a broader discussion about the precision of the callgraph. Could you open an issue about this please?

guppylang-internals/src/guppylang_internals/compiler/func_compiler.py
acl-cqc: Ooops yes I did and then forgot. Added. Also changed "incomplete" to "unsound".

guppylang-internals/src/guppylang_internals/checker/effects_checker.py
acl-cqc: Ah yes...I've seen least-upper-bound used times before where it was not *really* least, but you are quite right.

tests/test_callgraph.py
acl-cqc: Moreover, comment is just plain wrong (leaf *is* in the callgraph), so I've fixed that, and added an `@guppy.declare` alongside that *isn't* in the callgraph.

guppylang-internals/src/guppylang_internals/analysis/callgraph.py
acl-cqc: This is for targets that were never `register_call_graph_node`'d. Comment clarified.

guppylang-internals/src/guppylang_internals/analysis/effects.py
acl-cqc: Yeah, I did think maybe I should do that, thanks for the "reminder" 😉

guppylang-internals/src/guppylang_internals/analysis/callgraph.py
acl-cqc: Now `Topsort puts callers first - reverse so we compute callees ready for caller.`

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Ooops, done already, thanks

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
acl-cqc: good point that when passing this to `register_effects` it's a lot more obvious than it was when doing `register_call` ;). Only by shortening a little ("worst-case safe approximation"), but yes done.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
acl-cqc: Ok, updated other comments nearby to say why.

PR #2037, opened 2026-07-13
refactor: Remove inheritance of CompiledCallableDef from CallableDef

acl-cqc:
Tiny step towards https://github.com/Quantinuum/guppylang/discussions/1583 but may be demonstrative.

CompiledCallableDef inherits from CallableDef but not because it needs to be a subtype or because it needs `check_call` or `synthesize_call` methods (it doesn't) - only because it wants to inherit one data member (!...itself only a type refinement). That inheritance seems like a "code smell" and one data member is not much to repeat.

This also allows to drop a type:ignore that was silencing a (mildly scary) warning :)

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/value.py
+++ b/guppylang-internals/src/guppylang_internals/definition/value.py
@@ -35,7 +35,8 @@ class CompiledValueDef(ValueDef, CompiledDef):
 
 @dataclass(frozen=True)
 class CallableDef(ValueDef):
-    """Abstract base class for definitions that represent functions."""
+    """Abstract base class for definitions that represent functions and to which
+    calls can be type-checked."""
 
     ty: FunctionType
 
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/definition/value.py
```diff
@@ -55,8 +56,11 @@ def __call__(self, *args: Any, **kwargs: Any) -> Any:
         raise RuntimeError("Guppy functions can only be called in a Guppy context")
 
 
-class CompiledCallableDef(CallableDef, CompiledValueDef):  # type: ignore[misc, unused-ignore]
-    """Abstract base class a global module-level function."""
+class CompiledCallableDef(CompiledValueDef):
```
acl-cqc, line 59: The `type: ignore` was to silence a warning
```
guppylang-internals/src/guppylang_internals/definition/value.py:58: error: Definition of "__replace__" in base class "CallableDef" is incompatible with definition in base class "CompiledValueDef"  [misc]
Found 1 error in 1 file (checked 166 source files)
```
which...well, CallableDef and CompiledValueDef are siblings (both inheriting from ValueDef), I assume `__replace__` was defined by the `@dataclass`...so this is a multiple inheritance problem....but beyond that, well, I haven't really understood this yet, but hey, it's good not to have such a warning issued in the first place right ;)

maximilianruesch (approved):
We can break this easily if we need to later, so no problem with merging this imo

PR #2046, opened 2026-07-14
fix: Fix comptime inout updates for `state_output` (and other custom) functions

### LINKED ISSUE #1980: [Bug]: Comptime inout updates for custom functions?
### Description

Follow up to https://github.com/Quantinuum/guppylang/issues/1920 mostly to have an issue to flag up a potential point for issues: inout updates in comptime functions currently require an accurate function signature, custom functions don't always have one specified, and it might also include variable arguments which also complicates things. This has so far mostly been fine with custom functions like `panic`/`exit` as any linear values there are consumed anyway but need to keep in mind with any other future custom functions, maybe enforce adding signatures in some way? 

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

tatiana-s:
Also closes #1980 by adding general functionality for letting custom functions define input flags for their arguments

### DIFF
--- a/guppylang-internals/src/guppylang_internals/decorator/wasm.py
+++ b/guppylang-internals/src/guppylang_internals/decorator/wasm.py
@@ -4,6 +4,7 @@ import pathlib
 from typing import TYPE_CHECKING, ParamSpec, TypeVar, overload
 
 from guppylang_internals.definition.common import DefId
+from guppylang_internals.definition.custom import MarkAsOwnedChecker
 from guppylang_internals.definition.wasm import RawWasmFunctionDef
 from guppylang_internals.dummy_decorator import _dummy_custom_decorator, sphinx_running
 from guppylang_internals.engine import DEF_STORE
@@ -111,6 +112,7 @@ def _wasm_helper(fn_id: int | None, f: Callable[P, T]) -> GuppyFunctionDefinitio
         f,
         WasmCallChecker(),
         WasmModuleCallCompiler(f.__name__, fn_id),
+        MarkAsOwnedChecker(),
         True,
         signature=None,
         wasm_index=fn_id,
--- a/guppylang-internals/src/guppylang_internals/decorator/wasm.py
+++ b/guppylang-internals/src/guppylang_internals/decorator/wasm.py
@@ -4,7 +4,6 @@ import pathlib
 from typing import TYPE_CHECKING, ParamSpec, TypeVar, overload
 
 from guppylang_internals.definition.common import DefId
-from guppylang_internals.definition.custom import MarkAsOwnedChecker
 from guppylang_internals.definition.wasm import RawWasmFunctionDef
 from guppylang_internals.dummy_decorator import _dummy_custom_decorator, sphinx_running
 from guppylang_internals.engine import DEF_STORE
@@ -112,7 +111,6 @@ def _wasm_helper(fn_id: int | None, f: Callable[P, T]) -> GuppyFunctionDefinitio
         f,
         WasmCallChecker(),
         WasmModuleCallCompiler(f.__name__, fn_id),
-        MarkAsOwnedChecker(),
         True,
         signature=None,
         wasm_index=fn_id,
--- a/guppylang-internals/src/guppylang_internals/definition/custom.py
+++ b/guppylang-internals/src/guppylang_internals/definition/custom.py
@@ -109,6 +109,10 @@ class RawCustomFunctionDef(ParsableDef):
     call_checker: "CustomCallChecker"
     call_compiler: "CustomInoutCallCompiler"
 
+    # Provides functions to help with linearity checking if the function signature is
+    # missing.
+    linearity_checker: "CustomLinearityChecker"
+
     # Whether the function may be used as a higher-order value. This is only possible
     # if a static type for the function is provided.
     higher_order_value: bool
@@ -152,6 +156,7 @@ class RawCustomFunctionDef(ParsableDef):
             ty,
             self.call_checker,
             self.call_compiler,
+            self.linearity_checker,
             self.higher_order_value,
             GlobalConstId.fresh(self.name),
             sig is not None,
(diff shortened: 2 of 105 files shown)


guppylang-internals/src/guppylang_internals/decorator/wasm.py
```diff
@@ -111,6 +112,7 @@ def _wasm_helper(fn_id: int | None, f: Callable[P, T]) -> GuppyFunctionDefinitio
         f,
         WasmCallChecker(),
         WasmModuleCallCompiler(f.__name__, fn_id),
+        MarkAsOwnedChecker(),
```
tatiana-s: This should replicate previous behaviour but note sure if this is correct in general?

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -394,6 +401,53 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
         """
 
 
+class CustomLinearityChecker(ABC):
```
tatiana-s: Not sure if this is the best name for this as currently it is not actually checking any linearity and just defining flags for arguments so maybe there is a better name (on the other hand if we need any other custom functions related to linearity they could all go here in the future)?

guppylang-internals/src/guppylang_internals/definition/custom.py
```diff
@@ -394,6 +401,53 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
+    """Provides helper functions for linearity checking of custom functions."""
+
+    @abstractmethod
+    def compute_input_flags(self, args: list[ast.expr]) -> list[InputFlags]:
+        """Computes input flags for a custom call's arguments."""
+
+
+class DefaultLinearityChecker(CustomLinearityChecker):
```
tatiana-s: Maybe this is the wrong default is it could lead to more failing cases where actually a function would have been fine with previous behaviour and now it needs to explicitly use the `MarkAsOwnedChecker` but I think that is better than more "silent" failures

guppylang-internals/src/guppylang_internals/definition/custom.py
mark-koch: Agree that this doesn't really do linearity checking.

What about just moving this into the existing `CustomCallChecker` with a default implementation that matches `DefaultLinearityChecker`?

guppylang-internals/src/guppylang_internals/definition/custom.py
mark-koch: I think this is a good default 👍

Also this doesn't count as breaking since custom functions are internal anyways

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -220,38 +221,49 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
...
+
+    # Custom functions without a signature or incomplete signature (e.g. varargs)
+    # need to provide a linearity checker to compute the input flags.
+    elif (
+        isinstance(resolved_func, CustomFunctionDef)
+        and resolved_func.linearity_checker is not None
+    ):
+        input_flags = resolved_func.linearity_checker.compute_input_flags(arg_exprs)
+
+    if input_flags is None or len(input_flags) != len(args):
```
mark-koch: Should this just be `else`?

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -220,38 +221,49 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
...
+        resolved_func = result
+
+    input_flags: list[InputFlags] | None = None
+    if len(resolved_func.ty.inputs) == len(args):
+        input_flags = [inp.flags for inp in resolved_func.ty.inputs]
+
+    # Custom functions without a signature or incomplete signature (e.g. varargs)
+    # need to provide a linearity checker to compute the input flags.
+    elif (
+        isinstance(resolved_func, CustomFunctionDef)
```
mark-koch: Should we also check if `not resolved_func.has_signature`?

mark-koch (commented):
Nice! As you pointed out yourself, my high-level comment is that this doesn't really have much to do with custom linearity checking. In particular, would it make sense to just move this into the existing `CustomCallChecker`?

Can you also update `BBLinearityChecker.visit_GlobalCall` to respect the custom input flags?

Also, does this mean we can now delete all these ugly `AbortExpr`, `BarrierExpr`, `StateOutputExpr` nodes and turn them into custom compilers?

Finally, it is not entirely clear to me how this should interact with `CustomfunctionDef.has_signature`? Presumably, if the function specifies a signature, then we shouldn't allow a custom linearity checker?

tatiana-s:
Incorporating this into the call checker is better 👍

Regarding `has_signature`, it doesn't exclude needing `compute_input_flags` in the case of varargs, so I added a check for that. See also https://github.com/Quantinuum/guppylang/issues/2073 for thoughts on making this less reliant on just putting all the information in the decorator.

Regarding the custom nodes, happy to do a follow up for those, though is this the only reason we haven't changed them to custom compiler this? I can't quite remember

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -220,38 +221,49 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
...
-                success = update_packed_value(
-                    arg, GuppyObject(ty, inout_wire), state.dfg.builder
+        resolved_func = result
+
+    input_flags: list[InputFlags] | None = None
+    if len(resolved_func.ty.inputs) == len(args):
+        input_flags = [inp.flags for inp in resolved_func.ty.inputs]
+
+    # Custom functions without a signature or incomplete signature (e.g. varargs)
+    # need to provide a linearity checker to compute the input flags.
```
mark-koch: Can you update this comment since it's no longer a linearity checker?

mark-koch (approved):
Thanks, looks good to me!

> Regarding the custom nodes, happy to do a follow up for those, though is this the only reason we haven't changed them to custom compiler this? I can't quite remember

I think so? As far as I remember the only reason we had those custom nodes was to make linearity checking work

PR #2048, opened 2026-07-14
test: fixture for enabling experimental features selectively

### LINKED ISSUE #2028: [Feature]: distinguish tests which use experimental-features
### Problem Statement

Currently all experimental features are on for all tests. This means we cannot test (and tests cannot document) what can be achieved with experimental features turned off, which is the default!

### Proposed Solution

A pytest fixture that temporarily turns experimental-features on for any test that it is applied to

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

ferbetanzo:
Closes #2028. Includes the following changes:

- New `experimental_features_enabled` context manager for temporarily enabling experimental features. Used in some snapshot tests for enabling experimental features and making sure they are disabled even if the test fails.
    - We are using hardcoded lists to mark the ones that require experimental features. Another option would be to set a naming convention for snapshot tests that require them, but this seemed like a worst option. Happy to be convinced otherwise.
- New `use_experimental_features` fixture that uses above context manager. It is used in test cases directly or whole suites via a fixture with `autouse=True`.

### DIFF
--- a/justfile
+++ b/justfile
@@ -17,7 +17,7 @@ check:
 
 # Run the tests.
 test *PYTEST_FLAGS:
-    uv run pytest tests/ -n auto {{PYTEST_FLAGS}}
+    uv run pytest -n auto {{PYTEST_FLAGS}}
 
 # Export the integration test cases to a directory.
 export-integration-tests directory="guppy-exports":
--- a/justfile
+++ b/justfile
@@ -17,7 +17,7 @@ check:
 
 # Run the tests.
 test *PYTEST_FLAGS:
-    uv run pytest -n auto {{PYTEST_FLAGS}}
+    uv run pytest tests/ -n auto {{PYTEST_FLAGS}}
 
 # Export the integration test cases to a directory.
 export-integration-tests directory="guppy-exports":
--- a/tests/error/errors_on_usage/var_in_modifier4.py
+++ b/tests/error/errors_on_usage/var_in_modifier4.py
@@ -1,7 +1,7 @@
 import guppylang
 from guppylang import guppy
 from guppylang.std.builtins import power
-guppylang.enable_experimental_features()
+#guppylang.enable_experimental_features()
 
 @guppy
 def main() -> None:
--- a/tests/error/errors_on_usage/var_in_modifier4.py
+++ b/tests/error/errors_on_usage/var_in_modifier4.py
@@ -1,7 +1,5 @@
-import guppylang
 from guppylang import guppy
 from guppylang.std.builtins import power
-#guppylang.enable_experimental_features()
 
 @guppy
 def main() -> None:
--- a/tests/error/test_errors_on_usage.py
+++ b/tests/error/test_errors_on_usage.py
@@ -37,8 +37,4 @@ files_with_experimental_flag = [
 
 @pytest.mark.parametrize("file,needs_experimental_features", files_with_experimental_flag)
 def test_errors_on_usage(file: str, needs_experimental_features: bool, capsys, snapshot):
-    if needs_experimental_features:
-        with experimental_features_enabled():
-            run_error_test(file, capsys, snapshot)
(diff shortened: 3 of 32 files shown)


tests/error/test_errors_on_usage.py
```diff
@@ -16,7 +17,28 @@
...
-@pytest.mark.parametrize("file", files)
-def test_errors_on_usage(file, capsys, snapshot):
-    run_error_test(file, capsys, snapshot)
+@pytest.mark.parametrize("file,needs_experimental_features", files_with_experimental_flag)
+def test_errors_on_usage(file: str, needs_experimental_features: bool, capsys, snapshot):
+    if needs_experimental_features:
+        with experimental_features_enabled():
+            run_error_test(file, capsys, snapshot)
+    else:
+        run_error_test(file, capsys, snapshot)
```
hsemenenko: Could `run_error_test` be updated to take `needs_experimental_features` as input instead of duplicating the conditional logic in all of the test files?

tests/error/test_errors_on_usage.py
```diff
@@ -16,7 +17,28 @@
...
+    (file, any(case in file for case in tests_that_require_experimental_features))
+    for file in files
+]
 
-@pytest.mark.parametrize("file", files)
-def test_errors_on_usage(file, capsys, snapshot):
-    run_error_test(file, capsys, snapshot)
+@pytest.mark.parametrize("file,needs_experimental_features", files_with_experimental_flag)
+def test_errors_on_usage(file: str, needs_experimental_features: bool, capsys, snapshot):
+    if needs_experimental_features:
```
hsemenenko: Does the list of experimental files need to be a fixture?

```suggestion
    if file in files_with_experimental_flag:
```

justfile
```diff
@@ -17,7 +17,7 @@ check:
 
 # Run the tests.
 test *PYTEST_FLAGS:
-    uv run pytest -n auto {{PYTEST_FLAGS}}
+    uv run pytest tests/ -n auto {{PYTEST_FLAGS}}
```
hsemenenko: I do not like this change. It's quite natural to use `just test tests/integration` to select specific tests but this overrides the folder to always be `tests/`

tests/error/errors_on_usage/var_in_modifier4.py
hsemenenko, line 1: imo it is clearer which snapshot tests need experimental features if it is included in the script like it is here. I do not feel that strongly about it though.

hsemenenko (commented):
I have a few suggestions, but they are perhaps due to my misunderstanding of how enabling experimental features works.

I do feel strongly about the change to `just test` though and think it should be reverted 🙏 .

justfile
ferbetanzo: Yeah, I agree. I added it because I also have some `test_*.py` files laying around when I'm debugging issues, and those get picked up when doing `just test`. But that is a me problem and it does interfere with the ability to specify tests 😄. Will revert it.

tests/error/test_errors_on_usage.py
ferbetanzo: Yeah, this is a great suggestion. Implemented it in f6a370e7aad2d9323d6553417673dfc9d3a12817.

tests/error/test_errors_on_usage.py
ferbetanzo: It does not need to be a fixture, but by parametrizing it we get individual test cases for each snapshot check. Otherwise we would execute a single test for all the checks and if one fails the others would be skipped.

justfile
ferbetanzo: Done in ca6fff2b0cdc2bbdab5d143916ca40810b05f0b0

hsemenenko (approved):
Thanks for making the changes! This looks good to me.

PR #2049, opened 2026-07-14
feat: Smarter coercions in array constructor

### LINKED ISSUE #2036: [Bug]: Error compiling array of functions
### Description

The following program seems to compile on [v0.21](https://play.guppylang.org?guppy=fac8fc73-754d-56bc-dbde-52f83c5957a1&utm_source=sharing) but errors on `v1`.
```python
from guppylang import guppy
from guppylang.std.quantum import x, y, z

@guppy
def main() -> None:
    paulis = array(x,y,z)
```

```
5 | @guppy
6 | def main() -> None:
7 |     paulis = array(x,y,z)
  |                      ^ Expected expression of type `def x(q: qubit) -> None`, got
  |                        `def y(q: qubit) -> None`

```

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

mark-koch:
Fixes #2036

Before, calls to the array constructor were type-checked by inferring a type for the first element and then checking the remaining elements against that. This worked fine in the past, but now that functions need a coercion to be turned into opaque functions, this started breaking some existing code (see #2036).

This PR makes the array constructor smarter by first inferring types for as many elements as possible and checking whether they can be coerced into a common type.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/std/_internal/checker.py
+++ b/guppylang-internals/src/guppylang_internals/std/_internal/checker.py
@@ -1,4 +1,5 @@
 import ast
+from contextlib import suppress
 from dataclasses import dataclass
 from typing import ClassVar
 
@@ -17,8 +18,10 @@ from guppylang_internals.checker.expr_checker import (
     check_call,
     check_num_args,
     check_type_against,
+    coerce_to_common,
     synthesize_call,
     synthesize_comprehension,
+    try_coerce_to,
 )
 from guppylang_internals.definition.custom import (
     CustomCallChecker,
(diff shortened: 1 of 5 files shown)


guppylang-internals/src/guppylang_internals/std/_internal/checker.py
```diff
@@ -309,26 +317,51 @@ def synthesize(self, args: list[ast.expr]) -> tuple[ast.expr, Type]:
...
-                for i in range(len(rest)):
-                    rest[i], subst = checker.check(rest[i], ty)
-                    assert len(subst) == 0, "Array element type is closed"
-                result_ty = array_type(ty, len(args))
+                for i, ty in enumerate(tys):
+                    if ty is None:
+                        args[i], subst = checker.check(args[i], common_ty)
+                        assert len(subst) == 0, "common_ty is closed"
+                    elif ty != common_ty:
+                        coerced = try_coerce_to(ty, common_ty, args[i], self.ctx)
```
maximilianruesch, line 356: Is there a way we can avoid making the same coerce-check twice? For example by replacing the types with the coerced type in the loop above? Or is it not exactly the same coercion?

maximilianruesch (approved):
Happy for this to be merged regardless, it is an improvement!

guppylang-internals/src/guppylang_internals/std/_internal/checker.py
mark-koch: It might be a different coercion. For example `array(nat(42), int(42), float(42.0))` first checks if `nat` can be coerced to `int` and then if `int` can be coerced to `float`. So `common_ty` would be `float`, but the first coercion check only goes to `int`

PR #2050, opened 2026-07-14
fix: Correctly derive linearity of structs from their fields

### LINKED ISSUE #2047: [Bug]: Inconsistent behaviour with linear type parameters
### Description

Generic types given a linear type parameter are sometimes assumed to be linear, even if they have no linear attributes, but at other times work fine.

The following does not compile as `Case1[qubit]` is not droppable

```python
@guppy.struct(frozen=True)
class Case1[T]:
    @guppy
    def call(self, a: T) -> None:
        pass


@guppy
def main() -> None:
    t = Case1[qubit]()

```

However, when adding an argument, the following does compile

```python
@guppy.struct(frozen=True)
class Case2[T]:
    func_arg: Function[[T], None]

    @guppy
    def call(self, a: T) -> None:
        func_arg(a)


@guppy
def main() -> None:
    t = Case2(x)

```
Adding a nonlinear attribute to Case1 allows it to compile again.

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

mark-koch:
Fixes #2047

This bug has been in the code base for a long while: the linearity of structs was determined by the linearity of their generic parameters (logic inherited from `ParametrizedTypeBase`). But that's clearly wrong since structs can have params that are not used in fields which should not influence linearity.

Maybe we should fully get rid of this `intrinsically_copyable` and `intrinsically_droppable` business since it seems more confusing than useful...

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/ty.py
+++ b/guppylang-internals/src/guppylang_internals/tys/ty.py
@@ -842,6 +842,9 @@ class StructType(ParametrizedTypeBase):
 
     defn: "CheckedStructDef"
 
+    intrinsically_copyable: bool = field(default=True, init=False)
+    intrinsically_droppable: bool = field(default=True, init=False)
+
     @cached_property
     def fields(self) -> list["CheckedField"]:
         """The fields of this struct type."""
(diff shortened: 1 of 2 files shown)


croyzor (changes requested):
Please update docstrings for `intrinsically_{copyable,droppable}` in `ParametrizedTypeBase`

croyzor:
Do we still need `intrinsically_copyable` and `intrinsically_droppable`? I find it quite confusing

croyzor (approved):

PR #2061, opened 2026-07-16
feat(guppylang)!: Raise an error if non daggerable classical functions are used in dagger cotext

nicolaassolini-qntm:
To avoid silently introducing measurement and allocation, all the functions in a dagger context must be `daggarable=true`. This is different from the controlled context where classical functions do not have any constraints.

This extra constraint is added to avoid unexpected behaviour when the execution order is changed

BREAKING CHANGE: previous guppy program that used classical function in dagger context may be non valid anymore


guppylang-internals/src/guppylang_internals/checker/unitary_checker.py
```diff
@@ -96,30 +97,50 @@ def _check_classical_args(self, args: list[ast.expr]) -> bool:
...
-        flag_ok = self.flags in ty.unitary_flags
-        if not classic_args and not flag_ok:
+
+        # If we are under any modifier, we cannot allocate qubits
+        if contain_qubit_ty(call_ty.output) and self.flags != UnitaryFlags.NoFlags:
+            err = UnitaryCallError(node, self.flags, missing_keyword_hint=False)
+            err.add_sub_diagnostic(UnitaryCallError.QubitAllocationNote(None))
+            raise GuppyError(err)
+
+        # If the function is has quantum i/o, the flags must be compatible with the
```
cqc-alec:
```suggestion
        # If the function has quantum i/o, the flags must be compatible with the
```

guppylang-internals/src/guppylang_internals/checker/unitary_checker.py
```diff
@@ -175,8 +190,32 @@ def visit_BarrierExpr(self, node: BarrierExpr) -> None:
...
+            raise GuppyTypeError(
+                UnitaryCallError(
+                    node,
+                    self.flags,
+                    missing_keyword_hint=True,
+                )
+            )
+
+    def visit_AbortExpr(self, node: AbortExpr) -> None:
+        # panics and exits are not allowed under dagger, since the execution order
```
cqc-alec, line 205: Do we not want to allow panics, e.g. from array accesses? I think we don't care too much about the ordering of panics.

nicolaassolini-qntm: I tested and is not working (the panic disappear) either under control and dagger, I'm 90% sure that it is because order edges are removed. Since in the short term we plan to patch the order edges only under control I preferred not to allow these operations under dagger to avoid bugs.

guppylang-internals/src/guppylang_internals/checker/unitary_checker.py
```diff
@@ -96,30 +97,50 @@ def _check_classical_args(self, args: list[ast.expr]) -> bool:
...
         `func`: it's only used for a better error message when the call is a GlobalCall.
         Is None for LocalCall and TensorCall.
         """
-        classic_args = self._check_classical_args(node.args)
-        flag_ok = self.flags in ty.unitary_flags
-        if not classic_args and not flag_ok:
+
+        # If we are under any modifier, we cannot allocate qubits
+        if contain_qubit_ty(call_ty.output) and self.flags != UnitaryFlags.NoFlags:
+            err = UnitaryCallError(node, self.flags, missing_keyword_hint=False)
```
cqc-alec, line 109: Do we have a test for this?

It's not clear to me how this distinguishes QAlloc operations from other operations with a qubit output?

nicolaassolini-qntm: we are more general with this check, since a function cannot take ownership of a qubit, if the function is returning a qubit, that means that there is some qubit allocation happening in the function body

cqc-alec (approved):

guppylang-internals/src/guppylang_internals/checker/unitary_checker.py
nicolaassolini-qntm: the tests are the files in `tests/error/modifier_errors` named `flag_*somenthing*_qubit" here tests/error/modifier_errors

PR #2062, opened 2026-07-16
chore: Bump selene to 0.3.0a2

jake-arkinstall:
Selene 0.3.0a2 includes:
- Zero-length arg support for argreader (entrypoint params)
- An env reader API (if wanted) that supports different values during multiprocessing

selene-core remains at 0.3.0a1 and plugins that are built for 0.3.0 will remain compatible.

It is likely that this will change minimally for the final 0.3.0 release.


tatiana-s:
@jake-arkinstall Looks like you need to remove one of the tests now that zero-arg length support is available? `test_emulator_args.py::test_empty_array_arg`

tatiana-s (approved):

PR #2066, opened 2026-07-16
feat(guppylang)!: Raise an error if non daggerable classical functions are used in dagger cotext

aborgna-q:
Backport of #2061

BREAKING CHANGE: previous guppy program that used classical function in dagger context may be non valid anymore


nicolaassolini-qntm (approved):
:shipit:

PR #2071, opened 2026-07-17
feat: Revert "feat: Add quantum pass to default opt level (#2063)" but keep tket dependency bump

tatiana-s:
This reverts part of commit 2970cb5ce9445896c12e3cbc84d76368dd7f4d7d, but keeps the tket version change. See https://github.com/Quantinuum/tket2/issues/1856 describing the issue motivating this revert.

### DIFF
--- a/guppylang/src/guppylang/optimizer.py
+++ b/guppylang/src/guppylang/optimizer.py
@@ -56,7 +56,7 @@ class OptimizationLevel(Enum):
             case OptimizationLevel.Default | OptimizationLevel.Classical:
                 from tket import passes
 
-                return [passes.NormalizeGuppy()]
+                return [passes.Normalize()]
             case OptimizationLevel.Minimal:
                 return []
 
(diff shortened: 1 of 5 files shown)


aborgna-q (approved):

guppylang/src/guppylang/optimizer.py
```diff
@@ -53,18 +53,10 @@ class OptimizationLevel(Enum):
...
-                from pytket.passes import RemoveRedundancies
+            case OptimizationLevel.Default | OptimizationLevel.Classical:
                 from tket import passes
 
-                return [passes.Normalize(), passes.PytketHugrPass(RemoveRedundancies())]
-            case OptimizationLevel.Classical:
-                from tket import passes
-
-                return [passes.Normalize()]
+                return [passes.NormalizeGuppy()]
```
aborgna-q:
```suggestion
                return [passes.Normalize()]
```

PR #2072, opened 2026-07-17
feat: Revert "feat: Add quantum pass to default opt level" but keep tket dependency bump

tatiana-s:
(cherry-pick of efb0e83)


aborgna-q (approved):
ci error looks like github being github

PR #2081, opened 2026-07-21
fix: Unitary checker skips invalid function calls in arguments

nicolaassolini-qntm:
When a non-allowed function call is hidden as an argoument of a valid function call, the non-allowed function was not detected. 

Fixing this bug, require to update some tests (after #2061)


tatiana-s (approved):
I don't think your PR title is very accurate, as far as I understand this is not actually about nested functions but rather about making sure that all arguments to a call are checked by the unitary checker? Please adjust it it to something more descriptive (so it's then clear on the change log what was fixed) before merging but otherwise looks good to me

PR #962, opened 2025-05-15
feat: Stop emitting non-local edges

mark-koch:
Temporary hack. We should revert this once https://github.com/CQCL/hugr/issues/1234 is resolved

See #963

PR #2090, opened 2026-07-24
feat: re-enable non-local edges (revert #962)

ss2165:
Reverts the temporary #962 hack now that qis-compiler (0.4.x) lowers non-local edges again — see Quantinuum/hugr#3169.

Non-copyable places go back on non-local edges instead of being packed through the `Conditional`/`TailLoop` signatures. Partial win on its own (alloca −2.3–3.9x, opt-2 compile −1.6–2.3x on an `or`-chain); the CFG block-signature O(W²) is handled in the PR stacked on top.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
@@ -194,18 +194,24 @@ def choose_vars_for_tuple_sum(
         [[v.ty.to_hugr(dfg.ctx) for v in var_row] for var_row in output_vars]
     )
 
-    # We pass all values into the conditional instead of relying on non-local edges.
-    # This is because we can't handle them in lower parts of the stack yet :/
-    # TODO: Reinstate use of non-local edges.
-    #  See https://github.com/quantinuum/guppylang/issues/963
-    all_vars = {v.id: dfg[v] for var_row in output_vars for v in var_row}
-    all_vars_wires = list(all_vars.values())
-    all_vars_idxs = {x: i for i, x in enumerate(all_vars.keys())}
+    # Non-copyable types must be passed into the conditional since we can't use
+    # inter-graph (non-local) edges to feed them in implicitly. Copyable values are
+    # read directly from the enclosing DFG via non-local edges instead of being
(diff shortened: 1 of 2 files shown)


guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
```diff
@@ -194,18 +194,24 @@ def choose_vars_for_tuple_sum(
...
-    # TODO: Reinstate use of non-local edges.
-    #  See https://github.com/quantinuum/guppylang/issues/963
-    all_vars = {v.id: dfg[v] for var_row in output_vars for v in var_row}
-    all_vars_wires = list(all_vars.values())
-    all_vars_idxs = {x: i for i, x in enumerate(all_vars.keys())}
-
-    with dfg.builder.add_conditional(unit_sum, *all_vars_wires) as conditional:
+    # Non-copyable types must be passed into the conditional since we can't use
+    # inter-graph (non-local) edges to feed them in implicitly. Copyable values are
+    # read directly from the enclosing DFG via non-local edges instead of being
```
acl-cqc, line 199: There is an alternative here. Rather than building (inside BB0, that then branches to either BB1 or BB2) a conditional containing Case1 and Case2, we generate two new basic blocks BB1' and BB2', that unconditionally branch to BB1 and BB2. Each of these takes the same inputs, i.e. all of the outputs of BB0 (the union of inputs required by BB1 and BB2 - could still use nonlocal (Dom) edges as both new blocks are dominated by their unique predecessor BB0), and discards those not needed. BB0 then branches to either BB1' or BB2', without needing the Conditional, SumType, or big Tag. BB(1/2)' can also be skipped if BB(1/2) uses all the inputs itself, i.e. generalizing the existing check before the *call* to `choose_vars_for_unit_sum`: https://github.com/Quantinuum/guppylang/blob/6fcf84ca0dd4e149700ab33d34d2b293ca889467/guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py#L126-L130

acl-cqc: Tried this here: https://github.com/Quantinuum/guppylang/pull/2311

PR #2091, opened 2026-07-24
feat: de-thread copyable CFG values via non-local Dom edges

ss2165:
Stacked on #2090.

Copyable values live across CFG blocks but only used past a dominating block were threaded through every intervening block signature — O(W²) in the width of e.g. an `if a or b or ...:` chain, which blows up LLVM (SLP/SCCP) to ~90min compiles in the worst cases.

Emit a non-local Dom edge from the defining block instead, gated on a strict single-def-dominates-all-uses check (linear, multi-def and loop-carried values stay threaded). Block-signature width goes O(W²)→O(W); on a 16×`if` / 48-bool repro opt-2 compile drops 20s→2.2s.

`Measurement` is carved out — it's future-like and the Helios QIS lowers it to a linear type, which can't ride a non-local edge. Tracked in Quantinuum/tket2#1883.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/cfg/analysis.py
+++ b/guppylang-internals/src/guppylang_internals/cfg/analysis.py
@@ -7,6 +7,9 @@ from guppylang_internals.cfg.bb import BB, VariableStats, VId
 # Type variable for the lattice domain
 T = TypeVar("T")
 
+# Type variable for basic blocks (used by the structural analyses below)
+BBT = TypeVar("BBT", bound=BB)
+
 # Analysis result is a mapping from basic blocks to lattice values
 Result = dict[BB, T]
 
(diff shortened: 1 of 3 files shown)


acl-cqc:
Discussed this with @nicolaassolini-qntm and @aborgna-q. We love the idea of using Dom edges but think the implementation should look something like:
* An analysis pass immediately after Live Variable Analysis to compute the variables (can be just the subset of live ones) at the start of each block B for which the defining block is known (i.e., the same for all control flow paths up to B).
    * Whilst such a defining block will necessarily be a dominator of B if it exists, this is not necessarily the easiest way to compute it
    * This is easily expressed as a forwards intersection dataflow analysis (Live Variables is backwards union). In fact it pretty much exists already: guppy's `AssignmentAnalysis` computes the definite-assignment, so this just needs extending to include the defining block (as an `Option[Block]` or similar: intersection of two different defining blocks gives an unknown but the variable is still definitely assigned).
* Then, store this map (Variable -> defining Block) for each `BB` and remove from the `cfg.live_before` or at least from each BB's `inputs` and `outputs`: https://github.com/Quantinuum/guppylang/blob/dbfbced48eeb486665784b3bcfc403775083584e/guppylang-internals/src/guppylang_internals/checker/cfg_checker.py#L393-L397. (Your PR seems to compute this in `compile_cfg` so that is one approach.)
* When compiling each BB and initializing variables: https://github.com/Quantinuum/guppylang/blob/dbfbced48eeb486665784b3bcfc403775083584e/guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py#L101-L102 we need to initialize places for the dom inputs as well as the regular ones, for the former using whatever value/hugr-Node the dominating block *exits* with. (This corresponds somewhat with the changes to `compile_bb` in this PR.)

acl-cqc:
Comment moved to https://github.com/Quantinuum/guppylang/issues/2361 so closing this as superceded by that issue

Issue #2086, opened 2026-07-22
[Feature]: Range reverse


GBoydQ:
### Problem Statement

A method for reversing a range would be helpful. I would be more than happy to implement this but there are a few options

1. Add a `Range.reverse` method
2. Do the more pythonic `reversed(r: Range)`, but have it only work for range before the iter protocol is implemented
3. Wait for the proper `reversed(i: Iterable)`

Also looking at the range implementation it seems that currently iterating hangs when `step=0`

### Effort Estimate

None

mark-koch:
Python's [`reversed` function](https://docs.python.org/3/library/functions.html#reversed) relies on the `Sequence` protocol which we can actually already define, so maybe that would be a good way to go?

Also, Python defines a `__reversed__` magic method that we could define?

nicolaassolini-qntm:
First step in solving this issue is done in: https://github.com/Quantinuum/guppylang/pull/2092 (the PR adds `__reversed__` to `Range`)

PR #2092, opened 2026-07-24
feat: Add `Range.__reversed__ ` and panic on range step=0

thewill-i-am:
Add a `reverse_in_place` method to the `Range` struct that reverses the iteration direction in place by negating `_step` and recomputing `_next` and `_stop` so the reversed range yields the same elements in opposite order. Mirrors the existing `array.reverse_in_place` pattern.

Fix a hang where iterating `range(a, b, 0)` would loop forever: the `_next += 0` cursor never advanced. The public `_range3` constructor now panics with "range() arg 3 must not be zero", and `reverse_in_place` panics on a zero-step range as a defensive guard. `panic` is imported deferred from `guppylang.std.platform` to avoid a circular import since `iter` loads early via `builtins`/`array`.

Tests cover 9 reverse cases (positive/negative/uneven/empty ranges) checked against `reversed(builtins.range(...))`, HUGR validation, double-reverse idempotency, and runtime panics for both `step=0` paths.

### DIFF
--- a/guppylang/src/guppylang/std/iter.py
+++ b/guppylang/src/guppylang/std/iter.py
@@ -73,6 +73,35 @@ class Range:
         self._next += self._step
         return some((actual_next, self))
 
+    @guppy
+    @no_type_check
+    def reverse_in_place(self: Self) -> None:
+        if self._step == 0:
+            panic("range.reverse_in_place: step is zero")
+        if self._step > 0:
+            diff = self._stop - self._next
+            if diff > 0:
+                count = (diff + self._step - 1) // self._step
+                last = self._next + (count - 1) * self._step
+                self._stop = self._next - self._step
+            else:
+                self._stop = self._next
+                last = self._next
+            self._next = last
+            self._step = -self._step
+        else:
+            diff = self._next - self._stop
+            if diff > 0:
+                neg = -self._step
+                count = (diff + neg - 1) // neg
+                last = self._next + (count - 1) * self._step
+                self._stop = self._next - self._step
+            else:
+                self._stop = self._next
+                last = self._next
+            self._next = last
+            self._step = -self._step
+
 
 @guppy
 @no_type_check
@@ -89,6 +118,8 @@ def _range2(start: int, stop: int) -> Range:
 @guppy
 @no_type_check
 def _range3(start: int, stop: int, step: int) -> Range:
+    if step == 0:
+        panic("range() arg 3 must not be zero")
     return Range(start, stop, step)
 
 
@@ -111,4 +142,12 @@ def range(start: int, stop: int = 0, step: int = 1) -> Range:
     ``start`` defaults to ``0`` and ``step`` defaults to ``1``. If the provided ``stop``
     value is comptime known, then the returned iterator will have a static size
     annotation and may for example be used inside array comprehensions.
+
+    Iterating with a ``step`` of ``0`` raises a runtime panic. The
+    ``Range.reverse_in_place`` method reverses the iteration direction in place.
     """
+
+
+# Delayed import to avoid cyclic import since `iter` is loaded very early via
+# `builtins`/`array` (which import `SizedIter` from this module).
+from guppylang.std.platform import panic  # noqa: E402
--- a/guppylang/src/guppylang/std/iter.py
+++ b/guppylang/src/guppylang/std/iter.py
@@ -75,7 +75,7 @@ class Range:
 
     @guppy
     @no_type_check
-    def reverse_in_place(self: Self) -> None:
+    def __reversed__(self: Self) -> None:
         if self._step == 0:
             panic("range.reverse_in_place: step is zero")
 
--- a/guppylang/src/guppylang/std/iter.py
+++ b/guppylang/src/guppylang/std/iter.py
@@ -78,29 +78,23 @@ class Range:
     def reverse_in_place(self: Self) -> None:
         if self._step == 0:
             panic("range.reverse_in_place: step is zero")
-        if self._step > 0:
-            diff = self._stop - self._next
-            if diff > 0:
-                count = (diff + self._step - 1) // self._step
-                last = self._next + (count - 1) * self._step
-                self._stop = self._next - self._step
-            else:
-                self._stop = self._next
-                last = self._next
-            self._next = last
-            self._step = -self._step
+
+        diff = self._stop - self._next
+
+        # The range is empty when diff and step have different signs,
+        # or when the difference is zero.
+        if diff == 0 or (diff < 0) != (self._step < 0):
+            last = self._next
+            self._stop = self._next
         else:
-            diff = self._next - self._stop
-            if diff > 0:
-                neg = -self._step
-                count = (diff + neg - 1) // neg
-                last = self._next + (count - 1) * self._step
-                self._stop = self._next - self._step
-            else:
-                self._stop = self._next
-                last = self._next
-            self._next = last
-            self._step = -self._step
+            distance = diff if diff >= 0 else -diff
+            step = self._step if self._step >= 0 else -self._step
+            count = (distance + step - 1) // step
+            last = self._next + (count - 1) * self._step
+            self._stop = self._next - self._step
+
+        self._next = last
+        self._step = -self._step
 
 
 @guppy
@@ -143,8 +137,10 @@ def range(start: int, stop: int = 0, step: int = 1) -> Range:
     value is comptime known, then the returned iterator will have a static size
     annotation and may for example be used inside array comprehensions.
 
-    Iterating with a ``step`` of ``0`` raises a runtime panic. The
-    ``Range.reverse_in_place`` method reverses the iteration direction in place.
+    Iterating with a ``step`` of ``0`` raises a runtime panic.
+
+    Use the ``Range.reverse_in_place`` method to reverse the iteration direction
+    in place.
     """
 
 
--- a/guppylang/src/guppylang/std/iter.py
+++ b/guppylang/src/guppylang/std/iter.py
@@ -77,7 +77,7 @@ class Range:
     @no_type_check
     def __reversed__(self: Self) -> None:
         if self._step == 0:
-            panic("range.reverse_in_place: step is zero")
+            panic("Range.__reversed__: step is zero")
 
         diff = self._stop - self._next
 
@@ -138,9 +138,6 @@ def range(start: int, stop: int = 0, step: int = 1) -> Range:
     annotation and may for example be used inside array comprehensions.
 
     Iterating with a ``step`` of ``0`` raises a runtime panic.
-
-    Use the ``Range.reverse_in_place`` method to reverse the iteration direction
-    in place.
     """
 
 
--- a/tests/integration/test_range.py
+++ b/tests/integration/test_range.py
@@ -83,7 +83,7 @@ def test_static_generic_size(validate):
     validate(main.compile_function())
 
 
-def test_range_reverse(run_int_fn):
+def test_range_reverse(run_int_fn, validate):
     """Reversing a range iterates the same elements as `reversed(range(...))`."""
 
     @guppy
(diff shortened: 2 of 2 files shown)


tests/integration/test_range.py
```diff
@@ -78,3 +81,110 @@ def main() -> None:
...
+def test_range_reverse(run_int_fn):
+    """Reversing a range iterates the same elements as `reversed(range(...))`."""
+
+    @guppy
+    def main(start: int, stop: int, step: int) -> int:
+        total = 0
+        r = range(start, stop, step)
+        r.reverse_in_place()
+        for x in r:
+            total += x + 100  # Make the initial 0 obvious
```
maximilianruesch: (not a full review) How does this tell you in what order your elements are present? As I understand them, this test asserts that all elements are iterated over, but not what order they are in.

thewill-i-am: Hi @maximilianruesch 

I added this test because `reverse_in_place` mutates the internal `Range` state and needs
runtime coverage across edge cases like negative steps, uneven steps, empty ranges, and
off-by-one boundaries. The goal was to compare against Python's
`reversed(builtins.range(...))`.

You're right that the current sum-based check does not verify order.

thewill-i-am: @maximilianruesch 

I updated the test to use an order-sensitive accumulator instead, so sequences like
`3, 1` and `1, 3` now produce different results. 


The test is now order-sensitive because the accumulator is no longer a simple sum.

Previously, the test used:
```python
total += x + 100
That only verifies that the same values are produced, because addition is commutative:
1 + 3 == 3 + 1
So the test could pass even if reverse_in_place produced the right elements in the wrong order.
Now the test uses an order-sensitive fold:
total = total * 100 + x + 50
This makes the accumulated result depend on the exact sequence of values.
For example, for range(1, 5, 2), the original sequence is:
1, 3
The reversed sequence should be:
3, 1
With the new accumulator, the correct reversed sequence gives:
# 3, 1
total = 0
total = 0 * 100 + 3 + 50   # 53
total = 53 * 100 + 1 + 50  # 5351
But the wrong order gives a different result:
# 1, 3
total = 0
total = 0 * 100 + 1 + 50   # 51
total = 51 * 100 + 3 + 50  # 5153
So 5351 != 5153, and the test would fail if the range were not actually reversed.
The expected value uses the same encoding over Python's reference behavior:
return encode(reversed(builtins.range(a, b, s)))
```

So the test compares Guppy's reverse_in_place result against the order produced by
Python's reversed(range(...)).

guppylang/src/guppylang/std/iter.py
```diff
@@ -73,6 +73,35 @@ def __next__(self: Self @ owned) -> Option[tuple[int, Self]]:
         self._next += self._step
         return some((actual_next, self))
 
+    @guppy
+    @no_type_check
+    def reverse_in_place(self: Self) -> None:
```
CalMacCQ: In Python, `list.reverse()` also acts in place. Therefore maybe we don't need to call it `reverse_in_place`. What do you think @nicolaassolini-qntm ?

guppylang/src/guppylang/std/iter.py
```diff
@@ -111,4 +142,12 @@ def range(start: int, stop: int = 0, step: int = 1) -> Range:
     ``start`` defaults to ``0`` and ``step`` defaults to ``1``. If the provided ``stop``
     value is comptime known, then the returned iterator will have a static size
     annotation and may for example be used inside array comprehensions.
+
+    Iterating with a ``step`` of ``0`` raises a runtime panic. The
+    ``Range.reverse_in_place`` method reverses the iteration direction in place.
```
nicolaassolini-qntm:
```suggestion
    Iterating with a ``step`` of ``0`` raises a runtime panic. 
    
    Use the ``Range.reverse_in_place`` method to reverse the iteration direction in place.
```

guppylang/src/guppylang/std/iter.py
```diff
@@ -73,6 +73,35 @@ def __next__(self: Self @ owned) -> Option[tuple[int, Self]]:
...
+            if diff > 0:
+                neg = -self._step
+                count = (diff + neg - 1) // neg
+                last = self._next + (count - 1) * self._step
+                self._stop = self._next - self._step
+            else:
+                self._stop = self._next
+                last = self._next
+            self._next = last
+            self._step = -self._step
```
nicolaassolini-qntm: Just a refactor to make the function easier to follow (IMO)

I hope indentation is fine :)

```suggestion
    def reverse_in_place(self: Self) -> None:
        if self._step == 0:
            panic("range.reverse_in_place: step is zero")

        diff = self._stop - self._next

        # The range is empty when diff and step have different signs,
        # or when the difference is zero.
        if diff == 0 or (diff < 0) != (self._step < 0):
            last = self._next
            self._stop = self._next
        else:
            step = self._step
            # Change diff and step sign to ensure that count is correct
            if diff < 0:
                diff = -diff
                step = -step

            count = (diff + step - 1) // step
            last = self._next + (count - 1) * self._step
            self._stop = self._next - self._step

        self._next = last
        self._step = -self._step
 ```

File: tests/integration/test_range.py [hunk 5]
```diff
@@ -78,3 +81,122 @@ def main() -> None:
...
+        r.reverse_in_place()
+        for x in r:
+            total = total * 100 + x + 50
+        return total
+
+    def encode(xs) -> int:
+        total = 0
+        for x in xs:
+            total = total * 100 + x + 50
+        return total
```
nicolaassolini-qntm, line 103: Maybe a brief comment to explain the logic here?

tests/integration/test_range.py
```diff
@@ -78,3 +81,122 @@ def main() -> None:
...
+    @guppy
+    def main(start: int, stop: int, step: int) -> int:
+        total = 0
+        r = range(start, stop, step)
+        r.reverse_in_place()
+        for x in r:
+            total += x
+        return total
+
+    validate(main.compile_function())
```
nicolaassolini-qntm: I think this is redundant after the change I suggested above (or below chronologically on GitHub)

tests/integration/test_range.py
```diff
@@ -78,3 +81,122 @@ def main() -> None:
...
+
+    def encode(xs) -> int:
+        total = 0
+        for x in xs:
+            total = total * 100 + x + 50
+        return total
+
+    def expected(a: int, b: int, s: int) -> int:
+        return encode(reversed(builtins.range(a, b, s)))
+
```
nicolaassolini-qntm, line 107:
```suggestion

validate(main.compile_function())
```

nicolaassolini-qntm (changes requested):
Thanks!
I think the implementation is good. Here some suggestions

thewill-i-am:
@nicolaassolini-qntm @CalMacCQ , Thank you for your time spent on the review; I'll implement the suggestions. I think they should be ready by Monday.

tests/integration/test_range.py
thewill-i-am: Yes, agreed. Since `test_range_reverse` already compiles and emulates the reversed range
across multiple cases, the validate-only test is redundant now. I removed the whole
`test_range_reverse_validate` test rather than only the `validate(main.compile_function())`

guppylang/src/guppylang/std/iter.py
thewill-i-am: @nicolaassolini-qntm. 

I tried the suggested refactor locally, but it seems to change the behavior for negative-step ranges.

For example, `range(5, 0, -1)` yields `5, 4, 3, 2, 1`, so after `reverse_in_place()` I would expect it to yield `1, 2, 3, 4, 5`, matching `reversed(builtins.range(5, 0, -1))`.

With the suggested refactor, that case becomes empty instead. I think the issue comes from using one shared `diff = stop - next` formula for both positive and negative steps, which makes the signed division/modulo cases tricky.

Could you confirm that the expected behavior for `range(5, 0, -1).reverse_in_place()` is `1, 2, 3, 4, 5`?

thewill-i-am: @nicolaassolini-qntm More explanation here: 

For `range(5, 0, -1).reverse_in_place()`:

- The current/original implementation produces: `1, 2, 3, 4, 5`
- The suggested refactor produces: an empty range

For `range(1, 5, 2)` reversed twice:

- The current/original implementation produces: `1, 3`
- The suggested refactor produces: an empty range

Could you confirm which behavior you expect? My understanding is that `reverse_in_place`
should match `reversed(builtins.range(...))`, and reversing twice should restore the
original sequence.

guppylang/src/guppylang/std/iter.py
nicolaassolini-qntm: I agree.

I also think that having range.reserve() is compatible with a future reversed(i: Iterable) implemented through protocols. (@mark-koch  ?)

guppylang/src/guppylang/std/iter.py
nicolaassolini-qntm: My bad, I updated the suggestion, now it should be fine (the issue was `count = (diff + step - 1) // step` with negative `diff` and `step`

thewill-i-am: it's working now @nicolaassolini-qntm !!!

I removed a redundant `count` assignment. The first `count = ...` was immediately
overwritten by the next line, so it had no effect, and it could be misleading for negative
steps because `step` could still be negative there. The remaining formula uses
`distance = abs(diff)` and `step = abs(self._step)`, which is the intended calculation.

guppylang/src/guppylang/std/iter.py
thewill-i-am: That makes sense; I'm happy to rename the method, but I think that's a longer discussion. I'll wait for you to decide. I'll be waiting for your decision. @nicolaassolini-qntm @CalMacCQ @mark-koch

tests/integration/test_range.py
thewill-i-am: `# Use an order-sensitive fold to distinguish e.g. [3, 1] from [1, 3].` Done

tests/integration/test_range.py
thewill-i-am: I think this suggestion is outdated / misplaced now. This line is inside the
`expected(...)` helper, so replacing it with `validate(main.compile_function())` would
break the test because `expected` would no longer return the Python reference value.

The validate-only test was removed separately because `test_range_reverse` already compiles
and emulates the function through `run_int_fn`.  Please correct me if I am wrong

nicolaassolini-qntm: I'm suggesting to add at line 107 `validate(main.compile_function())` before the list of `run_int_fn` just to see if a potential failing test is due to validation or due to an emulator fail

thewill-i-am: Oh, makes sense. Done !
It's fixed now !  -> Line 108

nicolaassolini-qntm (commented):
Is good for me now; we'll let you know about the function name. Thanks again!

nicolaassolini-qntm (approved):

guppylang/src/guppylang/std/iter.py
nicolaassolini-qntm: Hi @thewill-i-am, We discussed and the proper name should be `__reversed__`. Thanks for the help!

Also, I edited your PR since this PR is not closing completely https://github.com/Quantinuum/guppylang/issues/2086 (your contribution is a good help for us in implementing that feature 😄 )

thewill-i-am: @nicolaassolini-qntm Awesome! It was very interesting working on this, let me know if there's anything else I can help with.

nicolaassolini-qntm:
@thewill-i-am I renamed the method and requested a Copilot review for a final check. If everything is fine, as I suppose, I will merge your contribution.

Thanks again. If you are willing to contribute again to our projects (guppy, but also tket2 and hugr), feel free to have a look at our open issues. We also use the label `P-help-wanted` on the issues that are more suitable for external contributors.

thewill-i-am:
> @thewill-i-am I renamed the method and requested a Copilot review for a final check. If everything is fine, as I suppose, I will merge your contribution.
> 
> Thanks again. If you are willing to contribute again to our projects (guppy, but also tket2 and hugr), feel free to have a look at our open issues. We also use the label `P-help-wanted` on the issues that are more suitable for external contributors.

@nicolaassolini-qntm  Thanks! My plan is to keep contributing because I’m a big fan of the work you did with Guppy. :)

By the way, some tests are failing after the rename. Would you like me to take a look at that?

PR #2093, opened 2026-07-25
feat: Better Error when calling method on Struct class  

### LINKED ISSUE #1603: [Feature]: Better Error when calling method on Struct class
### Problem Statement

The program: 
```python
@guppy.struct
class MyStruct:
    field: int

    @guppy
    def foo(self: "MyStruct") -> None:
        pass

@guppy
def main() -> None:
    MyStruct.foo()
```

raise:
```
Error: Attribute not found (at <In[8]>:13:13)
   | 
11 | @guppy
12 | def main() -> None:
13 |     MyStruct.foo()
   |              ^^^ `int -> MyStruct` has no attribute `foo`

```

We should print a hint suggesting that Struct must be initialised to be able to call the method on it

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

rmfisher26:
Closes #1603

## Summary
- Add a new `InstanceMethodOnClassError` diagnostic that fires when an instance method
  is called directly on a struct or enum class (e.g. `MyStruct.foo()`) instead of on an
  instance (`MyStruct(...).foo()`), replacing the previous cryptic "Attribute not found"
  message with one that names the method, the type, and shows a corrected example.
- Applies to both structs and enum classes (`MyEnum.method()` → suggests
  `MyEnum.Left(...).method(...)`), while leaving legitimate enum variant construction
  (`MyEnum.Left()`) unaffected.
- Fixes a parameter-shadowing edge case surfaced while implementing the above: a local
  parameter that shadows an unrelated global struct/enum name (e.g.
  `def main(StructA: StructB) -> int: return StructA.foo()`) was being misdiagnosed as
  "instance method called on class" instead of resolving to the local instance.

## Test plan
- [x] Added `tests/error/struct_errors/instance_method_on_class.py` reproducing the
      exact example from #1603, with its `.err` snapshot showing the new message.
- [x] Updated `tests/error/enum_errors/use_enum_class4.err` and `use_enum_class5.err`
      snapshots to reflect the improved message for the equivalent enum case.
- [x] Added `tests/integration/test_struct.py::test_method_call_on_param_shadowing_struct_name`
      as a regression test for the parameter-shadowing fix above.
- [x] Manually verified the exact code sample from the issue now produces the intended
      error + help message.
- [x] Full test suite passes (`uv run pytest tests/`).


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -649,6 +670,31 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                    and name_id in self.ctx.globals
+                ):
+                    defn = self.ctx.globals[name_id]
+                    if (
+                        isinstance(defn, TypeDef)
+                        and not isinstance(defn, ParsedEnumDef | CheckedEnumDef)
+                        and node.attr != "__new__"
+                        and ENGINE.get_instance_func(defn, node.attr) is not None
+                    ):
+                        self._raise_instance_method_on_class(
```
maximilianruesch: I do not think it is worthwhile to extract this into a separate function that only contains three statements, especially since they are very clear in intention and only used twice here throughout.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -725,13 +771,21 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
-            elif (method_w_ty := self._check_method(ty, node)) and not isinstance(
-                node.value, GlobalName
-            ):
+            elif method_w_ty := self._check_method(ty, node):
+                if isinstance(node.value, GlobalName):
+                    # `EnumClass.method` — the method exists, but was looked up on the
+                    # enum class itself rather than on an instance/variant of it. Raise
+                    # a specific error instead of falling through to the generic
+                    # "no variant" message below, which would be misleading here since
+                    # the method does in fact exist.
```
```suggestion
                    # Method exists, but on enum instances (since class methods are currently unsupported). Fail with a helpful error instead of the generic "no variant" message below.
```
Perhaps need to deal with overly long lines, they are not easy to judge from the GH UI.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -649,6 +670,31 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                    and name_id not in self.ctx.generic_param_inst
+                    and name_id in self.ctx.globals
+                ):
+                    defn = self.ctx.globals[name_id]
+                    if (
+                        isinstance(defn, TypeDef)
+                        and not isinstance(defn, ParsedEnumDef | CheckedEnumDef)
+                        and node.attr != "__new__"
+                        and ENGINE.get_instance_func(defn, node.attr) is not None
+                    ):
```
maximilianruesch: This is a fairly complicated expression. Could you explain why you need each of these conditions? Perhaps a comment would help here?

maximilianruesch (dismissed):
Nice error message! A few questions remain.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
rmfisher26: Thanks for your suggestion and that makes sense to make it clearer here so I have removed the helper function.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
rmfisher26: Thank you for your documentation suggestion on this.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
rmfisher26: Sure! I have added some comments to explain what's happening so let me know if this what you are looking for. 

- The outer condition check handles shadowing: `name_id `is only treated as the global class/enum reference if it isn't shadowed by a local variable or a generic parameter of the same name (e.g. `def f(MyStruct: Other)` or a `nat_var `named `MyStruct`) — both take priority, matching `_check_name_id `below.

- The inner condition check narrows to a "instance method called on class": `defn `must be a type (excluding enums — `get_instance_func `can't tell a method apart from a variant constructor there, so that case is handled separately further down)

I have also added two regression tests covering the generic-param-shadowing and field-vs-method cases, which weren't covered before.

maximilianruesch: This has a lot of overlap with `_check_name_id`. Even though it would be hard to refactor into one helper method (and perhaps not helpful), I encourage you to put this into a `_find_instance_func(self, ...) -> Definition | None` to make the intention way clearer (and keep this already convoluted function somewhat more tidy).

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -669,6 +670,45 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+                        err = InstanceMethodOnClassError(
+                            attr_span, defn.name, node.attr
+                        )
+                        err.add_sub_diagnostic(
+                            InstanceMethodOnClassError.CallOnInstanceHelp(
+                                None, f"{defn.name}(...).{node.attr}(...)"
+                            )
+                        )
+                        raise GuppyTypeError(err)
```
nicolaassolini-qntm: this logic is relevant only when the type is a struct, can we move the check below under `isinstance(ty, StructType)` similar to what is done for the enum case?

tests/error/struct_errors/field_access_on_class.err
```diff
@@ -0,0 +1,8 @@
+Error: Attribute not found (at $FILE:18:20)
+   | 
+16 |     # the ordinary "attribute not found" error rather than being misdiagnosed as an
+17 |     # instance method called on the class.
+18 |     return MyStruct.field
+   |                     ^^^^^ `int -> MyStruct` has no attribute `field`
```
nicolaassolini-qntm: Is it possible to have a better error here? Somenthing saying "`field` is an instance flied of `MyStruct`, not a static field"

nicolaassolini-qntm (changes requested):
Hi, I have some suggestions

nicolaassolini-qntm:
Hi @rmfisher26, I left some suggestions for you

rmfisher26:
Hi @nicolaassolini-qntm, thanks for your input. I'm still intending to work on this so apologies for the delay.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
rmfisher26: @maximilianruesch , I have extracted with a new function  `_find_instance_member_on_class(self, node: ast.Attribute)`, so let me know what you think of that if that's closer to what would be better here. This also combines the efforts for trying to improve the error message to support `field `and `method` accesses

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
rmfisher26: Hi @nicolaassolini-qntm, I extracted it into `_find_instance_member_on_class` (see reply above) to improve the readability and included a comment at the call site explaining why I wasn't able to do same as the enum case.

tests/error/struct_errors/field_access_on_class.err
rmfisher26: > Is it possible to have a better error here? Somenthing saying "`field` is an instance flied of `MyStruct`, not a static field"

Hi @nicolaassolini-qntm , I have extracted with a new function  `_find_instance_member_on_class(self, node: ast.Attribute)`,  so this combines the efforts for trying to improve the error message to support `field `and `method` accesses. Let me know if that is closer to something more useful here.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -746,6 +773,50 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
...
+        `MyStruct.field`) rather than on an instance, or `None` otherwise.
+
+        Excludes enum methods: `get_instance_func` can't tell an instance method
+        apart from a variant constructor there, so this would misfire on legal
+        variant construction like `MyEnum.Left()`. That case is handled
+        separately by the `EnumType` branch further down. Enums have no
+        class-level fields, so no such exclusion is needed for those.
+        """
+        if not isinstance(node.value, ast.Name):
+            return None
```
nicolaassolini-qntm: we call this function only when node.value is an ast.name right? better do an assertion, or change the signature of the function

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -650,6 +651,23 @@ def visit_Attribute(self, node: ast.Attribute) -> tuple[ast.expr, Type]:
             # visit_Name (that is called through synthesize)
             ty = get_type_opt(node.value)
             if ty is None:
+                # Can't be scoped under `isinstance(ty, StructType)` like the enum
+                # case is below: a bare struct/alias reference resolves (via
+                # `_check_name_id`) to its `__new__` constructor's `FunctionType`,
+                # never to `StructType`, so we check it here instead.
```
nicolaassolini-qntm: I'm not sure about this. We're doing the check when `ty` is `None`, i.e. when we haven't set the `node.type` attribute yet, but `ty == None` isn't a good flag for a potential error, IMO. Since this error happens only with structure classes, and `_check_name_id` types it as a function, can we move this check to after visiting `node.value` if the type found is a function returning a struct?
Then the checks inside `_find_instance_member_on_class` should be enough to distinguish a guppy struct class from a function returning a structure, I hope :).

nicolaassolini-qntm (commented):
Thanks for the work, the new error messages are good!
I still have a question, though.

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
rmfisher26: Hi @nicolaassolini-qntm, sure I committed refactoring to try to address your questions, so I hope this may be closer to what you would like to see.

For the first part, I have moved it as a branch alongside the enum case (`isinstance(ty, FunctionType) and ty.output` being a struct.

So the check also confirms `node.value `points at the struct's own `__new__ `  which is the same identity `_check_global` uses when it resolves a bare class name.

I also had some new tests for `attribute_on_function_returning_struct.py` to show that a global `extern `value returning a struct isn't confused with the struct's own class reference. 

And `attribute_on_local_function_returning_struct.py` the checker still correctly falls through to the plain `Attribute not found` error instead of misfiring as `Instance method accessed on class` for a local parameter.

nicolaassolini-qntm (commented):
LGTM, just a final little refactor, I would move:

- `attribute_on_function_returning_struct`
- `attribute_on_local_function_returning_struct`
- `generic_param_shadows_struct`

In `tests/error/misc_errors` since they  are not testing specifically struct.

Thanks again for your contribution!

nicolaassolini-qntm (approved):
All good, thanks a lot!

PR #2096, opened 2026-07-27
fix!: Stop treating protocol bounds as always copyable and droppable

### LINKED ISSUE #2087: [Bug]: Protocol bounds have an implicit `Copy, Drop` bound
### Description

```python
@guppy.protocol
class HasLength:
    @guppy.require
    def __len__(self) -> int: ...

@guppy
def my_len(value: HasLength) -> int:
    return value.__len__()

@guppy
def main(xs: array[int, 10]) -> int:
    return my_len(xs)

main.check_function()
```

fails with

```
Error: Expected a copyable type (at file.py:173:18)
    | 
171 | @guppy
172 | def main(xs: array[int, 10]) -> int:
173 |     return my_len(xs)
    |                   ^^ Expected a copyable type, got type `array[int, 10]` which is
    |                      not implicitly copyable
```

It's behaving as if `value` had an extra `Copy` bound (the same also applies for `Drop`).

We should change the type parsing logic to stop setting `must_be_copyable` and `must_be_droppable` when parsing type bounds.

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

Guppy Compiler

### Environment

```markdown

```

### Additional Context

_No response_

mark-koch:
Fixes #2087 

BREAKING CHANGE: Protocol arguments are now non-copyable and non-droppable by default. Add an explicit `Copy` or `Drop` bound to make them copyable or droppable.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
@@ -856,7 +856,15 @@ def immediate_child_places(place: Place) -> list[Place]:
 
 def is_inout_var(place: Place) -> TypeGuard[Variable]:
     """Checks whether a place is a borrowed variable."""
-    return isinstance(place, Variable) and InputFlags.Inout in place.flags
+    return (
+        isinstance(place, Variable)
+        and InputFlags.Inout in place.flags
+        # Only affine types are considered for inout. Normally, copyable types wouldn't
+        # have an inout flag, however it can show up when generic functions are
+        # instantiated with protocol instances that are copyable. In those cases, we
+        # just consider the variable as not inout.
+        and not place.ty.copyable
+    )
 
 
 def has_explicit_copy(ty: Type) -> bool:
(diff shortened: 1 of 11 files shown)


guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -856,7 +856,15 @@ def immediate_child_places(place: Place) -> list[Place]:
...
     """Checks whether a place is a borrowed variable."""
-    return isinstance(place, Variable) and InputFlags.Inout in place.flags
+    return (
+        isinstance(place, Variable)
+        and InputFlags.Inout in place.flags
+        # Only affine types are considered for inout. Normally, copyable types wouldn't
+        # have an inout flag, however it can show up when generic functions are
+        # instantiated with protocol instances that are copyable. In those cases, we
+        # just consider the variable as not inout.
+        and not place.ty.copyable
```
croyzor, line 866: Nice, I think this was piece I was missing in order to implement this. But it's adding a bit of complexity to the flag system that I don't like

croyzor (approved):
🚀

PR #2098, opened 2026-07-27
fix: Correctly parse modifier protocol bounds

### LINKED ISSUE #2095: [Bug]: Modifier protocol bounds are not parsed correctly
### Description

For example, 

```python
@guppy(daggerable=True)
def apply_dagger[F: Daggerable[[qubit], None]](f: F, q: qubit) -> None:
    f(q)

@guppy
def main(q: qubit) -> None:
    with dagger:
        apply_dagger(h, q)

main.check_function()
```

fails with

```
Error: Invalid type argument (at file.py:325:31)
    | 
323 | 
324 | @guppy(daggerable=True)
325 | def apply_dagger[F: Daggerable[[qubit], None]](f: F, q: qubit) -> None:
    |                                ^^^^^^^ Not a valid type argument
```

It works, if we instead define `def apply_dagger[F: ](f: Daggerable[[qubit], None], q: qubit) -> None`

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

mark-koch:
Fixes #2095

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tys/parsing.py
+++ b/guppylang-internals/src/guppylang_internals/tys/parsing.py
@@ -532,10 +532,16 @@ def parse_bound(
         arg_nodes = (
             bound.slice.elts if isinstance(bound.slice, ast.Tuple) else [bound.slice]
         )
-        # Special case for the `Callable` protocol
+        # Special case for the `Callable` and modifiable function protocols
         if isinstance(proto_defn, CallableProtocolDef):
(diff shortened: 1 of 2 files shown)


tatiana-s (approved):

maximilianruesch:
It seems like this PR has to be updated with respect to https://github.com/Quantinuum/guppylang/pull/2096

PR #2099, opened 2026-07-27
docs: Document expected qubits in emulator API

maximilianruesch:
Part way to https://github.com/Quantinuum/guppy-docs/issues/159 which we should resolve in this repository.

Integrated preview:
image

### DIFF
--- a/guppylang/src/guppylang/emulator/__init__.py
+++ b/guppylang/src/guppylang/emulator/__init__.py
@@ -14,7 +14,7 @@ outputs.
 Basic emulation
 -----------------
 
-Calling ``.emulator`` on a Guppy function compiles it into an
+Calling ``.emulator()`` on a Guppy function compiles it into an
 :py:class:`EmulatorInstance`. The method has a required parameter
 for the number of qubits to allocate. This cannot be inferred from the program
 fully automatically as it can request an arbitrary number of qubits at runtime.
--- a/guppylang/src/guppylang/emulator/__init__.py
+++ b/guppylang/src/guppylang/emulator/__init__.py
@@ -37,8 +37,8 @@ Calling ``.run()`` on the instance runs the emulation, returning an
 
 .. hint::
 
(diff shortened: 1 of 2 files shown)


guppylang/src/guppylang/emulator/__init__.py
```diff
@@ -16,9 +16,8 @@
 
 Calling ``.emulator`` on a Guppy function compiles it into an
```
cqc-alec:
```suggestion
Calling ``.emulator()`` on a Guppy function compiles it into an
```

guppylang/src/guppylang/emulator/__init__.py
```diff
@@ -36,6 +35,22 @@ def foo() -> None:
 
     foo.emulator(n_qubits=1).run()
 
+.. hint::
+
+    You can omit `n_qubits` from the `.emulator` call if you add the `@expected_qubits`
+    decorator to `foo`:
```
cqc-alec:
```suggestion
    You can omit ``n_qubits`` from the ``.emulator`` call if you add the ``@expected_qubits``
    decorator to ``foo``:
```

guppylang/src/guppylang/emulator/__init__.py
```diff
@@ -36,6 +35,22 @@ def foo() -> None:
...
+
+    .. code-block:: python
+
+        from guppylang.decorator import expected_qubits
+
+        @guppy
+        @expected_qubits
```
cqc-alec: Doesn't this need a numerical value?
```suggestion
        @expected_qubits(1)
```

cqc-alec (changes requested):

guppylang/src/guppylang/emulator/__init__.py
maximilianruesch: 😬

cqc-alec (approved):

PR #2104, opened 2026-07-28
refactor: do tracing during checking

### LINKED ISSUE #1592: [Feature]: Split compilation of comptime functions into a separate checking phase
### Problem Statement

Currently, checking and hugr compilation of comptime functions happens in one go during the compilation stage. We should split this up to first run tracing and checking, and then build the Hugr in a separate step.

This will make comptime functions more consistent with the rest of the compiler.

### Proposed Solution

_No response_

### Component

Guppy Compiler

### Effort Estimate

Hard - Significant effort required

### Additional Context

_No response_

acl-cqc:
closes #1592

The strategy is to record the ops, calls, loads and load_functions that we'll need to perform on the Hugr builder later (during compilation), but only to check these now.

The "99%" performance degradation is for a benchmark that just calls `check()` so since this now executes the user's python code (whereas it didn't before)....that's expected. (Albeit 😆 🤣 )


guppylang-internals/src/guppylang_internals/tracing/state.py
```diff
@@ -1,33 +1,41 @@
...
 
 
 @dataclass
 class TracingState:
     """Internal state that is used during the tracing phase of comptime functions."""
 
-    #: Reference to the global compilation context.
-    ctx: CompilerContext
+    #: Context used to translate Guppy types into HUGR types.
+    ctx: ToHugrContext
```
acl-cqc, line 22: We might wanna get rid of this first; `None` implements `ToHugrContext` 😆

guppylang-internals/src/guppylang_internals/tracing/state.py
```diff
@@ -68,3 +76,172 @@ def set_tracing_state(state: TracingState) -> Iterator[None]:
         yield
     finally:
         _STATE.reset(token)
+
+
+TraceableCall: TypeAlias = GlobalCall | AbortExpr | StateOutputExpr
+
+
+@dataclass(frozen=True)
```
acl-cqc: Ok, so a bunch of these dataclasses were written by AI...I have tidied them up a bit (and fixed its TraceCall which was totally wrong ;-) )...I think they are correct now!

acl-cqc: Moreover, these might be better in a separate `tracing/builder.py`; I did start off with that and had circular-import problems but these could be solved (just, all the four methods in TraceBuilder will have to locally import from state.py)

guppylang-internals/src/guppylang_internals/compiler/expr_compiler.py
```diff
@@ -559,6 +548,10 @@ def _visit_output_tag(self, tag: Const, loc: ast.expr) -> str:
         return tag_value
 
     def visit_AbortExpr(self, node: AbortExpr) -> Wire:
+        returns = self.compile_abort(node)
+        return self._pack_returns(returns, get_type(node))
+
+    def compile_abort(self, node: AbortExpr) -> list[Wire]:
```
acl-cqc: These helpers are a bit ugly....could do with a neater way

guppylang-internals/src/guppylang_internals/tracing/state.py
acl-cqc: Now renamed and moved into `recorder.py`

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -212,15 +238,60 @@ def compile_call(
 
     @override
     def compile_inner(self, ctx: CompilerContext) -> None:
-        """Compiles the body of the function by tracing it."""
-        from guppylang_internals.tracing.function import trace_function
+        """Replays the trace recorded while the function was checked."""
```
acl-cqc, line 226: Possibly this should go in `recorder.py`?

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -188,20 +217,36 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
...
     with capture_guppy_errors():
-        # Try to turn args into `GuppyObjects`
+        # Try to turn args into `GuppyObjects`, each containing a `ComptimeVariable`.
+        # (We only really need the variables for arguments that turn out to be inout,
+        # so that we can update the variable with the new port after the call,
+        # but we do it for all arguments for simplicity.)
+        new_vars: list[tuple[ComptimeVariable, TraceWire]] = []
+
+        def assign_var(obj: GuppyObject, arg: Any) -> GuppyObject:
+            # We can get into trace_call more than once for e.g. __radd__.
```
acl-cqc, line 218: This comment applied before the PR too

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -188,20 +217,36 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
...
+            # If the first call fails, we still mark GuppyObjects as used,
+            # but thankfully such methods only happen for Copyable types (ATM).
+            w = obj._use_wire(func)
+            if isinstance(w, ComptimeVariable):
```
acl-cqc, line 222: The alternative is to make GuppyObject *only* contain `ComptimeVariable` (rather than `TraceWire | ComptimeVariable`), and the various other classes/unpacking methods etc. too. However, *construction* of GuppyObjects would then also need to return the `new_vars` map, which would be a bit more complex. But maybe cleaner....?

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -248,34 +289,31 @@ def trace_call(func: CallableDef, *args: Any) -> Any:
             "`compute_input_flags` in the call checker."
         )
 
-    for flags, arg, var in zip(input_flags, args, arg_vars, strict=True):
+    # Record call to compile later
+    call = state.builder.record_call(call_node, new_vars)
+
+    # Since all inputs are GuppyObjects identifying ComptimeVariables (varieties
+    # of Place), ExprCompiler will update the Places to map to the output wires
+    # of the call. Here we just write the GuppyObjects with those Places back to
```
acl-cqc, line 288: This is only necessary because we copied the input GuppyObjects above, but we had to do that because of getting here twice, as per [this thread](https://github.com/Quantinuum/guppylang/pull/2104#discussion_r3710741932)

acl-cqc: (However this avoids the need, that we discussed, to second-guess the variables ids resulting from ExprCompiler, or anything like that 👍)

guppylang-internals/src/guppylang_internals/tracing/function.py
acl-cqc: Because of the `suppress(Exception)` here, which suppresses the Exception but not the side-effects: https://github.com/Quantinuum/guppylang/blob/1ace5b92f0848ee7455b35ca5d617a6423e96b97/guppylang-internals/src/guppylang_internals/checker/expr_checker.py#L830-L854

guppylang-internals/src/guppylang_internals/tracing/function.py
acl-cqc: I experimented with this using AI but eventually realized it was doomed: code building `GuppyObject`s is finely interwoven with adding of unpacking ops, so if we are aborting (inside that `suppress(Exception)` ) then not only do we want to avoid registering the new ComptimeVariables (and ObjectUse's of existing ones) but also avoid registering *all* the ops. That suggests nested TraceRecorders that can be merged (after the point where any exceptions might be thrown, heh)...quite a big change and the issue that we should replace `suppress(Exception)` with some kind of transaction is longstanding (unless I've made it worse??)

guppylang-internals/src/guppylang_internals/tracing/state.py
```diff
@@ -5,29 +5,30 @@
...
 
-    #: Reference to the global compilation context.
-    ctx: CompilerContext
+    #: Context used to translate Guppy types into HUGR types.
+    ctx: ToHugrContext
 
-    #: The current dataflow graph under construction.
-    dfg: DFContainer
+    #: The trace of operations performed during the comptime execution of a function.
+    builder: "TraceRecorder"
```
mark-koch: Suggest renaming this field to something else since we already use `builder` for Hugr stuff. Maybe `TracingState.trace`?

guppylang-internals/src/guppylang_internals/tracing/recorder.py
```diff
@@ -0,0 +1,192 @@
...
+
+#: A value usable as an input to an op/call in a trace, i.e. that will be resolved
+#: to a Hugr `Wire` during compilation. (`ComptimeVariable`s are resolved via the
+#: state held in the DFContainer.)
+TraceOutput: TypeAlias = TraceWire | ComptimeVariable
+
+
+@dataclass(frozen=True)
+class TraceOperation:
+    """A primitive dataflow operation emitted while tracing."""
```
mark-koch: We could consider splitting this up into `TraceMakeTuple`, `TraceUnpackTuple`, `TraceMakeArray`, `TraceUnpackArray` since these are the only Hugr ops that we actually use. Then we could also get rid of the `_operation_output_count` helper, and maybe even `TraceNode`?

guppylang-internals/src/guppylang_internals/tracing/recorder.py
```diff
@@ -0,0 +1,192 @@
...
+    inputs: tuple[TraceOutput, ...]
+    output_count: int
+    node: AstNode | None
+
+
+@dataclass(frozen=True)
+class TraceLoad:
+    """A HUGR value loaded while tracing."""
+
+    value: val.Value | DefId
```
mark-koch: Does this need to be a Hugr value, or can we delay that conversion until we build the actual Hugr?

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -88,20 +101,31 @@ def params(self) -> Sequence[Parameter]:
...
         generic_args = {
             param.name: arg
             for param, arg in zip(self.ty.params, type_args, strict=True)
         }
+        from guppylang_internals.tracing.function import trace_function
+
+        trace = trace_function(
+            self.python_func,
+            mono_ty,
+            sum(InputFlags.Comptime not in inp.flags for inp in mono_ty.inputs),
```
mark-koch: Would prefer to just compute this inside `trace_function` based on the passed `mono_ty`

guppylang-internals/src/guppylang_internals/definition/traced.py
mark-koch: Yes imo good idea to move this

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -71,26 +75,47 @@ class TracingReturnError(Error):
...
+    func_def: "TracedFunctionDef",
+) -> Trace:
     """Kicks off tracing of a function.
 
     Invokes the passed Python callable and constructs the corresponding Hugr using the
     passed builder.
     """
-    state = TracingState(ctx, DFContainer(builder, ctx, {}), node, func_def)
+
+    def const_argument_to_python_value(arg: Argument) -> Any:
```
mark-koch: Does this need to be a local function? As far as I can see it doesn't capture anything?

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -71,26 +75,47 @@ class TracingReturnError(Error):
...
+        """Extracts a Python value from the given generic argument."""
+        match arg:
+            case ConstArg(ConstValue(value=v)):
+                assert v is not None
+                return v
+            case (
+                ConstArg(BoundConstVar())
+                | ConstArg(ExistentialConstVar())
```
mark-koch: Existential variables should never show up, those should be internal errors

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -71,26 +75,47 @@ class TracingReturnError(Error):
...
+                # This means we are building the arguments with which to trace
+                # an uninstantiated generic function. So, avoid tracing such...
+                raise RequiresMonomorphizationError
+            case TypeArg(ty=ExistentialTypeVar()):
+                raise InternalGuppyError("Shouldn't happen?!")
+            case _:
+                # TODO: We don't have a comptime representation of types yet, so we can
+                #  only translate const arguments into Python values for now. In the
+                #  future, drop this restriction and support all kinds of arguments.
+                return None
```
mark-koch, line 102: Careful with returning `None` if `None` could also be a valid const value. Maybe use a custom sentinel value instead?

mark-koch (commented):
Oof I did not consider the issues with `supress(GuppyError)` when we were discussing this before... I didn't have too much time to think about this today, so will have another look tomorrow. For now, I just have some nits and small comments

Also wondering if it would be feasible to move all references to Hugr out if tracing (e.g. no hugr types or values) and only convert during compilation?

guppylang-internals/src/guppylang_internals/tracing/function.py
acl-cqc: Sorry - not there - but at least, because of the suppress here: https://github.com/Quantinuum/guppylang/blob/6b86190ca456431761e06540aaf226f26ff2c67f/guppylang-internals/src/guppylang_internals/tracing/object.py#L87-L99

Possibly elsewhere too, I'm not sure if that's the only source.

guppylang-internals/src/guppylang_internals/tracing/state.py
acl-cqc: Yes, oops 🤦, it should just be `recorder` I think

guppylang-internals/src/guppylang_internals/definition/traced.py
acl-cqc: Yes, done, thanks

guppylang-internals/src/guppylang_internals/tracing/function.py
acl-cqc: I felt it was a long way away and it was getting too specific/quirky so better hidden...can move it back if you feel strongly?

guppylang-internals/src/guppylang_internals/tracing/function.py
acl-cqc: Yes, thank you, indeed!

guppylang-internals/src/guppylang_internals/tracing/function.py
acl-cqc: Hmmm, ok, have changed to return the ConstValue, and then take the `.value` in the caller if not None. I'm not sure I've seen what possible constant might be represented as None (not `nothing(): Option` surely?), but I think this is safe now regardless ;)

guppylang-internals/src/guppylang_internals/tracing/recorder.py
acl-cqc: Yeah. Removing TraceNode makes it worthwhile, it's about even on line count when I do that. A mildly funky overload on `TraceRecorder._add` as a final tada ;)

guppylang-internals/src/guppylang_internals/tracing/recorder.py
acl-cqc: Yes, but had to split into TraceLoad and TraceLoadVal as the latter could be an Any (the former is a Def)

guppylang-internals/src/guppylang_internals/definition/traced.py
acl-cqc: Ok I haven't actually moved this, because now there is no `hugr` in `tracing` 🎉. This code is where we convert between the hugr and tracing APIs so I think it's reasonable for it to live outside (as now) or I guess I could put it in its own file `tracing/compile.py`?

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -88,20 +105,29 @@ def params(self) -> Sequence[Parameter]:
     def check(self, type_args: Inst, globals: Globals) -> "TracedMonoFunctionDef":
         """Monomorphizes the function for the given type arg instantiation.
 
-        The actual checking happens during tracing while constructing the HUGR.
+        Executes the Python body while recording a replayable HUGR trace.
```
mark-koch, line 93: Can we drop the Hugr here since the tracing is no longer tied to hugr?

guppylang-internals/src/guppylang_internals/definition/traced.py
mark-koch: Slight preference for `tracing/compile.py` but I don't feel super strongly

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -212,15 +238,84 @@ def compile_call(
...
+                    outputs = [builder.load(hugr_val)[0]]
+                    output_count = 1
+                case TraceLoad(v_def, node):
+                    defn = ctx.build_compiled_def(v_def.id, type_args=())
+                    # TypeDefs should already have been converted to LoadFunctions
+                    # of their *constructors* during checking/tracing.
+                    if not isinstance(defn, CompiledValueDef):
+                        def_kind = defn.description.capitalize()
+                        err = f"{def_kind} `{defn.name}` is not a value"
+                        raise GuppyComptimeError(err)
```
mark-koch: Any chance we can catch this during tracing? Don't worry about it if it would be annoying to change though

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -212,15 +238,84 @@ def compile_call(
...
+                    outputs = [defn.load(DFContainer(builder, ctx), ctx, node)]
+                    output_count = 1
+                case TraceFunctionLoad(def_id, type_args, node):
+                    defn = ctx.build_compiled_def(def_id, type_args)
+                    assert isinstance(defn, CompiledValueDef)
+                    outputs = [defn.load(DFContainer(builder, ctx), ctx, node)]
+                    output_count = 1
+                case TraceCall(call_node, input_places):
+                    # but keep this:
```
mark-koch: Is this comment still relevant?

guppylang-internals/src/guppylang_internals/definition/traced.py
```diff
@@ -212,15 +238,84 @@ def compile_call(
...
+                    for arg, val in input_places:
+                        dfg[arg] = get_wire(val)
+                    outputs = [comp.compile(call_node, dfg)]
+                    output_count = 1
+                case _:
+                    assert_never(entry)
+
+            assert len(outputs) == output_count
```
mark-koch: `output_count` is not used?

guppylang-internals/src/guppylang_internals/tracing/function.py
mark-koch: > I'm not sure I've seen what possible constant might be represented as None

Guppy supports the type `None` (lowers to the Hugr unit type) and it's value `None`. Afaik we don't actually support `None` const values when parsing types (would be a pretty useless feature anyways...) but we might add that for consistency in the future

guppylang-internals/src/guppylang_internals/tracing/function.py
mark-koch: Don't feel strongly, feel free to keep it as is

mark-koch (approved):
Thanks! Didn't end up coming up with better solution for the transaction problem, but I think the way it now should be ok for now.

Just a couple of final nits, happy to merge after 👍

guppylang-internals/src/guppylang_internals/definition/traced.py
acl-cqc: We've already checked that it's a ValueDef, which is all that we can do during tracing/checking - we won't have CompiledValueDef's available at that stage will we?

Or, ah, given we've already checked it's a ValueDef, maybe this is unnecessary (InternalGuppyError or even assert) - I think it should never fire

guppylang-internals/src/guppylang_internals/definition/traced.py
acl-cqc: Detritus, thanks!

guppylang-internals/src/guppylang_internals/definition/traced.py
acl-cqc: Changed to an assert, added a dummy variable declaration so the type-checker will moan if it's not a ValueDef

PR #2118, opened 2026-07-30
fix!: Type args to Either/Result factory methods `right`/`err` same way round

### LINKED ISSUE #2113: [Bug]: type error trying to build "Either" values
### Description

First attempt was:
```
@guppy
def max(i: int, f: float) -> Either[int, float]:
     return left(i) if float(i)>f else right(f)
```
this gives:
```
E           guppylang_internals.error.GuppyTypeInferenceError: Error: Cannot infer type (at /Users/alanlawrence/repos/guppylang/tests/integration/tracing/test_option_either.py:88:15)
E              | 
E           86 |     @guppy
E           87 |     def max(i: int, f: float) -> Either[int, float]:
E           88 |         return left(i) if float(i)>f else right(f)
E              |                ^^^^^^^ Cannot infer type variables in expression of type
E              |                        `Either[int, ?R]`
E           
```
poor, but perhaps just a limitation rather than a bug. So I tried:
```
@guppy
    def max(i: int, f: float) -> Either[int, float]:
        return left[int,float](i) if float(i)>f else right[int,float](f)
```
which bizarrely says:
```
55 |     @guppy
56 |     def max(i: int, f: float) -> Either[int, float]:
57 |         return left[int,float](i) if float(i)>f else right[int,float](f)
   |                                                                       ^ Expected argument of type `int`, got `float`
```

However, this works:
```
@guppy
def max(i: int, f: float) -> Either[int, float]:
   r: Either[int, float] = right(f)   
   return left[int,float](i) if float(i)>f else r
```

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

acl-cqc:
[First commit](97d138e9479d7467d8fae28c65b0dca6b948f782) adds tests of explicit type args to `right`, which have to be the wrong way round to get the tests to [pass](https://github.com/Quantinuum/guppylang/actions/runs/30568185841/).

(Yes, these tests are perhaps a little synthetic. But with type inference only working within basic blocks, not across them, there are more complex cases where manually specifying type args is the *only* way to make something compile...)

Declaring the type params "the correct way round" is only possible in/after python 3.12, but since we have said that is now the minimum supported version, we can do that.... and Result is also affected/corrected.

The "downside" is that although we don't support python 3.10, we still have `ruff` checking against this in `ruff.toml`, and fixing that...is a much bigger PR. But also only visible to developers of guppy, whereas this affects those *using* guppy. So for now I've whitelisted std/either.py and std/err.py for ruff to check against py312, and left the rest for [a cleanup issue](#2119).

closes #2113 

BREAKING CHANGE: Either and Result no longer inherit Generic, and calls to `right` or `err` with explicit type args will need the args reversed

### DIFF
--- a/guppylang-internals/src/guppylang_internals/std/_internal/compiler/either.py
+++ b/guppylang-internals/src/guppylang_internals/std/_internal/compiler/either.py
@@ -61,13 +61,7 @@ class EitherConstructor(EitherCompiler, CustomCallCompiler):
 
     def compile(self, args: list[Wire]) -> list[Wire]:
         ty = self.either_ty
-        if self.tag == 1:
-            # In the `right` case, the type args are swapped around since `R` occurs
-            # first in the signature :(
-            ty.variant_rows = [ty.variant_rows[1], ty.variant_rows[0]]
-        # For the same reason, the type of the input corresponds to the first type
-        # variable
-        inp_arg = self.type_args[0]
+        inp_arg = self.type_args[self.tag]
         assert isinstance(inp_arg, TypeArg)
         [inp] = args
         # Unpack the single input into a row
--- a/guppylang/src/guppylang/std/either.py
+++ b/guppylang/src/guppylang/std/either.py
@@ -3,7 +3,7 @@
 Type `Either[L, R]` represents a value of either typr `L` ("left") or `R` ("right").
 """
 
-from typing import Generic, no_type_check
+from typing import no_type_check
 
 from guppylang_internals.decorator import custom_function, custom_type
 from guppylang_internals.std._internal.compiler.either import (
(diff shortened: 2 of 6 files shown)


croyzor (changes requested):
> So for now I've whitelisted std/either.py and std/err.py for ruff to check against py312

I don't see any changes to `ruff.toml` – it looks like this PR is missing some commits!

croyzor (approved):
nice! 🚀

acl-cqc:
Also added a test to Result (one where explicit type args are *necessary* :-( )

PR #2121, opened 2026-07-31
docs: Optimization module docs

aborgna-q:
Adds module docs for the optimization level configuration.

Part of https://github.com/Quantinuum/guppy-docs/issues/154. I will make another PR there with changes to the sphinx config.

image

### DIFF
--- a/guppylang/src/guppylang/optimizer.py
+++ b/guppylang/src/guppylang/optimizer.py
@@ -1,6 +1,8 @@
 """Optimization configuration for Guppy compilation.
 
-Guppy applies a predefined set of optimization passes when compiling a program.
+Guppy applies a predefined set of optimization passes when compiling a program
+using the TKET compiler.
+
 These passes clean up artifacts introduced by the compiler and may simplify both
 classical and quantum operations when calling ``compile()``,
 ``compile_function()``, or ``emulator()``.
--- a/guppylang/src/guppylang/optimizer.py
+++ b/guppylang/src/guppylang/optimizer.py
@@ -18,11 +18,20 @@ Pass a member of :py:class:`OptimizationLevel` to ``with_opt_level()``:
 .. code-block:: python
 
     from guppylang import OptimizationLevel, guppy
+    from guppylang.std.builtins import output
+    from guppylang.std.quantum import h, measure, qubit
 
     @guppy
     def main() -> None:
-        _x = 2 + 2
+        q = qubit()
+        h(q)
+        h(q)
+        if measure(q):
+            output("result", 2 + 2)
+        else:
+            output("result", 3 + 3)
 
+    # Classical optimization will keep the self-inverse Hadamard gates.
     package = main.with_opt_level(OptimizationLevel.Classical).compile()
 
 The available levels are:
@@ -47,7 +56,7 @@ It disables optional optimizations on the program.
 
 .. code-block:: python
 
-    emulator = main.with_minimal_opt().emulator(n_qubits=0)
+    emulator = main.with_minimal_opt().emulator(n_qubits=1)
 
 
 Running custom passes
--- a/guppylang/src/guppylang/optimizer.py
+++ b/guppylang/src/guppylang/optimizer.py
@@ -139,7 +139,7 @@ class OptimizationLevel(Enum):
     """
 
     def passes(self) -> list[ComposablePass]:
-        """Return the HUGR passes that implement this optimization level."""
+        """Return the list of HUGR passes ran by this optimization level."""
         match self:
             case OptimizationLevel.Default | OptimizationLevel.Classical:
                 from tket import passes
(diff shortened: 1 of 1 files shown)


guppylang/src/guppylang/optimizer.py
```diff
@@ -1,4 +1,84 @@
...
+* :py:attr:`OptimizationLevel.Default` applies Guppy's standard optimization
+  level. This may include both classical and quantum optimizations that do
+  not alter the program's gateset. Calling ``main.compile()`` or
+  ``main.emulator(...)`` directly uses this level.
+* :py:attr:`OptimizationLevel.Classical` restricts optimization to classical
+  operations. The program will execute the same quantum operations as the original
+  source, but may have a simplified control flow structure.
+* :py:attr:`OptimizationLevel.Minimal` applies only structural rewrites needed
+  to produce executable output. This is useful for low-level program analysis or
+  when more control over the optimization passes is desired.
```
aborgna-q, line 50: @acl-cqc I expanded a bit the definitions for these. Do you agree with them?

guppylang/src/guppylang/optimizer.py
```diff
@@ -1,4 +1,84 @@
...
+
+    @guppy
+    def main() -> None:
+        _x = 2 + 2
+
+    package = main.with_opt_level(OptimizationLevel.Classical).compile()
+
+The available levels are:
+
+* :py:attr:`OptimizationLevel.Default` applies Guppy's standard optimization
```
CalMacCQ, line 41: I think a certain kind of reader is likely to ask "What specific quantum optimizations are you doing to my program?!". I think we should explicitly say that we do no quantum optimization at the moment and then change this statement if/when we include the `RemoveRedundancies` pass. For now we could say... "we may include quantum optimization in the future".

As for the classical structural optimizations, I think it would make sense to crosslink to the `Normalize` pass in the tket-py docs.

guppylang/src/guppylang/optimizer.py
```diff
@@ -1,4 +1,84 @@
...
+Choosing an optimization level
+------------------------------
+
+Pass a member of :py:class:`OptimizationLevel` to ``with_opt_level()``:
+
+.. code-block:: python
+
+    from guppylang import OptimizationLevel, guppy
+
+    @guppy
```
CalMacCQ: Could we have a slightly less trivial `main` function? Maybe one that involves classical and quantum ops.

guppylang/src/guppylang/optimizer.py
```diff
@@ -1,4 +1,84 @@
...
+Running custom passes
+---------------------
+
+Use :py:meth:`OptimizerInstance.with_optimization` to append any HUGR
+``ComposablePass`` to an optimization pipeline. For example, the following
+starts with minimal optimization and then runs tket's function-inlining pass:
+
+.. code-block:: python
+
+    from tket import passes
```
CalMacCQ: How about this? I think it fits better to have the call take a single line.

```suggestion
    from tket.passes import InlineFunctions

    # Apply a tket pass to inline Guppy functions
    package = main.with_minimal_opt().with_optimization(InlineFunctions()).compile()    
```

CalMacCQ (commented):
Thanks @aborgna-q ! Just a few comments.

guppylang/src/guppylang/optimizer.py
aborgna-q: My worry with describing the internals is that it may limit us when fine-tuning the passes down the line.

If we promise an optimization level runs a specific set of passes, would adding a new optimization there be a breaking change?

Maybe we can add a description in the enum itself with the "current set of optimizations", but no promise on stability.

CalMacCQ: I don't think we'd be making a new promise by documenting what is already in the source code. 

We could perhaps explicitly promise to not convert to a new gateset in `main.compile()`.

My main concern is that benchamarkers etc need to know what quantum optimization is done by default so that they know if they need to turn it off.

Maybe linking to the source of `OptimizationLevel` would make sense?

I mean the sphinx [source] link rather than the github link. Maybe I sound paranoid, I just want to make sure the behaviour is transparent.

CalMacCQ (approved):

guppylang/src/guppylang/optimizer.py
aborgna-q: I added details in the enum definition, and linked it from the module docs.

image

guppylang/src/guppylang/optimizer.py
```diff
@@ -1,4 +1,98 @@
-"""Optimization configuration for Guppy compilation."""
+"""Optimization configuration for Guppy compilation.
+
+Guppy applies a predefined set of optimization passes when compiling a program.
```
CalMacCQ: "By using the TKET compiler"?

cqc-alec (approved):

PR #2124, opened 2026-07-31
docs: fix emulator argument formatting

ss2165:
## Summary

Correct the indentation of wrapped `Args:` descriptions on the emulator API so Sphinx/Napoleon renders each argument as one entry.

## Root cause

Wrapped description lines were aligned with argument names, which Napoleon interpreted as separate parameter entries.

## Validation

- `uv run ruff format --check guppylang/src/guppylang/defs.py`
- `uv run ruff check guppylang/src/guppylang/defs.py`
- `just build-docs`
- Confirmed Napoleon emits exactly four parameter fields for `emulator()`


CalMacCQ (approved):
Yikes, thanks for the fix.

Introduced napoleon recently and didn't notice this.

cqc-alec:
Do you mean to add the two files in the `.github` directory?

cqc-alec (approved):

PR #2126, opened 2026-08-03
feat: Add optional `debug_mode` argument to `compile` and `emulator`

### LINKED ISSUE #1607: Allow specifying whether debug info should be attached as `compile` argument

tatiana-s:
Closes #1607

### DIFF
--- a/guppylang-internals/src/guppylang_internals/metadata/common.py
+++ b/guppylang-internals/src/guppylang_internals/metadata/common.py
@@ -5,6 +5,7 @@ from hugr.debug_info import DebugRecord
 from hugr.metadata import HugrDebugInfo, Metadata, NodeMetadata
 from hugr.utils import JsonType
 
+from guppylang_internals.debug_mode import debug_mode_enabled
 from guppylang_internals.diagnostic import Fatal
 from guppylang_internals.error import GuppyError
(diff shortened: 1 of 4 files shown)


guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -100,6 +101,8 @@ def add_metadata(
     if metadata is not None:
         metadata_dict = metadata.as_dict()
         for key in metadata_dict:
+            if key == HugrDebugInfo.KEY and not debug_mode_enabled():
```
tatiana-s, line 104: Needed as `FunctionMetadata` (unintentionally?) persists between compilation calls

tatiana-s:
This PR also assumes that we still need the option to set the mode with `_DEBUG_MODE_ENABLED = os.getenv("GUPPYLANG_FORCE_DEBUGINFO") == "1"` (if that is not the case let me know @cgh-qtnm)

qartik (approved):

cgh-qtnm:
> This PR also assumes that we still need the option to set the mode with `_DEBUG_MODE_ENABLED = os.getenv("GUPPYLANG_FORCE_DEBUGINFO") == "1"` (if that is not the case let me know @cgh-qtnm)

That's correct

PR #2130, opened 2026-08-03
feat: Add RemoveRedundancies pass to default compilation

aborgna-q:
Re-enable the RemoveRedundancies pass on the default guppy program compilations.

Originally enabled in #2063 but reverted in #2071 due to a tket bug.

Requires https://github.com/Quantinuum/tket2/pull/1894 to be released in `tket-py 0.15.6`.


CalMacCQ:
I think we should also update the module level documentation as quantum optimization is now being applied here right?

tatiana-s (approved):
Approving as I assume this has been discussed, but are we okay with that performance decrease in the benchmarks? @aborgna-q

aborgna-q:
Yes, we'll be doing some performance optimizations that should hopefully improve these timings.

PR #2137, opened 2026-08-03
fix: address missing API docs members from the std.collections module.

### LINKED ISSUE #179: The guppylang-v1-rc1 API docs are incomplete for the `guppylang.std.collections` module
The structs `Queue`, `PriorityQueue` and `Stack` are missing from the new v1rc1 API docs here. ~The module level docstring is also incorrect~. 

Image

Seems like a regression in how the API docs are generated becuase the docs did show up in 0.21.16

Image

The module level docstring looks incorrect here too though.

CalMacCQ:
closes https://github.com/Quantinuum/guppy-docs/issues/179

required a fix to the dummy decorator which we use for sphinx.

Disclosure: I used chatgpt to help me pin down the source of this omission and then built the docs using this commit ref to ensure the fix worked.

Screenshot 2026-08-03 at 11 59 55

### DIFF
--- a/guppylang-internals/src/guppylang_internals/dummy_decorator.py
+++ b/guppylang-internals/src/guppylang_internals/dummy_decorator.py
@@ -16,7 +16,9 @@ class _DummyGuppy:
         return f
 
     def struct(self, *args: Any, **kwargs: Any) -> Any:
-        return self
+        if args:
+            return args[0]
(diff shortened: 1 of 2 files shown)


aborgna-q (approved):
Conditional approval on the doc checks passing

PR #2140, opened 2026-08-03
fix: Make most builtin functions + struct constructors daggerable

croyzor:
Explicitly mark stdlib functions as daggerable so that they can be used in a daggerable context, like a function declared `@guppy(daggerable=True)`

### DIFF
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -32,6 +32,7 @@ from guppylang_internals.std._internal.compiler.frozenarray import (
 )
 from guppylang_internals.tys import Effect
 from guppylang_internals.tys.builtin import array_type_def, frozenarray_type_def
+from guppylang_internals.tys.ty import UnitaryFlags
 
 from guppylang import guppy
 from guppylang.std.err import Result, err, ok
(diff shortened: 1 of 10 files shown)


tests/integration/test_modifier.py
```diff
@@ -587,3 +587,23 @@ def compile_to_sig(guppy_func: GuppyFunctionDefinition) -> str:
...
+
+    @guppy.comptime(daggerable=True)
+    def test_arr() -> array[int, n]:
+        return array(test(i) for i in range(n))
+
+    @guppy
+    def main() -> None:
+        test_arr[3]()
+
+    validate(main.compile())
```
nicolaassolini-qntm, line 621: I'm wondering if there is a way to have a more comprehensive test here, I would add at least something with array operations (e.g. take, put, ...). If possible also with arrays of qubits (if the linearity checker allows that).

guppylang/src/guppylang/std/array.py
```diff
@@ -219,7 +245,7 @@ def try_put(self: array[L, n], elem: L @ owned, idx: int) -> Result[None, L]:
         self.put(elem, idx)
         return ok(None)
 
-    @custom_function(ArrayDiscardAllUsedCompiler())
+    @custom_function(ArrayDiscardAllUsedCompiler(), unitary_flags=UnitaryFlags.Dagger)
```
nicolaassolini-qntm: not sure if makes sense, since you cannot discard stuff under a dagger context

nicolaassolini-qntm (commented):
Regarding `Either`, `Option` and `Error`. I'm not sure about having these objects inside a dagger context. Since we cannot return and branch under a dagger the only point of using these guys is to check and panic, I guess, and dagger is breaking panics ordering (sometimes also deleting some panics). I wuold avoid making daggarable operations related to these operators; I would prefer having a bad comptime error rather than a silent bug that deletes panics during compilation.

tests/integration/test_modifier.py
```diff
@@ -587,3 +587,35 @@ def compile_to_sig(guppy_func: GuppyFunctionDefinition) -> str:
         hashes.add(sig)
 
     assert len(hashes) == 1
+
+
+def test_arith(validate):
```
nicolaassolini-qntm:
```suggestion
def test_std(validate):
```

nicolaassolini-qntm (approved):
just a naming suggestion

PR #2166, opened 2026-08-04
fix: Prevent mutation of projections of moved values

### LINKED ISSUE #2117: [Bug]: A mutable struct instance can be implicitly copied without error
### Description

Whilst updating the docs I found some unexpected behaviour. 

Specifically, in Guppy v1 structs are supposed to be affine by default meaning that they cannot be implcitly copied.

However I seem to be able to run this code snippet using (guppylang v1rc2) without error

```python
from guppylang import guppy

@guppy.struct
class PauliString:
    xs: array[bool, 3]
    zs: array[bool, 3]

@guppy
def implicit_copy() -> None:
    my_pauli = PauliString(array(True, False, True), array(False, True, False))

    my_new_pauli = my_pauli

    # Trying to access my_pauli after a move should give an error but doesn't.
    my_pauli.zs = array(False, False, False)

implicit_copy.check() # No copy violation error given upon type check
```

### Steps to Reproduce

Run the snippet above using guppylang v1rc2.

### Expected Behaviour

A copy violation error

### Actual Behaviour

The snippet type checks with no error.

### Component

_No response_

### Environment

```markdown
- guppy version v1rc2
```

### Additional Context

_No response_

mark-koch:
Fixes #2117

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
@@ -145,7 +145,7 @@ class Use(NamedTuple):
 
 
 class Mutation(NamedTuple):
-    """Records data associated wih the mutation of a place."""
+    """Records data associated with the mutation of a place."""
 
     #: The AST node corresponding to mutation
     node: AstNode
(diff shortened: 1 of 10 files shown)


guppylang-internals/src/guppylang_internals/checker/linearity_checker.py
```diff
@@ -144,6 +144,21 @@ class Use(NamedTuple):
     origin_place: Place
 
 
+class Mutation(NamedTuple):
+    """Records data associated wih the mutation of a place."""
```
hsemenenko:
```suggestion
    """Records data associated with the mutation of a place."""
```

hsemenenko (approved):

PR #2180, opened 2026-08-06
feat: add target platform configuration for optimizer and emulator

### LINKED ISSUE #2042: [Feature]: Specify the target platform for optimization when compiling a program
### Problem Statement

See #2017.

We would add a `OptimizerInstance.with_target_platform` option to be chained before the compilation, which would influence the passes ran on the hugr.

This needs some discussion on whether it's a good idea or not.

nicolaassolini-qntm:
closes #2042

### DIFF
--- a/guppylang/src/guppylang/defs.py
+++ b/guppylang/src/guppylang/defs.py
@@ -398,8 +398,8 @@ class GuppyFunctionDefinition(GuppyDefinition, GuppyCompilableProgram, Generic[P
         return self.with_opt_level(OptimizationLevel.Minimal)
 
     def with_target_platform(self, platform: Platform) -> "OptimizerInstance[P, Out]":
-        """Configure the target platform used when compiling this function."""
-        return OptimizerInstance(self, target_platform=platform)
+        """Configure the target platform used when building an emulator for this function."""
+        return self.with_opt_level(OptimizationLevel.Default).with_target_platform(platform)
 
     def compile(self, debug_mode: bool = False) -> Package:
         """
--- a/guppylang/src/guppylang/defs.py
+++ b/guppylang/src/guppylang/defs.py
@@ -289,8 +289,11 @@ class GuppyFunctionDefinition(GuppyDefinition, GuppyCompilableProgram, Generic[P
         Returns:
             An `EmulatorInstance` that can be used to run the function in an emulator.
         """
-        return self.with_opt_level(OptimizationLevel.Default).emulator(
-            n_qubits, builder, libs, platform, debug_mode
+        return (
+            self
+            .with_opt_level(OptimizationLevel.Default)
+            .with_target_platform(platform)
+            .emulator(n_qubits, builder, libs, debug_mode=debug_mode)
         )
 
     def _emulator(
@@ -398,8 +401,10 @@ class GuppyFunctionDefinition(GuppyDefinition, GuppyCompilableProgram, Generic[P
         return self.with_opt_level(OptimizationLevel.Minimal)
 
     def with_target_platform(self, platform: Platform) -> "OptimizerInstance[P, Out]":
-        """Configure the target platform used when building an emulator for this function."""
-        return self.with_opt_level(OptimizationLevel.Default).with_target_platform(platform)
+        """Configure the target platform used when building the emulator."""
+        return self.with_opt_level(OptimizationLevel.Default).with_target_platform(
+            platform
+        )
 
     def compile(self, debug_mode: bool = False) -> Package:
         """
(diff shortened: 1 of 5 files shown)


guppylang/src/guppylang/optimizer.py
```diff
@@ -207,22 +207,37 @@ class OptimizerInstance(Generic[P, Out]):
...
     definition: GuppyFunctionDefinition[P, Out]
     passes: list[ComposablePass] = field(default_factory=list)
+    target_platform: Platform = field(default="helios")
 
     def with_optimization(
         self, optimization: ComposablePass
     ) -> OptimizerInstance[P, Out]:
         """Add an additional optimization pass to run while compiling the program."""
-        return OptimizerInstance(self.definition, [*self.passes, optimization])
+        return OptimizerInstance(
```
aborgna-q: `.replace`
https://docs.python.org/3/library/dataclasses.html#dataclasses.replace

guppylang/src/guppylang/optimizer.py
```diff
@@ -207,22 +207,37 @@ class OptimizerInstance(Generic[P, Out]):
 
     definition: GuppyFunctionDefinition[P, Out]
     passes: list[ComposablePass] = field(default_factory=list)
+    target_platform: Platform = field(default="helios")
```
aborgna-q: This should rather be `Platform | None`, with `None` as default.

That way we can keep a generic-target `.compile` by default, unless the user specifically sets it to `helios`.

`.emulator` should specify that it default to `helios` if unset.

nicolaassolini-qntm: Makes sense.

> .emulator should specify that it default to helios if unset.

As it was

guppylang/src/guppylang/optimizer.py
```diff
@@ -94,12 +94,31 @@ def main() -> None:
...
+
+.. code-block:: python
+
+    emulator = (
+        main.with_opt_level(OptimizationLevel.Classical)
+        .with_target_platform("sol")
+        .emulator(n_qubits=1)
+    )
+
+Passing ``platform`` directly to ``emulator()`` overrides the configured target.
```
nicolaassolini-qntm, line 114: Not sure if leaving this line, since we should (I suppose) discourage using the platform in .emulator()

nicolaassolini-qntm:
I also requested the review from the robot 🤖

nicolaassolini-qntm:
> This PR contains breaking changes to the public Python API. Please deprecate the old API instead (if possible), or mark the PR with a `!` to indicate a breaking change.
> 
> Breaking changes summary

I think this is not right; we are changing the default, but the semantics in case of default is preserved

guppylang/src/guppylang/defs.py
```diff
@@ -397,6 +397,10 @@ def with_minimal_opt(self) -> "OptimizerInstance[P, Out]":
         """
         return self.with_opt_level(OptimizationLevel.Minimal)
 
+    def with_target_platform(self, platform: Platform) -> "OptimizerInstance[P, Out]":
+        """Configure the target platform used when compiling this function."""
+        return OptimizerInstance(self, target_platform=platform)
```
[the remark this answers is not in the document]
nicolaassolini-qntm: copilot suggested this, which makes sense. But at the same time pose an issue. If I do main.with_target_platform I cannot set the optimizazion level anymore, do we need a new class to be able to do that?

guppylang/src/guppylang/optimizer.py
aborgna-q: Yeah, I'd remove it.

guppylang/src/guppylang/defs.py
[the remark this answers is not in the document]
nicolaassolini-qntm: Currently, we are doing as Copilot suggested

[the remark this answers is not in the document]
aborgna-q: Yeah, since now we are building an `OptimizerInstance` without setting an optimization level, we'll need to add `with_opt_level`/`with_minimal_opt` to the class too.

Also, `OptimizerInstance.passes` should set the default list on its `default_factory` definition.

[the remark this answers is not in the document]
nicolaassolini-qntm: > Also, OptimizerInstance.passes should set the default list on its default_factory definition.

Do you mean that the default value is Default optimisation instead of an empty list?

nicolaassolini-qntm:
@aborgna-q I did some changes. Also, I'm wondering if we should add a new Protocol for the common method between OptimizerInstance and GuppyFunctionDefinition (or change the GuppyCompilableProgram

aborgna-q (approved):
LGTM

A common protocol (with the default `with_minimal_opt` impl) wouldn't be a bad idea.

Issue #1792, opened 2026-06-02
[Bug]: Bad span for missing return diagnostic


mark-koch:
### Description

```python
@guppy
def main(b: bool) -> int:
    while (b := b):
        return 0
```

produces the error

```
Error: Expected return statement (at ...)
   | 
94 | @guppy
95 | def main(b: bool) -> int:
96 |     while (b := b):
   |            ^^^^^^ Expected return statement
```

### Expected Behaviour

The error span should highlight the entire while statement, not just the condition:

```
   | 
94 | @guppy
95 | def main(b: bool) -> int:
96 |     while (b := b):
   |     ^^^^^^^^^^^^^^^
97 |         return 0
   | ^^^^^^^^^^^^^^^^ Expected return statement
```

Also, it would be nice to get the usual help message

```
Note:
   | 
95 | def main(b: bool) -> int:
96 |     while (b := b):
   |            ------ Consider adding a return statement if this expression is
   |                   `False`
```

### Environment

```markdown

```

PR #2181, opened 2026-08-06
fix: Bad span for missing return diagnostic with walrus loop condition

thewill-i-am:
#1792

Fixes the bad span for the "Expected return statement" diagnostic when a `while` loop uses a walrus condition. Given:

```python
@guppy
def main(b: bool) -> int:
    while (b := b):
        return 0
```

### Before
```
   |     while (b := b):
   |            ^^^^^^ Expected return statement
```
The error highlighted only the condition `b := b` instead of the entire `while` statement, and the "Consider adding a return statement if this expression is `False`" help note was missing entirely.

### After
```
   |     while (b := b):
   |     ^^^^^^^^^^^^^^^
   |         return 0
   | ^^^^^^^^^^^^^^^^ Expected return statement

Note:
   |     while (b := b):
   |            ------ Consider adding a return statement if this expression is
   |                   `False`
```

## Root cause

`ExprBuilder.visit_NamedExpr` desugars a walrus `(b := b)` into a synthetic `ast.Assign` appended to the current basic block (the loop header `head_bb`), returning the bare `Name('b')` as the `branch_pred`. So `head_bb` ends up with **both** `statements = [Assign(b := b)]` and `branch_pred = Name('b')`.

`find_missing_return_point` walks ancestors looking for the nearest statement-bearing block; on finding one it uses `statements[-1]` as the error span and only looks for a branch predicate in ancestors *above* that block (deliberately skipping the block itself via `itertools.islice(..., 1, None)`). It never noticed the statement-bearing block was itself the branch header — the predicate was right there — and returned `(Assign, None)`, which (a) made the span cover only `b := b` and (b) suppressed the `MissingBranch` help note.

For non-walrus loops (`while i < n:`), `head_bb.statements` stays empty, so the function falls through to the second code path that returns `(None, (branch_pred, 0))` — which is why those cases already worked.

## Fix

Added a guard in `find_missing_return_point` (`guppylang-internals/src/guppylang_internals/cfg/builder.py`) right after computing `final_statement`. When the statement-bearing ancestor **also** owns a `branch_pred` and the last statement is the desugared walrus `Assign` whose single `Name` target matches the `branch_pred` (same `id`, same `lineno`), it returns `(None, (final_statement, 0))` — routing through the same path as non-walrus loops. The caller then uses `nodes[-1]` (the whole `while` statement) as the error span and attaches `MissingBranch(Assign, False)` pointing at the walrus.

The guard requires five conditions simultaneously, so it only matches the walrus-in-header desugar and doesn't affect any other case — all existing `no_return*` golden `.err` snapshots are unchanged.

The function's second return-element type was widened from `tuple[ast.expr, int]` to `tuple[ast.AST, int]` since the walrus `Assign` (a statement, not an expression) can now flow through; `MissingBranch` accepts any `ToSpan = ast.AST | Span`.

## Test

Added a regression test `tests/error/misc_errors/no_return_loop_walrus.{py,err}`, auto-picked-up by the parametrised `tests/error/test_misc_errors.py` runner. The `.err` golden was generated with `--snapshot-update`.

## Verification

- `uv run ruff check guppylang guppylang-internals tests` — All checks passed
- `uv run ruff format --check guppylang guppylang-internals tests` — 315 files already formatted
- `uv run mypy guppylang guppylang-internals` — no issues found in 169 source files
- `uv run pytest -n auto --ignore=tests/integration/test_notebooks.py` — **1846 passed, 25 skipped, 8 xfailed**
- The 15 `tests/integration/test_notebooks.py` failures are pre-existing environment issues (`ModuleNotFoundError: No module named 'guppylang'` in the Jupyter kernel) unrelated to this change.

## Files changed

- `guppylang-internals/src/guppylang_internals/cfg/builder.py` — fix + type widening
- `tests/error/misc_errors/no_return_loop_walrus.py` — new regression test
- `tests/error/misc_errors/no_return_loop_walrus.err` — new golden snapshot

### DIFF
--- a/guppylang-internals/src/guppylang_internals/cfg/builder.py
+++ b/guppylang-internals/src/guppylang_internals/cfg/builder.py
@@ -822,7 +822,7 @@ def make_assign(lhs: list[ast.AST], value: ast.expr) -> ast.Assign:
 
 def find_missing_return_point(
     final_bb: BB, cfg: CFG
-) -> tuple[BBStatement | None, tuple[ast.expr, int] | None]:
+) -> tuple[BBStatement | None, tuple[ast.AST, int] | None]:
     """Finds the last statement from the final BB or its predecessors, and the branch
     condition. It walks up the ancestors of the given final basic block to locate the
     nearest block containing statements, and attempts to find the most relevant branch
@@ -853,6 +853,25 @@ def find_missing_return_point(
             # we can stop the search and look for the condition in the branch
             # that leads to final_bb.
             final_statement = fbb_ancestor.statements[-1]
+            # A statement-bearing ancestor that also owns a branch predicate
+            # is a loop/if header whose condition got desugared into a
+            # side-effecting statement (e.g. a walrus `(b := b)` whose
+            # `Assign` ends up in the header BB next to the bare-`b`
+            # `branch_pred`). The last statement alone is a poor error span
+            # (it only covers the condition), so fall back to the loop-header
+            # path: the caller uses the whole enclosing statement
+            # (`nodes[-1]`) as the span and we point the help note at the
+            # desugared condition with `truth_value=False`.
+            if (
+                fbb_ancestor.branch_pred is not None
+                and isinstance(final_statement, ast.Assign)
+                and len(final_statement.targets) == 1
+                and isinstance(final_statement.targets[0], ast.Name)
+                and isinstance(fbb_ancestor.branch_pred, ast.Name)
+                and final_statement.targets[0].id == fbb_ancestor.branch_pred.id
+                and final_statement.lineno == fbb_ancestor.branch_pred.lineno
+            ):
+                return None, (final_statement, 0)
             # To have a better error message, we also look for the condition
             # of the branch without return.
             # However, there may be nested branches without returns.
--- a/guppylang-internals/src/guppylang_internals/cfg/builder.py
+++ b/guppylang-internals/src/guppylang_internals/cfg/builder.py
@@ -844,6 +844,22 @@ def find_missing_return_point(
     """
     final_statement = None
 
+    def branch_index(branch_bb: BB, target_bb: BB) -> int | None:
+        """Finds the unique branch leading to a target without looping back."""
+        ancestors = {target_bb}
+        worklist = [target_bb]
+        while worklist:
+            for predecessor in worklist.pop().predecessors:
+                if predecessor is not branch_bb and predecessor not in ancestors:
+                    ancestors.add(predecessor)
+                    worklist.append(predecessor)
+        branches = [
+            i
+            for i, successor in enumerate(branch_bb.successors)
+            if successor in ancestors
+        ]
+        return branches[0] if len(branches) == 1 else None
+
     # walk up the ancestors in the CFG
     # to find the nearest block with statements
     # (ancestors contain the final_bb as first element)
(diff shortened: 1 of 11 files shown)


croyzor:
Hi @thewill-i-am, thanks for the contribution! Sorry for the slow reply, it took me a while to get my head around this part of the codebase. This is a nice error message improvement, but I think it's not sufficient in its current form. Thinking about what other forms of surface syntax would be affected by this, both
```python
if b:
    return
```
and
```python
if True if b else False:
    return
```
get the improved error message, but it falls over when we have 
```python
if not b:
    return
```
or
```python
if False if b else True:
    return
```
because the code is always suggesting to add a return if `b` is False, which isn't always appropriate.

I'm happy for you to consider this and submit an updated version of the PR if you're still interested 🙂

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -853,6 +853,25 @@ def find_missing_return_point(
...
+            # is a loop/if header whose condition got desugared into a
+            # side-effecting statement (e.g. a walrus `(b := b)` whose
+            # `Assign` ends up in the header BB next to the bare-`b`
+            # `branch_pred`). The last statement alone is a poor error span
+            # (it only covers the condition), so fall back to the loop-header
+            # path: the caller uses the whole enclosing statement
+            # (`nodes[-1]`) as the span and we point the help note at the
+            # desugared condition with `truth_value=False`.
+            if (
+                fbb_ancestor.branch_pred is not None
```
croyzor: nitpick: this is redundant because we also check `isinstance(fbb_ancestor.branch_pred, ast.Name)`

guppylang-internals/src/guppylang_internals/cfg/builder.py
```diff
@@ -853,6 +853,25 @@ def find_missing_return_point(
...
+            # A statement-bearing ancestor that also owns a branch predicate
+            # is a loop/if header whose condition got desugared into a
+            # side-effecting statement (e.g. a walrus `(b := b)` whose
+            # `Assign` ends up in the header BB next to the bare-`b`
+            # `branch_pred`). The last statement alone is a poor error span
+            # (it only covers the condition), so fall back to the loop-header
+            # path: the caller uses the whole enclosing statement
+            # (`nodes[-1]`) as the span and we point the help note at the
+            # desugared condition with `truth_value=False`.
+            if (
```
croyzor, line 874: I think a slightly more flexible condition here would be
```python
isinstance(fbb_ancestor.branch_pred, ast.Name)
and isinstance(final_statement, ast.Assign)
and fbb_ancestor.branch_pred.id
in [
    tgt.id
    for tgt in final_statement.targets
    if isinstance(tgt, ast.Name)
]
```
because we could then catch something like
```python
a, b = foo()
if a:
    ...
```

guppylang-internals/src/guppylang_internals/cfg/builder.py
thewill-i-am: @croyzor  Hi, Thanks for the recommendation; I'll fix it sometime today. I did this a while back and don't quite remember what it was doing, haha.

guppylang-internals/src/guppylang_internals/cfg/builder.py
thewill-i-am: @croyzor 

Thanks for the suggestion! I broadened the condition to support assignments with multiple or nested targets.
I used ast.walk over each assignment target and matched only ast.Name nodes with an ast.Store context. Python represents a, b = foo() as a single ast.Tuple target, so this also handles the tuple-unpacking example while avoiding false matches such as obj in obj.field = value.
I also added a regression test for:
```
a, b = values
if a:
    return 0
```
The diagnostic now highlights the if statement and points the help note at a.

croyzor (approved):
Nice work, thank you! :rocket:

thewill-i-am:
> Nice work, thank you! 🚀

@croyzor @aborgna-q @CalMacCQ  Thanks ! Hey, I'd like to keep collaborating; is there a (ticket or issue) you'd like me to work on?

PR #2173, opened 2026-08-05
chore: update changelog and package versions to latest release

tatiana-s:

PR #2185, opened 2026-08-10
chore: Sync versions in init file

maximilianruesch:
PR #2173 forgot to update these. Arguably we should have a check for this (and it should be somewhat easy to write, since we can programmatically extract from the toml and the init file).

### DIFF
--- a/guppylang-internals/src/guppylang_internals/__init__.py
+++ b/guppylang-internals/src/guppylang_internals/__init__.py
@@ -1,3 +1,3 @@
 # This is kept in sync with the project version by the release workflow
 # (.github/workflows/release-pr.yml).
-__version__ = "1.0.0-a8"
+__version__ = "1.0.1"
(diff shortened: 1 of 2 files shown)


aborgna-q:
> Arguably we should have a check for this

Could we instead have an automated PR to update versions / CHANGELOG in `main` when there is a release from a branch? (keeping the latest, in case `main` is already ahead).

aborgna-q (approved):

maximilianruesch:
> Could we instead have an automated PR to update versions / CHANGELOG in main when there is a release from a branch? (keeping the latest, in case main is already ahead).

An automation is a bit more complicated, since if main already released the next alpha (which should be done from main), the changelog needs to be inserted before that section, so it is not a trivial "push to the top". I will create an issue for this enhancement, it seems to be slightly more complicated.

EDIT: Created #2186

PR #2188, opened 2026-08-10
feat: Add error message for missing self args

### LINKED ISSUE #2103: [Feature]: Improve error message for missing self args
### Problem Statement

dummy example
```python
from guppylang import guppy
from guppylang.std.builtins import nat, qubit, array
from guppylang.std.quantum import discard_array

@guppy.struct(frozen=True)
class StatePrep[N: nat]:
    @guppy
    def prep() -> array[qubit, N]:
        return array(qubit() for _ in range(N))

@guppy
def main() -> None:
    sp = StatePrep[3]()
    discard_array(sp.prep())


main.compile()
```
fails with 
```
Error: Variable not defined (at /Users/craig.roy/work/guppy/test_thing.py:8:31)
  | 
6 | class StatePrep[N: nat]:
7 |     @guppy
8 |     def prep() -> array[qubit, N]:
  |                                ^ `N` is not defined

```
when the real issue is that `prep` is missing a `self` arg as its first argument!

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

_No response_

### Additional Context

_No response_

SaashaJoshi:
Adds error message for missing `self` argument in methods. 
Closes #2103

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/func_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/func_checker.py
@@ -72,6 +72,12 @@ class MissingArgAnnotationError(Error):
 
 
 @dataclass(frozen=True)
+class MissingSelfError(Error):
+    title: ClassVar[str] = "Missing self argument"
+    span_label: ClassVar[str] = "Method requires a self argument"
+
+
+@dataclass(frozen=True)
 class RecursiveSelfError(Error):
     title: ClassVar[str] = "Recursive self annotation"
     span_label: ClassVar[str] = (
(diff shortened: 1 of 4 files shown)


qartik:
> Note: FYI, still working on passing the tests.

Thanks, I marked the PR as draft, please mark ready when you're done.

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -335,6 +341,18 @@ def check_signature(
     inputs = []
     ctx = TypeParsingCtx(globals, param_var_mapping, allow_free_vars=True)
     has_parent = def_id is not None and def_id in DEF_STORE.type_member_parents
+    not_wasm_func = def_id not in DEF_STORE.wasm_functions
+
+    # Check if method doesn't have any arguments.
+    # Excluding WASM functions where first argument must be a WASM module.
+    if (
+        has_parent
+        and not_wasm_func
```
croyzor:
```suggestion
```

guppylang-internals/src/guppylang_internals/checker/func_checker.py
```diff
@@ -335,6 +341,18 @@ def check_signature(
     inputs = []
     ctx = TypeParsingCtx(globals, param_var_mapping, allow_free_vars=True)
     has_parent = def_id is not None and def_id in DEF_STORE.type_member_parents
+    not_wasm_func = def_id not in DEF_STORE.wasm_functions
+
+    # Check if method doesn't have any arguments.
+    # Excluding WASM functions where first argument must be a WASM module.
```
croyzor:
```suggestion
```
The WASM module argument to WASM functions is just another `self` arg, so there's no need to make special cases for it. 

Run the tests with `--snapshot-update` to update the old error message for the wasm test to your new one 😄

croyzor (changes requested):

guppylang-internals/src/guppylang_internals/checker/func_checker.py
SaashaJoshi: This makes more sense.

croyzor (approved):
Nice, thank you for the contribution! 😄

PR #2191, opened 2026-08-11
fix: chore: Update `guppylang/pyproject.toml` to use the proper `README.md`

nicolaassolini-qntm:
Hatchling (now resolved to 1.32.0) enforces that the readme path must live inside the project directory. Since in guppylang we point to the repo-root README  validation raises: `ValueError: Readme path must be within the project directory: ../README.md`

To fix this issue, we changed the README in `guppylang/pyproject.toml` to use the `guppylang` root one. We also updated the info in that README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ def teleport(src: qubit @ owned, tgt: qubit) -> None:
     if measure(tmp):
         x(tgt)
 
+
 teleport.check()
 ```
 
--- a/guppylang/README.md
+++ b/guppylang/README.md
@@ -48,7 +48,7 @@ teleport.check()
 
 📒 [Example notebooks][examples]
 
-[examples]: ../examples/
+[examples]: https://github.com/quantinuum/guppylang/tree/main/examples
 [guide]: https://docs.quantinuum.com/guppy/language_guide/language_guide_index.html
 [website]: https://guppylang.org
 
--- a/guppylang/README.md
+++ b/guppylang/README.md
@@ -1,3 +1,6 @@
+![](https://raw.githubusercontent.com/quantinuum/guppylang/refs/heads/main/assets/guppy_logo.svg)
+
+
 # Guppy
 
 [![pypi][]](https://pypi.org/project/guppylang/)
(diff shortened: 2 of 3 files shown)


nicolaassolini-qntm:
Using `hatch-fancy-pypi-readme` is a Claude suggestion. I also saw that we can avoid it by simply making the readme dynamic  (declaring `dynamic = ["readme"]` in the .toml) and then creating a Python script named package/hatch_build.py right next to pyproject.toml containing:

```python
from hatchling.metadata.plugin.interface import MetadataHookInterface
from pathlib import Path

class ReadmeHook(MetadataHookInterface):
    def update(self, metadata):
        # Navigates up from the current script to find the root README
        readme_path = Path(__file__).parent.parent / "README.md"
        
        if readme_path.exists():
            metadata["readme"] = {
                "text": readme_path.read_text(encoding="utf-8"),
                "content-type": "text/markdown"
            }

# Hatchling requires this entry point registration inside the file
def get_metadata_hook():
    return ReadmeHook
```

aborgna-q (changes requested):
Why are we using `../README` instead of `guppylang/README.md`?

In general, is nice to have a root README for the repository, with development instructions, and a pypi README for users of the library.

Also `dynamic` pyprojects make everything more complicated.

nicolaassolini-qntm:
> Why are we using `../README` instead of `guppylang/README.md`?

I have no idea; the two ReadMe files are basically the same (the only difference is that the root one has the logo)

> Also dynamic pyprojects make everything more complicated.

As far as I know, it is the only way of making hatchling working now

I can change the README so we can use the guppylang root one on PyPi

aborgna-q (approved):

README.md
```diff
@@ -35,6 +35,7 @@ def teleport(src: qubit @ owned, tgt: qubit) -> None:
     if measure(tmp):
         x(tgt)
 
+
```
aborgna-q, line 38: ?


```suggestion
```

guppylang/README.md
```diff
@@ -33,12 +36,21 @@ def teleport(src: qubit @ owned, tgt: qubit) -> None:
...
+## Documentation
+
+🌐 [Guppy website][website]
+
+📖 [Language guide][guide]
+
+📒 [Example notebooks][examples]
 
-[examples]: https://github.com/quantinuum/guppylang/tree/main/examples
+[examples]: ../examples/
```
aborgna-q: This link won't work on pypi

README.md
nicolaassolini-qntm: ruff

PR #2195, opened 2026-08-13
test: Pull out error test case collection to common utility

maximilianruesch:
Pulls out a very common pattern for error tests to improve maintenance burden and reduce code duplication. I got too annoyed with this apparently.

Note that no actual test cases have been modified.

### DIFF
--- a/tests/error/util.py
+++ b/tests/error/util.py
@@ -23,6 +23,8 @@ def collect_error_test_cases(
     relative to the parent directory of this file.
 
     Returns a list of file paths that can be processed by pytest.mark.parametrize."""
+    # __file__ works since the current file is in the same directory as all the error
+    # test cases. If it is moved, this definition needs to be updated.
     path = pathlib.Path(__file__).parent.resolve() / directory
     files = [
         x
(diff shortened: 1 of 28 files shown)


tests/error/util.py
```diff
@@ -13,6 +14,28 @@
...
+def collect_error_test_cases(
+    directory: str,
+    *,
+    exclude: Callable[[pathlib.Path], bool] = lambda _: False,
+) -> list[str]:
+    """Collects all error test cases (*.py, but not __init__.py) in the given directory
+    relative to the parent directory of this file.
+
+    Returns a list of file paths that can be processed by pytest.mark.parametrize."""
+    path = pathlib.Path(__file__).parent.resolve() / directory
```
croyzor, line 28: does pytest.mark.parametrize mean special behaviour for the `__file__` variable? I don't really understand why this would work, so a comment would be good.

OR: I'm big python not-knower and don't know how `__file__` works beyond the context of a single-module program. Maybe it's obvious

maximilianruesch: No, this just happens to work because the `__file__` refers to the current `util.py` file, which happens to be in the same directory as all of the usage sites. I will add a comment.

croyzor (approved):
You're doing the lord's work 🙏

PR #2203, opened 2026-08-14
refactor: Remove redundant checks for function type unitary flags

### LINKED ISSUE #2196: [Refactor]: Remove useless unitary flags checks with function types
I'm pretty sure that https://github.com/Quantinuum/guppylang/blob/b47cf816cacc6ea852f3033eba6816e8e07173a7/guppylang-internals/src/guppylang_internals/checker/expr_checker.py#L1270 is not more used after https://github.com/Quantinuum/guppylang/issues/2095.

After #2095, ModifiableFunctionProtocolInst.check_implemented_by should be used before checking the unitary flags in `check_type_against`.

We should investigate to be sure that this is the case and then delete the redundant check

nicolaassolini-qntm:
closes #2196

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/errors/type_errors.py
+++ b/guppylang-internals/src/guppylang_internals/checker/errors/type_errors.py
@@ -41,26 +41,6 @@ class TypeMismatchError(Error):
 
 
 @dataclass(frozen=True)
-class UnitaryFlagMismatchError(Error):
-    title: ClassVar[str] = "Unitary flag mismatch"
-    span_label: ClassVar[str] = (
-        "Expected function with unitary flags: `{rendered_expected}`,"
-        " got: `{rendered_actual}`"
-    )
-
-    expected: UnitaryFlags
-    actual: UnitaryFlags
-
-    @property
-    def rendered_expected(self) -> str:
-        return self.expected.hint_rendering()
-
-    @property
-    def rendered_actual(self) -> str:
-        return self.actual.hint_rendering()
-
-
-@dataclass(frozen=True)
 class UnitaryFlagMismatchHint(Note):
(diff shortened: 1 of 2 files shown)


qartik (commented):
From Gemini 3.1 Pro:

> The logic and intent of this PR are completely sound!
> 
> Since PR #2095 turned modifiers into protocols, unitary flag mismatches are correctly handled by `ModifiableFunctionProtocolInst.check_implemented_by` (raising a protocol error with a `UnitaryFlagMismatchHint`). As a result, the removed check in `check_type_against` was essentially dead code, as user-provided `"Function[..., ...]"` annotations always parse to `NoFlags`, and `"Unitary[..., ...]"` parses to a protocol-bound parameter.
> 
> However, I found one minor cleanup that was missed:
> The `UnitaryFlagMismatchError` class in `guppylang-internals/src/guppylang_internals/checker/errors/type_errors.py` is now completely unused across the codebase. Since you removed its only usage and import in `expr_checker.py`, you should also delete the class definition to keep the error catalog clean. (`UnitaryFlagMismatchHint` is still used in `tys/builtin.py` and should be kept).
> 
> Great cleanup otherwise!

nicolaassolini-qntm:
> However, I found one minor cleanup that was missed:
The UnitaryFlagMismatchError class in guppylang-internals/src/guppylang_internals/checker/errors/type_errors.py is now completely unused across the codebase. Since you removed its only usage and import in expr_checker.py, you should also delete the class definition to keep the error catalog clean. (UnitaryFlagMismatchHint is still used in tys/builtin.py and should be kept).

The robot is right

> Great cleanup otherwise!

Thanks robot 🤖

qartik (approved):

Issue #2205, opened 2026-08-14
[Feature]: Support default method implementations in protocols


ss2165:
## Problem Statement

Protocols currently only support abstract `@guppy.require` members. Common derived behavior must be repeated by every concrete implementation, even when it can be expressed once in terms of required protocol methods.

This also prevents protocols from providing ownership-safe, unitary-compatible convenience operations over their required primitives.

## Proposed Solution

Allow ordinary `@guppy` methods with bodies in `@guppy.protocol` definitions as default method implementations.

- Keep `@guppy.require` members abstract.
- Prefer a compatible concrete method over a protocol default.
- Permit defaults to call required/default members through `self`.
- Preserve `Self`, generic parameters, ownership, and function modifiers.
- Diagnose ambiguity when multiple protocol bounds provide competing defaults.

## Component

Guppy Compiler

## Effort Estimate

Hard - Significant effort required

## Additional Context

Parent tracking issue: #1968

Example desired usage:

```python
@guppy.protocol
class Incrementable:
    @guppy.require
    def increment(self: Self @ owned, x: int) -> tuple[Self, int]: ...

    @guppy
    def increment_twice(self: Self @ owned, x: int) -> tuple[Self, int]:
        self, x = self.increment(x)
        return self.increment(x)
```

Acceptance coverage should include:

- required primitive plus derived default;
- concrete override precedence;
- generic `Self` and parameters;
- linear/owned defaults;
- unitary defaults under `control` and `dagger`;
- no default access outside a protocol bound;
- ambiguity across competing defaults;
- diagnostics for unsupported members, missing required members, and incompatible overrides.

PR #2206, opened 2026-08-14
feat(quantum): add abstract qubit protocol

ss2165:
## Summary

- add the public `AbstractQubit` protocol with descriptive (more verbose than stdlib functions), `Self`-typed quantum operations
- make the built-in `qubit` implement the protocol and retain the existing short-form `std.quantum` API as generic delegates
- generalize functional and array quantum operations, with integration coverage for a distinct protocol implementation
- ideally should follow: https://github.com/Quantinuum/guppylang/issues/2119

## Follow-up

Derived quantum operations such as Toffoli are currently required protocol members. Providing defaults for these operations is tracked by #2205; the protocol contains a matching TODO.

### DIFF
--- a/ruff.toml
+++ b/ruff.toml
@@ -1,5 +1,5 @@
 # See https://docs.astral.sh/ruff/rules/
-target-version = "py310"
+target-version = "py312"
 
 line-length = 88
 
--- a/tests/integration/test_quantum.py
+++ b/tests/integration/test_quantum.py
@@ -1,6 +1,6 @@
 """Various tests for the functions defined in `guppylang.prelude.quantum`."""
 
-from typing import no_type_check
+from typing import Self, no_type_check
 
 from guppylang.std.angles import angle
 
@@ -16,6 +16,7 @@ from guppylang.std.quantum import (
     measure_array,
     discard_array,
     Measurement,
+    AbstractQubit,
 )
 from guppylang.std.quantum.functional import (
     cx,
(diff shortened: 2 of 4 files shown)


ruff.toml
```diff
@@ -1,5 +1,5 @@
 # See https://docs.astral.sh/ruff/rules/
-target-version = "py310"
+target-version = "py312"
```
croyzor, line 2: this is outside the scope of the PR, you can make an exception for your test files at the bottom of ruff.toml

ss2165: yes sorry pushed local change accidentally

tests/integration/test_quantum.py
```diff
@@ -44,6 +45,118 @@
...
+        @guppy(unitary=True)
+        def rotate_y(self, theta: angle) -> None:
+            q.ry(self.inner, theta)
+
+        @guppy(unitary=True)
+        def controlled_rotate_z(self, target: Self, theta: angle) -> None:
+            q.crz(self.inner, target.inner, theta)
+
+        @guppy(unitary=True)
+        def toffoli(self, control2: Self, target: Self) -> None:
```
croyzor, line 122: toffoli would make more sense as a method that takes something implementing AbstractProtocol.

Generally, we need to think about what the minimal api surface for a qubit should be if we're going to add it to std. Something that's way too broad like seems like a big pain for abstracting logical qubits

tests/integration/test_quantum.py
```diff
@@ -44,6 +45,118 @@
...
+            q.toffoli(self.inner, control2.inner, target.inner)
+
+        @guppy(unitary=True)
+        def controlled_hadamard(self, target: Self) -> None:
```
croyzor, line 126: if we have unitary hadamard, we don't need this, we can write
```
with control(...):
    q.hadamard()

tests/integration/test_quantum.py
ss2165: yeah - hence why I'm unsure about this without defaults

ss2165: going extreme we could take tket1's approach: implement CX and an arbitrary single qubit gate and that's it (I guess also alloc/measure/discard)

croyzor: I would prefer that approach, it just makes things useless for codes that have a different set of native gates (e.g. if `CX` is costly but `CZ` is cheap). Maybe we shouldn't spend too much time worry about these cases, since writing around the stdlib for high performance code is the norm

PR #2208, opened 2026-08-17
feat: Add better error message for distinct nested functions.

### LINKED ISSUE #2123: [Feature]: Better error message for distinct functions
### Problem Statement

We should better explain that all top level functions have different types in error messages. A particularly nasty case is:
```python
@guppy 
def foo() -> Function[[], None]:
    if e:
        def bar() -> None: ...
    else:
        def bar() -> None: ...
    return bar
```
the error message complains about a type mismatch between
```
def bar() -> None
```
and 
```
def bar() -> None
```
which is quite confusing!

Nicely though, it gives the solution (adding type annotations)

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

SaashaJoshi:
Closes #2123

### DIFF
--- a/tests/integration/notebooks/errorsdemo.ipynb
+++ b/tests/integration/notebooks/errorsdemo.ipynb
@@ -327,12 +327,13 @@
       " 4 |         def bar() -> int:\n",
       "   |         -----------------\n",
       " 5 |             return 0\n",
-      "   | -------------------- This is of type `def bar() -> int`\n",
+      "   | -------------------- This is a distinct function `bar` of type `def bar() -> int`\n",
       " 6 |     else:\n",
       " 7 |         def bar() -> bool:\n",
       "   |         ------------------\n",
       " 8 |             return False\n",
-      "   | ------------------------ This is of type `def bar() -> bool`\n",
+      "   | ------------------------ This is a distinct function `bar` of type `def bar() ->\n",
+      "   |                          bool`\n",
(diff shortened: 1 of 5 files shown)


qartik:
Deferring the review to @croyzor as you are already involved in the original issue.

tests/integration/notebooks/errorsdemo.ipynb
```diff
@@ -1396,7 +1398,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.12.13"
+   "version": "3.14.3"
```
croyzor, line 1401: Is this change necessary?

SaashaJoshi: I think this happened when I explicitly regenerated the `.ipynb` notebook during testing. Since testing is done with Python 3.14, notebook metadata is updated as well. Is there another way to update `.ipynb` notebooks during testing? (--snapshot-update did not work)

croyzor: You can select the python version with `uv venv --python 3.12.13`, then regenerate maybe? I don't think this is a huge deal, just checking 🙂

croyzor (approved):
Nice change, thank you!! 🙌

PR #2210, opened 2026-08-17
fix: check+trace more functions before compilation starts

acl-cqc:
Even after #2104, `build_compiled_def` still calls `get_checked` which can cause more typechecking and tracing (during compilation). Specific cases addressed by this PR:
* Comptime functions with generic types
* functions that are only *loaded* (not called)
* Function Decls

...all of which are now checked before compilation. (The first two are tested.)

The "fix" in all cases is to call `ENGINE.register_generic_use` during checking of the call(er), via factory methods `make_type_apply`, `make_global_call` and `make_global_name`.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -274,8 +274,7 @@ class ExprChecker(AstVisitor[tuple[ast.expr, Subst]]):
             )
             # Apply instantiation of quantified type variables
             if inst:
-                assert isinstance(syn_ty, (FunctionType, FunctionDefType))
-                expr = instantiate_poly2(expr, syn_ty, inst)
+                expr = make_type_apply(expr, inst)
             return with_type(ty.substitute(subst), expr), subst
 
         # Otherwise, invoke the visitor
@@ -453,8 +452,8 @@ class ExprChecker(AstVisitor[tuple[ast.expr, Subst]]):
 
         # Apply instantiation of quantified type variables
         if inst:
-            assert isinstance(synth, (FunctionType, FunctionDefType))
-            node = instantiate_poly2(node, synth, inst)
+            node = make_type_apply(node, inst)
+
         return node, subst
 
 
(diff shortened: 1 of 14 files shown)


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1668,6 +1667,21 @@ def check_inst(func_ty: FunctionType, inst: Inst, node: AstNode) -> None:
         param.check_arg(arg, node)
 
 
+def make_type_apply(node: ast.expr, inst: Inst) -> ast.expr:
+    """Makes a new TypeApply node, registering that the target is used with the
+    specified type arguments."""
+    # TODO perhaps we can combine this with instantiate_poly, but that
```
acl-cqc: I did try this in 507ceb0ad21e948e6eb4e090faf33d60a7f4f727 but wasn't convinced that was an improvement. Thoughts welcome, or I can just drop the comment....

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1668,6 +1667,21 @@ def check_inst(func_ty: FunctionType, inst: Inst, node: AstNode) -> None:
         param.check_arg(arg, node)
 
 
+def make_type_apply(node: ast.expr, inst: Inst) -> ast.expr:
```
acl-cqc, line 1685: Might this name cause confusion when there is also a quite-different `check_type_apply` ??

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
tatiana-s: I think it's fine to keep them separate, avoids doing unnecessary work in places you only want the `make_type_apply` logic and not the whole `instantiate_poly`?

tatiana-s (approved):
Seems reasonable to me. Does this lead to any duplicate checking though where you do it both during checking and then again where you previously did it during compilation?

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
acl-cqc: I think they can be combined, perhaps more neatly - this might be necessary in the followup, so will see how that looks before merging this...

acl-cqc: Whilst they *can* be combined, I think it makes sense to keep `make_type_apply` as a minimal factory method paralleling `make_global_call` and `make_global_name`

acl-cqc:
> Seems reasonable to me. Does this lead to any duplicate checking though where you do it both during checking and then again where you previously did it during compilation?

The `to_check_worklist` will get filled up, but we'll check that they're not already in `CompilationEngine.checked` before actually checking again, so that should only happen once: see https://github.com/Quantinuum/guppylang/blob/085c4136637bdb23bdb62dd951d515e190eec93e/guppylang-internals/src/guppylang_internals/engine.py#L450-L451 and https://github.com/Quantinuum/guppylang/blob/085c4136637bdb23bdb62dd951d515e190eec93e/guppylang-internals/src/guppylang_internals/engine.py#L281-L290

acl-cqc:
I think this is now checking everything before compilation but don't have a definitive test of that, will try in another PR (think that's a "nice to have" but not essential). Can you have another look @tatiana-s? 🙏

tatiana-s (approved):
looks good to me 👍

PR #2211, opened 2026-08-17
fix: Unsupported inheritance error for guppy-defined bases

### LINKED ISSUE #2201: [Bug]: `TypeError` when trying to use (unsupported) struct inheritance.
### Description

When trying to use python inheritance in struct definitions an unrelated error is raised:

```py
TypeError: GuppyDefinition.__init__() takes 2 positional arguments but 4 were given
```

### Steps to Reproduce

[playpond link](https://play.guppylang.org?guppy=4ecd1396-7d37-31aa-a707-e95e6aef0ed0&n_shots=1&simulator=Coinflip&utm_source=sharing)

```py
from guppylang import guppy
from typing import Self

@guppy.struct
class MyStruct:
    @guppy
    def foo(self) -> int:
        return 0

@guppy.struct
class OtherStruct(MyStruct):
    pass

@guppy
def main() -> None:
    p = OtherStruct()
    output("v", p.foo())
```

### Expected Behaviour

A helpful error is raised, indicating the feature is not supported.

### Actual Behaviour

Unrelated error:

```py
PythonError: Traceback (most recent call last):
  File "/lib/python313.zip/_pyodide/_base.py", line 597, in eval_code_async
    await CodeRunner(
    ...<9 lines>...
    .run_async(globals, locals)
  File "/lib/python313.zip/_pyodide/_base.py", line 411, in run_async
    coroutine = eval(self.code, globals, locals)
  File "<exec>", line 41, in <module>
  File "/playpond/workspace.py", line 11, in <module>
    class OtherStruct(MyStruct):
        pass
TypeError: GuppyDefinition.__init__() takes 2 positional arguments but 4 were given
```

### Component

Guppy Compiler

### Environment

```markdown
- guppy 1.0.1
```

### Additional Context

_No response_

tatiana-s:
Trying to inherit from a guppy defined struct (as opposed to a python class as struct inheritance error tests we've had so far were testing for) wasn't reaching the inheritance unsupported error handling as we would get an error during parsing of the struct name instead. Resolving the guppy definition base as a python class allows us to reach the expected error. 

Closes #2201

### DIFF
--- a/guppylang-internals/src/guppylang_internals/tracing/object.py
+++ b/guppylang-internals/src/guppylang_internals/tracing/object.py
@@ -541,6 +541,18 @@ class TracingDefMixin(DunderMixin):
 
     wrapped: Definition
 
+    def __mro_entries__(self, bases: tuple[type, ...]) -> tuple[type, ...]:
+        """Resolve Guppy definitions used as Python class bases.
+
+        This lets class creation proceed so Guppy can report unsupported
+        inheritance with its regular diagnostics during struct/enum/protocol parsing.
+        """
+        if isinstance(self.wrapped, TypeDef):
+            python_class = getattr(self.wrapped, "python_class", None)
+            if isinstance(python_class, type):
+                return (python_class,)
+        return ()
+
     @property
     def id(self) -> DefId:
         return self.wrapped.id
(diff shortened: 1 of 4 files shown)


aborgna-q (approved):
ty!

Issue #707, opened 2024-12-10
Type inference doesn't work with `if`expressions


mark-koch:
E.g. in the line below, the compiler cannot infer the type of `[]` since the annotation doesn't flow through the `if` expression:

```python
x: list[int] = [] if cond() else [0]
```

This is hard to fix since `if` expressions are turned into control-flow before type checking.

PR #2213, opened 2026-08-17
fix(errors): error on loading polymorphic function

acl-cqc:
(Stumbled across this while investigating another issue, but nonetheless this seems an improvement.)

Currently if you assign a polymorphic function to a variable you get an error during *compilation*, not checking. Turned out we didn't have a test of this at all, so have added one...

(See new tests in [first commit](8acfaa72b5448f0c77cbae8ce52c31565853091d) which are passing on `main`)

In a couple of cases this leads to checking issuing a *different* error, and of course we never get to the one that would be issued later in compilation, so detecting the error earlier leads to a change in message.

I reason that assignment is the only case where we care:
* If the poly-function is used as an argument, type inference will be used, and infer an instantiation. We *could* do this for assignment (i.e. to find an instantiation where the variable is used only monomorphically), e.g. perhaps by searching across the basic block, so maybe #707 is in fact an understatement (type inference doesn't even work across a single BB yet...)
* For a top-level expression (`foo` where that's a poly-function), we don't need to find an instantiation, so it seems fine not to.

### DIFF
--- a/tests/error/type_errors/fun_item_ty_mismatch_generic.err
+++ b/tests/error/type_errors/fun_item_ty_mismatch_generic.err
@@ -1,20 +1,9 @@
-Error: Different types (at $FILE:21:11)
+Error: Unsupported (at $FILE:18:14)
    | 
-19 |     else:
-20 |         baz = bar
-21 |     return baz(42)
-   |            ^^^ Variable `baz` may refer to different types
-
-Notes:
-   | 
+16 | def main(b: bool) -> int:
 17 |     if b:
 18 |         baz = foo
-   |         --- This is of type `def foo(x: T) -> T`
-19 |     else:
-20 |         baz = bar
-   |         --- This is of type `def bar(x: T) -> T`
-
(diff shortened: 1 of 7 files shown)


tests/error/type_errors/fun_item_ty_mismatch_generic.err
```diff
@@ -1,20 +1,9 @@
...
 17 |     if b:
 18 |         baz = foo
-   |         --- This is of type `def foo(x: T) -> T`
-19 |     else:
-20 |         baz = bar
-   |         --- This is of type `def bar(x: T) -> T`
-
-Help: Consider adding type annotations to coerce them into opaque function
-values: `baz: Function[[?T], ?T] = ...`
+   |               ^^^ Polymorphic functions as dynamic higher-order values are not
```
acl-cqc, line 6: I mean, maybe this test is entirely redundant given the new one in `poly_callable.py` ?

PR #2224, opened 2026-08-18
refactor: Enforce barrier between checking and compilation stages

acl-cqc:
This is pure "code health": having a hard barrier between checking and compilation is required for callgraph analysis, but even then we don't strictly need to enforce that the barrier exists. However spotting violations early is good practice and will help ensure that any analysis done between stages (esp. interprocedural analysis, e.g. callgraph analysis) *is* seeing the whole, valid, program.


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -792,6 +796,10 @@ def visit_DesugaredListComp(self, node: DesugaredListComp) -> tuple[ast.expr, Ty
             node, node.generators, node.elt, self.ctx
         )
         result_ty = list_type(elt_ty)
+        # Check the append method now, before we need it during compilation
```
acl-cqc, line 856: Yes, two occurrences of this. It's tempting to put it inside `synthesize_comprehension` but (a) although it does specifically mention *list comprehension* in its doc-comment, and is only used for that, there's nothing in the code that actually uses the list type so far; (b) synthesize_comprehension is recursive (for flatmap's), so it'd be a bit more complex (or just registering more than once, unnecessarily...ok maybe that's harmless)

guppylang/src/guppylang/defs.py
```diff
@@ -150,6 +150,14 @@ def __getattr__(self, name: str) -> Any:
         # Handle attribute access when calling an enum variant constructor, like
         # `Enum.VariantA()`. In all other cases, we should not try create a new
         # attribute, so we directly raise the error.
+        from guppylang_internals.error import GuppyComptimeError
+        from guppylang_internals.tracing.state import tracing_active
+
+        if not tracing_active():
```
acl-cqc: This is a bit ugly, and makes me wonder if I need the same anywhere else for tests-never-written.

I tried a different approach - conceptually much better IMHO - in adda8bd617467d8f27a4cd6ff86420c6870a0b2a, but the error message is not nearly as good...thoughts?

guppylang-internals/src/guppylang_internals/error.py
```diff
@@ -124,19 +123,24 @@ def pretty_errors(f: FuncT) -> FuncT:
...
+        traceback: TracebackType | None,
     ) -> None:
         """Custom `excepthook` that intercepts `GuppyExceptions` for pretty printing."""
         if isinstance(err, GuppyError):
             sys.stderr.write(str(err))
             return
 
-        # If it's not a GuppyError, fall back to default hook
-        sys.__excepthook__(excty, err, traceback)
+        # If it's not a GuppyError, fall back to previous hook
```
acl-cqc, line 134: Previously this meant that if you called an `@pretty_errors` method *inside* a `@hide_trace` (or other exception hook), the outer one was lost for non-GuppyError's

tests/error/tracing_errors/call_python2.err
```diff
@@ -1,4 +1,4 @@
 Traceback (most recent call last):
   File "$FILE", line 7, in <module>
     Enum.VariantA(1)
-guppylang_internals.error.GuppyComptimeError: Function `VariantA` may only be called in a Guppy context
+guppylang_internals.engine.DefinitionStageError: Can only parse Type `Enum` during the `check` compiler stage
```
acl-cqc: This is not a fantastic error message, I'm hoping you feel it's good enough ;-).

1fa8554fe843eda521db16527c032a0698f959ab achieves a slightly better error message, but not much, and the code does not seem as good (we might need `if not tracing_active()`+Exception, in other places too?)

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -253,15 +258,20 @@ def reset(self) -> None:
...
         from guppylang_internals.checker.core import Globals
 
         if id in self.parsed:
             return self.parsed[id]
         defn = DEF_STORE.raw_defs[id]
-        if isinstance(defn, ParsableDef):
+        if self._stage != "check":
+            if not isinstance(
+                defn, (CheckableDef, CheckableGenericDef, CompilableDef, CompiledDef)
+            ):
```
ss2165: could be single condition

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -393,62 +407,69 @@ def check(self, def_ids: list[DefId], *, reset: bool = True) -> None:
...
-            entry_defn = self.get_parsed(def_id)
-            entry_params = (
-                entry_defn.params if isinstance(entry_defn, CheckableGenericDef) else []
-            )
-            entry_mono_args = tuple(param.to_bound() for param in entry_params)
-            try:
-                self.checked[def_id, entry_mono_args] = self.get_checked(
-                    def_id, entry_mono_args
+        assert self._stage == "none"
+        self._stage = "check"
```
ss2165: this pattern resembles contextvars to me, but not sure if necessary/good fit https://docs.python.org/3/library/contextvars.html

ss2165: ah sadly the most fitting feature is 3.14 and above :/ https://docs.python.org/3/library/contextvars.html#contextvars.ContextVar.set

tests/error/tracing_errors/call_python2.err
ss2165: this looks to be a regression in the error message

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Lost the comment, sadly ParsedDef is a union and that doesn't support `isinstance`

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: It pretty much is that, yes (modulo that we are careful only to change context from a known value to a new one, rather than to unconditionally override the old value). However, the feature that would make contextvars worth using is, indeed, new in 3.14; in 3.12 we'd still have:
```
   assert cv.get() == "old-value"
   tk = cv.set("new-value") # Replacing: cv = "new-value"
   try:
      ...
   finally:
      cv.set(tk.old_value) # Replacing: cv = "old-value"
```
which doesn't really look any better (sadly!)

ss2165: I still think it makes the usage and intention more narrow and therefore clearer for similar verbosity so would prefer

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -266,6 +266,13 @@ def get_parsed(self, id: DefId) -> ParsedDef:
         if id in self.parsed:
             return self.parsed[id]
         defn = DEF_STORE.raw_defs[id]
+        if self._stage == "none":
+            self._stage = "check"
+            try:
+                return self.get_parsed(id)
+            finally:
+                self._stage = "none"
+
```
ss2165: suggest encapsulating these state checks and updates in dedicated methods rather than updating state directly

guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -126,6 +126,7 @@ def build_compiled_def(self, def_id: DefId, type_args: Inst | None) -> CompiledD
 
         Might mutate the current Hugr if this definition has never been compiled before.
         """
+        assert ENGINE._stage == "compile"
```
maximilianruesch: This is accessing implementation details of `ENGINE`. Following the suggestion from @ss2165 on dedicated methods, I suggest you add a `assert_stage(...)` and a `CompilationStage(Enum)` rather than throwing around magic strings. The `update_stage` is then reserved for the `ENGINE` internals and can remain scoped this way.

A possible pattern is also to expose the current stage as read-only
```
@property
def stage(self) -> CompilationStage:
    return self._stage
```
and then do the assert here. The enum allows you to minimise the number of places you specify these literals, and makes refactors easier as well.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -481,88 +517,102 @@ def _compile(
...
-            ext
-            for ext in used_extensions_result.used_extensions.extensions
-            if ext.name not in std_ext_names
-        ]
-        return (
-            ModulePointer(
-                Package(modules=[graph.hugr], extensions=packaged_extensions), 0
-            ),
-            requested_defs,
+    def __init__(self, operation: str, defn: Definition, stage: str) -> None:
```
maximilianruesch: This could also take the enum for the stage

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -280,15 +298,26 @@ def get_parsed(self, id: DefId) -> ParsedDef:
...
+        if self._stage == "none":
+            self._stage = "check"
+            try:
+                return self.get_checked(id, mono_args)
+            finally:
+                self._stage = "none"
+
+        if self._stage != "check":
+            if not isinstance(defn, (CompiledDef, CompilableDef)):
+                raise DefinitionStageError("check", defn, "check")
```
maximilianruesch: Instead of doing this, when you have a helper method, you could add `self.assert_stage(CompilationStage.Check)` above the two explicit `check` calls below, so you only assert the stage when it is actually needed.

As it is right now, it may be practically the same, but the code is very confusing to me.

guppylang-internals/src/guppylang_internals/error.py
```diff
@@ -121,19 +120,24 @@ def pretty_errors[**P, T](f: Callable[P, T]) -> Callable[P, T]:
...
+        old_handler(excty, err, traceback)
 
-    @functools.wraps(f)
+    from functools import partial, wraps
```
maximilianruesch: Why add the import here instead of top-level, where it was previously?

guppylang-internals/src/guppylang_internals/error.py
maximilianruesch: I am confused as to how this code change changes this, since the only use of `hook` is in `pretty_errors_wrapped`, which immediately populates `old_handler` with the same value again (since the partial function application always runs). Could you elaborate on how this change benefits the situation?

maximilianruesch (changes requested):

guppylang-internals/src/guppylang_internals/compiler/core.py
ss2165: Would quite like to see the try/finally pattern extracted into a context manager actually

guppylang-internals/src/guppylang_internals/error.py
acl-cqc: `hide_trace` and `pretty_errors_wrapped` both use `exception_hook`, a contextmanager that backs up the old value of `sys.excepthook` in a *local* variable (`old_hook`), overrides `sys.excepthook` with a new value, does the wrapped function, and then restores `sys.excepthook`.

So if you `hide_trace` and within that execute something that is `pretty_errors_wrapped`, the handler from `hide_traced` is stored in `old_hook` and the pretty handler installed. If the code inside raises a GuppyError, the handler deals with this - fine. If the code inside raises any other kind of error, the pretty handler passes this back onto not `old_hook` but rather onto `sys.__except_hook__` (the handler from *outside* the `hide_trace`).

That is, given a *nest* of handlers, the old `pretty_errors_wrapped` forwards non-GuppyError exceptions onto *the outermost* handler, whereas the new forwards it onto the *next* handler out.

Does that help? We should discuss in person if not and see if we can come up with better wording for the comment.

guppylang-internals/src/guppylang_internals/error.py
acl-cqc: Well, previously we were doing `functools.wraps`, but I didn't like doing `functools.partial` so it seemed consistent to import both. I presumed that the old global `import functools` (rather than `import wraps from functools`) was because functools names are generic enough we don't want them in the global namespace...

I can use `functools.partial` here if you feel strongly, or (perhaps nicer?) `from functools import wraps as func_wraps, partial as func_partial` or something like that?

guppylang-internals/src/guppylang_internals/error.py
acl-cqc: Maybe the answer to your comment is to change:
```suggestion
        # If it's not a GuppyError, fall back to old_handler
```
perhaps ....`(the previous exception-hook)`

i.e. `hook` is meant as a description not a variable name?

guppylang-internals/src/guppylang_internals/compiler/core.py
acl-cqc: Yes, the context manager seems the right way to me, have done that (plus the `assert_stage` method)

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Yup ok fair. With the `assert_stage` method it's easy enough to call it a few times.

tests/error/tracing_errors/call_compile.py
maximilianruesch, line 1: Interesting error case here. An error is better than nothing, but since it can be user-triggered it should be a proper Guppy error with a help, etc. I don't know how feasible this is, and it does not need to be part of this PR. Do you have thoughts on this?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -248,27 +268,44 @@ def reset(self) -> None:
         self.generic_to_check_worklist = {}
         self.types_to_check_worklist = {}
 
+    def assert_stage(
+        self, operation: str, defn: Definition | None, stage: CompilationStage
+    ) -> None:
```
maximilianruesch: I suggest putting the `stage` argument first, and perhaps even making `operation` and `defn` keyword-only. Their purpose is not particularly clear from the method name, and you could even make them a generic `context: tuple[str, Definition | None]` or even `context: str`, which signals that the additional information you provide is just more context as to what triggered this assertion.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -480,8 +524,14 @@ def compile(self, def_ids: list[DefId], *, reset: bool = True) -> ModulePointer:
     def _compile(
         self, def_ids: list[DefId], *, reset: bool = True
     ) -> tuple[ModulePointer, list[CompiledDef]]:
+        self.assert_stage("compile", None, CompilationStage.NONE)
         self.check(def_ids, reset=reset)
```
maximilianruesch: How about we swap those two lines? The check call should perform its own assertions about the stage, and if you want to assert anything, it is that you are out of the check stage after it happened.

guppylang-internals/src/guppylang_internals/error.py
maximilianruesch: > functools names are generic enough we don't want them in the global namespace...

That should not matter, since we are (or should be) explicitly exporting the symbols we want from this file anyway.

> I can use functools.partial here if you feel strongly

I would consider myself to feel strongly about the position of the import, because unless it is technically required I think one should collect imports at the top of the file. Using `functools.partial` instead of `partial` is something I do because it gives the partial a bit more context, but I do not feel super strong about it. I would like to avoid import aliases though, so the second option of yours is no option to me.

guppylang-internals/src/guppylang_internals/error.py
maximilianruesch: I did get your intention the first time around, but I am confused as to how the code achieves this. Specifically, I am confused about the difference between:
```python
def my_func(my_value: str) -> str:
     return my_value

a = partial(my_func, "I am a donkey")
```
and 
```python
def a() -> str:
    return "I am a donkey"
```
which is the change that I think you are doing? `wraps(f)` is just a hint afaik, and does not influence the actual execution.

maximilianruesch (changes requested):
I am still a bit confused, but I like the enum and method changes a lot!

guppylang-internals/src/guppylang_internals/error.py
acl-cqc: The old code installed `hook` as handler, which called `sys.__excepthook__` i.e. the system default (root) handler.

The new code installs as handler `partial(hook, sys.excepthook)`. That is, it computes `hook` with `old_handler` set to the current `sys.excepthook`; then it installs *that* as handler. Then this new handler `partial(hook, sys.excepthook)`, calls `old_handler` rather than `sys.__excepthook__`.

The `wraps` really isn't significant (and hasn't changed).

Would it be clearer if I changed the `partial` (on line 140 below) to `partial(hook, old_handler=sys.excepthook)` ??

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: That means that calling this when you are not in CompilationStage.COMPILE would have side effects before it falls over; I'm not sure that's good?

tests/error/tracing_errors/call_compile.py
acl-cqc: The reason for non-GuppyError here is we don't want to show static source code (like for a typechecking error, i.e. where the error comes from looking statically at the source code) but a dynamic trace of how it is that we got to execute the forbidden `compile`...see e.g. `array_empty_call.err`.

guppylang-internals/src/guppylang_internals/error.py
acl-cqc: > Specifically, I am confused about the difference between:
> .... 
> and
> ....
> which is the change that I think you are doing?

Ah - I don't think I grokked the significance of this part of your comment. The point of the change is that `sys.excepthook` is not the same as `sys.__excepthook__`, and that `exception_hook` overwrites the former but not the latter.

maximilianruesch: Ah, yes that small difference is what I failed to spot. Thanks!

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: Hmm. Maybe checking both is best?

tests/error/tracing_errors/call_compile.py
acl-cqc: That is, because this error is in comptime (python) code, not guppy code

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Done. I did try the `context: tuple` idea but it seemed more verbose/awkward than just making the Definition optional. However, if we want to neaten further (to a simple `context: str` as you suggest) then I think this would be fine if we do #2243 first.

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Contextmanager method does this now.

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Yup, callers now format a `context: str`, given `Definition.__str__` this is much neater :)

guppylang-internals/src/guppylang_internals/engine.py
acl-cqc: Ok done, but I've made the second an `assert` because whereas assert-stage is about part of the compiler being run at the wrong time, this would be a bug in the stage-tracking logic itself

guppylang-internals/src/guppylang_internals/error.py
acl-cqc: Ok, `functools.partial` it is

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: Most wonderful!

maximilianruesch (approved):
🚀

PR #2229, opened 2026-08-20
fix: Add correct location debug info to calls inside comptime functions

### LINKED ISSUE #1606: Support location debug info inside comptime functions
Main issue: need to figure out how to pass the call AST node to the call tracing

tatiana-s:
Closes https://github.com/Quantinuum/guppylang/issues/1606

The alternative would have been trying to pass the correct AST node somehow instead of manually trying to find it here which is a bit more of a hack, however I think that would have been a lot more awkward refactoring.


guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -71,6 +73,46 @@ class TracingReturnError(Error):
...
+        Loc(calling_func_span.file, positions.end_lineno, positions.end_col_offset),
+    )
+    candidates = [
+        call
+        for call in ast.walk(calling_func_node)
+        if isinstance(call, ast.Call)
+        and (call_span := to_span(call)).start <= active_call_span.start
+        and active_call_span.end <= call_span.end
+    ]
+    # Choose inner-most call - this may be incorrect, however the debug information
```
tatiana-s: I don't like this bit, but unsure of how this could be improved

tatiana-s:
Adding the callsite computation seemed to worsen comptime checking time non-insignificantly, so it's now only done when debug mode is on since that will generally have worse performance for other reasons too. 
Screenshot 2026-08-24 at 15 36 29

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -71,6 +74,46 @@ class TracingReturnError(Error):
...
+        positions is None
+        or positions.lineno is None
+        or positions.col_offset is None
+        or positions.end_lineno is None
+        or positions.end_col_offset is None
+    ):
+        return calling_func_node
+
+    try:
+        calling_func_span = to_span(calling_func_node)
```
acl-cqc: can you not just use `get_file(calling_func_node)` here given that's all you need? Admittedly that can still return `None`

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -71,6 +74,46 @@ class TracingReturnError(Error):
...
+
+    active_call_span = Span(
+        Loc(calling_func_span.file, positions.lineno, positions.col_offset),
+        Loc(calling_func_span.file, positions.end_lineno, positions.end_col_offset),
+    )
+    candidates = [
+        call
+        for call in ast.walk(calling_func_node)
+        if isinstance(call, ast.Call)
+        and (call_span := to_span(call)).start <= active_call_span.start
```
acl-cqc: In fact the only use you make of active_call_span is its start/end, here;
* This could be `active_call_span in to_span(call)` using `Span.__contains__`
* But either way, this comparison of `class Loc`s uses the ordering generated by `@dataclass(order=True)` which looks like it would do alphabetical comparison of `file` before looking at start/end. So I wonder if you can just skip making `active_call_span` (with associated chance of failure) at all and just directly compare `positions.(start|end)_(lineno|col_offset)`

guppylang-internals/src/guppylang_internals/tracing/function.py
acl-cqc: I think just `Choose innermost call that spans the position given in the frame` would probably be fine here. You could say `best we can do` but when you say it may be incorrect I'm not really clear *why*...

tests/metadata/test_debug_info.py
```diff
@@ -204,6 +205,32 @@ def foo() -> None:
...
 
 
+def test_comptime_call_location():
+    @guppy.comptime
+    def foo() -> None:
+        bar()  # call 1  # MARKER:ct_call_bar
+        non_guppy_function()  # not a hugr call
+        comptime_bar()  # call 2  # MARKER:ct_call_comptime_bar
+        _ = (bar(), comptime_bar())  # MARKER:ct_calls_same_line
+
```
acl-cqc, line 217: Would also like to see a couple of cases like
`foo(foo())` where (a) the inner and (b) the outer calls fail....

tests/metadata/test_debug_info.py
```diff
@@ -204,6 +205,32 @@ def foo() -> None:
         assert (debug_info.line_no, debug_info.column) == expected_info[i]
 
 
+def test_comptime_call_location():
+    @guppy.comptime
+    def foo() -> None:
+        bar()  # call 1  # MARKER:ct_call_bar
```
acl-cqc, line 212: Is the idea that `bar()` and `comptime_bar()` are not defined? In which case what makes `comptime_bar` actually comptime? Ah no, I see them in `debug_info_example.py` but I have no idea how this testcase references that. Ok anyway....

tests/metadata/test_debug_info.py
```diff
@@ -204,6 +205,32 @@ def foo() -> None:
         assert (debug_info.line_no, debug_info.column) == expected_info[i]
 
 
+def test_comptime_call_location():
```
acl-cqc, line 209: Do you have a case where the logic in `_find_call_site` is incorrect (as your comment suggests it can be)?

acl-cqc (commented):
Thanks Tatiana! I'd say this looks pretty good, at least the bit where you're worried seems pretty good to me! (I have a couple more ideas for tests of that bit tho.) We may be able to simplify a little, too...

tests/metadata/test_debug_info.py
tatiana-s: Actually can't come up with one, I was thinking the nested case previously but I wasn't thinking about it correctly and it does work in the test

guppylang-internals/src/guppylang_internals/tracing/function.py
```diff
@@ -72,6 +75,42 @@ class TracingReturnError(Error):
...
+    if (
+        positions is None
+        or positions.lineno is None
+        or positions.col_offset is None
+        or positions.end_lineno is None
+        or positions.end_col_offset is None
+    ):
+        return calling_func_node
+
+    active_start = (positions.lineno, positions.col_offset)
```
acl-cqc, line 93: Might be worth adding a comment `use tuple rather than Loc so do we not need a filename`

acl-cqc (approved):
Looks great, thanks @tatiana-s :)

Issue #1842, opened 2026-06-11
[Bug]: Wrong metadata key used for unitary flags


nicolaassolini-qntm:
### Description

this issue requires an alias in tket2 (https://github.com/Quantinuum/tket2/pull/1669) to ensure compatibility.

### Environment

```markdown

```

PR #2230, opened 2026-08-20
perf: optimize dictionary substitution filtering

qartik:
### Summary
Optimize dictionary substitution filtering in the type-checking hot path by extracting the filtering logic into a reusable helper function.

### Changes
- Replaced three instances of dictionary comprehensions with a centralized `filter_subst()` helper
- Added early-exit optimization for empty subst or unsolved_vars sets
- Avoids unnecessary iterations and allocations during type synthesis

### Files Changed
- `guppylang-internals/src/guppylang_internals/checker/expr_checker.py`
  - Added `filter_subst()` helper function (line 1167)
  - Updated three call sites: lines 309, 1208, 1652

### Impact
- **Performance**: Reduced allocations in hot code paths (type checking and synthesis)
- **Backward Compatibility**: No API or behavior changes
- **Testing**: Uses existing test suite

### Original Pattern
```python
subst = {v: t for v, t in subst.items() if v in unsolved_vars}
```

### Optimized Pattern
```python
subst = filter_subst(subst, unsolved_vars)
```

This is part of a performance optimization initiative targeting high-impact, low-effort improvements in the type-checking system.

PR #2231, opened 2026-08-20
perf: optimize substitution filtering in expr_checker

qartik:
## Summary
- add `filter_subst_to_unsolved_vars` to centralize substitution filtering in `expr_checker.py`
- avoid creating a new dict when the substitution is already within the target unsolved variable set
- replace three hot-path dict comprehensions with the helper

## Testing
- `uv run pytest -q tests -k "expr or type"`

Related: #1842


qartik:
Closed in favor of #2230 which has more comprehensive documentation and context.

PR #2233, opened 2026-08-21
feat(emulator): Add Traces and Metrics to EmulatorResult

### LINKED ISSUE #2218: [Feature]: Trace program instructions using Emulator interface
### Problem Statement

Selene provides general operation tracing by executing the compiled program and recording: https://github.com/Quantinuum/selene/blob/0.3-series/selene-core/python/selene_core/trace.py

This is in particular used by circuit extraction: https://docs.quantinuum.com/selene/user_guide/simulation_logs.html

These capabilities are very useful for program analysis and should be available from the guppy emulator interface

### Proposed Solution

- Provide short-hands for circuit extraction, metrics, and Trace (wrappers around `.with_event_hook`)
- Add analysis examples to emulator module level docstring

### Component

Guppy Compiler

### Effort Estimate

### Additional Context

- For circuit extraction we should avoid or defer pytket import to avoid a tight dependency

ss2165:
Closes #2218 

Built with Codex iteration.

- Adds metric and trace enabling toggles to EmulatorInstance builder
- Creates invocation-local event hooks for selene
- Extracts traces and metrics and adds them to new fields in EmulatorResult
- New methods in EmulatorResult to extract Trace, pytket Circuit (from trace) and Metrics
- Example in emulator docs

### DIFF
--- a/guppylang/src/guppylang/emulator/result.py
+++ b/guppylang/src/guppylang/emulator/result.py
@@ -185,9 +185,9 @@ class EmulatorResult(QsysResult):
         <https://docs.quantinuum.com/tket/api-docs/circuit.html>`_ representing
         the stream of gates output by the user program at emulation time.
 
-        Circuits are extracted from each shot's trace in the compiler's compiled
-        gateset, typically the primitive gateset of the hardware platform being
-        emulated.
+        Circuits are extracted from each shot's trace in the gateset produced by
+        compilation, typically the primitive gateset of the hardware platform
+        being emulated.
 
         Enable trace collection with :meth:`EmulatorInstance.with_trace` before
         running the emulator. ``pytket`` availability is checked before conversion.
--- a/guppylang/src/guppylang/emulator/result.py
+++ b/guppylang/src/guppylang/emulator/result.py
@@ -182,8 +182,12 @@ class EmulatorResult(QsysResult):
 
     def circuits(self) -> list[Circuit]:
         """Return `pytket Circuit objects
-        <https://docs.quantinuum.com/tket/api-docs/circuit.html>`_ for each
-        emulated shot.
+        <https://docs.quantinuum.com/tket/api-docs/circuit.html>`_ representing
+        the stream of gates output by the user program at emulation time.
+
+        Circuits are extracted from each shot's trace in the compiler's compiled
+        gateset, typically the primitive gateset of the hardware platform being
+        emulated.
 
         Enable trace collection with :meth:`EmulatorInstance.with_trace` before
         running the emulator. ``pytket`` availability is checked before conversion.
--- a/guppylang/src/guppylang/emulator/result.py
+++ b/guppylang/src/guppylang/emulator/result.py
@@ -173,12 +173,9 @@ class EmulatorResult(QsysResult):
                 "Trace collection was not enabled; call `with_trace()` before running "
                 "the emulator."
             )
-        # Selene starts collection before parsing a shot, so an interrupted shot may
-        # have analysis data without a corresponding entry in ``self.results``.
-        return [
-            shot.get_trace()
-            for shot in self._circuit_extractor.shots[: len(self.results)]
-        ]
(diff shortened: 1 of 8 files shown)


guppylang/src/guppylang/emulator/result.py
```diff
@@ -127,3 +155,74 @@ def to_partial(
...
+                "the emulator."
+            )
+        # Selene starts collection before parsing a shot, so an interrupted shot may
+        # have analysis data without a corresponding entry in ``self.results``.
+        return [
+            shot.get_trace()
+            for shot in self._circuit_extractor.shots[: len(self.results)]
+        ]
+
+    def circuits(self) -> list[Circuit]:
```
jake-arkinstall, line 180: Within the selene event hook, we use 'user circuit' to make it clear that this "circuit" is the sequence of instructions that come directly from the user program, rather than e.g. that which may be spat out of the runtime (after any runtime compilation, reordering, eliding of gates etc).

But selene is quite bothered about the distinction between that which comes into the runtime, that which comes out of it, and so on. I don't know if that's necessarily the case in guppy land, as the primary interface focuses around the user's code. However, it's worth asking: in the context of the guppy emulator interface, is the name 'circuits' sufficiently clear?

tests/integration/test_emulator.py
```diff
@@ -62,6 +62,65 @@ def main() -> None:
...
+    assert [shot["user_program"]["qalloc_count"] for shot in metrics] == [1, 2]
+    assert [shot["user_program"]["measure_request_count"] for shot in metrics] == [1, 2]
+    assert [shot["post_runtime"]["measure_individual_count"] for shot in metrics] == [
+        1,
+        2,
+    ]
+
+    user_trace_events = [trace.get_user_program_trace().events for trace in traces]
+    assert [len(events) for events in user_trace_events] == [7, 12]
+    assert [[event.event.kind for event in events] for events in user_trace_events] == [
```
jake-arkinstall: I wonder if this should instead be a snapshot: it's sensitive to passes during compilation, and being able to just --snapshot-update is easier on the developer than rearranging these manually.

guppylang/src/guppylang/emulator/result.py
```diff
@@ -127,3 +155,74 @@ def to_partial(
...
+
+        Enable trace collection with :meth:`EmulatorInstance.with_trace` before
+        running the emulator.
+        """
+        if self._circuit_extractor is None:
+            raise RuntimeError(
+                "Trace collection was not enabled; call `with_trace()` before running "
+                "the emulator."
+            )
+        # Selene starts collection before parsing a shot, so an interrupted shot may
```
jake-arkinstall: Potential issue, but I will need to take another look at selene's behaviour in the scenario I have in mind to make sure I'm not misleading you here:

The scenario is that E.g. shot 9 fails due to a user program panic() call. completed_shots contains 8 shots.

IIRC we still do our best to populate event hooks for the failing shot too (including metrics and traces). If my memory is correct (and I am not certain, I will need to verify), then this interface would prevent access to the event data on the failing shot, whereas it might be the most important thing for the user to access.

ss2165: ah I was under the impression it was just empty - if it gets populated it is indeed useful to keep

ss2165: I ran some tests with a failing second that seems to produce empty

guppylang/src/guppylang/emulator/result.py
ss2165: tried to clarify in a future proof way in the docstring

tatiana-s (approved):
Looks good to me, subject to Jake being happy with the changes after his comments

guppylang/src/guppylang/emulator/result.py
ss2165: confirmed due to upstream behaviour in selene raising error before filling in failed shots event data, reverting to allow for future fix

PR #2234, opened 2026-08-24
fix: Order RNG context lifetimes with effects

qartik:
## Summary

- mark `NewRNGContext` and `DeleteRNGContext` with the shared `Effect.ANY`
- preserve source ordering between sequential global RNG lifetimes
- cover two seeded contexts and effect propagation through an uninlined helper

Related: Quantinuum/tket2#1945

## Testing

- `uv run pytest tests/integration/test_side_effect_ordering.py tests/integration/test_qsystem.py -q` (32 passed)
- `UV_CACHE_DIR=/private/tmp/guppylang-rng-uv-cache just check`


maximilianruesch (approved):
Thanks! I believe the RNG is already listed in our internal design doc to have a custom effect for, so that should all be sorted once we introduce them!

PR #2243, opened 2026-08-27
refactor: Extract stringifiers for internal definitions

acl-cqc:
Many places do `f"{somedefn.description} {somedefn.name}"` on a Definition, which is quite verbose. Meanwhile the `__str__` method of Definition is not used (changing it to throw an AssertionError does not fail any tests)...hence, define the latter to the former. Since there are also cases which explicitly capitalize the description, add `.caps_str()` for that (better names welcomed?).

The main impact is if you are debugging and printing out Definitions as part of that, but you can still get the same string as before back just via `repr(d)`, so the impact is quite small.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/definition/common.py
+++ b/guppylang-internals/src/guppylang_internals/definition/common.py
@@ -61,6 +61,13 @@ class Definition(ABC):
     name: str
     defined_at: ast.AST | None
 
+    def __str__(self) -> str:
+        return f"{self.description} {self.name}"
+
+    def caps_str(self) -> str:
+        """Returns a capitalized string representation of this definition."""
+        return f"{self.description.capitalize()} {self.name}"
+
     @property
     @abstractmethod
     def description(self) -> str:
--- a/guppylang-internals/src/guppylang_internals/definition/common.py
+++ b/guppylang-internals/src/guppylang_internals/definition/common.py
@@ -62,11 +62,11 @@ class Definition(ABC):
     defined_at: ast.AST | None
 
     def __str__(self) -> str:
-        return f"{self.description} {self.name}"
+        return f"{self.description} `{self.name}`"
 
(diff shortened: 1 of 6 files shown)


guppylang-internals/src/guppylang_internals/definition/common.py
```diff
@@ -61,6 +61,13 @@ class Definition(ABC):
     name: str
     defined_at: ast.AST | None
 
+    def __str__(self) -> str:
+        return f"{self.description} `{self.name}`"
+
+    def caps_str(self) -> str:
+        """Returns a capitalized string representation of this definition."""
+        return f"{self.description.capitalize()} `{self.name}`"
```
maximilianruesch, line 78: I suggest having `to_caps_str()`, `to_str()` (which does what `__str__` currently does) to mirror it, and have `__str__` instead call the latter. Then you can also add a docstring to `to_str` which points to `repr(defn)` if you want a full description back

guppylang-internals/src/guppylang_internals/tracing/object.py
```diff
@@ -577,8 +576,7 @@ def __call__(self, *args: Any) -> Any:
                 constructor_id := DEF_STORE.type_members[defn.id].get("__new__")
             ):
                 return TracingDefMixin(DEF_STORE.raw_defs[constructor_id])(*args)
-        err = f"{defn.description.capitalize()} `{defn.name}` is not callable"
-        raise GuppyComptimeError(err)
+        raise GuppyComptimeError(f"{defn.caps_str()} is not callable")
```
maximilianruesch: Perhaps we should have something that auto-capitalises all sentences that are put into these errors, so we do not need all of the extra functions to capitalise... But that should probably not be part of this PR.

maximilianruesch (approved):

PR #2245, opened 2026-08-28
chore: Update deprecated uv prerelease mode

nicolaassolini-qntm:
I was tired of getting this warning:

```
warning: The `if-necessary-or-explicit` pre-release mode is deprecated and will be removed in a future release. Use `if-necessary` instead.
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -21,7 +21,7 @@ test = [
 ]
 
 [tool.uv]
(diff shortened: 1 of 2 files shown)


maximilianruesch (commented):
This may require a certain minimum `uv` version to work correctly. Have you checked that this requirement is fulfilled / demanded in our development guides?

PR #2264, opened 2026-09-01
docs: incorrect URL in changelog

sburton84:


croyzor (approved):

aborgna-q:
Hi Simon. I saw you made a PR for this in `main`, thanks!

We do patch backports separatedly, so there is no need to submit a manual fix here.

Closing.

PR #2266, opened 2026-09-01
fix: Raise minimum version of tket to 0.15.7

maximilianruesch:
Backport of #2258 since it contains a bugfix for doubly-applied modifiers


nicolaassolini-qntm (approved):
🤖

Issue #2122, opened 2026-07-31
[Bug]: possible reordering of array index vs exit


acl-cqc:
### Description

```
    @guppy
    def foo(arr1: array[qubit, 3], i: int) -> None:
        h(arr1[i])
        exit("Index was ok", i)
```
produces this Hugr:
<img width="1396" height="876" alt="Image" src="https://github.com/user-attachments/assets/6cff9e5b-d977-4472-b99d-92d255ac37ab" />

There is no dependency between the array accesses (which may panic) and the exit, so these could be reordered in the compiler, changing the success/failure status of the program

### Expected Behaviour

We need an order edge or something!

### Environment

```markdown

```

acl-cqc:
Should be fixed as part of #1747 where we can do things properly via an appropriate partial order.

If we need a v1.0 fix, i.e. that does not use any effect code, we'd have to just make every array op side-effecting (on the single linear chain of order edges that v1.0 has), which might inhibit some optimization.

acl-cqc:
Note that following #2035 there are more cases where this bug *might* become observable - specifically, if the array index was inside a function call that was *not* inlined, then prior to #2035 the bug would have been masked, but (following #2035) this could show up.

acl-cqc:
Same is true of `arr.take(idx)` as well as `arr[idx]`: following #2328 both will be checked as having effects, but *in the quantum case* neither will be compiled with them. This is ripe for #1747.

`discard_all_taken` also appears to be compiled without effects but (in the quantum case only) can panic if some elements are not taken.

PR #2269, opened 2026-09-02
refactor: remove (unsafe) default effects for @hugr_op / @custom_func

acl-cqc:
We could have these default to `[Effect.ANY]` as (unlike present default) that would at least be safe/conservative; but discussion suggested that we want to explicitly specify it everywhere, at least in this repo.  (Using the default might imply that we haven't thought about it, and we have....at least a bit; hopefully, enough, but there are some legacy cases that need fixing, #2122).

So, adding said safe default, could be a two-line followup PR.

I used AI to update all the callsites, but then added a few fixes when reviewing myself....and skipped/noted a few. So, thoughts? I could be more consistent, but am still planning for #1747 to be the first attempt at definitiveness.

### DIFF
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -84,7 +84,7 @@ class array[T, n: nat](builtins.list[T]):
 
     @custom_function(
         NewArrayCompiler(),
-        NewArrayChecker(),
+        checker=NewArrayChecker(),
         higher_order_value=False,
         unitary_flags=UnitaryFlags.Dagger,
         effects=(),
@@ -103,7 +103,7 @@ class array[T, n: nat](builtins.list[T]):
 
     @custom_function(
         CopyInoutCompiler(),
-        ArrayCopyChecker(),
+        checker=ArrayCopyChecker(),
         unitary_flags=UnitaryFlags.Dagger,
         effects=(),
     )
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -87,6 +87,7 @@ class array[T, n: nat](builtins.list[T]):
         NewArrayChecker(),
         higher_order_value=False,
         unitary_flags=UnitaryFlags.Dagger,
+        effects=(),
     )
     def __new__(): ...
 
@@ -101,7 +102,10 @@ class array[T, n: nat](builtins.list[T]):
         return SizedIter(ArrayIter(self, 0))
 
     @custom_function(
-        CopyInoutCompiler(), ArrayCopyChecker(), unitary_flags=UnitaryFlags.Dagger
+        CopyInoutCompiler(),
+        ArrayCopyChecker(),
+        unitary_flags=UnitaryFlags.Dagger,
+        effects=(),
     )
     def copy[T: Copy, n: nat](self: array[T, n]) -> array[T, n]:
         """Copy an array instance. Will only work if T is a copyable type."""
@@ -110,6 +114,7 @@ class array[T, n: nat](builtins.list[T]):
         ArrayIsBorrowedCompiler(),
         checker=ArrayIndexChecker(),
         unitary_flags=UnitaryFlags.Dagger,
+        effects=(),
     )
     def is_borrowed[L, n: nat](self: array[L, n], idx: int) -> bool:
         """Checks if an element has been taken out of the array.
@@ -131,6 +136,7 @@ class array[T, n: nat](builtins.list[T]):
     @custom_function(
         ArrayGetitemCompiler(),
         checker=ArrayIndexChecker(),
+        effects=(),
     )
     def take[L, n: nat](self: array[L, n], idx: int) -> L:
         """Takes an element out of the array.
@@ -193,6 +199,7 @@ class array[T, n: nat](builtins.list[T]):
     @custom_function(
         ArraySetitemCompiler(elem_first=True),
         checker=ArrayIndexChecker(expr_index=2),
+        effects=(),
     )
     def put[L, n: nat](self: array[L, n], elem: L @ owned, idx: int) -> None:
         """Puts an element back into the array if it has been taken out previously.
@@ -246,7 +253,7 @@ class array[T, n: nat](builtins.list[T]):
         self.put(elem, idx)
         return ok(None)
 
-    @custom_function(ArrayDiscardAllUsedCompiler())
+    @custom_function(ArrayDiscardAllUsedCompiler(), effects=())
     def discard_all_taken[L, n: nat](self: array[L, n] @ owned) -> None:
         """Discards array assuming that all elements have been taken out, and panics if
         that is not the case.
--- a/guppylang/src/guppylang/std/array.py
+++ b/guppylang/src/guppylang/std/array.py
@@ -30,7 +30,6 @@ from guppylang_internals.std._internal.compiler.array import (
 from guppylang_internals.std._internal.compiler.frozenarray import (
     FrozenarrayGetitemCompiler,
 )
-from guppylang_internals.tys import Effect
 from guppylang_internals.tys.builtin import array_type_def, frozenarray_type_def
 from guppylang_internals.tys.ty import UnitaryFlags
 
@@ -64,15 +63,13 @@ class array[T, n: nat](builtins.list[T]):
     @custom_function(
         ArrayGetitemCompiler(),
         checker=ArrayIndexChecker(),
-        effects=[Effect.ANY],  # includes unwrap (compiled separately)
-    )
+    )  # includes unwrap (compiled separately)
     def __getitem__[L, n: nat](self: array[L, n], idx: int) -> L: ...
 
     @custom_function(
         ArraySetitemCompiler(),
         checker=ArrayIndexChecker(),
-        effects=[Effect.ANY],  # includes (compiled separately)
-    )
+    )  # includes unwrap (compiled separately)
     def __setitem__[L, n: nat](
         self: array[L, n], idx: int, value: L @ owned
     ) -> None: ...
@@ -329,7 +326,7 @@ class ArrayIter[L, n: nat]:
         return nothing()
 
 
-@custom_function(ArraySwapCompiler(), effects=[Effect.ANY])
+@custom_function(ArraySwapCompiler())  # unwraps for index-OOB
 def array_swap[L, n: nat](arr: array[L, n], idx: int, idx2: int) -> None:
     """Swap two elements in an array at indices idx and idx2.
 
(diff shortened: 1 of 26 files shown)


guppylang/src/guppylang/std/qsystem/helios/__init__.py
```diff
@@ -230,7 +230,8 @@ def _lazy_measure(q: qubit @ owned) -> Future[bool]: ...
 @custom_function(
     InoutMeasureCompiler(
         "LazyMeasureReset", ext=QSYSTEM_HELIOS_EXTENSION, return_future=True
-    )
+    ),
+    effects=(),
```
maximilianruesch: Is this said to have no effect since it actually does not yet force an ion reset? But it might enqueue one nonetheless...

guppylang/src/guppylang/std/qsystem/sol/__init__.py
```diff
@@ -260,7 +260,8 @@ def _lazy_measure(q: qubit @ owned) -> Future[bool]: ...
 @custom_function(
     InoutMeasureCompiler(
         "LazyMeasureReset", ext=QSYSTEM_SOL_EXTENSION, return_future=True
-    )
+    ),
+    effects=(),
```
maximilianruesch: See https://github.com/Quantinuum/guppylang/pull/2269/changes#r3935575424

guppylang/src/guppylang/std/array.py
```diff
@@ -64,15 +63,13 @@ class array[T, n: nat](builtins.list[T]):
     @custom_function(
         ArrayGetitemCompiler(),
         checker=ArrayIndexChecker(),
-        effects=[Effect.ANY],  # includes unwrap (compiled separately)
-    )
+    )  # includes unwrap (compiled separately)
     def __getitem__[L, n: nat](self: array[L, n], idx: int) -> L: ...
```
maximilianruesch: I am guessing the existence of the comment justifies to be explicit here? As well as in the second occurrence in this file

guppylang/src/guppylang/std/array.py
```diff
@@ -246,7 +250,7 @@ def try_put[L, n: nat](
         self.put(elem, idx)
         return ok(None)
 
-    @custom_function(ArrayDiscardAllUsedCompiler())
+    @custom_function(ArrayDiscardAllUsedCompiler(), effects=())
```
maximilianruesch: No effects is only correct if the compiler inserts the panic effects itelf...

guppylang/src/guppylang/std/array.py
```diff
@@ -193,6 +196,7 @@ def try_take[L, n: nat](self: array[L, n], idx: int) -> Option[L]:
     @custom_function(
         ArraySetitemCompiler(elem_first=True),
         checker=ArrayIndexChecker(expr_index=2),
+        effects=(),
```
maximilianruesch: Same here as in https://github.com/Quantinuum/guppylang/pull/2269/changes#r3935598499

guppylang/src/guppylang/std/array.py
```diff
@@ -110,6 +111,7 @@ def copy[T: Copy, n: nat](self: array[T, n]) -> array[T, n]:
         ArrayIsBorrowedCompiler(),
         checker=ArrayIndexChecker(),
         unitary_flags=UnitaryFlags.Dagger,
+        effects=(),
```
acl-cqc, line 114: Ok so actually this can panic if out-of-bounds

guppylang/src/guppylang/std/quantum/__init__.py
```diff
@@ -66,14 +66,16 @@ def __bool__(self: "Measurement") -> bool:
         return self.read()
 
 
-@hugr_op(quantum_op("TryQAlloc"))
+# Effect not because it panics (it doesn't) but because we want this to be
```
acl-cqc, line 69: This is one of a few fixes in the PR...

guppylang/src/guppylang/std/array.py
```diff
@@ -101,7 +102,10 @@ def __iter__[L, n: nat](self: array[L, n] @ owned) -> SizedIter[ArrayIter[L, n],
         return SizedIter(ArrayIter(self, 0))
 
     @custom_function(
-        CopyInoutCompiler(), ArrayCopyChecker(), unitary_flags=UnitaryFlags.Dagger
+        CopyInoutCompiler(),
+        checker=ArrayCopyChecker(),
+        unitary_flags=UnitaryFlags.Dagger,
+        effects=(),
```
acl-cqc, line 108: Only works for copyable elements so they are never borrowed so there is no question about that

guppylang/src/guppylang/std/list.py
```diff
@@ -42,19 +42,21 @@ def __getitem__(self: list[L], idx: int) -> L: ...
...
     def __new__(x): ...
 
-    @custom_function(NoopCompiler())  # TODO: define via Guppy source instead
+    @custom_function(
+        NoopCompiler(), effects=()
+    )  # TODO: define via Guppy source instead
     def __iter__(self: list[L] @ owned) -> list[L]: ...
 
-    @hugr_op(unsupported_op("pop"))
+    @hugr_op(unsupported_op("pop"), effects=[Effect.ANY])  # Effects unused
```
acl-cqc, line 56: hmmm, could wish that we didn't need to specify effects here! Any value would do so perhaps we should prefer `()` ?

maximilianruesch: I propose that you create special decorators for "unsupported hugr ops" (like usages in this file) and "unsupported builtin" (for usage in `guppylang/src/guppylang/std/unsupported.py` and similar like `guppylang/src/guppylang/std/string.py`). This would not only deduplicate code, but you can also explain *once* why you are fine with specifying any effects here (since they will never get picked up).

And / or: You create a special effect `NEVER`, where the framework errors if it detects its usage / propagation (as a sanity check). But I am also fine with just the above.

guppylang/src/guppylang/std/quantum/__init__.py
```diff
@@ -66,14 +66,16 @@ def __bool__(self: "Measurement") -> bool:
...
+# totally ordered with respect to other (non-try)QAlloc's for predictability.
```
maximilianruesch, line 70:
```suggestion
# totally ordered with respect to other (fallible) qubit allocations for predictability.
```

guppylang/src/guppylang/std/qsystem/utils.py
```diff
@@ -8,7 +8,8 @@
 from guppylang_internals.std._internal.util import external_op
 
 
-@hugr_op(external_op("GetCurrentShot", [], ext=QSYSTEM_UTILS_EXTENSION))
+# No effects: shot number is a constant (not state) for each run
```
maximilianruesch, line 11:
```suggestion
# No effects: result is constant for each particular shot
```

guppylang/src/guppylang/std/mem.py
```diff
@@ -14,7 +14,7 @@
 Out = guppy.type_var("Out", copyable=False, droppable=False)
 
 
-@custom_function(MemSwapCompiler())
+@custom_function(MemSwapCompiler(), effects=())
```
maximilianruesch, line 17: This has no effect since it compiles down to a wire swap?

guppylang/src/guppylang/std/array.py
```diff
@@ -110,6 +114,7 @@ def copy[T: Copy, n: nat](self: array[T, n]) -> array[T, n]:
         ArrayIsBorrowedCompiler(),
         checker=ArrayIndexChecker(),
         unitary_flags=UnitaryFlags.Dagger,
+        effects=(),  # Can panic, but preserving legacy behaviour (https://github.com/Quantinuum/guppylang/issues/2122)
```
maximilianruesch, line 117: This comment also applies to the `put` and `discard_all_taken` as well, no?

guppylang/src/guppylang/std/platform.py
```diff
@@ -181,6 +181,7 @@ def panic(msg: str, signal: int = 1, *args):
         NoneType(),
     ),
     has_var_args=True,
+    effects=(),
```
maximilianruesch, line 184: An exit should be ordered w.r.t. panics, so this should have effects imo. Or add a TODO comment with reference to the issue. Or record it in the issue. I just want to ensure that we do not miss it.

guppylang/src/guppylang/std/qsystem/helios/__init__.py
```diff
@@ -168,13 +169,13 @@ def measure_and_reset(q: qubit) -> Measurement:
     return lazy_measure_and_reset(q)
 
 
-@hugr_op(quantum_op("Reset", ext=QSYSTEM_HELIOS_EXTENSION))
+@hugr_op(quantum_op("Reset", ext=QSYSTEM_HELIOS_EXTENSION), effects=())
```
maximilianruesch, line 172: Why does this not have effects (or if it should, it should have a TODO comment)?

guppylang/src/guppylang/std/qsystem/random.py
```diff
@@ -66,17 +66,18 @@ def random_clifford_angle(self: "RNG") -> angle:
...
         """Discard the random number generator."""
 
-    @custom_function(RandomIntCompiler())
+    @custom_function(RandomIntCompiler(), effects=())
     @no_type_check
     def random_int(self: "RNG") -> int:
         """Generate a random 32-bit signed integer."""
 
-    @hugr_op(external_op("RandomFloat", [], ext=QSYSTEM_RANDOM_EXTENSION))
+    # No effects: all state is carried inside the borrowed `self`
```
maximilianruesch, line 74: This comment really only works with an internal guide to what "effect" means (i.e. some *externally observable mutation*)

guppylang/src/guppylang/std/num.py
maximilianruesch, line 1: I think some of these may panic, for example if there is a nat conversion from a negative int value. Could you fix these, or add a todo comment?

maximilianruesch (changes requested):
I had to keep myself from going over on questioning the effects. I still think we should already make an effort to record all of the cases we would want to look at again (or add todos so we can grep for them later).

Also, does this not close an issue already? I had a feeling it would, but I may also be wrong.

guppylang/src/guppylang/std/qsystem/helios/__init__.py
acl-cqc: it zeroes a qubit, right? qubit -> qubit, no side effects

maximilianruesch: Hmm, I thought that measuring induces a side effect. But perhaps I confused it with free, which has a side effect on the number of qubits active.

PR #2271, opened 2026-09-02
feat: Enable panic stack traces in the emulator

### LINKED ISSUE #1761: [Feature]: Update emulator interface to show stack traces if available
### Problem Statement

Once released selene has the stack trace contained within the exceptions it provides (requires debug information handling to be released with hugr and tket), this should be processed in the interface to show useful information about panic locations when available

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

tatiana-s:
Closes https://github.com/Quantinuum/guppylang/issues/1761

Note that the debug mode only works if no optimisation has been done on the HUGR (and there is only an automatic check and error for this when using it through the emulator interface, otherwise this is on the user).

Requires a selene release with https://github.com/Quantinuum/selene/pull/210 for CI / merging, so draft until then.


guppylang/src/guppylang/optimizer.py
```diff
@@ -257,7 +266,18 @@ def emulator(
         platform: Platform | None = None,
         debug_mode: bool = False,
     ) -> EmulatorInstance:
-        """Compile this function for emulation with the configured optimizations."""
+        """Compile this function for emulation with the configured optimizations.
+
+        Emulation in debug mode (enabling panic traces) requires minimal optimization.
```
aborgna-q, line 271: What happens when optimization is enabled?
Do stack traces get mangled, or do we lose **all** the info?

Could this be a warning instead (once we have those https://github.com/Quantinuum/guppylang/issues/1654)?

guppylang/src/guppylang/emulator/stack_trace.py
```diff
@@ -0,0 +1,60 @@
...
+    """
+    if stack_trace is None:
+        return None
+
+    source = SourceMap()
+    blocks: list[str] = []
+
+    def append_frame_snippet(span: Span, label: str | None, function_name: str) -> None:
+        renderer = DiagnosticsRenderer(source)
+        renderer.render_snippet(
```
aborgna-q, line 29: If the source code has been updated since compilation this may fail and hide the actual panic.

Could we put it in a `try` block?

guppylang/src/guppylang/emulator/exceptions.py
```diff
@@ -14,11 +14,36 @@ def __init__(
...
+        if stack_trace is not None:
+            from .stack_trace import render_stack_trace
+
+            message = getattr(
+                underlying_exception, "message", str(underlying_exception)
+            )
+            rendered = render_stack_trace(stack_trace, message)
+            if rendered is not None:
+                header = EmulatorError._panic_header(underlying_exception)
+                return f"{header}\n{rendered}"
```
aborgna-q, line 36: This hides any `--- stdout ---` or `--- stderr ---` in the underlying selene exception (coming after the stack trace).

aborgna-q (commented):
Looks great!
Just some suggestions

guppylang/src/guppylang/optimizer.py
tatiana-s: The main issues arising so far is stack traces being squashed due to LLVM tail calls and also https://github.com/Quantinuum/tket2/issues/1964 - I think at least until the latter is solved I think having it error is better as showing wrong code snippets just looks confusing even with a warning, but later I think keeping it as a warning and just showing partial traces when some of the info is optimised is an option.

aborgna-q: 👍 

I would add a non-doc comment explaining that, and linking to the issue.

aborgna-q (approved):
See my last comment about linking to the optimization issue, but otherwise LGTM

PR #2192, opened 2026-08-11
feat: Checking custom modifiers

nicolaassolini-qntm:
This PR introduces custom modifiers. They are implemented via the `@guppy.unitary` decorator, which allows us to associate a function with its controlled and daggered versions. Since we can currently only check these new Guppy objects, they are marked as experimental.

First step to complete #1724

PR #2274, opened 2026-09-03
feat: Compilation for custom modifier implementations

### LINKED ISSUE #1724: [Feature][Tracking]: Custom Modifier
see also: https://github.com/Quantinuum/guppylang/issues/585

TODO list:

#### CHECKING

- [x] First PR https://github.com/Quantinuum/guppylang/pull/2192 achiving:
  - [x] define decorator
  - [x] Discard the presence of extra methods inside the `@guppy.unitary` class
  - [x] checking if custom implementation are fine; we raise an error in the following cases: 
    - If a function has both `daggered` and `controlled` implementations, it must
        also have a `ctrl_daggered` implementation, unless the function is marked
        as `unitary=True`.
    - If a function is marked as `controllable=True` and has a `daggered`
      implementation, it must also have a `ctrl_daggered` implementation
      or the function is marked as `unitary=True`.
    - If a function is marked as `daggerable=True` and has a `controlled`
      implementation, it must also have a `ctrl_daggered` implementation
      or the function is marked as `unitary=True`.
    - If a function has a `ctrl_daggered` implementation, it must also have
      a `controlled` implementation, unless the function is marked as
      `controllable=True`.
  - [x] Refactor how the registration of custom implementation is done; this must happen in the DEF_STORE. Also, the checking must happen there.
  - [x] Type checking custom implementation signature
  - [x] Test new type errors added in decorator.py
  - [x] unitary checking must be flexible: if the unitary flag come from the custom implementation, then no check; if it comes form the flag on `__call__` then we need to check the body.

----- 
#### COMPILATION
PR: https://github.com/Quantinuum/guppylang/pull/2274

- [x] static analysis to know how many controlled versions are needed
- [x] Compilation of custom implementations
  - [x] Test custom methods are generated properly
  - [x] Test new metadata set in guppy
- [x] on tket2: https://github.com/Quantinuum/tket2/issues/1939
- [x] on tket2: update passes to handle customs: https://github.com/Quantinuum/tket2/issues/1613

nicolaassolini-qntm:
follows #2035
follows #2192 
closes #1724 

Allow the compilation of `@guppy.unitary` object


guppylang-internals/src/guppylang_internals/compiler/core.py
```diff
@@ -144,8 +162,91 @@ def build_compiled_def(self, def_id: DefId, type_args: Inst | None) -> CompiledD
                 defn = defn.compile_outer(self.module, self)
             self.compiled[def_id, mono_args] = defn
             self.worklist[def_id, mono_args] = None
+
+            if (def_id, mono_args) in self.custom_uses_by_call:
+                daggered, controlled, ctrl_daggered = (
+                    self._compile_custom_modifier_uses((def_id, mono_args))
+                )
```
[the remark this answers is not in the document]
nicolaassolini-qntm: I need to check this

guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -72,6 +116,23 @@ def set_debug_info(self, debug_info: DebugRecord) -> None:
     def set_expected_qubits(self, expected_qubits: int) -> None:
         self._node_metadata[MetadataExpectedQubitsHint] = expected_qubits
 
+    def set_modified_defs(
+        self,
+        *,
+        daggered: str | None,
+        controlled: list[str] | None,
```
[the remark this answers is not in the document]
nicolaassolini-qntm: Make sense

tests/error/test_modifier_errors.py
```diff
@@ -7,40 +7,21 @@
...
-    "custom_modified_missing2.py",
-    "custom_modified_missing3.py",
-    "custom_modifiers_missing_ctrl_daggered.py",
-    "custom_unitary_higher_order_requires_dagger.py",
-    "custom_unitary_higher_order_requires_unitary.py",
-    "dagger_loop3.py",
-    "power_arg_typecheck_inside.py",
-    "flags_nested_combined_outer.py",
-    "captured_classical_modified_nested.py",
     "captured_classical_modified.py",
```
nicolaassolini-qntm, line 33: I removed files that no longer require experimental features and sorted the rest alphabetically.

guppylang-internals/src/guppylang_internals/tys/subst.py
```diff
@@ -91,7 +96,12 @@ def _transform_BoundTypeVar(self, ty: BoundTypeVar) -> Type | None:
             ty.idx - len(self.inst),
             ty.copyable,
             ty.droppable,
-            [self.transform(impl) or impl for impl in ty.implements],
+            # NICOLA: TODO:
+            # fix: arg.bound_vars (`BoundTypeVar.bound_vars`(line 178)) evaluates {self}
+            # hashing the frozen dataclass.
+            # The bug was triggered by is_concrete_inst() implementation accesses
+            # arg.bound_vars
+            tuple(self.transform(impl) or impl for impl in ty.implements),
```
nicolaassolini-qntm, line 99: arg.bound_vars (`BoundTypeVar.bound_vars`(line 178)) evaluates {self} hashing the frozen dataclass.

When is_concrete_inst() accesses arg.bound_vars raised an error since implements was a list (not hashable)

pyproject.toml
```diff
@@ -32,7 +32,8 @@ guppylang-internals = { workspace = true }
 miette-py = { workspace = true }
 # Uncomment these to test the latest dependency version during development
 # hugr = { git = "https://github.com/quantinuum/hugr", subdirectory = "hugr-py", rev = "a4da8ef" }
-# tket = { git = "https://github.com/quantinuum/tket2", subdirectory = "tket-py", rev = "cdb47dd" }
+tket = { git = "https://github.com/quantinuum/tket2", subdirectory = "tket-py", rev = "cecdc83" }
+selene-hugr-qis-compiler = { git = "https://github.com/quantinuum/tket2", subdirectory = "qis-compiler", rev = "cecdc83" }
```
nicolaassolini-qntm: Waiting for [release qis-compiler 0.4.3 (#1984)](https://github.com/Quantinuum/tket2/pull/1984)

tests/integration/test_side_effect_ordering.py
```diff
@@ -412,25 +410,15 @@ def pure_function() -> None:
...
     ] == [True, False, False, True]
-    assert ["pure_function" in get_called_func_name(node) for node in main_calls] == [
-        False,
-        True,
-        True,
-        False,
-    ]
+    assert [
+        "pure_function" in get_called_func_name(hugr, node) for node in main_calls
+    ] == [False, True, True, False]
```
nicolaassolini-qntm: I fixed this test using the robot, in particular.
AI say:

The nested functions are now resolved as direct HUGR `Call` nodes, but the test still searched for the former `LoadFunc → CallIndirect` representation. I updated it to:

- Inspect direct `Call` nodes.
- Reuse the existing `get_called_func_name` helper.

Regarding the other change below:
AI say: Effect analysis also correctly identifies pure_function as effect-free, so only panic_call1 → panic_op → panic_call2 must be connected by order edges. The check_order input was updated accordingly, and the obsolete xfail for unresolved nested-function effects was removed. The specific behaviour changed in commit 69c9036e on September 2, 2026, and reached this branch through merge 7cb15ba0 on September 4. That commit converts a non-capturing nested function reference into a GlobalName. This lets the existing call-registration logic record the exact nested callee instead of treating it as an opaque function value with Effect.ANY.

nicolaassolini-qntm: Inspecting the hugr manually the AI comment makes sense, not sure why before the test was fine? I think it is an xfail on Alan branch

tests/integration/test_modifier.py
aborgna-q, line 1: We're not handling custom control definitions on higher order functions.

This test case fails:
```py

def test_controlled_impl_is_executed_through_higher_order_wrapper(
    use_experimental_features, run_int_fn
):
    @guppy.unitary
    class custom_gate:
        n = guppy.nat_var("n")

        @guppy
        def __call__(q: qubit) -> None:
            pass

        @guppy
        def controlled(q: qubit, _controls: array[qubit, n]) -> None:
            x(q)

    @guppy(controllable=True)
    def apply(f: Controllable[[qubit], None], q: qubit) -> None:
        f(q)

    @guppy
    def main() -> int:
        target = qubit()
        control_qubit = qubit()
        x(control_qubit)
        with control(control_qubit):
            apply(custom_gate, target)
        result = measure(target).read()
        discard(control_qubit)
        return 1 if result else 0

    run_int_fn(main, expected=1, num_qubits=2)
```

aborgna-q (dismissed):

tests/integration/test_modifier.py
nicolaassolini-qntm: I see...

guppylang-internals/src/guppylang_internals/analysis/modifier.py
maximilianruesch, line 1: This file can just be `modifiers.py`

tests/integration/test_modifier.py
nicolaassolini-qntm: Resolved: it was an issue with the implementation of the analysis, and I wasn’t propagating the context of local modifiers through the graph. It’s all sorted now.

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
```diff
@@ -75,6 +75,21 @@ class UnknownModifierError(Error):
...
+class RecursiveModifierControlCountError(Error):
+    title: ClassVar[str] = "Diverging controlled recursive custom modifier definition"
+    span_label: ClassVar[str] = (
+        "This recursive call increases the number of controls from "
+        "{previous_count} to {control_count}"
+    )
+    message: ClassVar[str] = (
+        "Custom controlled implementations that recursively increase the number of "
+        "controllers cannot be monomorphized. "
+    )
```
maximilianruesch, line 89:
```suggestion
    message: ClassVar[str] = (
        "Custom controlled implementations that recursively increase the number of "
        "controllers cannot be monomorphized."
    )
```
Also "monomorphized" is a word which not too many users will be familiar with. Is there some other word you can use?

guppylang-internals/src/guppylang_internals/checker/core.py
```diff
@@ -445,6 +449,9 @@ class Context(NamedTuple):
     together with the AST node that gives rise to said constraint."""
     current_caller: MonoDefId | None = None
 
+    #: Stores under which modifiers context a function is being called.
```
maximilianruesch:
```suggestion
    #: Active modifiers the function is being called with
```

tests/error/modifier_errors/custom_ctrl_daggered_missing_daggered.err
```diff
@@ -0,0 +1,15 @@
+Error: Invalid `@guppy.unitary` implementation (at $FILE:7:0)
+   | 
+ 5 | 
+ 6 | @guppy.unitary
+ 7 | class foo:
+   | ^^^^^^^^^^
+   | ...
+14 |         pass
```
maximilianruesch: Can we get this line a bit more context, i.e. the two above it as well? Just the "pass" is maybe a bit confusing. But this is not really urgent.

tests/error/modifier_errors/custom_ctrl_daggered_missing_daggered.err
```diff
@@ -0,0 +1,15 @@
...
+   | ^^^^^^^^^^^^
+
+A `@guppy.unitary` class declared as controllable and implementing
```
maximilianruesch: Is such a class declared as controllable iff there is a `__call__` method that is controllable? Are there other mechanisms? We should adjust this message to be a bit clearer, but I am so far unsure to what.

guppylang/src/guppylang/decorator.py
```diff
@@ -84,6 +81,7 @@
     from tket.metadata import InlineAnnotationValue
 
 type Decorator[S, T] = Callable[[S], T]
+type CustomModifierDefinitions = dict[CustomModifierKind, RawFunctionDef]
```
```suggestion
type _CustomModifierDefinitions = dict[CustomModifierKind, RawFunctionDef]
```

guppylang/src/guppylang/decorator.py
```diff
@@ -390,28 +388,33 @@ def main(q: qubit) -> None:
         object.__setattr__(call_raw_func, "name", cls.__name__)
 
         # Update the unitary metadata according to the custom implementations
-        custom_modified_definition = _get_custom_methods(cls)
+        custom_modified_definitions = _get_custom_methods(cls)
```
```suggestion
        custom_modifier_definitions = _get_custom_methods(cls)
```
Or the other way round, but we should be consistent about the naming here

guppylang/src/guppylang/decorator.py
```diff
@@ -941,20 +944,16 @@ def _get_unitary_call_def(cls: object) -> GuppyDefinition:
 
 def _get_custom_methods[T](
```
```suggestion
def _get_custom_modifier_definitions[T](
```

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -261,8 +258,8 @@ def check(self, type_args: Inst, globals: Globals) -> "CheckedFunctionDef":
             type_args,
             globals,
             mono_link_name,
-            def_id=self.id,
-            decorator_unitary_flags=self.decorator_unitary_flags,
+            self.id,
+            self.decorator_unitary_flags,
```
maximilianruesch: Why remove these?

tests/integration/test_poly_py312.py
```diff
@@ -439,7 +439,8 @@ def main(q: qubit) -> None:
         # We have to skip this test due to https://github.com/Quantinuum/guppylang/issues/2244
         # apply_daggerable_and_controllable(custom_unitary, q)
 
-    main.check()
+    validate(main.with_minimal_opt().compile_function())
+    validate(main.compile_function())
```
maximilianruesch, line 444: If there is a good reason to do it twice, please state the reason in a short comment above these two calls

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1421,8 +1421,16 @@ def function_def_value_to_function_value(
...
+        from guppylang_internals.experimental import check_capturing_closures_enabled
+
+        try:
+            check_capturing_closures_enabled(expr)
+            # Expression must be preserved to retain a possible closure.
+            return with_type(ty.sig, expr)
+        except GuppyError:
+            name = ENGINE.get_parsed(ty.def_id).name
+    else:
+        name = DEF_STORE.raw_defs[ty.def_id].name
```
maximilianruesch: Can you explain what this achieves? I suspect it is taken from https://github.com/Quantinuum/guppylang/pull/2279/changes. Why do you need it here?

guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -154,3 +202,36 @@ def add_unitary_metadata(
...
+    daggered: str | None = None,
+    controlled: list[str] | None = None,
+    ctrl_daggered: list[str] | None = None,
+) -> None:
+    """Adds the names of the functions implementing custom modifications, ensuring
+    reserved keys aren't overwritten."""
+    if daggered is not None:
+        if DAGGERED_KEY in node_metadata:
+            raise GuppyError(MetadataAlreadySetError(None, DAGGERED_KEY))
+        node_metadata[DAGGERED_KEY] = cast("JsonType", daggered)
```
maximilianruesch: Why is this cast required? Is it because of the `None`? We should add a `# type: ignore` here instead if we can.

guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -23,6 +23,50 @@ class MetadataUnitaryFlags(Metadata[int]):
...
+
+    KEY = "tket.num_control_qubits"
+
+
+# Metadata keys for modified definitions (daggered, controlled, ctrl-daggered)
+# To be removed when added to tket
+DAGGERED_KEY = DaggeredImplementation.KEY
+CONTROLLED_KEY = ControlledImplementations.KEY
+CTRL_DAGGERED_KEY = CtrlDaggeredImplementations.KEY
+NUM_CONTROL_QUBITS_KEY = NumControlQubits.KEY
```
maximilianruesch: These should just be inlined by now, and the comment is to be removed.

guppylang-internals/src/guppylang_internals/analysis/modifier_callgraph.py
```diff
@@ -0,0 +1,233 @@
...
+            raise GuppyError(
+                RecursiveModifierControlCountError(
+                    custom_call.call,
+                    previous_count,
+                    custom_use.control_count,
+                )
+            )
+
+
+def _call_ancestors[T](callee: T, callers: Mapping[T, set[T]]) -> set[T]:
```
maximilianruesch: This can be a member function of the callgraph, can it not?

guppylang-internals/src/guppylang_internals/tys/subst.py
```diff
@@ -27,6 +27,11 @@
 PartialInst = Sequence["Argument | None"]
 
 
+def is_concrete_inst(inst: Inst) -> bool:
+    """Whether an instantiation contains no opaque bound variables."""
+    return all(not arg.bound_vars for arg in inst)
```
maximilianruesch, line 32: In previous checks a recursive finder was employed to find the bound vars. Why is this not needed in this function?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -451,10 +469,7 @@ def register_generic_use(self, defn: CheckableGenericDef, type_args: Inst) -> No
 
         Adds the instantiation to the worklist and ensures that it will be checked.
         """
-        finder = BoundVarFinder()
-        for arg in type_args:
-            arg.visit(finder)
-        if not finder.bound_vars:
+        if is_concrete_inst(type_args):
```
maximilianruesch, line 476: See https://github.com/Quantinuum/guppylang/pull/2274/changes#r3949335664

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -369,24 +404,26 @@ def get_parsed(self, id: DefId) -> ParsedDef:
...
+                assert isinstance(parsed_custom_defn, ParsedFunctionDef)
+                _check_modified_def_signature(parsed_custom_defn, defn.ty)
+                # While parameterized custom methods (controlled and ctrl_daggered) are
+                # scheduled for check by `get_parsed`, (see
+                # ```
+                # 412| elif isinstance(defn, CheckableGenericDef) and defn.params:
+                # ```
+                # ), non-parameterized custom methods (daggered) are not scheduled
+                # thus we explicitly schedule them here. This ensure that all custom
+                # methods are checked even if not used.
```
maximilianruesch: Is this not rather something to fix inside `get_parsed`? Can you elaborate more on why you did it here?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -319,6 +348,12 @@ def register_call(
         self.call_graph[ctx.current_caller].append((callee.id, inst))
         if isinstance(callee, CallableEffects):
             self.register_effects((callee.id, inst), callee.call_effects)
+        elif isinstance(callee, CallableDef):
+            # Effects not known yet, will be computed.
```
maximilianruesch: This comment is not really fitting here

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -319,6 +348,12 @@ def register_call(
...
+            callee_mono_def_id: MonoDefId = (callee.id, inst)
+            edge = (ctx.current_caller, callee_mono_def_id)
+            modifier_contexts = self.local_modifiers_by_edge.setdefault(edge, {})
+            modifier_contexts.setdefault(ctx.modifier_ctx, call_node)
```
maximilianruesch: I believe this should be applied regardless of whether effects are applied? In which case this should be an `if` and not an `elif`.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -252,6 +270,13 @@ class CompilationEngine:
     # as calls are checked, to be then used for effects checking.
     call_graph: dict[MonoDefId, list[MonoDefId]]
     func_effects: dict[MonoDefId, set["Effect"]]
+    #: Distinct modifier contexts used on each monomorphized call-graph edge. The value
+    #: stores one representative call site for future diagnostics.
+    local_modifiers_by_edge: dict[CallGraphEdge, dict[ModifierContext, "AstNode"]]
```
maximilianruesch: Future diagnostics? Are any of them here yet? Also, if you only need the span, could it just be a `ToSpan` as a value instead of an AstNode?

guppylang-internals/src/guppylang_internals/analysis/modifier_callgraph.py
```diff
@@ -0,0 +1,233 @@
...
+        # The raw graph may contain the same callee once per call site. Modifier labels
+        # are already grouped by edge and local context, so visit each raw edge once.
+        for raw_callee in dict.fromkeys(raw_calls.get(caller, ())):
+            edge = (caller, raw_callee)
+            local_contexts = local_modifiers_by_edge.get(edge)
+            if local_contexts is None:
+                # `register_call` also records calls to definitions such as `panic`
+                # whose effects are supplied directly in `func_effects`. They have no
+                # checked body and therefore no local modifier label. Just preserve
+                # their edge: there is no body to analyze.
```
maximilianruesch: There is no need to talk about effects here, it is completely valid to record calls to functions that have no body (in the future, declarations will be included too).

guppylang-internals/src/guppylang_internals/analysis/modifier_callgraph.py
```diff
@@ -0,0 +1,233 @@
...
+    """A contextual call edge resolved to a concrete custom implementation."""
+
+    caller: ModifierCallState
+    callee: ModifierCallState
+    custom_use: ConcreteCustomUse
+    call: AstNode
+
+
+@dataclass(frozen=True)
+class ModifierCallGraphAnalysis:
```
maximilianruesch:
```suggestion
class ModifierAnalysisResult:
```

guppylang-internals/src/guppylang_internals/analysis/modifier.py
```diff
@@ -0,0 +1,233 @@
...
+    call: AstNode
+
+
+@dataclass(frozen=True)
+class ModifierCallGraphAnalysis:
+    """Result of propagating modifiers through a monomorphized call graph."""
+
+    expanded_calls: dict[MonoDefId, list[MonoDefId]]
+    resolved_calls: dict[EdgeWithModifierContext, MonoDefId]
+    custom_uses_by_mono_def: dict[MonoDefId, ConcreteCustomUse]
```
maximilianruesch, line 74: Please add some doc strings for these properties

guppylang-internals/src/guppylang_internals/analysis/modifier_callgraph.py
```diff
@@ -0,0 +1,233 @@
...
+                    # No custom implementation consumed the modifiers. The compiler
+                    # generates the modified callee, so its body inherits the context.
+                    next_state = ModifierCallState(raw_callee, effective_context)
+                    contextual_callers[next_state].add(state)
+                    if raw_callee in raw_calls:
+                        worklist.append(next_state)
+                    continue
+
+                # A custom implementation consumes the effective modifier context. Its
+                # body is consequently analyzed with an empty inherited context.
```
maximilianruesch: Can you explain how a custom implementation "consumes" the context? It is not clear to me why this is happening.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -586,7 +603,30 @@ def check(self, def_ids: list[DefId], *, reset: bool = True) -> None:
...
+        # fixed point.
+        self._drain_check_worklists()
+        while True:
+            modifier_analysis = analyze_modifier_calls(
+                entry_points,
+                self.call_graph,
+                self.local_modifiers_by_edge,
+                self._resolve_modified_call,
+            )
+            self.resolved_modified_calls = modifier_analysis.resolved_calls
```
maximilianruesch, line 623: The result here probably does not change for a specific key once set? In which case it may be worthwhile to pass the already derived results back into the analysis as a starting point?

BUT: This should be done as a follow up (so you can open an issue), not in this PR, even if my guess is right. It is a performance optimisation, and we do not know whether this has big performance implications yet.

tests/integration/test_custom_modifier_callgraph.py
```diff
@@ -0,0 +1,292 @@
...
+
+@guppy
+def _same_count_helper[n: nat](q: qubit, controls: array[qubit, n]) -> None:
+    with control(controls):
+        _same_count_recursive_gate(q)
+
+
+def test_effects_after_custom_modifier_resolution(
+    use_experimental_features,
+):
```
maximilianruesch:
```suggestion
def test_effects_after_custom_modifier_resolution():
```
These can go away, since this is not an experimental feature anymore

tests/integration/test_custom_modifier_callgraph.py
```diff
@@ -0,0 +1,292 @@
...
+    """Effects are computed according to custom modifier calls"""
```
maximilianruesch:
```suggestion
    """Effects are propagated through custom modifier calls."""
```

tests/integration/test_custom_modifier_callgraph.py
```diff
@@ -0,0 +1,292 @@
...
+    @guppy
+    def main(q: qubit, c: qubit) -> None:
+        with control(c):
+            custom_gate_2(q)
+
+    main.check()
+    [custom_use] = ENGINE.custom_uses_by_mono_def.values()
+    effects = compute_effects(CallGraph(ENGINE.call_graph), ENGINE.func_effects)
+    assert effects[custom_use.custom_def] == frozenset({Effect.ANY})
+    assert effects[main.id, ()] == frozenset({Effect.ANY})
```
maximilianruesch, line 105: Can this be a separate test case?

tests/metadata/test_metadata.py
```diff
@@ -94,6 +99,31 @@ def test_add_metadata_metadata_already_set():
...
+
+@pytest.mark.parametrize("key", [DAGGERED_KEY, CONTROLLED_KEY, CTRL_DAGGERED_KEY])
+def test_add_metadata_custom_implementation_names_reserved(key: str):
+    with pytest.raises(
+        GuppyError,
+        check=lambda e: (
+            isinstance(e.error, ReservedMetadataKeysError) and e.error.keys == {key}
+        ),
+    ):
+        add_metadata(NodeMetadata(), additional_metadata={key: "custom_name"})
```
maximilianruesch, line 180: Do both of these not have a respective other, one for already set, and reserved?

tests/integration/test_modifier.py
```diff
@@ -461,7 +461,286 @@ def main(q: qubit) -> None:
         apply_plain2(custom_unitary, q)
         apply_plain2(custom_call, q)
 
-    main.check()
+    validate(main.compile_function())
+
+
+def test_controlled_impl_is_executed(use_experimental_features, run_int_fn):
```
maximilianruesch:
```suggestion
def test_controlled_impl_is_executed(run_int_fn):
```
I suppose these can go as well

tests/integration/test_modifier.py
```diff
@@ -461,7 +461,286 @@ def main(q: qubit) -> None:
...
+        discard(control_qubit)
+        return 1 if result else 0
+
+    run_int_fn(main, expected=1, num_qubits=2)
+
+
+def test_controlled_impl_through_higher_order_call_is_executed(
+    use_experimental_features, run_int_fn
+):
+    """A propagated control selects custom metadata through a callable wrapper."""
```
maximilianruesch:
```suggestion
    """Control is propagated through higher order function calls."""
```

tests/integration/test_modifier.py
```diff
@@ -461,7 +461,286 @@ def main(q: qubit) -> None:
...
+
+        controlled_result = measure(controlled_target).read()
+        daggered_result = measure(daggered_target).read()
+        discard(control_qubit)
+        return (1 if controlled_result else 0) + (2 if daggered_result else 0)
+
+    run_int_fn(main, expected=3, num_qubits=3)
+
+
+def test_double_daggered_is_properly_solved(use_experimental_features, run_int_fn):
```
maximilianruesch:
```suggestion
def test_double_daggered(use_experimental_features, run_int_fn):
```

tests/integration/test_modifier.py
```diff
@@ -461,7 +461,286 @@ def main(q: qubit) -> None:
...
+        def __call__(q: qubit) -> None:
+            x(q)
+
+        @guppy
+        def daggered(q: qubit) -> None:
+            pass
+
+        @guppy
+        def ctrl_daggered(q: qubit, _controls: array[qubit, n]) -> None:
+            pass
```
maximilianruesch: Are these two necessary to implement for this gate?

maximilianruesch (changes requested):
Big PR, but largely understandable. Propagation of these values is a complex task, and while I am sure the whole compiler can do with a few good refactors, this is going into the weeds largely the right amount 👍

guppylang-internals/src/guppylang_internals/definition/function.py
nicolaassolini-qntm: we usually dont use keywords for non-optional args

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
nicolaassolini-qntm: Some merging noise, I suppose, I should be able to revert it

guppylang-internals/src/guppylang_internals/metadata/common.py
nicolaassolini-qntm: Not needed indeed, (only below due to "list[str]")

tests/integration/test_modifier.py
nicolaassolini-qntm: Not really

tests/integration/test_side_effect_ordering.py
nicolaassolini-qntm: The reason of the change was: https://github.com/Quantinuum/guppylang/pull/2274#discussion_r3949140642

tests/metadata/test_metadata.py
nicolaassolini-qntm: added

guppylang-internals/src/guppylang_internals/tys/subst.py
nicolaassolini-qntm: The recursion login is in`bound_vars`. `bound_vars` calls `bound_vars`recursevely on type elements.

guppylang-internals/src/guppylang_internals/engine.py
nicolaassolini-qntm: > should be an if and not an elif.

Yes, but since `callee: "CallableDef"` the if is useless

guppylang-internals/src/guppylang_internals/engine.py
nicolaassolini-qntm: This is a workaround to be sure to check the daggered method even when is not direclty call (see: `tests/error/type_errors/unitary_function_bad_method.py`). `controlled` and `ctrl_daggered` are already registered in generic_to_check_worklist since they are generic, maybe a better solution is to have an ad hoc list?

tests/error/modifier_errors/custom_ctrl_daggered_missing_daggered.err
nicolaassolini-qntm: This require changing the diagnostic logic (maybe introducing a flag to show more lines or changing the number of lines that are shown when the span is big?). But I wuold open a issue and leave it for a future PR

tests/error/modifier_errors/custom_ctrl_daggered_missing_daggered.err
nicolaassolini-qntm: > Is such a class declared as controllable iff there is a __call__ method that is controllable? 

Yes

guppylang-internals/src/guppylang_internals/checker/modifier.py
```diff
@@ -0,0 +1,68 @@
...
+        if self.daggered and self.control_sizes:
+            return CustomModifierKind.CTRL_DAGGERED
+        if self.daggered:
+            return CustomModifierKind.DAGGERED
+        if self.control_sizes:
+            return CustomModifierKind.CONTROLLED
+        return None
+
+
+NO_CALL_MODIFIERS = ModifierContext()
```
nicolaassolini-qntm, line 68: I think `UNMODIFIED_CALL` is better

guppylang-internals/src/guppylang_internals/analysis/modifier_callgraph.py
nicolaassolini-qntm: better now?

guppylang-internals/src/guppylang_internals/engine.py
nicolaassolini-qntm: Updated comment and type

guppylang-internals/src/guppylang_internals/checker/modifier.py
maximilianruesch: I actually disagree, `UNMODIFIED_CALL` makes it sound like the call itself is modified / rerouted / arguments are changing, etc. You only mean a specific thing of those, namely rerouting, and that was clearer with the previous constant name.

tests/error/modifier_errors/custom_ctrl_daggered_missing_daggered.err
maximilianruesch: Can you change the span to just point to the class name? Or do you think it is better if it covers the entire class?

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
maximilianruesch: Do you have suggestions for this?

guppylang/src/guppylang/decorator.py
maximilianruesch: By "the other way round", I meant using `custom_modified_definitions`. `custom_methods` is fairly ambiguous... The type alias name should be updated as well for consistency. Choose either "modifier" or "modified".

guppylang-internals/src/guppylang_internals/definition/function.py
maximilianruesch: But they were already there, are 80 lines removed from other changes you made, and do not hurt you. Please revert this change to keep your diff clean.

tests/integration/test_custom_modifier_callgraph.py
```diff
@@ -0,0 +1,294 @@
...
+    custom_main.check()
+    effects = compute_effects(CallGraph(ENGINE.call_graph), ENGINE.func_effects)
+    [custom_use] = ENGINE.custom_uses_by_mono_def.values()
+    assert effects[custom_use.unmodified_callee] == frozenset({Effect.ANY})
+    assert effects[custom_use.custom_def] == frozenset()
+    assert effects[custom_main.id, ()] == frozenset()
+
+
+def test_effects_after_custom_modifier_resolution_2():
+    """Effects are propagated through custom modifier calls 2"""
```
maximilianruesch: Just saying "2" does not add any value to the docstring. What is the difference to the previous test case? There must be something that is different, otherwise it would not be worth having two test cases.

tests/integration/test_custom_modifier_callgraph.py
```diff
@@ -0,0 +1,294 @@
...
+            custom_gate_2(q)
+
+    main.check()
+    [custom_use] = ENGINE.custom_uses_by_mono_def.values()
+    effects = compute_effects(CallGraph(ENGINE.call_graph), ENGINE.func_effects)
+    assert effects[custom_use.custom_def] == frozenset({Effect.ANY})
+    assert effects[main.id, ()] == frozenset({Effect.ANY})
+
+
+def test_recursive_custom_modifier_same_control_count(use_experimental_features):
```
maximilianruesch:
```suggestion
def test_recursive_custom_modifier_same_control_count():
```
Left overs? I suggest you check the entire file for these.

guppylang-internals/src/guppylang_internals/analysis/modifier.py
```diff
@@ -0,0 +1,242 @@
...
+                    next_state = ModifierCallState(raw_callee, effective_context)
+                    contextual_callers[next_state].add(state)
+                    if raw_callee in raw_calls:
+                        worklist.append(next_state)
+                    continue
+
+                # When we call custom implementation, the modifier context under we
+                # reset the modifier context since the custom call fully implement all
+                # the modification that are needed, and no further modifications are
+                # required to custom implementation body.
```
maximilianruesch:
```suggestion
                # When we call custom implementation, the modifier context under we
                # reset the modifier context since the custom call fully implements all
                # the modifications that are needed, and no further modifications are
                # required to custom implementation body.
```
The last bit of the sentence does not really make sense, what do you mean?

guppylang-internals/src/guppylang_internals/engine.py
nicolaassolini-qntm: You should be right. I opened https://github.com/Quantinuum/guppylang/issues/2291

guppylang-internals/src/guppylang_internals/checker/modifier.py
nicolaassolini-qntm: Change undone

guppylang-internals/src/guppylang_internals/metadata/common.py
```diff
@@ -23,6 +23,48 @@ class MetadataUnitaryFlags(Metadata[int]):
...
+    - `tests/test_guppy_decoupled.py:83`
+    - https://github.com/Quantinuum/guppylang/issues/1595"""
+
+    KEY = "tket.num_control_qubits"
+
+
+# DAGGERED_KEY = DaggeredImplementation.KEY
+# CONTROLLED_KEY = ControlledImplementations.KEY
+# CTRL_DAGGERED_KEY = CtrlDaggeredImplementations.KEY
+# NUM_CONTROL_QUBITS_KEY = NumControlQubits.KEY
```
maximilianruesch:
```suggestion
```

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -319,6 +349,11 @@ def register_call(
         self.call_graph[ctx.current_caller].append((callee.id, inst))
         if isinstance(callee, CallableEffects):
             self.register_effects((callee.id, inst), callee.call_effects)
+        # We record the modifier context under the call is performed.
```
maximilianruesch:
```suggestion
        # Record the modifier context under which the call is performed
```

guppylang-internals/src/guppylang_internals/engine.py
maximilianruesch: The line number is already outdated. And since this is part of `get_parsed` as well, I think you should rather reword this comment to explain why you want to check all custom methods even if not used.

So something like
```suggestion
                # While parameterized custom methods (controlled and ctrl_daggered) are
                # scheduled for check since they are generic, non-parameterized custom
                # methods (daggered) are not scheduled. We explicitly schedule them here
                # because <insert-reason>
```

guppylang-internals/src/guppylang_internals/analysis/modifier_callgraph.py
maximilianruesch: I think I kind of get what you mean, but please reread the comment and check for missing words. I am not entirely sure still.

maximilianruesch (changes requested):

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
nicolaassolini-qntm: I changed a bit title and message, should be better now.

tests/error/modifier_errors/custom_ctrl_daggered_missing_daggered.err
nicolaassolini-qntm: I prefer that the span cover the all class since the error is the entire class definition

nicolaassolini-qntm: NVM I changed my mind (added a function to get the header span of the class and update the errors)

guppylang/src/guppylang/decorator.py
nicolaassolini-qntm: done it

guppylang/pyproject.toml
maximilianruesch, line 1: This PR should wait until #2294 is merged in, or revert this change

guppylang/src/guppylang/decorator.py
```diff
@@ -1017,11 +1012,9 @@ def _set_unitary_metadata(
...
 ) -> UnitaryFlags:
     """Set unitary metadata based on the available custom implementations:
-    - `daggered`: The custom implementation for the daggered modifier, None if absent.
-    - `controlled`: The custom implementation for the controlled modifier,
-      None if absent.
-    - `ctrl_daggered`: The custom implementation for the ctrl_daggered modifier,
-      None if absent.
+    - `daggered`: The implementation of the custom daggered, None if absent.
+    - `controlled`: The implementation of the custom controlled, None if absent.
+    - `ctrl_daggered`: The implementation of the custom ctrl_daggered, None if absent.
```
```suggestion
    - `daggered`: The custom implementation of `daggered`, None if absent.
    - `controlled`: The custom implementation of `controlled`, None if absent.
    - `ctrl_daggered`: The custom implementation of `ctrl_daggered`, None if absent.
```

guppylang-internals/src/guppylang_internals/checker/errors/generic.py
maximilianruesch: Yes, much clearer to the user

tests/integration/test_arithmetic.py
maximilianruesch, line 1: Merge main?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -369,24 +404,28 @@ def get_parsed(self, id: DefId) -> ParsedDef:
...
+                # Ensure that all custom methods are checked, even if they are not
+                # directly used.
+                # Parameterized custom methods (`controlled` and `ctrl_daggered`) are
+                # generic, so they are always added to `generic_to_check_worklist` and
+                # checked later. Non-parameterized custom methods (`daggered`) are not
+                # generic and therefore are not added to `generic_to_check_worklist`.
+                # Normally, non-generic functions are added to `to_check_worklist` only
+                # when they are explicitly called. To ensure that `daggered` methods are
+                # checked even when they are not called, we explicitly add them to
+                # `to_check_worklist`.
```
maximilianruesch: This does not include the reason why we would need to check them if they are not called

guppylang-internals/src/guppylang_internals/span.py
```diff
@@ -125,6 +125,40 @@ def to_span(x: ToSpan) -> Span:
     return Span(start, end)
 
 
+def class_header_span(class_def: ast.ClassDef) -> Span:
```
maximilianruesch, line 128: I am surprised we do not already have this. How do errors for struct class definitions work?

guppylang-internals/src/guppylang_internals/analysis/modifier.py
```diff
@@ -0,0 +1,241 @@
...
+                    # generates the modified callee, so its body inherits the context.
+                    next_state = ModifierCallState(raw_callee, effective_context)
+                    contextual_callers[next_state].add(state)
+                    if raw_callee in raw_calls:
+                        worklist.append(next_state)
+                    continue
+
+                # When calling a custom implementation, we reset the modifier context
+                # because the custom call implements all required modifications. No
+                # further modifications need to be applied to its body.
```
maximilianruesch:
```suggestion
                # An custom implementation exists for this modifier context. A custom implementation requires resetting the modifier context since the body of the implementation should be taken as an as-is implementation, without any modifiers applied to it.
```
I obviously with the max line breaking that ruff dictates.

maximilianruesch (changes requested):
Almost there 🚀

guppylang-internals/src/guppylang_internals/span.py
nicolaassolini-qntm: There was no error pointing to the class header, strange but true

tests/integration/test_arithmetic.py
nicolaassolini-qntm: Cleaned the num.py changes (something went wrong with the merging).

nicolaassolini-qntm:
I also removed a couple of `use_experimental_features`

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -369,24 +404,30 @@ def get_parsed(self, id: DefId) -> ParsedDef:
...
+
+                assert isinstance(parsed_custom_defn, ParsedFunctionDef)
+                _check_modified_def_signature(parsed_custom_defn, defn.ty)
+                # Guppy normally checks only definitions that are called. Custom
+                # modifier methods, however, are resolved implicitly by the compiler.
+                # Thus, to prevent accepting a `@guppy.unitary` class containing a,
+                # ill-typed custom implementation, we check every custom modifier
+                # method. We enforce this because these methods are part of the class
+                # definition. and every custom method in a well-typed `@guppy.unitary`
+                # class must itself be well-typed.
```
maximilianruesch, line 423:
```suggestion
                # Guppy normally checks only definitions that are called. Custom
                # modifier methods, however, are resolved implicitly by the compiler.
                # Thus, to prevent accepting a `@guppy.unitary` class containing a
                # ill-typed custom implementation, we check every custom modifier
                # method. We enforce this because these methods are part of the class
                # definition and every custom method in a well-typed `@guppy.unitary`
                # class must itself be well-typed.
```
Much clearer, only punctuation left.

guppylang-internals/src/guppylang_internals/analysis/modifier.py
```diff
@@ -0,0 +1,242 @@
...
+                    next_state = ModifierCallState(raw_callee, effective_context)
+                    contextual_callers[next_state].add(state)
+                    if raw_callee in raw_calls:
+                        worklist.append(next_state)
+                    continue
+
+                # An custom implementation exists for this modifier context. A custom
+                # implementation requires resetting the modifier context since the body
+                # of the implementation should be taken as an as-is implementation,
+                # without any modifiers applied to it.
```
maximilianruesch:
```suggestion
                # A custom implementation exists for this modifier context. A custom
                # implementation requires resetting the modifier context since the body
                # of the implementation should be taken as an as-is implementation,
                # without any modifiers applied to it.
```
Seems like I typoed myself

maximilianruesch (approved):
Nice feature! 🚀

PR #2277, opened 2026-09-04
feat: Custom modifiers as struct and enum methods

### LINKED ISSUE #2262: [Feature]: Use custom modifier as structure methods
### Problem Statement

E.g.:
```python
@guppy.struct
class S:
    x: int

    @guppy.unitary
    class method:
        @guppy
        def __call__(self: "S") -> int:
            return self.x + 1

        @guppy
        def controlled[c: nat](self: "S", _ctrl: array[qubit, c]) -> int:
            return self.x - 1        


@guppy
def main() -> int:
    return S(42).method()

main.check()
```

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

### LINKED ISSUE #2276: [Feature]: Use custom modifier as enum methods
### Problem Statement

enum version for #2262 

### Proposed Solution

_No response_

### Component

_No response_

### Effort Estimate

None

### Additional Context

_No response_

nicolaassolini-qntm:
follows #2274
follows #2310 
close #2262 
close #2276 

As well as introducing the use of `@guppy.unitary` within structs and enums, I have updated the span in some error messages to ensure consistency


guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -225,6 +251,8 @@ def register_custom_modified_def(
         custom_defs = self.custom_modified_defs[parent_def_id]
         assert kind not in custom_defs, f"Custom {kind.value} already registered"
         custom_defs[kind] = custom_def_id
+        assert custom_def_id not in self.custom_modified_def_parents
+        self.custom_modified_def_parents[custom_def_id] = parent_def_id
```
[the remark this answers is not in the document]
nicolaassolini-qntm: @copilot explain better

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -187,17 +189,41 @@ def __init__(self) -> None:
...
+        member = self.raw_defs[member_id]
+        # Ordinary methods are defined directly in the type's class body, so their
+        # frame must be advanced out of that scope. A unitary method's `__call__` is
+        # defined one class scope deeper and needs to retain that frame to resolve
+        # unitary-class locals such as `n = guppy.nat_var("n")`.
+        is_unitary_call = (
+            isinstance(member, RawFunctionDef) and member.unitary_class_at is not None
+        )
+        self._register_type_member_parent(
+            ty_id, member_id, adjust_frame=not is_unitary_call
```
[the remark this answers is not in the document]
nicolaassolini-qntm: @copilot do an example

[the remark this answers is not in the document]
nicolaassolini-qntm: @copilot  your example is not supported in guppy

guppylang-internals/src/guppylang_internals/engine.py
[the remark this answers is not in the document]
nicolaassolini-qntm: @copilot this program is rejected after https://github.com/Quantinuum/guppylang/pull/2277#:~:text=%232274%0Afollows-,%232310,-close%20%232262%0Aclose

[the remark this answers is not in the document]
nicolaassolini-qntm: After https://github.com/Quantinuum/guppylang/pull/2277, such a thing is not possible, since we can only declare methods inside the class as @guppy methods.

guppylang-internals/src/guppylang_internals/engine.py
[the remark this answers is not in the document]
nicolaassolini-qntm: @copilot this program is not allowed, `n = guppy.nat_var("n")` cannot stay in @guppy.unitary
        class apply_h:

guppylang-internals/src/guppylang_internals/definition/struct.py
```diff
@@ -59,7 +59,7 @@
 class FieldFormHint(Help):
     message: ClassVar[str] = (
         "Struct can contain only fields of the form `name: Type` "
-        "or `@guppy` annotated methods"
+        "`@guppy`, annotated methods or `@guppy.unitary` classes"
```
qartik:
```suggestion
        "`@guppy` annotated methods or `@guppy.unitary` classes"
```

guppylang-internals/src/guppylang_internals/definition/struct.py
```diff
@@ -59,7 +59,7 @@
 class FieldFormHint(Help):
     message: ClassVar[str] = (
         "Struct can contain only fields of the form `name: Type` "
```
qartik:
```suggestion
        "Struct can contain only fields of the form `name: Type`, "
```

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -187,17 +189,41 @@ def __init__(self) -> None:
...
+        )
+
+        # When a `@guppy.unitary` class is used as method, the custom implementations
+        # are member too: their  first argument is the same `self` as the unmodified
```
qartik:
```suggestion
        # are members too: their first argument is the same `self` as the unmodified
```

qartik (commented):
Some feedback from Gemini 3.8 Flash:

> ### Medium Severity
> 
> **Incorrect scope frame for unitary methods in structs and enums**
>    - **File:** `guppylang-internals/src/guppylang_internals/engine.py:211-226`
>    - **Issue:** Passing `adjust_frame=False` retains the inner `@guppy.unitary` class frame for `__call__` and custom modifier implementations (`daggered`, `controlled`, etc.). When a struct/enum is defined inside a local scope (e.g. factory functions or test functions), `Globals` cannot resolve outer symbols such as explicit type annotations (e.g. `self: "S"` from #2262). Furthermore, PR #2310 already rejects non-method statements in `@guppy.unitary` classes, so the rationale in the comment is obsolete.
>    - **Fix:** Advance the frame past both the unitary class and the struct/enum class scopes so it reaches the enclosing definition scope.
> 
> ---
> 
> ### Low Severity / Test Coverage
> 
> **Missing test coverage for `self` in enum unitary methods**
>    - **Files:** `tests/integration/test_enum.py`, `tests/integration/test_modifier.py`
>    - **Issue:** In the added enum tests, `self` is ignored in every custom modifier implementation (`pass` or `x(q)`). None test accessing or pattern matching on `self` (e.g. `match self: case CustomGates.Enabled(): ...`).
>    - **Fix:** Add an integration test where an enum unitary method matches on `self`.

tests/error/enum_errors/unitary_class_overwrite_variant.err
```diff
@@ -0,0 +1,8 @@
+Error: Duplicate variant (at $FILE:8:4)
+  | 
+6 | 
+7 |     @guppy.unitary
+8 |     class name:
+  |     ^^^^^^^^^^^ Enum `MyEnum` already contains a variant named `name`
```
maximilianruesch: This is not *entirely* true, since it does not contain a variant but a member method named ` name`. Can your error message easily be improved, or is this more complicated? Not super important, but wondering about it nontheless.

guppylang-internals/src/guppylang_internals/definition/enum.py
```diff
@@ -99,19 +100,22 @@ def parse(self, globals: "Globals", sources: SourceMap) -> "ParsedEnumDef":
...
+        from guppylang_internals.definition.function import RawFunctionDef
+
         # We look for variants in the class body
         variants: dict[str, EnumVariant[UncheckedField]] = {}
-        used_func_names: dict[str, ast.FunctionDef] = {}
+        used_func_names: dict[str, ast.FunctionDef | ast.ClassDef] = {}
         variant_index = 0
         for i, node in enumerate(cls_def.body):
             match i, node:
-                # TODO: do we allow `pass` statements to define empty enum?
```
maximilianruesch, line 108: Does this have a conclusion / test?

guppylang-internals/src/guppylang_internals/definition/function.py
```diff
@@ -83,7 +83,13 @@ def default_func_link_name(raw_def: "RawFunctionDef | RawFunctionDecl") -> str:
     if (parent_ty_id := DEF_STORE.type_member_parents.get(raw_def.id)) is not None:
         parent = ENGINE.get_parsed(parent_ty_id)
         if isinstance(parent, ParsedStructDef | ParsedEnumDef):
-            return f"{parent.link_name_prefix}.{raw_def.python_func.__name__}"
+            if unitary_call_id := DEF_STORE.custom_modified_def_parents.get(raw_def.id):
+                unitary_call = DEF_STORE.raw_defs[unitary_call_id]
+                return (
+                    f"{parent.link_name_prefix}.{unitary_call.name}."
+                    f"{raw_def.python_func.__name__}"
```
maximilianruesch, line 90: Why not use `raw_def.name` here? Why unwrap the `python_func`?

tests/error/modifier_errors/invalid_custom_in_struct1.py
```diff
@@ -0,0 +1,30 @@
...
+
+@guppy.struct
+class FooStruct:
+    x: int
+
+    @guppy.unitary
+    class foo:
+
+        @guppy
+        def __call__(self,q: qubit) -> None:
```
maximilianruesch, line 16:
```suggestion
        def __call__(self, q: qubit) -> None:
```

tests/error/modifier_errors/invalid_custom_in_struct1.py
```diff
@@ -0,0 +1,30 @@
+"""Ensuring that custom implementations are properly checked"""
```
maximilianruesch, line 1:
```suggestion
```
Or make this more specific to say that this is in structs (and then also what distinguishes it from 2), but I think that it is not needed.

tests/error/modifier_errors/invalid_custom_in_struct2.py
```diff
@@ -0,0 +1,35 @@
+"""Ensuring that custom implementations are properly checked"""
```
maximilianruesch, line 1:
```suggestion
```
Same here

tests/error/modifier_errors/invalid_custom_in_struct2.py
```diff
@@ -0,0 +1,35 @@
...
+
+@guppy.struct
+class FooStruct:
+    x: int
+
+    @guppy.unitary
+    class foo:
+
+        @guppy
+        def __call__(self,q: qubit) -> None:
```
maximilianruesch, line 16:
```suggestion
        def __call__(self, q: qubit) -> None:
```

tests/integration/test_link_name.py
```diff
@@ -120,6 +141,34 @@ def main() -> None:
     }
 
 
+def test_struct_unitary_member_link_name_inferred(qualifier):
```
maximilianruesch, line 144: Could you also add a test using enum for this?

tests/integration/test_enum.py
```diff
@@ -345,3 +347,48 @@ def main() -> None:
         _, _, _ = chain(Shape.Circle(5.0))
 
     validate(main.compile_function())
+
+
+def test_enum_custom_modifier_impls_are_executed(validate):
```
maximilianruesch, line 352: There is no assertion of things actually being executed in this test. Either change the test name (I have a feeling it should be closer to the struct test in one of the related test files), or actually add some (like measuring a result). We probably do not want to duplicate the execution tests from `test_modifier.py`, so I prefer the former.

tests/integration/test_struct.py
```diff
@@ -142,6 +143,51 @@ def main(a: StructA, b: StructB) -> tuple[int, float]:
     validate(main.compile_function())
 
 
+def test_unitary_method_enclosing_scope(validate):
```
maximilianruesch, line 146:
```suggestion
def test_nested_unitary_method(validate):
```
Would that also be fitting? What do you mean with `enclosing_scope` exactly?

guppylang-internals/src/guppylang_internals/definition/enum.py
```diff
@@ -177,21 +181,37 @@ def parse(self, globals: "Globals", sources: SourceMap) -> "ParsedEnumDef":
         # Ensure that functions do not override enum variants
         # and that all functions are Guppy functions
         for func_name, func_def in used_func_names.items():
-            from guppylang.defs import GuppyDefinition
+            match func_def:
+                case ast.ClassDef() as node:
+                    header_span = class_header_span(node)
+                case ast.FunctionDef() as node:
+                    header_span = function_header_span(node)
```
maximilianruesch, line 188: 1. Please refactor this header span extraction into a helper function `extract_header_span(ast_node) -> ToSpan | None`, so you can use this in `guppylang-internals/src/guppylang_internals/definition/struct.py:151` as well
2. Please add a fallback, by stating something like `span = extract_header_span(func_def) or func_def`

guppylang-internals/src/guppylang_internals/definition/struct.py
```diff
@@ -290,3 +309,13 @@ def params_from_ast(nodes: Sequence[ast.expr], globals: Globals) -> list[Paramet
                 continue
         raise GuppyError(ExpectedError(node, "a type parameter"))
     return params
+
+
+def _raise_unexpected_struct_statement(node: ast.stmt) -> NoReturn:
```
maximilianruesch, line 314: I do not think this needs an extraction, it only has two use sites, and the code is very clear inline as well.

guppylang-internals/src/guppylang_internals/definition/struct.py
```diff
@@ -122,20 +143,18 @@ def parse(self, globals: Globals, sources: SourceMap) -> "ParsedStructDef":
...
         if overridden := used_field_names.intersection(used_func_names.keys()):
             x = overridden.pop()
-            raise GuppyError(
-                DuplicateFieldError(used_func_names[x], self.name, x, "struct")
+            error_ast = used_func_names[x]
+            error_span = (
+                function_header_span(error_ast)
+                if isinstance(error_ast, ast.FunctionDef)
+                else class_header_span(error_ast)
             )
```
maximilianruesch, line 156: You should be able to use the suggestion from https://github.com/Quantinuum/guppylang/pull/2277/changes#r4083095262 here

guppylang-internals/src/guppylang_internals/definition/enum.py
```diff
@@ -177,21 +181,37 @@ def parse(self, globals: "Globals", sources: SourceMap) -> "ParsedEnumDef":
...
+                    raise GuppyError(
+                        NonGuppyMethodError(
+                            header_span, self.name, func_name, "enum", "@guppy"
+                        )
                     )
-                )
+                case ast.ClassDef() if not (
+                    isinstance(v, GuppyDefinition)
+                    and isinstance(v.wrapped, RawFunctionDef)
+                    and v.wrapped.unitary_class_at is not None
```
maximilianruesch, line 206: This logic is complicated enough to warrant a helper function `is_guppy_unitary(...) -> bool` imo, this will make this statement much more readable. You can also use it in the struct case.

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -187,17 +189,43 @@ def __init__(self) -> None:
...
+        assert member_id not in self.type_member_parents, "Already a type member"
         self.type_member_parents[member_id] = ty_id
-        # Update the frame of the definition to the frame of the defining class
-        if member_id in self.frames:
+        # Advance past each class and its optional annotation scope.
+        if member_id not in self.frames:
+            return
+        for _ in range(class_scopes):
             frame = self.frames[member_id].f_back
             if frame:
```
maximilianruesch, line 230: Should this not be run once after the for loop?

guppylang-internals/src/guppylang_internals/engine.py
```diff
@@ -187,17 +189,43 @@ def __init__(self) -> None:
         self.sources = SourceMap()
         self.wasm_functions = {}
         self.custom_modified_defs = defaultdict(dict)
+        self.custom_modified_def_parents = {}
 
     def register_def(self, defn: RawDef, frame: FrameType) -> None:
         self.raw_defs[defn.id] = defn
         self.frames[defn.id] = frame
 
     def register_type_member(self, ty_id: DefId, name: str, member_id: DefId) -> None:
```
maximilianruesch, line 198: I am confused: Do we register the unitary class as a type member of the enum / struct, or rather the `__call__` function of the unitary class?

maximilianruesch (changes requested):
A bit of a confusing one, but I like the amount of testing 🚀

guppylang-internals/src/guppylang_internals/definition/enum.py
nicolaassolini-qntm: The comment was a left over, empty enum are fine: https://github.com/Quantinuum/guppylang/blob/c8c31954f814c93c876fd7691b95e7b10a651e50/tests/integration/test_enum.py#L37-L38

tests/error/modifier_errors/invalid_custom_in_struct1.py
nicolaassolini-qntm: Here we are testing that the custom methods signature is checked properly, in 2 we test that a bad defined `@guppy.unitary` (i.e. we miss some custom methods) is spotted also when it is inside struct method. I would keep this test, since we have the `self`; _2 may be not needed.

tests/integration/test_struct.py
nicolaassolini-qntm: robot name, enclosing_scope because the two class scopes are enclosed one in the other. I guess nested is better. We are testing that the method is getting both the type arg from Gates and from apply_h.

tests/integration/test_enum.py
nicolaassolini-qntm: an oversight, I dnt remeber why I added this in the first place, but your first suggestion is the right thing to do

guppylang-internals/src/guppylang_internals/engine.py
nicolaassolini-qntm: the '@guppy.unitary' decorator return the GuppyDefinition of the `__call__` method. Not sure if this answer your question.

guppylang-internals/src/guppylang_internals/engine.py
nicolaassolini-qntm: We do one `f_back` with normal functions and 2 with `.unitary` ones, I think is fine?

guppylang-internals/src/guppylang_internals/definition/enum.py
nicolaassolini-qntm: Why `unction extract_header_span(ast_node) -> ToSpan | None` I can do `def extract_header_span(node: ast.ClassDef | ast.FunctionDef) -> Span` and use an assert_naver to ensure only proper ast are passed

maximilianruesch: If you know at the point of use that no others are passed, even better. I was just not sure whether the type is restricted enough

qartik:
@codex review

PR #2279, opened 2026-09-04
fix: Resolve calls to nested functions iff capturing closures are disabled

### LINKED ISSUE #2272: [Bug]: Callgraph analysis does not resolve calls to nested functions
### Description

Using #2035 , even following #2116, calls to nested functions are not resolved to their true targets in the callgraph; instead we see these as indirect calls (and record the `Effect.ANY`).

The error occurs in `ExprChecker.visit_Call` / `ExprSynthesizer.visit_Call` where the call target (a `PlaceNode`) is correctly reported as having NestedFunctionDefType, causing `visit_Call` to use this that "forgets" the target is a nested function def: https://github.com/Quantinuum/guppylang/blob/b96b18fea772671f1d89747369e0ed25e9cd2920/guppylang-internals/src/guppylang_internals/checker/expr_checker.py#L1304-L1314

A possible fix that doesn't seem to break anything:
```
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -1332,8 +1332,16 @@ def function_def_value_to_function_value(
     their expression must be preserved to retain a possible closure.
     """
     if isinstance(ty, NestedFunctionDefType):
-        return with_type(ty.sig, expr)
-    name = DEF_STORE.raw_defs[ty.def_id].name
+        from guppylang_internals.experimental import check_capturing_closures_enabled
+
+        try:
+            check_capturing_closures_enabled(expr)
+            # Expression must be preserved to retain a possible closure.
+            return with_type(ty.sig, expr)
+        except GuppyError:
+            name = ENGINE.get_parsed(ty.def_id).name
+    else:
+        name = DEF_STORE.raw_defs[ty.def_id].name
     return with_type(ty.sig, with_loc(expr, make_global_name(name, ty.def_id)))
```

### Steps to Reproduce

_No response_

### Expected Behaviour

_No response_

### Actual Behaviour

_No response_

### Component

_No response_

### Environment

```markdown

```

### Additional Context

_No response_

acl-cqc:
fixes #2272

Existing `function_def_value_to_function_value` bailed out for nested functions in case they were not global names but closures. However this can only be the case if experimental features (specifically, capturing closures) are enabled: https://github.com/Quantinuum/guppylang/blob/702bcb09d2aae7c918c0311c6ff4bc5ed6abdb45/guppylang-internals/src/guppylang_internals/checker/func_checker.py#L215-L219

So for the meantime an easy fix is to allow resolving to a GlobalCall if-and-only-if experimental features are off. (When we enable captures, we may want to enable a better fix, for example, recording in the `CheckedNestedFunctionDef` whether that particular function had captures or not.)

### DIFF
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -1421,8 +1421,16 @@ def function_def_value_to_function_value(
     their expression must be preserved to retain a possible closure.
     """
     if isinstance(ty, NestedFunctionDefType):
-        return with_type(ty.sig, expr)
-    name = DEF_STORE.raw_defs[ty.def_id].name
+        from guppylang_internals.experimental import check_capturing_closures_enabled
+
+        try:
+            check_capturing_closures_enabled(expr)
+            # Expression must be preserved to retain a possible closure.
+            return with_type(ty.sig, expr)
+        except GuppyError:
+            name = ENGINE.get_parsed(ty.def_id).name
+    else:
+        name = DEF_STORE.raw_defs[ty.def_id].name
     return with_type(ty.sig, with_loc(expr, make_global_name(name, ty.def_id)))
 
 
--- a/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
+++ b/guppylang-internals/src/guppylang_internals/checker/expr_checker.py
@@ -1430,9 +1430,8 @@ def function_def_value_to_function_value(
             return with_type(ty.sig, expr)
         except GuppyError:
             # Capturing closures disabled
-            name = ENGINE.get_parsed(ty.def_id).name
-    else:
-        name = DEF_STORE.raw_defs[ty.def_id].name
+            pass
+    name = ENGINE.get_parsed(ty.def_id).name
     return with_type(ty.sig, with_loc(expr, make_global_name(name, ty.def_id)))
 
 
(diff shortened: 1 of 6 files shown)


guppylang-internals/src/guppylang_internals/checker/expr_checker.py
```diff
@@ -1421,8 +1421,17 @@ def function_def_value_to_function_value(
...
+        from guppylang_internals.experimental import check_capturing_closures_enabled
+
+        try:
+            check_capturing_closures_enabled(expr)
+            # We don't record whether individual nested functions capture anything,
+            # so conservatively assume it might.
+            return with_type(ty.sig, expr)
+        except GuppyError:
+            # Capturing closures disabled
+            pass
```
maximilianruesch: Please add a boolean function `are_capturing_closures_enabled`, which you can use to implement the `check_` as well. Let us avoid unnecessary try-except clauses when we have access to the source.

maximilianruesch (changes requested):

guppylang-internals/src/guppylang_internals/checker/expr_checker.py
acl-cqc: Fair, done, and yes, neater this way

maximilianruesch (approved):

PR #2283, opened 2026-09-07
feat: Support decimal places in rounding

### LINKED ISSUE #1561: [Feature]: Support decimal places in Guppy round for parity with Python round
### Problem Statement

Currently, the Guppy `round` function only accepts one argument: The number to round. This is compiled to calling the `__round__` function on the input type, which is a noop for `int` and `nat`, and defers to the `fround` Hugr op for `float`.

The Python version of `round` accepts either a single argument, doing a similar thing as above, or two arguments, where the second is the number of digits to round to.

This would be fine as a distinguishing feature. However, Guppy includes its `round` in the prelude, so even when not explicitly importing the Guppy version of the function, the "builtin" `round` dispatches to the Guppy version, thus erroring when a number of digits to round to is provided.

### Proposed Solution

We should try and include overloads for the rounding functions to accept a number of digits to round to, for the Guppy version to be a drop-in replacement for the Python builtin variant (as Guppy essentially forces your hand inside Guppy functions anyway).

After doing some research, the following steps seem to be necessary:
1. Making selene (or rather the QIS compiler) support rounding floats (See https://github.com/Quantinuum/hugr/pull/2949. Rudimentary integration and expected results up to Guppylang were manually verified). DONE: `hugr-llvm` supports emission of `FloatOps::fround` since `0.27.0`.
2. Pulling the QIS compiler update from `hugr-llvm` all the way into a release of the Python package `selene-hugr-qis-compiler` in the `tket2` repository. This has been released in `selene-hugr-qis-compiler==0.3.0`.
3. Allowing more than one fixed number of arguments in the `DunderChecker`, that is used to dispatch to dunder function calls
4. Supporting overloaded function targets in the `DunderChecker`
5. Implementing an overloaded `__round__` on `nat, int, float` that, for the first two, does implements two noop variants, and for the last, either dispatches to the `fround` op or shifts, dispatches, unshifts.
### Component

Standard library

### Effort Estimate

Hard - Coordination across multiple repositories and understanding of Rust required

### Additional Context

_No response_

maximilianruesch:
Closes #1561


maximilianruesch:
Hmm, I should probably pull the upgrade out of this PR

guppylang/src/guppylang/std/num.py
```diff
@@ -451,14 +467,23 @@ def __rmod__(self: float, other: float) -> float: ...
...
     @guppy
     @no_type_check
-    def __round__(self: float) -> int:
+    def ___round__no_digits(self: float) -> int:
         return int(self.___round__hugr())
 
+    @guppy
+    @no_type_check
+    def ___round__digits(self: float, ndigits: int) -> float:
+        factor = 10.0**ndigits
```
cqc-alec: We should gracefully handle cases where:
- the exponentiation overflows;
- `ndigits` is negative.
(I guess both should be errors.)

tests/integration/test_arithmetic.py
hsemenenko, line 1: Should this support rounding negative digits? In Python, `round(1,-1)` gives `0` but adding a test here gives `1`.

maximilianruesch: Good catch. I have added support, although I am not particularly happy about how it is implemented (conversion to float, applying its rounding, then converting back). Any thoughts?

hsemenenko: Probably not ideal. What about something like this
```python
    @guppy
    @no_type_check
    def ___round__digits(self: int, ndigits: int) -> int:
        if ndigits >= 0:
            return self
        factor = 10 ** (-ndigits)
        quotient = self // factor
        remainder = self % factor
        half = factor // 2
        if remainder > half or (remainder == half and quotient % 2 != 0):
            return (quotient + 1) * factor
        return quotient * factor
```

maximilianruesch: You are saying basically reimplementing `froundeven`? I am unsure if I like this...

guppylang/src/guppylang/std/num.py
```diff
@@ -531,11 +562,38 @@ def len(x): ...
 def pow(x, y): ...
 
 
+@guppy
+@no_type_check
+def ___round__inv(x: int, ndigits: int) -> int:
```
hsemenenko, line 567: Could consider making this `nat` to enforce being strictly positive.
```suggestion
def ___round__inv(x: int, ndigits: nat) -> int:
```

hsemenenko (approved):
Looks good, one comment about using `nat` that isn't strictly necessary.

guppylang/src/guppylang/std/num.py
maximilianruesch: This would cause extra function invocations, and since this function is private, enforcing this contract informally should be enough.

PR #2284, opened 2026-09-07
perf: Upgrade hugr to v0.18.6, tket to v0.15.8, qis-compiler to v0.4.3

maximilianruesch:
Upgrades the toolchain across the board for some performance improvements.

PR #2286, opened 2026-09-07
fix: Use froundeven for lowering float roundings

maximilianruesch:
Ensures that we use tie-to-even instead of tie-away-from-zero, for more consistency with Pythons `round`.

### DIFF
--- a/guppylang/pyproject.toml
+++ b/guppylang/pyproject.toml
@@ -30,7 +30,7 @@ classifiers = [
 dependencies = [
     "guppylang-internals==1.0.1",
     "numpy>=2.2.6,<3",
-    "selene-hugr-qis-compiler~=0.4.2",
+    "selene-hugr-qis-compiler~=0.4.3",
     "selene-sim~=0.3.0",
     "tqdm>=4.70.0",
     "pytket>=2.7.0",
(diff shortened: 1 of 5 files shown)


maximilianruesch:
This should be backported to v1.0.X as well, if we can do it (considering dependency constraints)

guppylang/pyproject.toml
```diff
@@ -30,7 +30,7 @@ classifiers = [
 dependencies = [
     "guppylang-internals==1.0.1",
     "numpy>=2.2.6,<3",
-    "selene-hugr-qis-compiler~=0.4.2",
+    "selene-hugr-qis-compiler~=0.4.3",
```
aborgna-q, line 33: This is fine for backporting, but for `guppy 1.1` we should bump it to `qis-compiler 0.5`

aborgna-q (approved):

guppylang/pyproject.toml
maximilianruesch: Yes, that upgrade will be done separately. I have now pulled out the actual dependency upgrades in https://github.com/Quantinuum/guppylang/pull/2284, so that this PR does not have a perf improvement attached that it should not have.

PR #2287, opened 2026-09-07
chore: Support decorator factories with `_dummy_custom_decorator`

hsemenenko:
When building sphinx docs for external modules that use `@link_name` I get an error:

```
@link_name("my_func")
TypeError: _dummy_custom_decorator.<locals>.<lambda>() takes 1 positional argument but 2 were given
```

This PR adds support for the decorator factories that return a callable decorator.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/decorator/gpu.py
+++ b/guppylang-internals/src/guppylang_internals/decorator/gpu.py
@@ -93,8 +93,4 @@ def _gpu_helper[**P, T](
 if not TYPE_CHECKING and sphinx_running():
     gpu_module = _dummy_custom_decorator
 
-    # Support both forms of the decorator
-    def gpu(*args: object, **kwargs: object):  # type: ignore[no-redef]
-        if len(args) == 1 and callable(args[0]) and not kwargs:
-            return args[0]
-        return _dummy_custom_decorator(*args, **kwargs)
+    gpu = _dummy_custom_decorator()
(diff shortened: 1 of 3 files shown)


maximilianruesch (commented):
Can you find an example of this use in the regular guppy docs? Perhaps one that would break without this fix?

maximilianruesch (changes requested):
I realised that this applies only to generating API docs for annotated functions, not actually tests / examples that execute them.

However I think you should take the opportunity add a sneaky test for this, that directly accesses the `_dummy_custom_decorator` function and tests that it outputs decorators that correctly return the underlying function.

cqc-alec:
The code [here](https://github.com/Quantinuum/guppylang/blob/3560b6ec088e7646bd137eeaee5fe97e220ad9c9/guppylang-internals/src/guppylang_internals/decorator/gpu.py#L96) could be cleaned up with this change.

hsemenenko:
@cqc-alec Good catch, I've cleaned up the `gpu` decorator and added a test that directly uses the `_dummy_custom_decorator`.

cqc-alec (approved):

maximilianruesch (approved):

guppylang-internals/src/guppylang_internals/decorator/gpu.py
```diff
@@ -93,8 +93,4 @@ def _gpu_helper[**P, T](
 if not TYPE_CHECKING and sphinx_running():
     gpu_module = _dummy_custom_decorator
 
-    # Support both forms of the decorator
-    def gpu(*args: object, **kwargs: object):  # type: ignore[no-redef]
-        if len(args) == 1 and callable(args[0]) and not kwargs:
-            return args[0]
-        return _dummy_custom_decorator(*args, **kwargs)
+    gpu = _dummy_custom_decorator()
```
maximilianruesch, line 95: Consider removing the space between the two. But otherwise 🚀

PR #2294, opened 2026-09-07
feat!: Upgrade to qis-compiler v0.5.0 and selene-sim v0.3.1

maximilianruesch:
Upgrades to the latest breaking version of the `selene-hugr-qis-compiler`, also upgrading selene with support for the new ops in the process.

BREAKING CHANGE: The seeding behaviour of the emulator has changed. Pass `mode = "legacy"` to `with_seed` to reproduce previous results.


guppylang/src/guppylang/emulator/instance.py
```diff
@@ -240,10 +241,17 @@ def with_timeout(self, value: datetime.timedelta | None) -> Self:
         Defaults to None (no timeout)."""
         return self._with_option(_timeout=value)
 
-    def with_seed(self, value: int | None) -> Self:
+    def with_seed(
+        self, value: int | None, mode: Literal["default", "legacy"] = "default"
```
maximilianruesch:
```suggestion
        self, value: int | None, *, mode: Literal["default", "legacy"] = "default"
```

tatiana-s (approved):

PR #2300, opened 2026-09-08
chore: Add llvmlite dependency constraints for non-macos platforms

maximilianruesch:
Recent release workflows have been battling resolving / building dependencies that are not yet fully compatible with setuptools v81.0.0 (such as llvmlite v0.45.1). We must add appropriate llvmlite constraint dependencies for now.

See https://github.com/Quantinuum/guppylang/actions/runs/34241378148/job/102112120510 for a succeeding "highest" run.

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -107,8 +107,8 @@ jobs:
           version: ${{ env.UV_VERSION }}
           enable-cache: true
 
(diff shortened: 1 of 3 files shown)


maximilianruesch:
We will also need to backport this onto the v1.0 branch

aborgna-q (approved):
Ouch -.-

PR #2311, opened 2026-09-09
perf(?): Basic blocks not conditionals when successors need different subsets of outputs

acl-cqc:
Not sure whether this is refactor (I mean it changes the Hugr), perf (effect on performance unclear) or feat (it changes the Hugr to reduce the amount of weird stuff??)....

But, presently, when a block in a CFG has successors that need different subsets of the outputs, we build a big SumType abstracting over the differences (actually just over *all* the nonlinear outputs, not just those ones that are different), to use as the branch predicate; and the predecessor block gains a Conditional which takes all of the nonlinear outputs and then builds one or other variant of that SumType with only the desired values.

Instead, insert "adaptor blocks" inbetween the predecessor and each successor which uses only some of the predecessor's outputs. Successors using all the outputs need no adaptor (when all successors use all the outputs, this was previously special-cased to avoid the Conditional, but this now falls out naturally / generally).

The nasty bit is we have to pass the adaptors up out of `compile_bb` so we can route the edges correctly in `compile_cfg`, but even that is just an extra return value, did not need to add anything to e.g. `class BB`.

### DIFF
--- a/guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
+++ b/guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
@@ -1,5 +1,5 @@
 import functools
-from collections.abc import Sequence
+from collections.abc import Mapping, Sequence
 
 from hugr import Wire
 from hugr import tys as ht
@@ -12,7 +12,7 @@ from guppylang_internals.checker.cfg_checker import (
     Row,
     Signature,
 )
-from guppylang_internals.checker.core import Place, Variable
+from guppylang_internals.checker.core import Place, PlaceId, Variable
 from guppylang_internals.compiler.builder import BlockBuilder, DFBuilder, ops
 from guppylang_internals.compiler.core import (
     CompilerContext,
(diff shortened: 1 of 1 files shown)


guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
```diff
@@ -112,50 +118,49 @@ def compile_bb(
...
-    else:
-        # Exit variables are not allowed to be sorted since their order corresponds to
-        # the function outputs
-        assert len(bb.successors) == 1, "Exit jumps are always unconditional"
+        # Identify all variables used by any successor block
+        var_map: Mapping[PlaceId, Place] = {
+            p.id: p for row in bb.sig.output_rows for p in row
+        }
+        outputs = sort_vars(list(var_map.values()))
+        # Can we use 'Dom` edges for droppable outputs? I think no - each output is
```
acl-cqc, line 141: I mean possibly we could but it would be kinda orthogonal to this

guppylang-internals/src/guppylang_internals/compiler/cfg_compiler.py
```diff
@@ -112,50 +118,49 @@ def compile_bb(
...
+        # used by some BB as an actual input, so we must pass them as explicit outputs.
+        for i, r in enumerate(bb.sig.output_rows):
+            if len(r) == len(var_map):
+                assert {p.id for p in r} == {p.id for p in var_map.values()}
```
acl-cqc, line 145: Maybe drop this? the `len` check seems like it should be both cheap and precise

PR #2314, opened 2026-09-10
feat: Accept bytes into emulator builder

maximilianruesch:
The underlying builder already accepts bytes, so this is literally just a type widening.

Putting this up for backport since it is very minimal but very useful for reducing uneccesary serialisation roundtrips in `guppyft`.

### DIFF
--- a/guppylang/src/guppylang/emulator/builder.py
+++ b/guppylang/src/guppylang/emulator/builder.py
@@ -133,7 +133,7 @@ class EmulatorBuilder:
 
     def build(
         self,
-        package: Package,
+        package: Package | bytes,
         n_qubits: int,
         arg_specs: Sequence[EntrypointArgSpec] = (),
     ) -> EmulatorInstance:


guppylang/src/guppylang/emulator/builder.py
aborgna-q, line 1: My only worry here is with ensuring extension compatibility for the pre-serialized package.

What happens if guppyft produces a hugr with a newer-but-compatible std extension that what selene can load?
(Just flagging this. It needs a longer design discussion for the whole hugr ecosystem, so happy to merge this as is).

aborgna-q (approved):

guppylang/src/guppylang/emulator/builder.py
maximilianruesch: > What happens if guppyft produces a hugr with a newer-but-compatible std extension that what selene can load?

Great question, it most likely blows up. I would guess that we will need to start serialising std extensions by default at some point, and only skip them if we know that the receiving target has sufficiently high versions, similar to what we are pulling in tket.

PR #2317, opened 2026-09-11
perf: Avoid single dispatch on hot code paths

maximilianruesch:
`@functools.singlemethoddispatch` is really slow compared to just invoking the functions directly via an `isinstance` check, and this is particularly relevant on hot code paths like type transformations. Thus, for these paths, prefer `isinstance`(or here, a `match` case, which is equivalent).


maximilianruesch:
Performance is improved on compilation workloads as well, since those include checking

aborgna-q (approved):

aborgna-q:
> +31.16%

Nice!

PR #2318, opened 2026-09-11
fix: Hide trace in @guppy.unitary decorator error

nicolaassolini-qntm:
before:
image
after:
image


maximilianruesch (commented):
Can you add a snapshot test for this change?

cqc-alec (approved):

nicolaassolini-qntm:
> Can you add a snapshot test for this change?

@maximilianruesch 
There was already a test for the error (https://github.com/Quantinuum/guppylang/blob/9624c438d510935067bc29e21aaf92702d30b452/tests/error/test_decorator_errors.py#L71). I moved it to a snapshot test. For coherence, I did the same with the other 2 similar tests

PR #2342, opened 2026-09-18
fix(compiler): Clear stale inout flags on copyable generic captures

### LINKED ISSUE #2248: [Bug]: Calling a generic function inside a `with control(...)` block fails to compile
### Description

Calling a generic function inside a `with control(q):` block, where the argument is the enclosing generic function's own type parameter, fails during Hugr compilation with an internal `ValueError` about mismatched output types. The type variable `T` appears to be leaking into the modified (control) block's output signature.

### Steps to Reproduce

```python
from guppylang import guppy, qubit
from guppylang.std.builtins import control, nat


@guppy
def goo[T](x: T) -> None:
    pass

@guppy
def fun[T](x: T) -> None:
    q = qubit()
    with control(q):
        goo(x)

    discard(q)

@guppy
def main() -> None:
    fun(1)

main.compile()
```

### Expected Behaviour

The program compiles successfully. Calling `goo(x)` inside a `with control(q):` block should behave the same as outside it — the generic argument `x: T` should be forwarded without affecting the control block's output signature.

### Actual Behaviour

Compilation fails with an internal error while setting the outputs of the modified block:

```
ValueError: The function has fixed output type [], but was given output wires with types
[ExtType(type_def=TypeDef(name='int', ...), args=[BoundedNatArg(n=6)])].
```

Truncated traceback:

```
  File ".../guppylang_internals/compiler/stmt_compiler.py", line 231, in visit_CheckedModifiedBlock
    compile_modified_block(node, self.dfg, self.ctx, self.expr_compiler)
  File ".../guppylang_internals/compiler/modifier_compiler.py", line 71, in compile_modified_block
    func_builder.set_outputs(*cfg)
  File ".../guppylang_internals/compiler/builder/__init__.py", line 236, in set_outputs
    super().set_outputs(*outputs)
  File ".../hugr/build/dfg.py", line 780, in set_outputs
    raise ValueError(error_message)
```

### Additional Context

The failure occurs in `compile_modified_block` when computing the control block's outputs. The value of type `T` (monomorphised to `int` from `main`'s `fun(1)` call) is incorrectly included in the modified block's output wires, whose fixed signature is empty (`[]`).

akshaysoftware:
Fixes #2248 

Generic values captured inside modifier blocks could retain a stale `Inout` flag after being instantiated with a copyable type, causing the generated block signature and CFG outputs to disagree.

This change clears `Inout` for copyable captures while preserving it for genuinely non-copyable values.

Added regression tests for both copyable and non-copyable generic instantiations.


tests/integration/test_modifier.py
```diff
@@ -134,6 +134,51 @@ def bar() -> None:
...
+        q = qubit()
+
+        with control(q):
+            generic_function(x)
+
+        discard(q)
+
+    @guppy
+    def main() -> None:
+        controlled_generic_call(1)
```
nicolaassolini-qntm, line 153: Can you add an error test where we call `controlled_generic_call(q: qubit)` and check that the checker properly rejects the function?

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
```diff
@@ -84,8 +85,8 @@ def _set_inout_if_non_copyable(var: Variable) -> Variable:
     """Set the `inout` flag if the variable is non-copyable."""
     if not var.ty.copyable:
         return var.add_flags(InputFlags.Inout)
-    else:
-        return var
+    # if copyable capture then not Inout, retaining any other flags
+    return replace(var, flags=var.flags & ~InputFlags.Inout)
```
nicolaassolini-qntm: instead of replace, can we add a `remove_flag` to the Variable class?

nicolaassolini-qntm (changes requested):
thanks for the helps, just few comments

tests/integration/test_modifier.py
akshaysoftware: Added an error test without the `@guppy(controllable=True)` annotation which passes

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
akshaysoftware: Added a `remove_flags` method to the `Variable` class and called it directly

akshaysoftware:
Thanks for the review @nicolaassolini-qntm! I have addressed your comments, do let me know if there is anything else 😄

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
```diff
@@ -84,8 +84,8 @@ def _set_inout_if_non_copyable(var: Variable) -> Variable:
     """Set the `inout` flag if the variable is non-copyable."""
     if not var.ty.copyable:
         return var.add_flags(InputFlags.Inout)
-    else:
-        return var
+    # if copyable capture then not Inout, retaining any other flags
```
nicolaassolini-qntm:
```suggestion
    # Copiable variables may retain the Inout flag.
```

nicolaassolini-qntm (changes requested):
All good, just a little comment improvent

guppylang-internals/src/guppylang_internals/checker/modifier_checker.py
akshaysoftware: I've applied the suggestion, thanks 😄

nicolaassolini-qntm (approved):
All good thanks!

